Issue #5, opened 2023-04-12
Create a Resource ID dimension


udam-f2:
Define how the resource id associated with the cost incurred should appear in cost data under dimensions.

Definition of done:

- [ ] Rationalize vendor-neutral, cross-cloud naming 
- [ ] Complete spec template and include naming (code name, display name), constraints, guidelines, compatibility with major providers etc.

PR #42, opened 2023-05-04
FOCUS #5: Resource ID spec changes and supporting documents

udam-f2:
Resource ID spec changes and supporting documents

### DIFF
--- a/specification/dimensions/resourceid.md
+++ b/specification/dimensions/resourceid.md
@@ -1,0 +1,30 @@
+# Resource ID
+
+A Resource ID is an identifier assigned to a resource by the provider. The Resource ID is commonly used for cost reporting, analysis, and allocation scenarios.
+
+The ResourceId column MUST be present in the billing data. This column MUST be of type String. The ResourceId value MAY be a nullable column as some cost data rows may not be associated with a resource. ResourceId MUST appear in the cost data if an identifier is assigned to a resource by the provider. ResourceId SHOULD be a fully-qualified identifier that ensures global uniqueness within the provider.
+
+## Column ID
+
+ResourceId
+
+## Display name
+
+Resource ID
+
+## Description
+
+Identifier assigned to a resource by the provider.
+
+## Content constraints
+
+| Constraint      | Value           |
+|-----------------|-----------------|
+| Column required | True            |
+| Data type       | String          |
+| Allows nulls    | True            |
+| Value format    | \<not specified> |
+
+## Introduced (version)
+
+0.5
--- a/supporting_content/dimensions/resourceid.md
+++ b/supporting_content/dimensions/resourceid.md
@@ -1,0 +1,59 @@
+# Column: ResourceId
+
+## Example provider mappings
+
+Current column mappings found in available data sets:
+
+| Provider  | Data set                 | Column                               |
+|-----------|--------------------------|--------------------------------------|
+| AWS       | CUR                      | line_item_resource_id                |
+| GCP       | Big Query Billing Export | resource.global_name                 |
+| Microsoft | Cost details             | ResourceId                           |
+
+## Example usage scenarios
+
+Current values observed in billing data for various scenarios:
+
+| Provider  | Scenario                                          | Pattern                                                                                               |
+|-----------|---------------------------------------------------|-------------------------------------------------------------------------------------------------------|
+| AWS       | ARN provided which includes a resource name or id | arn:partition:service:region:account-id:resource-id <br>arn:partition:service:region:account-id:resource-type/resource-id<br> arn:partition:service:region:account-id:resource-type:resource-id|
+| AWS       | Some cases, just a resource-id is provided        | I-0b21f4c4434558933 OR my-cur-bucket                                                                  |
+| AWS       | Tax, API call charges etc.                        | (null)                                                                                                |
+| GCP       | Resource based                                    | //container.googleapis.com/projects/<project_id>/locations/\<location>/clusters/\<cluster>            |
+| GCP       | Non-resource based (e.g. network egress)          | (null)                                                                                                |
+| Microsoft | Resource group resources                          | /subscriptions/\<guid>/resourceGroups/\<name>/providers/\<provider>/\<resource-type>/\<resource-name> |
+| Microsoft | Subscription resources                            | /subscriptions/\<guid>/providers/\<provider>/\<resource-type>/\<resource-name>                        |
+| Microsoft | Tenant resources                                  | /providers/\<provider>/\<resource-type>/\<resource-name>                                              |
+| Microsoft | Reservation purchases                             | /providers/Microsoft.Capacity/reservationOrders/\<guid>                                               |
+| Microsoft | Savings plan purchases                            | /providers/Microsoft.BillingBenefits/savingsPlanOrders/\<guid>                                        |
+| Microsoft | Marketplace and other purchases                   | (null)                                                                                                |
+
+### Documentation
+
+AWS:
+> [AWS User Guide documentation on ARNs](https://docs.aws.amazon.com/IAM/latest/UserGuide/reference-arns.html)
+
+Interesting note from above link on the "The resource identifier" regarding the last part of the ARN:
+
+"This is the name of the resource, the ID of the resource, or a [resource path](https://docs.aws.amazon.com/IAM/latest/UserGuide/reference-arns.html#arns-paths). Some resource identifiers include a parent resource (sub-resource-type/parent-resource/sub-resource) or a qualifier such as a version (resource-type:resource-name:qualifier)."
+
+Microsoft:
+> [Usage details](https://learn.microsoft.com/azure/cost-management-billing/automate/understand-usage-details-fields)
+
+GCP:
+> [Resource Names documentation](https://cloud.google.com/asset-inventory/docs/resource-name-format)
+
+## Discussion / Scratch space
+
+- Should there be a resource name and a separate fully qualified name? Use one column for resource ID and make a recommendation for the value to be globally unique within the provider (e.g. a fully qualified name)
+- Should this be prefixed with 'Provider' e.g. ProviderResourceID to show that it's a 1:1 mapping to a provider provided column.
+- Should governance say "Nullable where cost is not associated with a resource" to account for cost data not associated with a resource (e.g. API calls)
+- Do we make any statements about uniqueness? Or is this where we request a uniqueness composite key to be specified in a schema document
+  - Can we guarantee uniqueness and can it change over time?
+  - ResourceId can only be required to be unique within a provider not across providers (unless we dictate a format for fully qualifying the id e.g. ARN that includes provider).
+  - Also, note this is not the same as a line-item level uniqueness key. It's common for resource id's to show up many times in billing data for different time periods and different types of consumption
+- Empty values should not be allowed if a user or system generated ResourceId is null (should this be an attribute specified as a higher level attribute OR within this scope)?
+  - Should be specified in a higher-level attribute that applies spec-wide on how nulls/empty values and placeholders
+- Future attribute/dimension:
+  - Suggestion (Erik): Embrace a "Cloud Resource ID (CRID)" convention such as:
+    <br>  `Crid:provider:service-type:region:owner-account-id:resource-type:cloud-local-id`


specification/dimensions/resourceid.md
```diff
@@ -0,0 +1,30 @@
...
+
+ResourceId
+
+## Display name
+
+Resource ID
+
+## Description
+
+Identifier assigned to a resource by the provider
```
flanakin, line 17:
```suggestion
Identifier assigned to a resource by the provider.
```

specification/dimensions/resourceid.md
```diff
@@ -0,0 +1,30 @@
+# Resource ID
+
+A Resource ID is an identifier assigned to a resource by the provider. The Resource ID is commonly used for cost reporting, analysis and allocation scenarios.
```
flanakin, line 3: Do we need to agree on which style guide we're using for writing? For instance, should we use commas before the last item in lists, as shown here?
```suggestion
A Resource ID is an identifier assigned to a resource by the provider. The Resource ID is commonly used for cost reporting, analysis, and allocation scenarios.
```

specification/dimensions/resourceid.md
```diff
@@ -0,0 +1,30 @@
...
+
+The ResourceId column MUST be present in the billing data. This column MUST be of type String. The ResourceId value MAY be a nullable column as some cost data rows may not be associated with a resource. ResourceId MUST appear in the cost data if an identifier is assigned to a resource by the provider. ResourceId SHOULD be a fully-qualified identifier that ensures global uniqueness within the vendor.
```
flanakin, line 5: "...within the provider"?
```suggestion
The ResourceId column MUST be present in the billing data. This column MUST be of type String. The ResourceId value MAY be a nullable column as some cost data rows may not be associated with a resource. ResourceId MUST appear in the cost data if an identifier is assigned to a resource by the provider. ResourceId SHOULD be a fully-qualified identifier that ensures global uniqueness within the provider.
```

supporting_content/dimensions/resourceid.md
```diff
@@ -0,0 +1,55 @@
+# Dimension: ResourceId
+
+## Example mappings
+
+| Provider  | Data set                 | Vendor column                        |
```
flanakin, line 5: Should this be either "Provider column" or just "Column"?
```suggestion
| Provider  | Data set                 | Column                        |
```

supporting_content/dimensions/resourceid.md
```diff
@@ -0,0 +1,55 @@
+# Dimension: ResourceId
```
flanakin, line 1:
```suggestion
# Column: ResourceId
```

supporting_content/dimensions/resourceid.md
```diff
@@ -0,0 +1,55 @@
...
+
+## Example mappings
+
+| Provider  | Data set                 | Vendor column                        |
+|-----------|--------------------------|--------------------------------------|
+| AWS       | CUR                      | line_item_resource_id                |
+| GCP       | Big Query Billing Export | resource.name? resource.global_name? |
+| Microsoft | Cost details             | ResourceId                           |
+
+## Examples
```
flanakin, line 11: Do we need to qualify that these are existing examples and not suggested values when implementing FOCUS?

supporting_content/dimensions/resourceid.md
```diff
@@ -0,0 +1,55 @@
...
+| GCP       | Non-resource based (e.g. network egress)          | (null)                                                                                           |
+| Microsoft | Resource group resources                          | /subscriptions/<guid>/resourceGroups/<name>/providers/<provider>/<resource-type>/<resource-name> |
+| Microsoft | Subscription resources                            | /subscriptions/<guid>/providers/<provider>/<resource-type>/<resource-name>                       |
+| Microsoft | Tenant resources                                  | /providers/<provider>/<resource-type>/<resource-name>                                            |
+| Microsoft | Purchase charges                                  | (null)                                                                                           |
+| Microsoft | Reservation purchases                             | /providers/Microsoft.Capacity/reservationOrders/<guid>                                           |
+| Microsoft | Savings plan purchases                            | /providers/Microsoft.Capacity/savingsPlanOrders/<guid>                                           |
```
flanakin, line 24:
```suggestion
| Microsoft | Savings plan purchases                            | /providers/Microsoft.BillingBenefits/savingsPlanOrders/<guid>                                           |
```

supporting_content/dimensions/resourceid.md
```diff
@@ -0,0 +1,55 @@
...
+| AWS       | Tax, API call charges etc.                        | (null)                                                                                           |
+| GCP       | Resource based                                    | //container.googleapis.com/projects/<project_id>/locations/<location>/clusters/<cluster>         |
+| GCP       | Non-resource based (e.g. network egress)          | (null)                                                                                           |
+| Microsoft | Resource group resources                          | /subscriptions/<guid>/resourceGroups/<name>/providers/<provider>/<resource-type>/<resource-name> |
+| Microsoft | Subscription resources                            | /subscriptions/<guid>/providers/<provider>/<resource-type>/<resource-name>                       |
+| Microsoft | Tenant resources                                  | /providers/<provider>/<resource-type>/<resource-name>                                            |
+| Microsoft | Purchase charges                                  | (null)                                                                                           |
```
flanakin, line 22: nit: We might want to move this down after reservation and savings plan purchases, since it's the catch-all bucket.
```suggestion
| Microsoft | Marketplace and other purchases                                  | (null)                                                                                           |
```

supporting_content/dimensions/resourceid.md
```diff
@@ -0,0 +1,55 @@
...
+| Microsoft | Reservation purchases                             | /providers/Microsoft.Capacity/reservationOrders/<guid>                                           |
+| Microsoft | Savings plan purchases                            | /providers/Microsoft.Capacity/savingsPlanOrders/<guid>                                           |
+
+### Documentation
+
+AWS:
+> <https://docs.aws.amazon.com/IAM/latest/UserGuide/reference-arns.html>
```
flanakin, line 29: nit: Should this be in Markdown format?

supporting_content/dimensions/resourceid.md
```diff
@@ -0,0 +1,55 @@
...
+
+Interesting note on the "The resource identifier" which is at the end of the ARN: This is the name of the resource, the ID of the resource, or a [resource path](https://docs.aws.amazon.com/IAM/latest/UserGuide/reference-arns.html#arns-paths). Some resource identifiers include a parent resource (sub-resource-type/parent-resource/sub-resource) or a qualifier such as a version (resource-type:resource-name:qualifier)."
```
flanakin, line 31: Do we care to define ARN?

supporting_content/dimensions/resourceid.md
```diff
@@ -0,0 +1,55 @@
...
+
+Microsoft:
```
flanakin, line 33: https://learn.microsoft.com/azure/cost-management-billing/automate/understand-usage-details-fields

supporting_content/dimensions/resourceid.md
```diff
@@ -0,0 +1,55 @@
...
+
+GCP:
+> <https://cloud.google.com/apis/design/resource_names> <https://cloud.google.com/asset-inventory/docs/resource-name-format> 
+
+## Discussion / Scratch space
+
+- Use one column for resource ID and make strong recommendations for the value to be globally unique within the provider. Eg. Full ARN for AWS.
+
+- Should this be prefixed with 'Vendor' e.g. VendorResourceID to show that it's a 1:1 mapping to a vendor provided column.
```
flanakin, line 42: I believe this should be "provider", not "vendor".
```suggestion
- Should this be prefixed with 'Provider' e.g. ProviderResourceID to show that it's a 1:1 mapping to a provider-defined column. 
```

supporting_content/dimensions/resourceid.md
```diff
@@ -0,0 +1,55 @@
...
+> <https://docs.aws.amazon.com/IAM/latest/UserGuide/reference-arns.html>
+
+Interesting note on the "The resource identifier" which is at the end of the ARN: This is the name of the resource, the ID of the resource, or a [resource path](https://docs.aws.amazon.com/IAM/latest/UserGuide/reference-arns.html#arns-paths). Some resource identifiers include a parent resource (sub-resource-type/parent-resource/sub-resource) or a qualifier such as a version (resource-type:resource-name:qualifier)."
+
+Microsoft:
+
+GCP:
+> <https://cloud.google.com/apis/design/resource_names> <https://cloud.google.com/asset-inventory/docs/resource-name-format> 
+
+## Discussion / Scratch space
```
flanakin, line 38: nit: Should we move action items to issues and potential discussion points that don't have a specific action to discussions? Not sure if we want to commit these to the repo, where they won't be tracked/closed.

supporting_content/dimensions/resourceid.md
```diff
@@ -0,0 +1,55 @@
...
+
+- Use one column for resource ID and make strong recommendations for the value to be globally unique within the provider. Eg. Full ARN for AWS.
+
+- Should this be prefixed with 'Vendor' e.g. VendorResourceID to show that it's a 1:1 mapping to a vendor provided column. 
+
+- Should governance say "Nullable where cost is not associated with a resource"
+
+- Do we make any statements about uniqueness? Or is this where we request a uniqueness composite key to be specified in a schema document
+
+- Can uniqueness and can it change over time?
```
flanakin, line 48: Is this "Can we guarantee uniqueness..."? I don't remember.

supporting_content/dimensions/resourceid.md
```diff
@@ -0,0 +1,55 @@
...
+
+- Empty values should not be allowed if a user/systematically generated value is null (should this be an attribute specified as a higher level attribute OR within this scope)?
```
flanakin, line 50: Can we remove this given the null handling rules define empty values are not allowed?

flanakin (changes requested):

specification/dimensions/dimensions.mdpp
flanakin, line 1: Should we rename this folder to "columns"?

supporting_content/dimensions/resourceid.md
udam-f2: Section name says Example Mappings which should be clear hopefully.

supporting_content/dimensions/resourceid.md
udam-f2: I don't think so. This is really meant to capture scratch notes IMO. It's not going in the spec. I was considering not even converting to markdown to allow for it to be updated easily without needing to convert back-and-forth.

I've updated to indicate this is from the link above (which is official doc on ARN so should be sufficient)

supporting_content/dimensions/resourceid.md
udam-f2: My goal was to provide a sneak peek for the things we discussed / considered as a part of building this. It's not really to track anything. Agree on those needing to be issues in the repo.

udam-f2: Cleaned up more of the discussion items

supporting_content/dimensions/resourceid.md
udam-f2: Same as above, i think the goal is to show the things we discussed (sausage making process). Happy to discuss and change if we feel this doc / discussions space should be more buttoned down.

specification/dimensions/dimensions.mdpp
udam-f2: This will evolve as we add more. The idea would be to have specification folder, then dimensions, metrics and attributes folders. You might be suggesting a columns folder and then possibly separate dimensions and metrics in there? or consider everything a column. We can adjust as we go - if another approach provides more value/clarity.

supporting_content/dimensions/resourceid.md
flanakin: Hmm... if it's minutes, we're missing a bunch. I thought this was the open items we didn't close, which we should be tracking somewhere... 🤔

specification/dimensions/dimensions.mdpp
flanakin: We originally agreed not to use the term "dimension". I assumed that was a global determination. Since we're not building a BI system, there's really no difference between a "dimension" and a "metric" in the spec. We just have "columns" that have one type of data or another.

supporting_content/dimensions/resourceid.md
udam-f2: It's not minutes. This is just things we discussed and/or considered while building this. Minutes aren't a part of the PR. These are so people can get a sense for what was discussed and considered as a part of building the spec content.

If anything requires decisions beyond this work or needs to be tracked, they should be opened as an issue in the repo with the discussion topic label.

specification/dimensions/dimensions.mdpp
udam-f2: I've not heard a decision like that tbh. I might have missed it some how but they are definitely not just same with just a data type difference. This is fairly common data analytics terminology which is the primary use case for people dealing with cost data IMO.

Might be stating the obvious here but dimensions are qualitative values. Metrics/measures (we should decide the name now that we're working on the first cost metric) are quantitative values. You generally need at least one from each group to answer a question or tell a particular story with the data. As you know, there's other differences as well - like quantitative columns can be used for aggregations and other math operations while qualitative columns cannot.

It's clearer for our consumers IMO if we keep the different types of columns separated.

flanakin (approved):

Issue #23, opened 2023-04-19
Attribute: Create a convention for a column naming scheme


udam-f2:
### Description
Discussion about if we should use snake, camel or pascal case for column id naming?

### Proposed approach
Pascal case has the most traction for use in column ids.

Normalized values (e.g. Service Category values like Compute, Developer Tools etc.) are the other place this could apply but since we're allowing spaces in that context so there's no need to apply this for those values.

### Github issue or Reference
General spec-wide decision

### Context
Column id is different from the display name which is a UI friendly name for a column. E.g. ResourceName (Column id) vs Resource Name (display name)

flanakin:
My vote is for pascal-casing (EveryWordUppercased). Want me to take a stab at coming up with a proposal for column naming conventions? This would be good to review with everyone in a PR :-)

mike-finopsorg:
There is nothing official from the Linux Foundation on this. I also vote for PascalCase as well for column naming.

udam-f2:
Converting to issue for creating a new Attribute for column naming convention (not to be confused with column name spacing convention :) )

PR #44, opened 2023-05-07
FOCUS #23: Column naming convention attribute and supporting documents

udam-f2:
- Column naming convention
- Hook-in attributes section

### DIFF
--- a/specification/attributes/attributes.mdpp
+++ b/specification/attributes/attributes.mdpp
@@ -1,0 +1,10 @@
+# Attributes
+
+Attributes are requirements that apply to the billing datasets. Requirements on data content can include naming
+conventions, data types, formatting standardizations etc. Attributes may additionally introduce high-level requirements
+for data granularity, recency, frequency etc. Requirements defined in attributes are necessary for servicing 
+[FinOps capabilities][FODOFC] accurately and using a standard set of instructions regardless of the origin of the data.
+
+!INCLUDE "column_naming_convention.md",1
+
+[FODOFC]: https://www.finops.org/framework/capabilities/
--- a/specification/attributes/attributes.mdpp
+++ b/specification/attributes/attributes.mdpp
@@ -1,8 +1,8 @@
 # Attributes
 
 Attributes are requirements that apply to the billing datasets. Requirements on data content can include naming
-conventions, data types, formatting standardizations etc. Attributes may additionally introduce high-level requirements
-for data granularity, recency, frequency etc. Requirements defined in attributes are necessary for servicing 
+conventions, data types, formatting standardizations, etc. Attributes may introduce high-level requirements
+for data granularity, recency, frequency, etc. Requirements defined in attributes are necessary for servicing 
 [FinOps capabilities][FODOFC] accurately and using a standard set of instructions regardless of the origin of the data.
 
 !INCLUDE "column_naming_convention.md",1
--- a/specification/attributes/column_naming_convention.md
+++ b/specification/attributes/column_naming_convention.md
@@ -1,0 +1,34 @@
+# Column naming convention
+
+Column IDs provided in billing datasets following a consistent naming convention reduces friction for FinOps
+practitioners that consume the datasets for analysis, reporting, and other use cases.
+
+All columns defined in the FOCUS specification MUST follow the naming requirements listed below. Provider generated
+columns SHOULD adopt these same naming requirements over time.
+
+## Attribute ID
+
+ColumnNamingConvention
+
+## Attribute name
+
+Column Naming Convention
+
+## Description
+
+Naming Convention for columns appearing in billing data.
+
+## Requirements
+
+* Column IDs MUST use Pascal case
+* Column IDs MUST NOT use abbreviations
+* Column IDs SHOULD NOT use acronyms
+* Column IDs MUST be alphanumeric with no special characters
+
+## Exceptions
+
+* Identifiers will use the "Id" abbreviation since this is a standard pattern across the industry.
+
+## Introduced (version)
+
+0.5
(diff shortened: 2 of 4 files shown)


specification/attributes/attributes.mdpp
```diff
@@ -0,0 +1,10 @@
+# Attributes
+
+Attributes are requirements that apply to the billing datasets. Requirements on data content can include naming
```
flanakin, line 3: nit: I would recommend keeping paragraphs on a single line so insertions don't look like we're changing the entire paragraph, when we're just moving words to keep line lengths down.

specification/attributes/attributes.mdpp
```diff
@@ -0,0 +1,10 @@
...
+conventions, data types, formatting standardizations etc. Attributes may additionally introduce high-level requirements
```
flanakin, line 4:
```suggestion
conventions, data types, formatting standardizations, etc. Attributes may introduce high-level requirements
```

specification/attributes/attributes.mdpp
```diff
@@ -0,0 +1,10 @@
...
+for data granularity, recency, frequency etc. Requirements defined in attributes are necessary for servicing
```
flanakin, line 5:
```suggestion
for data granularity, recency, frequency, etc. Requirements defined in attributes are necessary for servicing 
```

specification/attributes/column_naming_convention.md
```diff
@@ -0,0 +1,34 @@
+# Column naming convention
+
+Column IDs provided in billing datasets following a consistent naming convention reduces friction for FinOps
```
flanakin, line 3: nit: I'm getting hung up on "billing" datasets. Technically, the columns we define will overlap with data that isn't directly related to the bill, like recommendations. Recommendations are about cost, not bills. Should we talk about "cost data" or "cost details" or something? Not a big deal. I just keep thinking about the related datasets that we'll overlap with and not all of them are related to getting billed and paying invoices.

specification/attributes/column_naming_convention.md
```diff
@@ -0,0 +1,34 @@
...
+practitioners that consume the datasets for analysis, reporting, and other use cases.
+
+All columns defined in the FOCUS specification MUST follow the naming requirements listed below. Provider generated
```
flanakin, line 6: Isn't "FOCUS specification" redundant? We should take an official position, as others have.
```suggestion
All columns defined in FOCUS MUST follow the naming requirements listed below. Provider generated
```

specification/attributes/column_naming_convention.md
```diff
@@ -0,0 +1,34 @@
...
+
+ColumnNamingConvention
+
+## Attribute name
+
+Column Naming Convention
+
+## Description
+
+Naming Convention for columns appearing in billing data.
```
flanakin, line 19:
```suggestion
Naming convention for columns appearing in billing data.
```

specification/attributes/column_naming_convention.md
```diff
@@ -0,0 +1,34 @@
...
+
+## Requirements
```
flanakin, line 21: Each of these bullets should end in a period.

specification/attributes/column_naming_convention.md
```diff
@@ -0,0 +1,34 @@
...
+
+* Column IDs MUST use Pascal case
```
flanakin, line 23: nit: Wikipedia refers to Pascal case as "upper camel case". Do we have an opinion on which we should use? I prefer Pascal case, but just calling it out to get others' perspective. Please resolve if no other comments 🙂

flanakin (approved):

specification/attributes/column_naming_convention.md
mike-finopsorg: We should just link to an online explanation of the format. I also like Pascal case.

specification/attributes/column_naming_convention.md
mike-finopsorg: Sorry what's the recommendation here. Instead of saying "billing datasets" use ....?

specification/attributes/column_naming_convention.md
mike-finopsorg: My recommendation is the AP Style as per: https://www.instructionalsolutions.com/blog/bulleted-list-punctuation

- Use a period or other full stop after every bullet that is a sentence.
- Use a period after bullet list that completes the opening stem sentence that introduces it.
- Don’t use a period after bullet lists that are not complete sentences or do not complete the opening stem sentence. 
- Don’t use semicolons to end punctuation. 
- Use either all full sentences in your bullet lists or all fragments. Avoid a mix.

mike-finopsorg: So in this case, yes to full stops.

specification/attributes/column_naming_convention.md
rileyjenk: Same I like Pascal case...but wow its origins are back to the Pascal language...was hoping to find a quick document on the format but haven't found anything particularly useful.

rileyjenk (approved):

specification/attributes/attributes.mdpp
udam-f2: I thought about this too. It's a really crappy experience for anyone trying to update/maintain text in a line that goes on for ever. Do we know if there's a standard way for doing this? I see the same pros and cons listed online. Some people break down by sentence. Open to changing in the future if the group feels strongly one way or the other.

specification/attributes/column_naming_convention.md
udam-f2: Updated to say cost data. Not sure about recommendations but saying cost data here seems like a fine change.

specification/attributes/column_naming_convention.md
udam-f2: Seems like this: https://techterms.com/definition/pascalcase is a decent one to link. Adding that.

specification/attributes/column_naming_convention.md
udam-f2: I think we mentioned in the docs that both are correct (even though the last S is for specification).. Updated

PR #47, opened 2023-05-09
FOCUS #10: Service Name related spec changes

udam-f2:
Service Name related spec changes

### DIFF
--- a/specification/dimensions/servicename.md
+++ b/specification/dimensions/servicename.md
@@ -1,0 +1,30 @@
+# Service Name
+
+Service Name is an identifier assigned to a service by the provider. The Service Name is commonly used for cost reporting, analysis and allocation scenarios.
+
+The Service Name column MUST be present in the cost data. This column MUST be of type String and MUST NOT contain null values.
+
+## Column ID
+
+ServiceName
+
+## Display name
+
+Service Name
+
+## Description
+
+Name of the service associated with the cost data.
+
+## Content constraints
+
+|    Constraint   |      Value      |
+|:----------------|:----------------|
+| Column required | True            |
+| Data type       | String          |
+| Allows nulls    | False            |
+| Value format    | \<not specified> |
+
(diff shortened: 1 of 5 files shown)


specification/dimensions/servicename.md
```diff
@@ -0,0 +1,30 @@
+# Service Name
+
+Service Name is an identifier assigned to a service by the provider. The Service Name is commonly used for cost reporting, analysis and allocation scenarios.
+
+The Service Name column MUST be present in the cost data. This column MUST be of type String and MUST NOT contain null values.
```
flanakin, line 5: In other specs, we used the column ID here. Do you have a preference? Either way, we should be consistent.
```suggestion
The ServiceName column MUST be present in the cost data. This column MUST be of type String and MUST NOT contain null values.
```

supporting_content/dimensions/servicename.md
```diff
@@ -0,0 +1,30 @@
...
+
+## Example usage scenarios
+
+Current values observed in billing data for various scenarios:
+
+| Provider  | Data set                 | Example Value                                         |
+|-----------|--------------------------|-------------------------------------------------------|
+| AWS       | CUR                      | AmazonS3 \| AmazonRDS                                 |
+| GCP       | Big Query Billing Export | Networking \| Cloud SQL \| BigQuery                   |
+| Microsoft | Cost details             | Virtual Machines \| Azure App Service I Azure Monitor |
```
flanakin, line 22: nit: Would it be simpler to use commas here? (Same for other examples.)
```suggestion
| Microsoft | Cost details             | Virtual Machines, Azure App Service, Azure Monitor |
```

flanakin (approved):

specification/dimensions/servicename.md
udam-f2: 100%. Let me update. I made the same comment on the charge type one PR :)

Issue #50, opened 2023-05-10
Automate the population of dimensions being added to the mdpp templates


mike-finopsorg:
### Description
Currently the inclusion of individual dimensions into the PDF requires manual entry into the mdpp files. This can lead to files being missed in the assembly process if the manual step is skipped. We should look at automating this process, but avoid having a situation where the addition of content in an order that makes sense is hard to manage.

### Definition of done
The manual step of linking dimensions into the mdpp file is now automated.

PR #51, opened 2023-05-11
issue #50 valdate for missing MD includes

mike-finopsorg:
Added validation step to PDF build to ensure we do not miss any MD files in the dimension and attribute folders. Any missing INCLUDES will cause the build to fail.

### DIFF
--- a/specification/validate_includes.py
+++ b/specification/validate_includes.py
@@ -1,0 +1,48 @@
+#!/usr/bin/env python3
+import argparse
+from os import listdir
+import sys
+import re
+from os.path import isfile, join
+
+def compile_includes(keyname):
+    """
+    This function will validate that a folder has not got any missing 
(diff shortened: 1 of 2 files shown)


specification/validate_includes.py
```diff
@@ -0,0 +1,48 @@
+#!/usr/bin/env python3
+import argparse
+from os import listdir
+import sys
+import re
+from os.path import isfile, join
+
+def compile_includes(keyname):
```
udam-f2, line 8: nit: we're no longer compiling these. Not a huge deal if we don't update.

udam-f2 (approved):

specification/validate_includes.py
mike-finopsorg: Added push to fix this. Good catch

PR #54, opened 2023-05-11
Issue52 - Added some extra spec tidy ups

mike-finopsorg:
In order to avoid added double paragraph spacing this change will remove the newline between the PDF anchor links. In addition the TOC created by markdownPP is very verbose and deep. This change will trim the third layer in the TOC out.


udam-f2 (approved):
LGTM

Issue #35, opened 2023-05-01
Attribute: Null handling for columns


udam-f2:
Define how null, empty values, n/a and not-set type values should be specified in cost data. Explain which one is the right one to use for different scenarios.

Some scenarios to think about are:
- Nullable columns vs non-nullable columns
- User defined columns such as tags where a user may set an empty string as a valid value
- Is there a scenario where we determine a column as not-nullable but a vendor cannot provide a value?

flanakin:
Thinking about null handling, should we include a placeholder in all column specs for how to identify why a charge has a null value for the specific field? I'm not sure if we can genericize this, but it might be worthwhile walking thru it. We've already identified a number of scenarios where columns are null today. This isn't a 0.5 thing. Just thinking about the issues I've seen when summarizing data based on nullable columns.

udam-f2:
@flanakin Could you please create a discussion topic for this? 
I worry that we're not going to define a complete set of cases where a column may have a null value. However, maybe we can just list a few common scenarios?

PR #56, opened 2023-05-14
FOCUS #35: Spec changes for NULL handling attribute

udam-f2:
Handling cases where a column doesn't have a valid value that can be presented.

### DIFF
--- a/specification/attributes/null_handling.md
+++ b/specification/attributes/null_handling.md
@@ -1,0 +1,33 @@
+# Null Handling
+
+Cost data rows that don’t have a value that can be presented for a column must be handled in a consistent way to reduce
+friction for FinOps practitioners that consume the data for analysis, reporting, and other use cases.
+
+All columns defined in FOCUS MUST follow the null handling requirements listed below. Provider generated columns SHOULD
+adopt these same null handling requirements over time.
+
+## Attribute ID
+
+NullHandling
+
+## Attribute name
+
+Null Handling
+
+## Description
+
+Indicates how to handle columns that don't have a value.
+
+## Requirements
+
+* Columns MUST use NULL when there isn't a value that can be specified for a nullable column.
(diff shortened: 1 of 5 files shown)


specification/attributes/null_handling.md
```diff
@@ -0,0 +1,33 @@
+# Null Handling
+
+Cost data rows that don’t have a value that can be presented for a column must be handled in a consistent way to reduce
```
flanakin, line 3: Do we have a stance on smart/curly quotes yet?
```suggestion
Cost data rows that don't have a value that can be presented for a column must be handled in a consistent way to reduce
```

supporting_content/attributes/null_handling.md
```diff
@@ -0,0 +1,29 @@
...
+
+## Example usage scenarios
+
+Current values observed in billing data for various scenarios:
+
+| Provider  | Data set                 | Scenario                                                                         |
+|-----------|--------------------------|----------------------------------------------------------------------------------|
+| AWS       | CUR                      | Product_operation shows the type of operation used by a consumer. <br>However, in scenarios where the line item represents tax, this column is set to null. |
+| GCP       | Big Query Billing Export | Credits.name is null when a credit is not applied                                |
+| Microsoft | Cost Details             | `TODO` need example                                                              |
```
flanakin, line 11: ResourceGroup is null when a charge is not from a resource deployed to a resource group.

flanakin (approved):

specification/attributes/null_handling.md
udam-f2: Did some digging on this yesterday. I'm not a 100% sure. From my research, they are generally considered a positive if its using smart quotes / considered more readable. Doesn't seem like the linter has support to fail based on one or the other (i'm sure we could create a rule if we wanted to). Tempted to say since most editors automatically does the conversion, we should stick to the 'smart' kind but very much open to ideas/suggestions. We can also go clean up later in bulk if needed.

mike-finopsorg (approved):

Issue #31, opened 2023-04-26
Create an Invoice Entity dimension


udam-f2:
Define how the invoicing entity name should appear in cost data under dimensions. Invoicing entity is the entity that is billing for the infra and services that was purchased. It may be the same as the publisher / provider or may be different based on marketplace type transactions and or 3rd party reseller use cases.
 
Definition of done:
- [ ] Rationalize vendor-neutral, cross-cloud naming
- [ ] Complete spec template and include naming (code name, display name), constraints, guidelines, compatibility with major providers etc.

PR #59, opened 2023-05-17
FOCUS #31: Invoicing Entity spec changes and supporting documents

ijurica:
Invoicing Entity spec changes and supporting documents


ijurica:
Closing due to commit using wrong user

Issue #30, opened 2023-04-26
Create a Publisher dimension


udam-f2:
Define how the publisher name should appear in cost data under dimensions. Publisher is the entity that created the infra and services a consumer was able to purchase on a particular platform (may be sold by a third-party).
 
Definition of done:
- [ ] Rationalize vendor-neutral, cross-cloud naming
- [ ] Complete spec template and include naming (code name, display name), constraints, guidelines, compatibility with major providers etc.

PR #60, opened 2023-05-17
FOCUS #30: Publisher spec changes and supporting documents

ijurica:
Publisher spec changes and supporting documents


ijurica:
Closing due to commit using wrong user

Issue #11, opened 2023-04-12
Create a Provider dimension


udam-f2:
Define how the provider name should appear in cost data under dimensions. Provider is the entity that is making the infra and services available for purchase.
 
Definition of done:
- [ ] Rationalize vendor-neutral, cross-cloud naming
- [ ] Complete spec template and include naming (code name, display name), constraints, guidelines, compatibility with major providers etc.

udam-f2:
Consider how transactions that happen in a marketplace get represented (separate dimension needed?)

PR #65, opened 2023-05-17
FOCUS #31: Invoicing Entity spec changes and supporting documents

ijurica:
Invoicing Entity spec changes and supporting documents

### DIFF
--- a/specification/dimensions/dimensions.mdpp
+++ b/specification/dimensions/dimensions.mdpp
@@ -6,6 +6,8 @@ dimensions to categorize, filter, and reveal details in your data when grouped w
 
 !INCLUDE "resourceid.md",1
 !INCLUDE "resourcename.md",1
+!INCLUDE "provider.md",1
+!INCLUDE "publisher.md",1
 !INCLUDE "invoiceissuer.md",1
 !INCLUDE "servicename.md",1
 
--- a/specification/dimensions/provider.md
+++ b/specification/dimensions/provider.md
@@ -1,0 +1,32 @@
+# Provider
+
+A Provider is an entity that made the resources and/or services available for purchase. It is commonly used for cost analysis and reporting scenarios.
+
+The Provider column MUST be present in the billing data. This column MUST be of type String and MUST NOT contain null values.
+
+See [Appendix: Origination of cost data](#originationofcostdata) section for examples of Provider, Publisher and Invoice Issuer values that can be used for various purchasing scenarios.
+
+## Column ID
+
+Provider
+
+## Display name
+
+Provider
+
+## Description
+
+The entity that made the resources and/or services available for purchase.
+
+## Content constraints
+
+| Constraint      | Value           |
+|-----------------|-----------------|
+| Column required | True            |
+| Data type       | String          |
+| Allows nulls    | False           |
+| Value format    | \<not specified> |
+
+## Introduced (version)
+
+0.5
--- a/specification/dimensions/publisher.md
+++ b/specification/dimensions/publisher.md
@@ -1,0 +1,32 @@
+# Publisher
+
+A Publisher is an entity that produced the resources and/or services that were purchased. It is commonly used for cost analysis and reporting scenarios.
+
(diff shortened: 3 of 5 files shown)


specification/dimensions/invoicingentity.md
```diff
@@ -0,0 +1,32 @@
+# Provider
+
+An Invoicing Entity is an entity responsible for invoicing for the infrastructure and/or services consumed. It is commonly used for cost analysis and reporting scenarios.
+
+The InvoicingEntity column MUST be present in the billing data. This column MUST be of type String and MUST NOT contain null values.
```
mike-finopsorg, line 5: should there be a space in InvoicingEntity?

specification/dimensions/invoicingentity.md
```diff
@@ -0,0 +1,32 @@
...
+
+See [Appendix: Origination of the cost data](../appendix/origination_of_cost_data.md) section for context (Provider/Publisher/Invoicing Entity) in various purchasing scenarios.
```
mike-finopsorg, line 7: We need to confirm how this link works in the spec PDF

specification/dimensions/invoicingentity.md
ijurica: discussed and chose to refer to the column ID, and not the display name in normative content, at least for now

specification/dimensions/invoicingentity.md
udam-f2: It won't. I added comment in the other PR for this. We need to use the anchor.

specification/dimensions/invoicingentity.md
udam-f2: So here where we're at with this (and definitely open to changing):

The first sentence is referring to the 'thing', thus using the column display name as a pro noun. I could be wrong in this. Michael Flanakin thought it should be lower case (e.g. "An invoice issuer is an entity responsible for". Alternatively we can use the Column ID. Which of the 3 options is the right one? Im not sure, but agree we need to just use one way. Can easily update other places once we decide.

The second sentence is normative content that uses the Column ID. It's attempting to refer to the column and some times the value inside the column (in the case of normalized columns where we say must be one of the following values). Again, open to an approach which we can then update in all others.

Any opinions (possibly from an expert) on this?

ijurica:
Short overview of the currently applied changes and leftovers...

**APPLIED CHANGES:**

- Due to dimension renaming (Invoicing Entity -> Invoice Issuer), the MD files were renamed and every mention of the terms Invoicing Entity and InvoicingEntity were replaced.
- Current ordering in the dimensions.mdpp file - invoiceissuer.md placed above servicename.md
- Two mentions of the term "infrastructure and/or services" were replaced with the term "resources and/or services".
- In the spec MD file, link to the corresponding appendix section uses anchor (#origination-of-cost-data). Can be easily changed to (#originationofcostdata).
- The proposed change, related to the wording of the reference to the appendix section in the supporting content MD file (see conversation in https://github.com/FinOps-Open-Cost-and-Usage-Spec/FOCUS_Spec/issues/11 Provider PR), has been implemented.

**LEFTOVERS:**

- Column ID vs. Display Name within normative content - see unresolved conversation in https://github.com/FinOps-Open-Cost-and-Usage-Spec/FOCUS_Spec/issues/11 Provider PR
- Note regarding dimension's common usage scenarios - see unresolved conversation in https://github.com/FinOps-Open-Cost-and-Usage-Spec/FOCUS_Spec/issues/11 Provider PR

specification/dimensions/invoicingentity.md
ijurica: changed to (#originationofcostdata)

udam-f2 (approved):
LGTM

specification/dimensions/invoicingentity.md
udam-f2: Will fix this more broadly in all docs given the decision from ChargeType PR

udam-f2: Will fix this more broadly in all docs given the decision from ChargeType PR

Issue #80, opened 2023-05-23
Increase document content width


mike-finopsorg:
### Description
Current rendering of the spec has a very narrow width. 

### Definition of done
Increased rendering width

Issue #79, opened 2023-05-23
Support watermarks and branch specific includes


mike-finopsorg:
### Description
> In order to make the release cycle easier it will be nice to generate the spec output with a watermark to highlight the status of the document. Also there are introduction sections that need to be swapped out for release versions

### Definition of done
> Watermarks supported and conditional includes work per branch

PR #81, opened 2023-05-24
Better formatting for specification Issues #79 #80

mike-finopsorg:
Support CSS styles to better format specification output
Added render width to better fit specification content
Added FOCUS Logo and formatted intro page
Updated builds to support optional include of draft watermark
Updated builds to export html and PDF including needed external files


udam-f2 (approved):
Looks great!

Issue #84, opened 2023-05-25
Issue with truncated last line in builds on OSX


mike-finopsorg:
### Description
When the make build runs on OSX it truncates the final line in the markdown

### Definition of done
No final line removal issue

mike-finopsorg:
The issue is found to be a problem with how we are using sed and the included version of sed on OSX. This Is causing more effort to maintain and I am making the suggestion we install GNU sed on OSX via brew so it matches the type of sed used in the real GitHub Actions build.

PR #85, opened 2023-05-25
Fix issue with OSX builds missing last line #84

mike-finopsorg:
Removed dependancy on OSX default sed
Added steps to install gsed (GNU Sed) on OSX for development This gsed on OSX is the same GNU sed used on GitHub Actions Possibly a different version but generally the same features. This simplifies the Makefile and avoids the duplicate effort to maintain.


udam-f2 (approved):
LGTM

Issue #83, opened 2023-05-24
Repository maintenance task


udam-f2:
### Description
Consider the following changes: 
- Limit logo size or add a fixed size. On wider browser windows, the image might take up the whole screen
- Font for PDF seems good but looks tiny for HTML. Not sure if its an issue on my browser
- Consider if headings should be made smaller

### Definition of done
Ensure both PDF (fixed width) and HTML (variable width) mediums show content in a readable manner.

PR #90, opened 2023-05-26
Issue #83 clean up HTML

mike-finopsorg:
Make it possible to use different CSS between PDF and HTML
Make it possible to use different CSS for different builds (foreshadowing publication)


udam-f2 (approved):
LGTM

Issue #18, opened 2023-04-17
Create a Sub Account Name dimension


udam-f2:
Define how a provider-defined logical resource container name should appear in cost data under dimensions. Accounts, projects and subscriptions are common provider defined grouping constructs used by vendors. 

'Resource group' or resource container are particularly bad names for this task since many providers have a separate resource group construct that can have a many -> one relationship with accounts/subscriptions etc. See: [Azure/AWS account structure](https://www.linkedin.com/pulse/account-structure-comparison-between-aws-azure-richard-lenan-zhao)

Definition of done:

- [ ] Rationalize provider-neutral, cross-cloud naming 
- [ ] Complete spec template and include naming (code name, display name), constraints, guidelines, compatibility with major providers etc.

udam-f2:
Related to https://github.com/FinOps-Open-Cost-and-Usage-Spec/FOCUS_Spec/issues/19

PR #96, opened 2023-05-30
FOCUS #18: Sub Account Name dimension

MichaelArkoosh-vega:
FOCUS #18: Sub Account Name dimension


rileyjenk (changes requested):
Sub Accounts can generate invoices.

MichaelArkoosh-vega:
> Sub Accounts can generate invoices.

Where they can generate and invoice (meaning usage in the account can result in an invoice (example I purchase an RI in sub account A.) I think we may need to change the terminology to "Sub accounts must be associated with a billing account as they do not **receive** invoices." - Thoughts?

MichaelArkoosh-vega:
> @udam-f2 They do roll up to the billing account - but that is why I was wondering if we wanted to move from generate to receive? As the underlying usage the invoice is generated from is in the the sub account but the invoice is received at the billing account.

udam-f2:
> > @udam-f2 They do roll up to the billing account - but that is why I was wondering if we wanted to move from generate to receive? As the underlying usage the invoice is generated from is in the the sub account but the invoice is received at the billing account.

Ah, got it now.. then yes, that makes sense. The only odd case we may need to verify is MSP type cases? Seems like they would be fine as well since the invoice is received by the MSP from the billing account. They may generate a different invoice for their underlying customers.

MichaelArkoosh-vega:
Updated the verbiage in the spec document to state "Sub Accounts do not receive an invoice" as, even though the invoice may show in the sub account it is the billing account that will receive/pay said invoice.

specification/dimensions/subaccountname.md
```diff
@@ -0,0 +1,41 @@
...
+Spec Content
+------------
+
+### Sub Account Name
+
+A sub account is an optional provider-supported construct for organizing resources and services connected to a billing account. Some common scenarios for using sub accounts include grouping based on organizational constructs and access management boundaries. Sub accounts must be associated with a billing account as they do not receive invoices.
+
+A sub account name is a display name assigned to a sub account. It is commonly used for cost allocation, reporting, and chargeback purposes.
+
+The SubAccountName column MUST be present in the billing data. This column MUST be of type String and MUST NOT contain null values. Where a provider doesn't support a sub account construct (only has a billing account), use the BillingAccountName value.
```
flanakin, line 14: Does this convey what we discussed?
```suggestion
The SubAccountName column MUST be present in the billing data. This column MUST be of type String and SHOULD NOT contain null values. If a provider doesn't support a sub account construct (only has a billing account), use the BillingAccountName value. If a provider does support a sub account construct, but the charge does not apply to a sub account, the SubAccountName column may be null.
```

specification/dimensions/subaccountname.md
```diff
@@ -0,0 +1,41 @@
+Work item: [Issues #18](https://github.com/FinOps-Open-Cost-and-Usage-Spec/FOCUS_Spec/issues/18)
+
+Owner: John Grubb / Michael Arkoosh
+
+Spec Content
+------------
+
```
ijurica, line 7: a few notes:

- selected block should be removed
- Sub Account Name should be Heading L1 and all others Heading L2

supporting_content/dimensions/subaccountname.md
```diff
@@ -0,0 +1,20 @@
+Work item: [Issues #18](https://github.com/FinOps-Open-Cost-and-Usage-Spec/FOCUS_Spec/issues/18)
+
+Owner: John Grubb / Michael Arkoosh
+
+Supporting content
+------------------
+
+#### Example provider mappings
```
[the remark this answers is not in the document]
rileyjenk: After pulling some examples and looking various options for account structures in AWS here is what was found. 
1.) There are multiple ways to have multiple invoice generating accounts in AWS regardless if the are stand-alone, master-payer, or linked accounts.  This means that from an AWS side an account can be a sub account (linked account) to a higher order account (master payer account) and generate its own invoice and will not be included on the master payer.

That being said as far as FOCUS is concerned , any account that generates an invoice is considered a billing account. The provider may also specify it as a sub account. In the data there won't be a link in the data for true relationship.

PR #97, opened 2023-05-31
FOCUS #16 : Create  billing account name dimension spec

amitw-g:
FOCUS #16 : Billing account name dimension


rileyjenk (changes requested):
1 ) I question us including any account name since they can be changed. Account IDs do not change. I would suggest that we don't include name at all. If we do have it then we have to understand that it is subject to change. 

2.)This has the same issue as https://github.com/FinOps-Open-Cost-and-Usage-Spec/FOCUS_Spec/pull/99#pullrequestreview-1454175668.

specification/dimensions/billingaccountname.md
```diff
@@ -0,0 +1,34 @@
+# Billing Account Name
+
+A billing account is a container for resources and/or services that are billed together in an invoice. Common scenarios for using billing accounts include grouping based on organizational constructs and hierarchy-based cost allocation approach.
+
+A billing account name is a display name assigned to a billing account. It is commonly used for cost reporting, invoice reconciliation, and cost allocation scenarios.
```
flanakin, line 5: Should we say this is used to differentiate charges from different billing accounts and providers? Cost reporting is too generic, invoice recon isn't completely accurate given this alone doesn't help with that, and allocation is also arguably generic.

specification/dimensions/billingaccountname.md
```diff
@@ -0,0 +1,34 @@
...
+
+The BillingAccountName column MUST be present in the billing data. This column MUST be of type String. The BillingAccountName MAY be a nullable column as some providers do not provide the ability to set a display name and in that case it MUST NOT duplicate the same value as the BillingAccountId. The BillingAccountName MUST NOT be null if a display name can be assigned to a billing account . BillingAccountName MUST be unique when a customer has more than one billing account.
```
flanakin, line 7: Why do we want to allow nulls? If someone groups by this, then it'll be strange to see that there's a charge that doesn't have a billing account name. Shouldn't we require it, even if the provider falls back to the ID? An ID is better than a null.

flanakin, line 7: There's an extra space before the second to last period.

flanakin, line 7: We cannot guarantee uniqueness in a display name. We should remove that. If we add that as a requirement, then we'll force providers to add the ID into the display name to ensure uniqueness, which we should avoid.

specification/dimensions/billingaccountname.md
```diff
@@ -0,0 +1,34 @@
...
+
+See [Appendix: Grouping of resources and/or services](https://docs.google.com/document/u/0/d/1U24it_zByg-jKB9iOmiFbrULtzj5G5wUtawvntyeJuA/edit) section for billing accounts, sub-/resource accounts and hierarchy-based cost allocation approach scenarios.
```
flanakin, line 9: Don't link to Google Docs.

specification/dimensions/dimensions.mdpp
```diff
@@ -11,5 +11,6 @@ dimensions to categorize, filter, and reveal details in your data when grouped w
 !INCLUDE "resourceid.md",1
 !INCLUDE "resourcename.md",1
 !INCLUDE "servicename.md",1
+!INCLUDE "billingaccountname.md",1
```
flanakin, line 14: I believe this list should be sorted alphabetically.

supporting_content/dimensions/billingaccountname.md
```diff
@@ -0,0 +1,26 @@
...
+| Provider  | Data set                 | Column                                                    |
+|-----------|--------------------------|-----------------------------------------------------------|
+| AWS       | CUR                      | Not available (bill_payer_account_id maps to BillingAccountId) |
+| GCP       | BigQuery Billing Export  | Not available (billing_account_id maps to BillingAccountId) |
+| Microsoft | Cost details             | EA: BillingAccountName, MCA: BillingProfileName, MOSA: SubscriptionName
+
+## Documentation
+
+- GCP: [Resource Hierarchy](https://cloud.google.com/resource-manager/docs/cloud-platform-resource-hierarchy#resource-hierarchy-detail)
+- Azure: [Resource Hierarchy](https://learn.microsoft.com/en-us/azure/governance/management-groups/overview), [Organizing resources](https://learn.microsoft.com/en-us/azure/cost-management-billing/manage/view-all-accounts)
```
flanakin, line 16: nit: You don't need the MG docs link.

flanakin (changes requested):

udam-f2:
> 1 ) I question us including any account name since they can be changed. Account IDs do not change. I would suggest that we don't include name at all. If we do have it then we have to understand that it is subject to change.
> 
> 2.)This has the same issue as [#99 (review)](https://github.com/FinOps-Open-Cost-and-Usage-Spec/FOCUS_Spec/pull/99#pullrequestreview-1454175668).



> 1 ) I question us including any account name since they can be changed. Account IDs do not change. I would suggest that we don't include name at all. If we do have it then we have to understand that it is subject to change.
> 
I think billing data is always going to be a point in time view of the current state of things right? We could say the same about resource names. Not against mentioning this may change over time.

Ultimately, the data is more understandable imo if its there. Also many teams will bake in some business context into the names (e.g. prod/int/dev or app/team names so its easy to group OR run rules to generate additional dimensions. For those reasons, I'd vote for keeping it. Others have thoughts on this?

> 2.)This has the same issue as [#99 (review)](https://github.com/FinOps-Open-Cost-and-Usage-Spec/FOCUS_Spec/pull/99#pullrequestreview-1454175668).

specification/dimensions/billingaccountname.md
udam-f2: Sounds good to me. More specific is better.

Curious, as it's defined here, why is invoice recon not accurate? When you get a bill for the usage captured here, thats what our finance / business office folks used to use to reconcile / process payments.

specification/dimensions/billingaccountname.md
udam-f2: This is saying if a customer has more than one, it should be unique within the customers group of accounts (not globally unique).

MichaelArkoosh-vega: I would agree with the not allowing nulls for Billing Account Name, especially where this is a more macro name vs a resource where there could be thousands of individual resources. We also are not allowing nulls on the sub account name so I think it makes sense not to allow nulls here either.

amitw-g: Fixed.

amitw-g: How about 
```
"The BillingAccountName column MUST be present in the billing data. This column MUST be of type String. The BillingAccountName MUST NOT be null if a display name can be assigned to a billing account. BillingAccountName MUST be unique when a customer has more than one billing account." 
```
?

udam-f2 (approved):

rileyjenk (approved):

Issue #17, opened 2023-04-17
Create a Billing Account ID dimension


udam-f2:
Define how the billing/invoiced account ID should appear in cost data under dimensions. Some vendors call this a 'management account' or a 'billing account'. 

Definition of done:

- [ ] Rationalize vendor-neutral, cross-cloud naming 
- [ ] Complete spec template and include naming (code name, display name), constraints, guidelines, compatibility with major providers etc.

udam-f2:
https://github.com/FinOps-Open-Cost-and-Usage-Spec/FOCUS_Spec/issues/16

Issue #91, opened 2023-05-26
Create an appendix for grouping of resources and/or services (Billing Account and Sub/Resource Account)


ijurica:
### Type
> Appendix section

### Description
> Need an appendix section that describes various dimensions used for grouping of resources and/or services. Currently, there are 2 constructs used for grouping of resources and/or services. These are Billing Account and Sub/Resource Account. Both are assigned an ID and a Name and thus the billing data contains 4 dimensions - Billing Account ID, Billing Account Name, Sub/Resource Account ID and Sub/Resource Account Name. This appendix section should be linked from all 4 dimensions.

amitw-g:
Also need to add a link to various dimension specs linking to the appendix.

PR #99, opened 2023-05-31
Focus #17 : Create billing account id dimension spec

amitw-g:
FOCUS #17: Billing Account Id dimension


rileyjenk (approved):

rileyjenk (dismissed):
The problem with this is that it conflicts with Sub accounts where separate invoices are generated. The other Account Column is Sub Account (not finished), which would be the same as the sub account and not include the higher order account. Example: Say you have a master payer account in AWS and then a sub account that has its own linked account and is setup to have its own invoice.  In the CUR there is a column  called PayerAccount which specifies the master payer account id, and then another column is the UsageAccountID, which is the sub account where the usage occurred...but it also the account id that shows up on the invoice.

udam-f2:
> @rileyjenk and I discussed this. The distinct property of a billing account is that it is what you get an invoice for. Could argue we should name this Invoiced Account instead :). Sub Accounts are an optional construct that allows grouping below billing account level (for resource grouping or access management based on organizational constructs). Riley suggested there be a higher level dimension that allows these billing accounts to be grouped further (which would be optional as well). @ijurica has also suggested the need for this. I have added a new issue to track this work in a future release.

specification/dimensions/billingaccountid.md
```diff
@@ -0,0 +1,34 @@
+# Billing Account ID
+
+A billing account is a container for resources and/or services that are billed together in an invoice. Common scenarios for using billing accounts include grouping based on organizational constructs and hierarchy-based cost allocation approach.
+
+A billing account id is an provider assigned identifier for a billing account. It is commonly used for cost reporting, invoice reconciliation and cost allocation scenarios.
+
+The BillingAccountId column MUST be present in the billing data. This column must be of type String and MUST NOT contain null values. BillingAccountId SHOULD be a globally unique identifier within a provider. BillingAccountId MUST be unique when a customer has more than one billing account.
+
+See [Appendix: Grouping of resources and/or services](https://docs.google.com/document/u/0/d/1U24it_zByg-jKB9iOmiFbrULtzj5G5wUtawvntyeJuA/edit) section for billing accounts, sub-/resource accounts and hierarchy-based cost allocation approach scenarios.
```
flanakin, line 9: Don't link to Google Docs.

specification/dimensions/dimensions.mdpp
```diff
@@ -11,5 +11,6 @@ dimensions to categorize, filter, and reveal details in your data when grouped w
 !INCLUDE "resourceid.md",1
 !INCLUDE "resourcename.md",1
 !INCLUDE "servicename.md",1
+!INCLUDE "billingaccountid.md",1
```
flanakin, line 14: Sort alphabetically. 

FYI - You're going to hit a conflict with this change when the BillingAccountName file goes in.

amitw-g: Fixed the sorting, I will merge the above the code into that branch since I own both PRs.

specification/dimensions/billingaccountid.md
amitw-g: I am removing the content around Appendix since its not added yet. I have added a comment in https://github.com/FinOps-Open-Cost-and-Usage-Spec/FOCUS_Spec/issues/91 to add the link once its ready.

specification/dimensions/billingaccountid.md
```diff
@@ -0,0 +1,34 @@
+# Billing Account ID
+
+A billing account is a container for resources and/or services that are billed together in an invoice. Common scenarios for using billing accounts include grouping based on organizational constructs and hierarchy-based cost allocation approach.
+
+A billing account id is an provider assigned identifier for a billing account. It is commonly used for cost reporting, invoice reconciliation and cost allocation scenarios.
```
flanakin, line 5:
```suggestion
A billing account ID is an provider assigned identifier for a billing account. It is commonly used for cost reporting, invoice reconciliation, and cost allocation scenarios.
```

specification/dimensions/billingaccountid.md
```diff
@@ -0,0 +1,32 @@
+# Billing Account ID
+
+A billing account is a container for resources and/or services that are billed together in an invoice. Common scenarios for using billing accounts include grouping based on organizational constructs and hierarchy-based cost allocation approach.
+
+A billing account ID is an provider assigned identifier for a billing account. It is commonly used for cost reporting, invoice reconciliation and cost allocation scenarios.
```
udam-f2, line 5:
```suggestion
A billing account ID is a provider assigned identifier for a billing account. It is commonly used for invoice reconciliation and cost allocation scenarios.
```

specification/dimensions/billingaccountid.md
```diff
@@ -0,0 +1,32 @@
...
+
+The BillingAccountId column MUST be present in the billing data. This column must be of type String and MUST NOT contain null values. BillingAccountId SHOULD be a globally unique identifier within a provider. BillingAccountId MUST be unique when a customer has more than one billing account.
```
udam-f2, line 7:
```suggestion
The BillingAccountId column MUST be present in the billing data. This column must be of type String and MUST NOT contain null values. BillingAccountId MUST be a globally unique identifier within a provider.
```
Had some discussions. I will change my vote and go with some of the others that suggested simplifying this to say MUST be globally unique.

specification/dimensions/billingaccountid.md
```diff
@@ -0,0 +1,32 @@
+# Billing Account ID
+
+A billing account is a container for resources and/or services that are billed together in an invoice. Common scenarios for using billing accounts include grouping based on organizational constructs and hierarchy-based cost allocation approach.
+
+A billing account ID is a provider assigned identifier for a billing account. It is commonly used for invoice reconciliation and cost allocation scenarios.
```
ijurica, line 5: I suggest that we combine these two usage-related sentences ("_Common scenarios..._" and "_It is commonly used..._") into one. 
We could add the _invoice reconciliation_ reference to the first one and leave out the second?

Note: The same applies to the **billing account name spec** and corresponding **appendix section**.

udam-f2: Just had this discussion with Michael Flanakin as well. Technically, the first paragraph is for the billing account and the second is for billing account id. I think maybe its best to do what you're saying here and merge to the top paragraph and then people can select if they want to use the billing account id or name to do these things.

udam-f2 (approved):
Ready for merge after update to description.

specification/dimensions/billingaccountid.md
ijurica: As @rileyjenk mentioned, unlike billing account ID, billing account name (i.e., name dimensions in general) are subject to change and thus we would prefer using IDs for many of these UC scenarios (cost allocation in particular). 
We might consider including this as a constraint or at least in the supporting content?

specification/dimensions/billingaccountid.md
```diff
@@ -0,0 +1,32 @@
+# Billing Account ID
+
+A billing account is a container for resources and/or services that are billed together in an invoice. Common scenarios for using billing accounts include grouping based on organizational constructs and hierarchy-based cost allocation approach.
```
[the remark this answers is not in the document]
ijurica: OR implies one or the other?
In Udam's suggestion the reference to the invoice reconciliation separates the two and resolves the confusion?

specification/dimensions/billingaccountid.md
rileyjenk: As long as we say it may change I think it is ok and does improve immediate human readability. Over time is where we need to specify that it can be subject to change.

rileyjenk (approved):

rileyjenk (approved):

specification/dimensions/billingaccountid.md
amitw-g: Resolved by Udam's change

Issue #89, opened 2023-05-25
HTML and PDF improvement candidates


ijurica:
### Description
A few more changes to consider:
- adjusting the max-width in the HTML version for readability (I guess 75ch would be way too narrow, especially for the problematic table in the appendix)
- using bold weights for L1 and L2 headings as well in both PDF and HTML version
- page numbers in the PDF version (and in TOC as well)
- page breaks in the PDF version once the content is finalized

### Definition of done
Improved navigation and layout (PDF) and readability (HTML).

mike-finopsorg:
I have merged in a lot of the feedback here, Some items are much harder to address like page numbers in the TOC. Due to the way the TOC is generated prior to the PDF generation today. This may change in the future but the current assembly process gives us other benefits that are more valuable. As the spec reaches a slower pace of development this will likely change and we should reassess the build process.

PR #101, opened 2023-06-05
Formatting improvements - Vendored markdown-pp #89

mike-finopsorg:
Vendored in the markdown-pp package so we can modify it to
work as we need, versus the previous approach of trying to
modify the output to match our requirements. This has greatly
reduced the complexity of sed operations and made it possible to
use css more effectively. Table of contents is now a HTML UL which
can be styled. Made some of the recommendations in Issue #89.


udam-f2 (approved):
LGTM

Issue #75, opened 2023-05-21
Attribute: Currency code format


udam-f2:
Define the the way currency should be specified in FOCUS columns. Provider columns are encouraged to use the same for consistency.

mike-finopsorg:
Discussion in Slack the current thinking is to use
`Decimal / Extended Precision` as the field type and then add a section in Attributes about accuracy of numbers. Explaining the importance of the spec to have accuracy for FinOps.

Related:
https://en.wikipedia.org/wiki/Extended_precision

mike-finopsorg:
In addition when describing the currency that a charge is in aligning to the `code` type in  https://en.wikipedia.org/wiki/ISO_4217 is preferred.

udam-f2:
There may be two things here that need to be defined. One is the billed currency which is necessary for 0.5. The other part of this is 'priced currency' which may match up with the public pricing that's advertised by the provider (e.g. website pricing currency)

PR #119, opened 2023-06-20
FOCUS #75: Currency Code attribute and related content

udam-f2:
Currency Code attribute related changes.

Co-authored-by: Sanjna Srivatsa <ssrivatsa@vmware.com>
Co-authored-by: Eleni Rundle <elenirundle@google.com>

### DIFF
--- a/specification/attributes/currency_code_format.md
+++ b/specification/attributes/currency_code_format.md
@@ -1,0 +1,29 @@
+# Currency Code Format
+
+Columns that contain currency information in cost data following a consistent format reduces friction for FinOps practitioners that consume the data for analysis, reporting, and other use cases.
+
+All currency related columns defined in the FOCUS specification MUST follow the formatting requirements listed below. Provider generated currency-related columns SHOULD adopt the same format requirements over time.
+
+## Attribute ID
+
+CurrencyCodeFormat
+
+## Attribute name
+
+Currency Code Format
+
+## Description
+
+Formatting for currency columns appearing in billing data.
+
+## Requirements
+
+* Currency related columns MUST be represented as a three-letter alphabetic code as dictated in the governing document [ISO 4217:2015](https://www.iso.org/standard/64758.html).
+
+## Exceptions
+
+None
+
+## Introduced (version)
+
+0.5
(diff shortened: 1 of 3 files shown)


mikepolson-vmware (commented):
Should we mention that the three-digit alpha code values must be from ISO-4217?  https://en.wikipedia.org/wiki/ISO_4217

specification/attributes/currency_code_format.md
```diff
@@ -0,0 +1,29 @@
+# Currency Code Format
+
+Columns that contain currency information in cost data following a consistent format reduces friction for FinOps practitioners that consume the data for analysis, reporting, and other use cases.
+
+All currency related columns defined in the FOCUS specification MUST follow the formatting requirements listed below. Provider generated currency-related columns SHOULD adopt the same format requirements over time.
```
flanakin, line 5: nit: "FOCUS specification" is redundant. Should we use "schema" or just leave it out?
```suggestion
All currency related columns defined in the FOCUS schema MUST follow the formatting requirements listed below. Provider generated currency-related columns SHOULD adopt the same format requirements over time.
```

supporting_content/attributes/currency_code_format.md
```diff
@@ -0,0 +1,39 @@
...
+| GCP      | Big Query Billing Export | Currency       |
+| Azure    | Cost details             | BilledCurrency |
+
+## Example usage scenarios
+
+Current values observed in billing data for various scenarios:
+
+| Provider | Data set                 | Example Value |
+|:---------|:-------------------------|:--------------|
+| AWS      | CUR                      | USD           |
```
flanakin, line 19: uber-nit: I would suggest using a non-USD currency as the example.

flanakin (approved):

udam-f2:
> Should we mention that the three-digit alpha code values must be from ISO-4217? https://en.wikipedia.org/wiki/ISO_4217

Hey @mikepolson-vmware, the requirements section says: "* Currency related columns MUST be represented as a three-letter alphabetic code as dictated in the governing document [ISO 4217:2015](https://www.iso.org/standard/64758.html)." Does that seem sufficient? That happens to be the only requirement right now for currency code but in the future we may have more so I want to as a pattern just refer to the requirements rather than naming individual ones. Let me know.

mike-finopsorg (approved):

PR #166, opened 2023-08-27
FOCUS #149: CommitmentDiscountId spec changes and supporting content

### LINKED ISSUE #149: CommitmentId column
Define a column to identify the commitment-based discount that is applied to a charge. This column will include the unique identifier for the commitment-based discount.

See also: #150 

> ℹ️ _This column name is a placeholder and is not final._

flowerinthenight:
Closes #149


specification/dimensions/commitmentid.md
```diff
@@ -0,0 +1,32 @@
+# Commitment ID
+
+A commitment is a contractual commitment to use a minimum amount of spending on resources for a fixed term in exchange for discounts on those resources. Commitments can cover various resources such as virtual machines, or databases. A provider can offer various pricing models, payment options, and term durations for its commitment offerings.
```
flanakin, line 3: A few thoughts:
1. Thinking out loud: Should we say "A commitment-based discount is..."? There are other commitments that aren't "commitment-based discounts". Not sure if we want to generalize this to include those or not.
2. Does "minimum" make it sound like they're only valuable if you have 100% utilization?
3. We should work in a reference to differentiate usage-based and cost-based commitments. (Added something, but I feel like it could be better.)
```suggestion
A commitment-based discount is a contractual commitment to use a certain amount of a specific type of usage for a fixed term in exchange for a discounted unit price. Commitment-based discounts can cover either usage or cost of various resources such as virtual machines or databases. A provider can offer various pricing models, payment options, and term durations for commitment-based discounts.
```

specification/dimensions/commitmentid.md
```diff
@@ -0,0 +1,32 @@
...
+
+A Commitment ID is the identifier assigned to a commitment by the provider.
```
flanakin, line 5:
```suggestion
A Commitment ID is the identifier assigned to a commitment-based discount by the provider.
```

specification/dimensions/commitmentid.md
```diff
@@ -0,0 +1,32 @@
...
+
+The identifier assigned to a commitment by the provider.
+
+## Content constraints
+
+|    Constraint   |      Value      |
+|:----------------|:----------------|
+| Column required | True            |
+| Data type       | String          |
+| Allows nulls    | False            |
```
flanakin, line 27: Need to support nulls for charges that aren't related to CBDs.
```suggestion
| Allows nulls    | True             |
```

specification/dimensions/commitmentid.md
```diff
@@ -0,0 +1,32 @@
+# Commitment ID
+
+A commitment is a contractual commitment to use a minimum amount of spending on resources for a fixed term in exchange for discounts on those resources. Commitments can cover various resources such as virtual machines, or databases. A provider can offer various pricing models, payment options, and term durations for its commitment offerings.
+
+A Commitment ID is the identifier assigned to a commitment by the provider.
+
+The CommitmentId column MUST be present in the billing data. This column must be of type String and MUST NOT contain null values. CommitmentId MUST be a globally unique identifier within a provider.
```
flanakin, line 7: Saying "globally unique identifier" sounds like we require it to be a GUID. 
```suggestion
The CommitmentId column MUST be present in the billing data. This column must be of type String and MUST NOT contain null values when a charge is related to a commitment-based discount. When a charge is not associated with a commitment-based discount, the column MUST be null. CommitmentId MUST be unique within the provider.
```

specification/dimensions/commitmentid.md
```diff
@@ -0,0 +1,32 @@
...
+
+CommitmentId
+
+## Display name
+
+Commitment ID
+
+## Description
+
+The identifier assigned to a commitment by the provider.
```
flanakin, line 19:
```suggestion
The identifier assigned to a commitment-based discount by the provider.
```

supporting_content/dimensions/commitmentid.md
```diff
@@ -0,0 +1,44 @@
...
+|-----------|--------------------------|----------------------------------------------------------|
+| AWS       | CUR                      | reservation/ReservationARN<br>savingsPlan/SavingsPlanArn |
+| GCP       | Big Query Billing Export | credits.id                                               |
+| Microsoft | Cost details             | ReservationId (old)<br>BenefitId (new)                   |
+
+
+## Documentation
+
+- GCP: [Structure of Standard data export](https://cloud.google.com/billing/docs/how-to/export-data-bigquery-tables/standard-usage), [Committed Use Discounts](https://cloud.google.com/docs/cuds)
+- Azure: [Azure Savings Plans](https://learn.microsoft.com/en-us/azure/cost-management-billing/savings-plan/savings-plan-compute-overview), [Azure Reservations](https://learn.microsoft.com/en-us/azure/cost-management-billing/reservations/save-compute-costs-reservations)
```
flanakin, line 17:
```suggestion
- Microsoft: [Azure Savings Plans](https://learn.microsoft.com/azure/cost-management-billing/savings-plan/savings-plan-compute-overview), [Azure Reservations](https://learn.microsoft.com/azure/cost-management-billing/reservations/save-compute-costs-reservations)
```

supporting_content/dimensions/commitmentid.md
```diff
@@ -0,0 +1,44 @@
...
+- AWS: [Reserved Instances](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/ec2-reserved-instances.html), [Savings Plans](https://docs.aws.amazon.com/savingsplans/latest/userguide/what-is-savings-plans.html)
```
flanakin, line 18: nit: Move to be first in the list so it's alphabetized.

supporting_content/dimensions/commitmentid.md
```diff
@@ -0,0 +1,44 @@
...
+Current values observed in billing data for various scenarios:
+
+| Provider  | Data set                   | Scenario                                                         |
+|-----------|----------------------------|------------------------------------------------------------------|
+| AWS       | CUR                        | arn:partition:service:region:account-id:reserved-instances/\<id> |
+| AWS       | CUR                        | arn:partition:savingsplans:[region]:account-id:savingsplan/\<id> |
+| GCP       | Big Query Billing Export   | \<alphanumeric identifier>                                       |
+| GCP       | Big Query Billing Export   | \<credit description>, i.e. "Committed Use Discount: CPU"        |
+| Microsoft | Commitment purchase        | Commitment order GUID                                            |
+| Microsoft | Amortized usage            | Commitment instance GUID                                         |
```
flanakin, line 32:
```suggestion
| Microsoft | Cost details               | Commitment order GUID for purchase records                       |
| Microsoft | Cost details               | Commitment instance GUID for amortized usage                     |
```

specification/dimensions/commitmentid.md
```diff
@@ -0,0 +1,32 @@
+# Commitment ID
+
+A commitment-based discount is a contractual commitment to use a certain amount of a specific type of usage for a fixed term in exchange for a discounted unit price. Commitment-based discounts can cover either usage or cost of various resources such as virtual machines or databases. A provider can offer various pricing models, payment options, and term durations for commitment-based discounts.
+
+A Commitment ID is the identifier assigned to a commitment-based discount by the provider.
```
flanakin, line 5: Mention the scenarios that the column can be used to address, like commitment utilization and savings.

specification/dimensions/commitmentid.md
```diff
@@ -0,0 +1,32 @@
+# Commitment ID
```
flanakin, line 1: Can we change this to "Commitment Discount ID"? (Same for the name column.)

flanakin (changes requested):

caseydoran (approved):
LGTM!

specification/dimensions/commitmentid.md
flowerinthenight: You mean any references to "Commitment" should be changed to "Commitment Discount", right? (include filenames as well: `commitmentdiscountid.md`, etc.)

specification/dimensions/commitmentid.md
```diff
@@ -0,0 +1,32 @@
...
+
+A commitment-based discount is a contractual commitment to use a certain amount of a specific type of usage for a fixed term in exchange for a discounted unit price. Commitment-based discounts can cover either usage or cost of various resources such as virtual machines or databases. A provider can offer various pricing models, payment options, and term durations for commitment-based discounts.
```
rodneyjoyce, line 3: I really like this paragraph, but it is nothing specifically to do with the ID. ie. it applies to all Commitment fields - it just happens to be on the top of this one to give it context. Does this mean all Commitment Discount fields should have it? Or should it be abstracted into a base class?

specification/dimensions/commitmentid.md
```diff
@@ -0,0 +1,32 @@
...
+
+The CommitmentId column MUST be present in the billing data. This column must be of type String and MUST NOT contain null values when a charge is related to a commitment-based discount. When a charge is not associated with a commitment-based discount, the column MUST be null. CommitmentId MUST be unique within the provider.
+
+## Column ID
+
+CommitmentId
+
+## Display name
+
+Commitment ID
```
rodneyjoyce, line 15: Just double-checking that we're 100% sure this should not be Id?

specification/dimensions/dimensions.mdpp
```diff
@@ -13,6 +13,7 @@ dimensions to categorize, filter, and reveal details in your data when grouped w
 !INCLUDE "chargeperiodend.md",1
 !INCLUDE "chargeperiodstart.md",1
 !INCLUDE "chargetype.md",1
+!INCLUDE "commitmentid.md",1
```
rodneyjoyce, line 16: change file name to commitmentdiscountid if @flanakin 's suggestion goes ahead.

rodneyjoyce (commented):
Minor comments to be considered - nothing major

specification/dimensions/commitmentid.md
flowerinthenight: Renamed.

specification/dimensions/dimensions.mdpp
flowerinthenight: Updated with the latest commit.

specification/dimensions/commitmentid.md
flowerinthenight: There's [this](https://github.com/FinOps-Open-Cost-and-Usage-Spec/FOCUS_Spec/blob/working_draft/specification/glossary.md) file. It's empty though.

joe-ferrero (commented):
Looks good to me.  One question about the word "specific" in the first sentence.  Could specific apply to a group of similar SKUs, rather than a single SKU?  I've seen a "cross-service discount" applied as "15% across all of our consumption SKUs".  Would "defined" be a better choice than "specific"?

specification/dimensions/commitmentid.md
flowerinthenight: Pretty much patterned with BillingAccount[Id|Name], Resource[Id|Name], and SubAccount[Id|Name] columns.

jasonkellyAP (approved):

flanakin (approved):

specification/dimensions/commitmentdiscountid.md
```diff
@@ -0,0 +1,32 @@
+# Commitment Discount ID
+
+A commitment-based discount is a contractual commitment to use a certain amount of a specific type of usage for a fixed term in exchange for a discounted unit price. Commitment-based discounts can cover either usage or cost of various resources such as virtual machines or databases. A provider can offer various pricing models, payment options, and term durations for commitment-based discounts.
+
+A Commitment Discount ID is the identifier assigned to a commitment-based discount by the provider.
```
flanakin, line 5: Add scenarios:
```suggestion
A Commitment Discount ID is the identifier assigned to a commitment-based discount by the provider. Commitment Discount ID is commonly used for scenarios like calculating utilization and savings per commitment-based discount. 
```

specification/dimensions/commitmentid.md
flanakin: Made an explicit suggestion to address this feedback in the latest iteration. Closing.

specification/dimensions/commitmentid.md
flanakin: We'll mention this to Karl as part of the glossary effort! We'll clean this up as needed once we have the right guidance. Closing this for now.

BenShy (approved):

specification/dimensions/commitmentdiscountid.md
```diff
@@ -0,0 +1,32 @@
+# Commitment Discount ID
+
+A commitment-based discount is a contractual commitment to use a certain amount of a specific type of usage for a fixed term in exchange for a discounted unit price. Commitment-based discounts can cover either usage or cost of various resources such as virtual machines or databases. A provider can offer various pricing models, payment options, and term durations for commitment-based discounts.
```
udam-f2, line 3: @kk09v @flanakin We need a single definition here (that hopefully also aligns with the FinOps Foundation definition - or that we suggest back to the foundation). CommitmentType PR has a different blurb describing this same thing.

Karl, could we prioritize Commitment-discounts in the glossary first since it's actively being worked on and referenced in many PRs.

specification/dimensions/commitmentdiscountid.md
```diff
@@ -0,0 +1,32 @@
...
+
+A Commitment Discount ID is the identifier assigned to a commitment-based discount by the provider. Commitment Discount ID is commonly used for scenarios like calculating utilization and savings per commitment-based discount.
```
udam-f2, line 5:
```suggestion
A Commitment Discount ID is the identifier assigned to a commitment-based discount by the provider. Commitment Discount ID is commonly used for scenarios like calculating utilization and savings due to commitment-based discounts.
```

specification/dimensions/commitmentdiscountid.md
udam-f2: Actually, to reduce inter-dependencies, @flanakin could we settle on one definition and go with it and Karl can do a sweep and move things into the glossary as he's doing the other ones as well.

aqu-erp (approved):

specification/dimensions/commitmentdiscountid.md
flanakin, line 5: Remove trailing space.
```suggestion
A Commitment Discount ID is the identifier assigned to a commitment-based discount by the provider. Commitment Discount ID is commonly used for scenarios like calculating utilization and savings per commitment-based discount.
```

udam-f2 (approved):

rileyjenk:
I fear that while this handles the case of providing an identifier for commitment based discounts it forces our hand in providing an additional identifier column for commitments that provide no discount yet adhere to nearly everything else. I would consider removing discount from the name. Specific examples of this are on demand capacity reservations. They behave very similarly as reservations and savings plans. 


See AWS: https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/capacity-reservations-pricing-billing.html
See Azure: https://learn.microsoft.com/en-us/azure/virtual-machines/capacity-reservation-overview

rileyjenk (changes requested):
I fear that while this handles the case of providing an identifier for commitment based discounts it forces our hand in providing an additional identifier column for commitments that provide no discount yet adhere to nearly everything else. I would consider removing discount from the name. Specific examples of this are on demand capacity reservations. They behave very similarly as reservations and savings plans.

See AWS: https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/capacity-reservations-pricing-billing.html
See Azure: https://learn.microsoft.com/en-us/azure/virtual-machines/capacity-reservation-overview

udam-f2:
> I fear that while this handles the case of providing an identifier for commitment based discounts it forces our hand in providing an additional identifier column for commitments that provide no discount yet adhere to nearly everything else. I would consider removing discount from the name. Specific examples of this are on demand capacity reservations. They behave very similarly as reservations and savings plans.
> 
> See AWS: https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/capacity-reservations-pricing-billing.html See Azure: https://learn.microsoft.com/en-us/azure/virtual-machines/capacity-reservation-overview

Do ODCRs charge for the reserved amount and negate the actual usage that applies in that period? or just return rows for actual usage and then a single entry for the unused portion? If its the latter, feels like we don't really need the concept of a commitment for those. The unused portion would just have the ODCR id as the resource id? @rileyjenk

aqu-erp:
> I fear that while this handles the case of providing an identifier for commitment based discounts it forces our hand in providing an additional identifier column for commitments that provide no discount yet adhere to nearly everything else. I would consider removing discount from the name. Specific examples of this are on demand capacity reservations. They behave very similarly as reservations and savings plans.
> 
> See AWS: https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/capacity-reservations-pricing-billing.html See Azure: https://learn.microsoft.com/en-us/azure/virtual-machines/capacity-reservation-overview

I'm not sure I agree that AWS on demand capacity reservations (without discounts) meet the threshold of similar attributes for a commitment given:
1. It does not have a term. 
2. In the AWS billing data, the capacity reservation costs are driven by the instance, rather than the reservation itself.
    - The unused portion of such a reservation is more akin to a separate but identically priced type of usage.

jackyljgoog (approved):

flanakin:
Given capacity reservations overlap with commitment discounts, we can't use the same column. Let me add another column to the backlog for tracking capacity reservations.

rileyjenk (approved):

rileyjenk:
@flanakin posted in the other PR but additional columns identifying on demand capacity reservations is merited. While very similar they differ enough and can occur simultaneously for the same resource at the same time = two pieces of data needed to know details for both.

rileyjenk (approved):

specification/dimensions/commitmentdiscountid.md
```diff
@@ -0,0 +1,32 @@
...
```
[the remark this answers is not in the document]
flowerinthenight: Updated.

specification/dimensions/commitmentdiscountid.md
udam-f2, line 3:
```suggestion
A commitment-based discount is a contractual commitment for an amount of usage or spend, for a specified term in exchange for a discounted unit price. Commitment-based discounts can cover either usage or cost of various resources such as virtual machines or databases. A provider can offer various pricing models, payment options, and term durations for commitment-based discounts.
```

specification/dimensions/commitmentdiscountid.md
```diff
@@ -0,0 +1,32 @@
+# Commitment Discount ID
+
+A commitment-based discount is a contractual commitment for an amount of usage or spend, for a specified term, in exchange for a discounted unit price. Commitment-based discounts can cover either usage or cost of various resources such as virtual machines or databases. A provider can offer various pricing models, payment options, and term durations for commitment-based discounts.
+
+A Commitment Discount ID is the identifier assigned to a commitment-based discount by the provider. Commitment Discount ID is commonly used for scenarios like chargeback for commitments and savings per commitment-based discount.
+
+The CommitmentDiscountId column MUST be present in the billing data. This column must be of type String and MUST NOT contain null values when a charge is related to a commitment-based discount. When a charge is not associated with a commitment-based discount, the column MUST be null. CommitmentDiscountId MUST be unique within the provider.
```
marc-perreaut, line 7: > when a charge is related to a commitment-based discount

Maybe: 'when a record is related to a commit-based discount'.

Indeed, in case of a resource covered by commitment-based discount, the charge (BilledCost) would be 0.
Maybe it could be explictly mentioned: the CommitDiscountId value must not be null if the record relates to the purchase of a commitment-based discount or if the record relates to the usage of a resource covered by a commitment-based discount.

marc-perreaut, line 7: Maybe it can be mentioned that CommitmentDiscountID works in pair with CommitmentDiscountName.

aqu-erp: While that is logically true for human-reading, an ID still retains its value as a dimension independent of the name. Would also look into how other ID/name dimensions are presented with this change, and I think lack of reference to each other makes sense.

specification/dimensions/commitmentdiscountid.md
```diff
@@ -0,0 +1,32 @@
+# Commitment Discount ID
+
+A commitment-based discount is a contractual commitment for an amount of usage or spend, for a specified term, in exchange for a discounted unit price. Commitment-based discounts can cover either usage or cost of various resources such as virtual machines or databases. A provider can offer various pricing models, payment options, and term durations for commitment-based discounts.
```
aqu-erp, line 3:
```suggestion
A commitment-based discount is a contractual commitment for a minimum amount of usage or spend throughout a specified term, in exchange for discounted unit pricing on that amount. The commitment may be based on quantities of resource units or monetary value, with various payment options and timeframes.
```

- Additional clarity on discount (can be many unit prices, and only for the amount committed)
- Removed reference to specific resources in case providers do not or stop offering commitments on those resources.

aqu-erp (changes requested):

specification/dimensions/commitmentdiscountid.md
udam-f2: I would prefer if we didn't have to say 'minimum amount' to make it more general- might be true today cloud commitments but if we think about the broader context, I could see someone saying you had to average over a period of time (e.g. for people, average 40 hours/week of a consultants time over some period)

specification/dimensions/commitmentdiscountid.md
udam-f2: I'm fairly sure we've referred to the record / line-item / row as a 'charge' in other parts of the doc. Does a charge have to be a non-zero number? We need to update other places if we want to start referring to the line items as a record.

Examples are:
- ChargePeriod / ChargeType applies to all rows
- "Billing Currency is an identifier that represents the currency that a charge for resources and/or services was billed in."
- "If a provider does not support a sub account construct (only has a billing account) or does support a sub account construct, but the charge does not apply to a sub account, the SubAccountId column MUST be null."

specification/dimensions/commitmentdiscountid.md
ljadvey: Rather than minimum, it’s spend committed for a time period (not minimum spend).  I liked the original language before minimum was added.

specification/dimensions/commitmentdiscountid.md
ljadvey: I don’t think a charge needs to be non-zero for this. I would still want to see that charge.

specification/dimensions/commitmentdiscountid.md
flowerinthenight: Something like:

> A commitment-based discount is a contractual commitment for an amount of usage or spend committed for a time period, in exchange for discounted unit pricing on that amount. The commitment may be based on quantities of resource units or monetary value, with various payment options and timeframes.

?

specification/dimensions/commitmentdiscountid.md
flowerinthenight: This is in keeping with our existing id/name pairs such as BillingAccountId, ResourceId, and SubAccountId. We might need to update those items as well.

specification/dimensions/commitmentdiscountid.md
aqu-erp: Updated the suggestion to remove the "minimum", to mitigate marginal changes to the spec if providers add additional terms to their definition. 

There is an important distinction to emphasize when removing "minimum", that the `term` of the entire commitment is not necessarily the `term` of the committed usage or spend amount. Hopefully "throughout" makes that clear.

In @udam-f2 's scenario, if applied to compute, then there is no distinction for commit term and CBD term : average of $40/hr Compute over a 1-year period (trued-up at year end). Rather than the current commit for an absolute $40/hr Compute over a 1-year period (enforceable per hour).

flowerinthenight: By the way, should it be `time frame` instead of `timeframe`?

specification/dimensions/commitmentdiscountid.md
marc-perreaut: Thanks for the clarification.
Indeed, I had not noticed that before a 'charge' was a synonym for a record / line-item / row.

marc-perreaut: Yes indeed, nothing is mentioned for the id fields in FOCUS v0.5.
So this could be done later overall and in a consistent way.

timdcook123 (approved):

specification/dimensions/commitmentdiscountid.md
udam-f2: Yes, this was discussed earlier on - and we are going to move metrics and dimensions under a single 'columns' section and then add a section where we link 'related' columns and attributes. This will be done at the end - as a final pass. Good call-out though.

aqu-erp (approved):

PR #167, opened 2023-08-27
FOCUS #150: CommitmentDiscountName spec changes and supporting content

### LINKED ISSUE #150: CommitmentName column
Define a column to identify the commitment-based discount that is applied to a charge. This column will include the display name for the commitment-based discount.

See also: #149 

> ℹ️ _This column name is a placeholder and is not final._

flowerinthenight:
Closes #150

### DIFF
--- a/specification/dimensions/commitmentdiscountname.md
+++ b/specification/dimensions/commitmentdiscountname.md
@@ -4,7 +4,7 @@ A commitment-based discount is a contractual commitment to use a certain amount 
 
 A Commitment Discount Name is the display name assigned to a commitment-based discount.
 
-The CommitmentDiscountName column MUST be present in the billing data. This column must be of type String. The CommitmentDiscountName value MAY be null if a display name cannot be assigned to a commitment-based discount. CommitmentDiscountName MUST NOT be null if a display name can be assigned to a commitment-based discount.
+The CommitmentDiscountName column MUST be present in the billing data. This column must be of type String. The CommitmentDiscountName value MUST be null if the charge is not related to a commitment-based discount and MAY be null if a display name cannot be assigned to a commitment-based discount. CommitmentDiscountName MUST NOT be null if a display name can be assigned to a commitment-based discount.
 
 ## Column ID
 
--- a/specification/dimensions/commitmentdiscountname.md
+++ b/specification/dimensions/commitmentdiscountname.md
@@ -1,6 +1,6 @@
 # Commitment Discount Name
 
-A commitment-based discount is a contractual commitment for an amount of usage or spend, for a specified term, in exchange for a discounted unit price. Commitment-based discounts can cover either usage or cost of various resources such as virtual machines or databases. A provider can offer various pricing models, payment options, and term durations for commitment-based discounts.
+A commitment-based discount is a contractual commitment for an amount of usage or spend throughout a specified term, in exchange for discounted unit pricing on that amount. The commitment may be based on quantities of resource units or monetary value, with various payment options and time frames.
 
 A Commitment Discount Name is the display name assigned to a commitment-based discount.
 
--- a/specification/dimensions/commitmentdiscountname.md
+++ b/specification/dimensions/commitmentdiscountname.md
@@ -1,18 +1,18 @@
-# Commitment Name
+# Commitment Discount Name
 
 A commitment-based discount is a contractual commitment to use a certain amount of a specific type of usage for a fixed term in exchange for a discounted unit price. Commitment-based discounts can cover either usage or cost of various resources such as virtual machines or databases. A provider can offer various pricing models, payment options, and term durations for commitment-based discounts.
 
-A Commitment Name is the display name assigned to a commitment-based discount.
+A Commitment Discount Name is the display name assigned to a commitment-based discount.
 
-The CommitmentName column MUST be present in the billing data. This column must be of type String. The CommitmentName value MAY be null if a display name cannot be assigned to a commitment-based discount. CommitmentName MUST NOT be null if a display name can be assigned to a commitment-based discount.
+The CommitmentDiscountName column MUST be present in the billing data. This column must be of type String. The CommitmentDiscountName value MAY be null if a display name cannot be assigned to a commitment-based discount. CommitmentDiscountName MUST NOT be null if a display name can be assigned to a commitment-based discount.
 
 ## Column ID
 
-CommitmentName
+CommitmentDiscountName
 
 ## Display name
 
-Commitment Name
+Commitment Discount Name
 
 ## Description
 
--- a/specification/dimensions/commitmentdiscountname.md
+++ b/specification/dimensions/commitmentdiscountname.md
@@ -1,6 +1,6 @@
 # Commitment Discount Name
 
-A commitment-based discount is a contractual commitment to use a certain amount of a specific type of usage for a fixed term in exchange for a discounted unit price. Commitment-based discounts can cover either usage or cost of various resources such as virtual machines or databases. A provider can offer various pricing models, payment options, and term durations for commitment-based discounts.
+A commitment-based discount is a contractual commitment for an amount of usage or spend, for a specified term, in exchange for a discounted unit price. Commitment-based discounts can cover either usage or cost of various resources such as virtual machines or databases. A provider can offer various pricing models, payment options, and term durations for commitment-based discounts.
 
 A Commitment Discount Name is the display name assigned to a commitment-based discount.
 
--- a/specification/dimensions/commitmentname.md
+++ b/specification/dimensions/commitmentname.md
@@ -2,9 +2,9 @@
 
 A commitment is a contractual commitment to use a minimum amount of spending on resources for a fixed term in exchange for discounts on those resources. Commitments can cover various resources such as virtual machines, or databases. A provider can offer various pricing models, payment options, and term durations for its commitment offerings.
 
-A Commitment ID is the identifier assigned to a commitment by the provider.
+A Commitment Name is the display name assigned to a commitment by the provider.
 
-The CommitmentId column MUST be present in the billing data. This column must be of type String and MUST NOT contain null values. CommitmentId MUST be a globally unique identifier within a provider.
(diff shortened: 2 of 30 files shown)


specification/dimensions/commitmentname.md
```diff
@@ -0,0 +1,32 @@
...
+
+CommitmentName
+
+## Display name
+
+Commitment Name
+
+## Description
+
+The display name assigned to a commitment by the provider.
```
flanakin, line 19:
```suggestion
The display name assigned to a commitment-based discount by the provider.
```

supporting_content/dimensions/commitmentname.md
```diff
@@ -0,0 +1,42 @@
...
+|-----------|--------------------------|--------------------------------------------|
+| AWS       | CUR                      | Not present as a separate column in data.  |
+| GCP       | Big Query Billing Export | credits.full_name                          |
+| Microsoft | Cost details             | ReservationName (old)<br>BenefitName (new) |
+
+
+## Documentation
+
+- GCP: [Structure of Standard data export](https://cloud.google.com/billing/docs/how-to/export-data-bigquery-tables/standard-usage), [Committed Use Discounts](https://cloud.google.com/docs/cuds)
+- Azure: [Azure Savings Plans](https://learn.microsoft.com/en-us/azure/cost-management-billing/savings-plan/savings-plan-compute-overview), [Azure Reservations](https://learn.microsoft.com/en-us/azure/cost-management-billing/reservations/save-compute-costs-reservations)
```
flanakin, line 17:
```suggestion
- Microsoft: [Azure Savings Plans](https://learn.microsoft.com/azure/cost-management-billing/savings-plan/savings-plan-compute-overview), [Azure Reservations](https://learn.microsoft.com/azure/cost-management-billing/reservations/save-compute-costs-reservations)
```

supporting_content/dimensions/commitmentname.md
```diff
@@ -0,0 +1,42 @@
...
+
+Current values observed in billing data for various scenarios:
+
+| Provider  | Data set                   | Scenario                                       |
+|-----------|----------------------------|------------------------------------------------|
+| AWS       | CUR                        | Not present as a separate column in data.      |
+| GCP       | Big Query Billing Export   | Examples: "Spend-based committed use discount" |
+| GCP       | Big Query Billing Export   | Empty if credits.id is a description type id.  |
+| Microsoft | Commitment purchase        | Commitment order name (string)                 |
+| Microsoft | Amortized usage            | Commitment instance name (string)              |
```
flanakin, line 31:
```suggestion
| Microsoft | Cost details               | Commitment order name for purchases            |
| Microsoft | Cost details               | Commitment instance name for amortized usage   |
```

specification/dimensions/commitmentname.md
```diff
@@ -0,0 +1,32 @@
+# Commitment Name
+
+A commitment is a contractual commitment to use a minimum amount of spending on resources for a fixed term in exchange for discounts on those resources. Commitments can cover various resources such as virtual machines, or databases. A provider can offer various pricing models, payment options, and term durations for its commitment offerings.
+
+A Commitment Name is the display name assigned to a commitment by the provider.
```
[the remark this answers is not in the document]
flowerinthenight: Agree we should just leave it out.

specification/dimensions/commitmentname.md
```diff
@@ -0,0 +1,32 @@
+# Commitment Name
+
+A commitment-based discount is a contractual commitment to use a certain amount of a specific type of usage for a fixed term in exchange for a discounted unit price. Commitment-based discounts can cover either usage or cost of various resources such as virtual machines or databases. A provider can offer various pricing models, payment options, and term durations for commitment-based discounts.
```
rodneyjoyce, line 3: Hmm tying in with my comment from CommitmentId - should we be duplicating this on every Cxxx column? Glossary?

specification/dimensions/dimensions.mdpp
```diff
@@ -13,6 +13,7 @@ dimensions to categorize, filter, and reveal details in your data when grouped w
 !INCLUDE "chargeperiodend.md",1
 !INCLUDE "chargeperiodstart.md",1
 !INCLUDE "chargetype.md",1
+!INCLUDE "commitmentname.md",1
```
rodneyjoyce, line 16: commitmentdiscountname (if rename goes ahead)

specification/dimensions/commitmentname.md
```diff
@@ -0,0 +1,32 @@
...
+
+A Commitment Name is the display name assigned to a commitment-based discount.
+
+The CommitmentName column MUST be present in the billing data. This column must be of type String. The CommitmentName value MAY be null if a display name cannot be assigned to a commitment-based discount. CommitmentName MUST NOT be null if a display name can be assigned to a commitment-based discount.
```
rodneyjoyce, line 7: Does it add any value to say that it does not need to be unique? ie. what do FinOps Practioners use this field for?

rodneyjoyce (commented):
Minor comments to be considered

specification/dimensions/dimensions.mdpp
flowerinthenight: Renamed.

BenShy (approved):

specification/dimensions/commitmentname.md
flanakin: For now, yes.

specification/dimensions/commitmentdiscountname.md
```diff
@@ -0,0 +1,32 @@
+# Commitment Discount Name
+
+A commitment-based discount is a contractual commitment to use a certain amount of a specific type of usage for a fixed term in exchange for a discounted unit price. Commitment-based discounts can cover either usage or cost of various resources such as virtual machines or databases. A provider can offer various pricing models, payment options, and term durations for commitment-based discounts.
+
+A Commitment Discount Name is the display name assigned to a commitment-based discount.
+
+The CommitmentDiscountName column MUST be present in the billing data. This column must be of type String. The CommitmentDiscountName value MAY be null if a display name cannot be assigned to a commitment-based discount. CommitmentDiscountName MUST NOT be null if a display name can be assigned to a commitment-based discount.
```
flanakin, line 7:
```suggestion
The CommitmentDiscountName column MUST be present in the billing data. This column must be of type String. The CommitmentDiscountName value MUST be null if the charge is not related to a commitment-based discount and MAY be null if a display name cannot be assigned to a commitment-based discount. CommitmentDiscountName MUST NOT be null if a display name can be assigned to a commitment-based discount.
```

flanakin (approved):

udam-f2 (approved):

rileyjenk (dismissed):
I fear that while this handles the case of providing an identifier for commitment based discounts it forces our hand in providing an additional identifier column for commitments that provide no discount yet adhere to nearly everything else. I would consider removing discount from the name. Specific examples of this are on demand capacity reservations. They behave very similarly as reservations and savings plans.

See AWS: https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/capacity-reservations-pricing-billing.html
See Azure: https://learn.microsoft.com/en-us/azure/virtual-machines/capacity-reservation-overview

specification/dimensions/commitmentdiscountname.md
```diff
@@ -0,0 +1,32 @@
+# Commitment Discount Name
+
+A commitment-based discount is a contractual commitment for an amount of usage or spend, for a specified term, in exchange for a discounted unit price. Commitment-based discounts can cover either usage or cost of various resources such as virtual machines or databases. A provider can offer various pricing models, payment options, and term durations for commitment-based discounts.
+
+A Commitment Discount Name is the display name assigned to a commitment-based discount.
+
+The CommitmentDiscountName column MUST be present in the billing data. This column must be of type String. The CommitmentDiscountName value MUST be null if the charge is not related to a commitment-based discount and MAY be null if a display name cannot be assigned to a commitment-based discount. CommitmentDiscountName MUST NOT be null if a display name can be assigned to a commitment-based discount.
```
marc-perreaut, line 7: Same comments as for CommitmentDiscountId:

> when a charge is related to a commitment-based discount
Maybe: 'when a record is related to a commit-based discount'.

Indeed, in case of a resource covered by commitment-based discount, the charge (BilledCost) would be 0.
Maybe it could be explictly mentioned: the CommitDiscountId value must not be null if the record relates to the purchase of a commitment-based discount or if the record relates to the usage of a resource covered by a commitment-based discount.

ljadvey (approved):

specification/dimensions/commitmentdiscountname.md
marc-perreaut: As per Udam's comment in CommitDiscountId PR, the word 'charge' in FOCUS refers to record / line item / row.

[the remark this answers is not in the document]
marc-perreaut: Indeed, there is no cross-reference in ID/name dimensions in FOCUS v0.5. So, cross-referencing could be done later globally and in a consistent way.

Furthermore and interestingly, FOCUS v0.5 states in the description of Resource Name:

> Resources not provisioned interactively or only have a system generated ResourceId MUST NOT duplicate the same value as the ResourceName.

Do we want to adopt the same principle for CommitmentDiscountName?

timdcook123 (approved):

specification/dimensions/commitmentdiscountname.md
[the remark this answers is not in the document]
udam-f2: Good catch @marc-perreaut. The only thing I'm struggling with is with resources, we have the case where they are provisioned automatically by systems and are transient in nature. I guess commitment discounts aren't quite like that though obviously could be created by systems still. I'm torn. We do have the statement about MUST NOT be null if a display name can be assigned. Should it just say MUST be null if a name cannot be or was not assigned? Any others have thoughts on this? 

Seems like a minor thing.. would be happy if we can get this checked in soon :)

udam-f2 (approved):

specification/dimensions/commitmentdiscountname.md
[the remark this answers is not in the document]
flanakin: They should be consistent, but moreso than that, we should pull that logic into an attribute so it's consistent across all ID/Name columns. We have a separate task to clean that up, so I'm personally not overly concerned. Happy to go whichever way here.

PR #171, opened 2023-09-06
FOCUS #172: Resource Type dimension

cnharris10:
Closes https://github.com/FinOps-Open-Cost-and-Usage-Spec/FOCUS_Spec/issues/172

# Preview

image

### DIFF
--- a/specification/dimensions/resourcetype.md
+++ b/specification/dimensions/resourcetype.md
@@ -2,7 +2,7 @@
 
 Resource Type describes the kind of resource for which you are being charged.  A Resource Type is commonly used for scenarios like identifying cost changes in groups of similar resources and may include values like Virtual Machine, Data Warehouse, and Load Balancer.
 
-The ResourceType column MUST be present within billing data.  This column MUST be of type String and MUST NOT be null when a corresponding ResourceId is not null.  When a corresponding ResourceId value is null, the ResourceType column value MUST also be null.  Providers MUST use a consistent value-format and a set of values for ResourceType values within the dataset.
+The ResourceType column MUST be present within billing data.  This column MUST be of type String and MUST NOT be null when a corresponding ResourceId is not null.  When a corresponding ResourceId value is null, the ResourceType column value MUST also be null.  Providers MUST use a consistent value-format and a set of values for ResourceType values within their cost and usage datasets.
 
 ## Column ID
 
--- a/specification/dimensions/resourcetype.md
+++ b/specification/dimensions/resourcetype.md
@@ -2,7 +2,9 @@
 
 Resource Type describes the kind of resource for which you are being charged.  A Resource Type is commonly used for scenarios like identifying cost changes in groups of similar resources.
 
-The ResourceType column MUST be present within billing data.  ResourceType MUST be of type String and MUST NOT be NULL when a corresponding ResourceId is not NULL.  When a corresponding ResourceId value is NULL, the ResourceType column value MUST be NULL. Providers SHOULD choose a consistent value-format for all values.
+The ResourceType column MUST be present within billing data.  ResourceType MUST be of type String and MUST NOT be NULL when a corresponding ResourceId is not NULL.  When a corresponding ResourceId value is NULL, the ResourceType column value MUST be NULL. Providers SHOULD choose a consistent values and a value-format that accounts for additional values added.
+
+Some potential values MAY include Virtual Machine, Data Warehouse, and Load Balancer.
 
 ## Column ID
 
--- a/specification/dimensions/resourcetype.md
+++ b/specification/dimensions/resourcetype.md
@@ -1,6 +1,6 @@
 # Resource Type
 
-Resource Type describes the kind of resource for which you are being charged.  A Resource Type is commonly used for scenarios like identifying cost changes in groups of similar resources and MAY include values like Virtual Machine, Data Warehouse, and Load Balancer.
+Resource Type describes the kind of resource for which you are being charged.  A Resource Type is commonly used for scenarios like identifying cost changes in groups of similar resources and may include values like Virtual Machine, Data Warehouse, and Load Balancer.
 
 The ResourceType column MUST be present within billing data.  This column MUST be of type String and MUST NOT be null when a corresponding ResourceId is not null.  When a corresponding ResourceId value is null, the ResourceType column value MUST also be null.  Providers MUST use a consistent value-format and a set of values for ResourceType values within the dataset.
 
--- a/specification/dimensions/resourcetype.md
+++ b/specification/dimensions/resourcetype.md
@@ -2,9 +2,7 @@
 
 Resource Type describes the kind of resource for which you are being charged.  A Resource Type is commonly used for scenarios like identifying cost changes in groups of similar resources.
 
-The ResourceType column MUST be present within billing data.  ResourceType MUST be of type String and MUST NOT be NULL when 
-a corresponding ResourceId is not NULL.  When a corresponding ResourceId value is NULL, the ResourceType column value MUST be NULL. Providers SHOULD choose a consistent value-format for all values.
-
+The ResourceType column MUST be present within billing data.  ResourceType MUST be of type String and MUST NOT be NULL when a corresponding ResourceId is not NULL.  When a corresponding ResourceId value is NULL, the ResourceType column value MUST be NULL. Providers SHOULD choose a consistent value-format for all values.
 
 ## Column ID
 
--- a/specification/dimensions/resourcetype.md
+++ b/specification/dimensions/resourcetype.md
@@ -1,6 +1,6 @@
 # Resource Type
 
-A Resource Type is a classifiable representation of a specific kind of cloud service or corresponding resource that can be provisioned, configured, and managed by users. A Resource Type is commonly used for scenarios like identifying changes in cost over time or building a resource inventory.  While Resource ID and Resource Name identifies a unique instance of a resource, the Resource Type describes the identity of the resource.
+Resource Type describes the kind of resource for which you are being charged.  A Resource Type is commonly used for scenarios like identifying cost changes in groups of similar resources.
 
 The ResourceType column MUST be present within billing data.  ResourceType MUST be of type String and MUST NOT be NULL when a corresponding ResourceId is not NULL.  When a corresponding ResourceId value is NULL, the ResourceType column value MUST be NULL.
 
@@ -14,7 +14,7 @@ Resource Type
 
 ## Description
 
-A classifiable representation of a specific kind of cloud service or corresponding resource that can be provisioned, configured, and managed by users.
+The kind of resource for which you are being charged.
 
 ## Normalized?
 
--- a/specification/dimensions/resourcetype.md
+++ b/specification/dimensions/resourcetype.md
@@ -15,11 +15,6 @@ Resource Type
 ## Description
 
 The kind of resource for which you are being charged.
-
-## Data type
-
-String
-
 ## Content Constraints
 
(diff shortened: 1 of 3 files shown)


specification/dimensions/resourcetype.md
```diff
@@ -0,0 +1,53 @@
+# Resource Type
+
+A Resource ID is an identifier assigned to a resource by the provider. The Resource ID is commonly used for cost reporting, analysis, and allocation scenarios.  While Resource ID and Resource Name identifies a unique instance of a resource, the Resource Type describes the identity of the resource. Resource Type is commonly used for cost reporting, analysis, and allocation scenarios.
```
flanakin, line 3: nit: Do we need to speak to resource ID and name? We can probably jump right into Resource Type, right?

The resource type isn't really about the identity of the resource. It's more about the kind of resource it is. I added what Bing Chat says below. Feel free to take it or leave it :-) 

For scenarios, we should probably be more specific since everything can be used for reporting and analysis.
```suggestion
A Resource Type is a logical representation of a specific kind of cloud service or resource that can be provisioned, configured, and managed by users. Resource Type is commonly used for scenarios like identifying changes in cost over time or building a resource inventory.
```

specification/dimensions/resourcetype.md
```diff
@@ -0,0 +1,53 @@
...
+
+ResourceType
+
+## Display Name
+
+Resource Type
+
+## Description
+
+Classifiable description of a corresponding ResourceId
```
flanakin, line 17: We usually make this the same as the intro paragraph (minus scenarios). I like this description tho. Would be good to work this into the one above. Maybe use "Resource ID" in this context.

specification/dimensions/resourcetype.md
```diff
@@ -0,0 +1,53 @@
...
+| Column required | True            |
+| Data type       | String          |
+| Allows nulls    | True            |
+| Value format    | \<not specified> |
+
+## Introduced (version)
+
+1.0
+
+## Supporting content
```
flanakin, line 40: Supporting content should be in a separate file, right?

specification/dimensions/resourcetype.md
```diff
@@ -0,0 +1,53 @@
...
+
+The following is a list of potential Resource Type values that could exist across AWS, Azure, and GCP.
```
flanakin, line 42: Might also be good to include some notes about our discussions around having a normalized resource type column.

specification/dimensions/resourcetype.md
cnharris10: I like the rewrite but I also like demonstrating the juxtaposition of the Resource Type vs. the Resource ID/Resource Name.

How about?

```
A Resource Type is a logical representation of a specific kind of cloud service or resource that can be provisioned, configured, and managed by users. Resource Type is commonly used for scenarios like identifying changes in cost over time or building a resource inventory.  While Resource ID and Resource Name identifies a unique instance of a resource, the Resource Type describes the identity of the resource.
```

specification/dimensions/resourcetype.md
cnharris10: I'm not sure, but if so, would this be the appendix?

specification/dimensions/resourcetype.md
cnharris10: To foreshadow that FOCUS might go this direction in the future?

specification/dimensions/resourcetype.md
cnharris10: Edit above and duplicated here, thoughts?
```
A classifiable representation of a specific kind of cloud service or corresponding resource that can be provisioned, configured, and managed by users.
```

specification/dimensions/resourcetype.md
flanakin: WOrks for me. My only comment would be the "type" isn't really an "identity" of the resource.

flanakin (approved):

ijurica (approved):

rileyjenk (approved):

rileyjenk:
@ijurica  I added a tab for this on the standardized sample: 
https://docs.google.com/spreadsheets/d/14dMUdo4-QAbrj9AxFHmvfknpIFRmS5k30kokHMokTBg/edit#gid=2089451315
See tab called #172 ResourceType

specification/dimensions/resourcetype.md
```diff
@@ -0,0 +1,38 @@
+# Resource Type
+
+Resource Type describes the kind of resource for which you are being charged.  A Resource Type is commonly used for scenarios like identifying cost changes in groups of similar resources.
```
flanakin, line 3: @cnharris10 Is the intent of this to be a human-readable display name (e.g., "Virtual machines") or to be the identifier the provider uses to indicate the resource's type (e.g., "Microsoft.Compute/virtualMachines")? If we don't care, this is fine as is. If you want to force it to be human-readable, we should call that out explicitly. We should also determine if we want to allow for the ID (even if after 1.0), given that's what providers (at least Microsoft) make available today.

supporting_content/dimensions/resourcetype.md
```diff
@@ -0,0 +1,34 @@
...
+|:------------------------- |:------------------------- |:------------------------------------- |:--------------------------------- |
+| Virtual Machine           | EC2 Virtual Machine       | Azure Virtual Machine                 | Compute Engine Virtual Machine    |
+| Disk                      | EBS Volume                | Managed Disk                          | Persistent Disk                   |
+| Serverless Function       | Lambda Function           | Azure Function                        | Cloud Functions Function          |
+| Blob Storage Bucket       | S3 Bucket                 | Blob Storage Container                | GCS Bucket                        |
+| File System               | EFS File System           | File Storage                          | File Store                        |
+| Load Balancer             | Elastic Load Balancer     | Public Load Balancer                  | Cloud Load Balancer               |
+| Relational Database       | Amazon Aurora             | Database for MySQL/Postgres Cluster   | CloudSQL Cluster                  |
```
udam-f2, line 25: Should probably say Aurora Cluster
```suggestion
| Relational Database       | Amazon Aurora Cluster     | Database for MySQL/Postgres Cluster   | CloudSQL Cluster                  |
```

specification/dimensions/resourcetype.md
udam-f2: Yeah, I think we have to provide some guidance for the value providers should use given that no one seems to have exactly what we're proposing in supporting content.

cnharris10: It's meant to be open-ended and up to the provider to define. Human-readable or coded is fine. I will add 2 points around:

- up to the provider to create values
- value SHOULD be consistent in one form (ex: human-readable or coded) but not mixed.

ljadvey (approved):

specification/dimensions/resourcetype.md
marcin-walkow, line 1: Should we briefly mention what a Resource Type can be, but not limited to, like a Virtual Machine, Storage or SQL DB? (even if it is not normalized and the supporting content is there?)
_Resource Type describes the kind of resource for which you are being charged.  A Resource Type is commonly used for scenarios like identifying cost changes in groups of similar resources. **A few examples of Resource Type include Virtual Machine, Storage or SQL DB**_

I am thinking about PricingUnit that we added such examples to give a brief understanding of what it _**can**_ be

cnharris10: I can add a line with some potential examples but I do think @flanakin has a good point that we should not bound Resource Type to be human-readable text.  However, the provider wants to surface these values is up to them but they SHOULD remain consistent.

tobrien (approved):

specification/dimensions/resourcetype.md
```diff
@@ -0,0 +1,40 @@
...
+
+Resource Type MUST be present within billing data.  Resource Type MUST be of type String and MUST NOT be NULL when a corresponding ResourceId is not NULL.  When a corresponding ResourceId value is NULL, the ResourceType column value MUST also be NULL.  Resource Type MUST contain consistent values and a consistent value-format that accounts for additional values added.
+
+Some potential values MAY include Virtual Machine, Data Warehouse, and Load Balancer.
```
udam-f2, line 7: I really don't see the value in this. Either we provide a recommended set or we leave this out of normative content. If we really want to show some example values, how about between sentence one and two in the first paragraph?

If we want to put in normative content, we should probably provide a more comprehensive list and have something stronger like if the resource is one of these types, these values MUST be used. It doesn't seem that's the intention here.

specification/dimensions/resourcetype.md
flanakin: Should we ask practitioners what they want here?

specification/dimensions/resourcetype.md
flanakin: Fwiw, I'm not opposed to making it human readable. That's what we've done in every other column. My only point was to be explicit about it if we want it to be. Maybe that's something we need to define as a general attribute across the board, tho. I suppose I only asked because Microsoft's value is more of a code. I'm sure someone else will use a code in another column that I'd expect would be human readable, so there's probably value in an all-encompassing attribute 🤔

specification/dimensions/resourcetype.md
```diff
@@ -0,0 +1,34 @@
+# Resource Type
+
+Resource Type describes the kind of resource for which you are being charged.  A Resource Type is commonly used for scenarios like identifying cost changes in groups of similar resources and MAY include values like Virtual Machine, Data Warehouse, and Load Balancer.
+
+The ResourceType column MUST be present within billing data.  This column MUST be of type String and MUST NOT be null when a corresponding ResourceId is not null.  When a corresponding ResourceId value is null, the ResourceType column value MUST also be null.  Providers MUST use a consistent value-format and a set of values for ResourceType values within the dataset.
```
udam-f2, line 5:
```suggestion
The ResourceType column MUST be present within billing data.  This column MUST be of type String and MUST NOT be null when a corresponding ResourceId is not null.  When a corresponding ResourceId value is null, the ResourceType column value MUST also be null.  Providers MUST use a consistent value-format and a set of values for ResourceType values within their cost and usage datasets.
```

specification/dimensions/resourcetype.md
```diff
@@ -0,0 +1,34 @@
+# Resource Type
+
+Resource Type describes the kind of resource for which you are being charged.  A Resource Type is commonly used for scenarios like identifying cost changes in groups of similar resources and MAY include values like Virtual Machine, Data Warehouse, and Load Balancer.
```
udam-f2, line 3:
```suggestion
Resource Type describes the kind of resource for which you are being charged.  A Resource Type is commonly used for scenarios like identifying cost changes in groups of similar resources and may include values like Virtual Machine, Data Warehouse, and Load Balancer.
```

specification/dimensions/resourcetype.md
```diff
@@ -0,0 +1,34 @@
...
+Resource Type
+
+## Description
+
+The kind of resource for which you are being charged.
+
+## Data type
+
+String
+
```
udam-f2, line 22: Where are we getting these random headings from :) ? Data type is already in the content constraints below.

udam-f2, line 22:
```suggestion
```

specification/dimensions/resourcetype.md
```diff
@@ -0,0 +1,40 @@
...
+Resource Type
+
+## Description
+
+The kind of resource for which you are being charged.
+
+## Normalized?
+
+No
+
```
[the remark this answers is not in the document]
cnharris10: Removing it

marcin-walkow (approved):

Issue #151, opened 2023-08-18
CommitmentDiscountType column


flanakin:
Define a column to distinguish the type of commitment-based discount (e.g., committed usage vs. committed cost).

This column will be normalized so it is consistent across providers.

> ℹ️ _This column name is a placeholder and is not final._

PR #173, opened 2023-09-09
FOCUS #151: Commitment Discount Category

rodneyjoyce:
First draft for review


specification/dimensions/commitmentdiscounttype.md
```diff
@@ -0,0 +1,39 @@
+# Commitment Discount Type
+
+A commitment-based discount is a contractual commitment in exchange for a discounted unit price. Commitment discounts are available as spend-based or resource-based discounts, depending on the CSP offering them and the particular service.
```
udam-f2, line 3:
```suggestion
A commitment-based discount is a contractual commitment in exchange for a discounted unit price. Commitment discounts are available as spend-based or resource-based discounts, depending on the Provider's offering.
```

udam-f2: Also, not a fan of resource-based 'discounts'. Its really a discount due to a commitment for an amount of spend or resource consumption. 

FinOps documentation refers to this as resource-based commitments vs spend-based commitments.

udam-f2:
```suggestion
A commitment-based discount is a contractual commitment in exchange for a discounted unit price. These discounts may be available for spend-based commitments or resource-based commitments, depending on the Provider's offerings.
```

specification/dimensions/commitmentdiscounttype.md
```diff
@@ -0,0 +1,39 @@
...
+
+This column determines whether the commitment is based on a minimum spend or minimum usage.
+
+The CommitmentDiscountType column MUST be present in the billing data. This column must be of type String and MUST NOT contain null values when a charge IS related to a commitment-based discount. When a charge IS NOT associated with a commitment-based discount, the column MUST be null. The CommitmentDiscountType MUST be one of the allowed values.
```
udam-f2, line 7:
```suggestion
The CommitmentDiscountType column MUST be present in the billing data. This column must be of type String and MUST NOT contain null values when a charge is related to a commitment-based discount. When a charge is associated with a commitment-based discount, the column MUST be null. The CommitmentDiscountType MUST be one of the allowed values.
```

specification/dimensions/commitmentdiscounttype.md
```diff
@@ -0,0 +1,39 @@
+# Commitment Discount Type
```
udam-f2, line 1: Do we need to clarify how this is not the same as a negotiated discount you get when you commit to a spend-based commitment to a provider - as a part of your contract negotiation? 'Contractual commitment' below could be ambiguous.

supporting_content/dimensions/commitmentdiscounttype.md
```diff
@@ -0,0 +1,17 @@
+# Column: CommitmentDiscountType
+
+## Documentation
```
udam-f2, line 3: Could we add the provider mappings and examples sections please. Many of the providers may not have this equivalent but then we can say that in the table (AWS for example has a bunch of columns for RIs and different columns for savings plans - therefore based on which ones they populate, you can tell what kind of discount it was. That's good context to provide even though its not a direct mapping for this column)

supporting_content/dimensions/commitmentdiscounttype.md
```diff
@@ -0,0 +1,17 @@
...
+- Microsoft
+  - Azure:  Understand usage details fields: https://learn.microsoft.com/en-us/azure/cost-management-billing/automate/understand-usage-details-fields
+- GCP
+  - Google Commitment Types:  https://cloud.google.com/docs/cuds#spend_versus_resource_commitments (This is the most similar to our selected implementation)
+- AWS
+  - Amazon: Reservation details - https://docs.aws.amazon.com/cur/latest/userguide/reservation-columns.html
+
+## Discussion Topics
```
udam-f2, line 11: It would be good to add some notes about our discussions on if this should be a broader 'Discount Type' or specifically a 'Commitment Discount Type' and why we chose to go this direction (and possibly why a general discount type might still be useful.

specification/dimensions/commitmentdiscounttype.md
```diff
@@ -0,0 +1,39 @@
...
+
+A commitment-based discount is a contractual commitment in exchange for a discounted unit price. Commitment discounts are available as spend-based or resource-based discounts, depending on the CSP offering them and the particular service.
+
+This column determines whether the commitment is based on a minimum spend or minimum usage.
```
marc-perreaut, line 5: Does this column apply for:
- Lines corresponding to commitment purchases (when a RI/SP is purchased)
- Lines corresponding to commitment usages (when a resource is used and the RI/SP discount applies)
- Both?

supporting_content/dimensions/commitmentdiscounttype.md
```diff
@@ -0,0 +1,17 @@
...
+TBC: Another column that describes the type of CUD in the terms of the CSP, eg.
```
marc-perreaut, line 12: I understand there is a discussion about a new dimension PricingModel, whose values could be maybe OnDemand, SavingsPlan, Reservation, Spot, CUD. Would it match the need?

Moreover, if PricingModel is confirmed, the CommitmentDiscountType could be deducted from the PricingModel:
- If PricingModel is SavingsPlan, then CommitmentDiscountType is SpendBased
- If PricingModel is Reservation or CUD, then CommitmentDiscountType is ResourceBased
- If PricingModel is Spot or OnDemand, then CommitmentDiscountType is NULL
So in that case, would the CommitmentDiscountType be still needed as a dedicated explicit dimension?

udam-f2: @rodneyjoyce feel free to chime in but I believe there were discussions to try to limit the pricing model to a 'single responsibility' (e.g. CommitmentDiscount rather than the specific program) and then have the type gives the type and a program name for providing the provider specific column for the name given by provider (e.g. Reserved Instances or Committed Use Discounts).

specification/dimensions/commitmentdiscounttype.md
flanakin: Can we move these comments to the CommitmentDiscountId PR? Given that's further along, I'd like to close the wording there. This PR should just copy what it has when finalized.

@rodneyjoyce along the same lines, if there's anything you feel is missing there, can you add a comment as well?

specification/dimensions/commitmentdiscounttype.md
flanakin: nit: Follow the typical "A <column name> is..." pattern. Might be good to say reference the CommitmentDiscountId column, since this is in reference to that.

flanakin, line 5: nit: "Spend" is jargon. Do we want to use that? We haven't taken a stance on whether we'll use jargon or not. Admittedly, "cost" doesn't roll off the tongue in the same way, but we do have "cost" columns and we do not have "spend" columns. Food for thought...

specification/dimensions/commitmentdiscounttype.md
flanakin: Missed a very important "not"...
```suggestion
The CommitmentDiscountType column MUST be present in the billing data. This column must be of type String and MUST NOT contain null values when a charge is related to a commitment-based discount. When a charge is not associated with a commitment-based discount, the column MUST be null. The CommitmentDiscountType MUST be one of the allowed values.
```

specification/dimensions/commitmentdiscounttype.md
```diff
@@ -0,0 +1,39 @@
...
+| Data type       | String           |
+| Allows nulls    | True             |
+| Value format    | List of values   |
+
+Allowed values:
+
+| Value      | Description                                                                                                                                                                   |
+|:---------------|:--------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
+| Spend-based    | Commitments are purchased and measured in terms of the dollars per hour of equivalent on-demand spend.                                                                    |
+| Resource-based | Commitments are purchased and measured in terms of the underlying resources.                                                                                              |
```
flanakin, line 35: Technically, both quantity and cost are based on resources. It's probably more accurate to refer to this as either a "Quantity-Based" or "Usage-Based". 

Just to throw out another idea: We could use the column name that is committed. I'm not sure that's ideal, but just wanted to share in case that sparks something.
```suggestion
| Usage-Based | Commitments are purchased and measured in terms of the underlying resource usage.                                                                                              |
```

specification/dimensions/commitmentdiscounttype.md
```diff
@@ -0,0 +1,39 @@
...
+| Column required | True             |
+| Data type       | String           |
+| Allows nulls    | True             |
+| Value format    | List of values   |
+
+Allowed values:
+
+| Value      | Description                                                                                                                                                                   |
+|:---------------|:--------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
+| Spend-based    | Commitments are purchased and measured in terms of the dollars per hour of equivalent on-demand spend.                                                                    |
```
flanakin, line 34: I believe `-Based` should be capitalized.

supporting_content/dimensions/commitmentdiscounttype.md
rodneyjoyce: Ok, will get some examples for this

supporting_content/dimensions/commitmentdiscounttype.md
rodneyjoyce: Ok, will add

supporting_content/dimensions/commitmentdiscounttype.md
rodneyjoyce: Yes, the problem with the Pricing Model is that we could not agree on the list of values, so we decided to start with these more low-level/specific columns. It's a bit of a chicken and egg situation - once we get these columns signed off we will take a look at Pricing Model with fresh eyes.

WRT the point about inferring the Discount Type from another column (e.g. Pricing Model) - this is a balancing act between convenience and redundancy I guess. Practitioners should be able to solve the most common problems and answer the most common questions with as little effort as possible and without having to consider additional business logic...

The other column hinted at here is capturing the Discount Type Name from each cloud provider which would be specific to their implementation (but will still be a specific Commitment Discount Type.

specification/dimensions/commitmentdiscounttype.md
rodneyjoyce: Yes, this makes sense. What is the term we used for the account-level, negotiated discounts?

specification/dimensions/commitmentdiscounttype.md
rodneyjoyce: I am not 100% sure on this, but I would imagine it only applies where the ChargeType is Purchase - ie. the record where you bought the ComittmentDiscount.  I guess this depends on which column records the unit price discount that has been in given in exchange for the comittment.

rodneyjoyce: If Spend and Cost can be used interchangeably then it's probably wise it standardise on Cost and not introduce another term. Anyone else have thoughts on this?

flanakin: I would expect this to be on the usage records as well.

aqu-erp: Tough call. `Spend` here is part of a descriptor of a commitment, representing a commitment of monetary-amount per time-unit. Is the question, "Do we allow jargon in field values?" 

If we standardised on "cost": the "cost" **_at the basis of the_** commitment is a different term from "cost" **_of the_** commitment. Not clear to me yet if this is problematic.

rodneyjoyce: Ok, but it would only apply where the CommitmentDiscountId is not null

specification/dimensions/commitmentdiscounttype.md
rodneyjoyce: Is this something I should put in a poll to a wider audience to get a quorum?

specification/dimensions/commitmentdiscounttype.md
flanakin: @udam-f2, I'm assuming you're referring to the CBD description in the first paragraph. Since we're sourcing that from the CommitmentDiscountId PR, can you review the description there and leave a comment if you feel it's not covering anything in particular? I'll resolve this for this PR and we can update as needed.

supporting_content/dimensions/commitmentdiscounttype.md
flanakin: Microsoft has PricingModel with "SavingsPlan" and "Reservation" values.

I think AWS has a PurchaseOption column that is similar.

specification/dimensions/commitmentdiscounttype.md
flanakin: That might be quickest. If we feel we have agreement here, then there's no need, but polls do tend to get relatively quick closure.

specification/dimensions/commitmentdiscounttype.md
flanakin: I feel this needs to be part of our core design principles. Let me bring this to the practitioner squad and see what they say. Leave this comment open, but let's not block on it. If anyone has any opinions, please share them here.

supporting_content/dimensions/commitmentdiscounttype.md
aqu-erp: GCP's equivalent here is the  "Credit Type" column

specification/dimensions/commitmentdiscounttype.md
marc-perreaut: Could a ChargeType be a Purchase of something that is not a commitment discount? I believe ChargeType can be Purchase for Marketplace. In this case, CommitmentDiscountId would not apply.

Furthemore, a CommitmentDiscountType could also apply where the ChargeType is UnusedCommittment (in Azure, UnusedCommittment is actually either UnusedReservation or UnusedSavingsPlan).

The specification says later "When a charge is not associated with a commitment-based discount, the column MUST be null." So, maybe the column CommitmentDiscountType must not be null if and only if the rows related to  a commitment  discount, whatever its ChargeType. Or maybe we need to have a vision or direction on structuring dimensions ChargeType (and PricingModel to me), which drives the behavior of other dimensions like CommitmentDiscountType and CommitmentDiscountId.

marc-perreaut: The word "spend" is used by AWS ("Savings Plans is a flexible pricing model that can help you reduce your bill by up to 72% compared to On-Demand prices, in exchange for a one- or three-year hourly spend commitment"), by GCP ("Spend-based committed use discounts provide a discount in exchange for your commitment to spend a minimum amount for a product or service."), and by Azure ("[Savings Plan] Get lower prices by agreeing to spend a fixed hourly amount on compute services for one or three years.").

Moreover, "spend-based commitment discount" sounds much clearer than "cost-based commitment discount".

flanakin: All `CommitmentDiscount*` columns should be specified when the charge (any type of charge) is related to a commitment-based discount.

We haven't closed unused commitment handling yet. I forwarded you an invite in case you can make it :-) 

+1 on your last paragraph. `PricingModel` convos are pending. If you have thoughts, we could setup time to chat!

flanakin (approved):

udam-f2:
@flanakin @rodneyjoyce What's the github issue attached to this work?

specification/dimensions/commitmentdiscounttype.md
rodneyjoyce: Ok, going with Spend

specification/dimensions/commitmentdiscounttype.md
rodneyjoyce: I am now leaning more towards Usage-Based after reading through some of the other polls. I'll still put it to a poll to see if there's something we have missed if we change to this

rodneyjoyce:
> @flanakin @rodneyjoyce What's the github issue attached to this work?

How do I specifically link the PR to the Project work Item? 
image

specification/dimensions/commitmentdiscounttype.md
rodneyjoyce: Actually, no, I think we agree on this - no one else has commented, so I'll switch to Usage-Based and anyone can comment here if they feel strongly about it.

supporting_content/dimensions/commitmentdiscounttype.md
rodneyjoyce: Updated

supporting_content/dimensions/commitmentdiscounttype.md
rodneyjoyce: > Microsoft has PricingModel with "SavingsPlan" and "Reservation" values.
> 
> I think AWS has a PurchaseOption column that is similar.

Would these not be for the NEW column - ie. the new, non-normalised column that is the Program name?

rodneyjoyce: Ok, please review my changes @udam-f2

flanakin: > Would these not be for the NEW column - ie. the new, non-normalised column that is the Program name?

Yes, just calling out the closest thing to the purpose of this column.

aqu-erp (approved):

flowerinthenight (approved):
LGTM

specification/dimensions/commitmentdiscounttype.md
```diff
@@ -0,0 +1,41 @@
+# Commitment Discount Type
+
+A commitment-based discount is a contractual commitment in exchange for a discounted unit price. These discounts may be available for spend-based commitments or resource-based commitments, depending on the Provider's offerings.
```
udam-f2, line 3:
```suggestion
A commitment-based discount is a contractual commitment for an amount of usage or spend throughout a specified term, in exchange for discounted unit pricing on that amount. The commitment may be based on quantities of resource units or monetary value, with various payment options and time frames.
```

udam-f2: Brining over the definition that was finalized in Commitment Discount ID/Name

specification/dimensions/commitmentdiscounttype.md
```diff
@@ -0,0 +1,41 @@
...
+
+This column determines whether the commitment is based on a minimum spend or minimum usage.
```
udam-f2, line 5: We decided to remove the word minimum to make this more general / future-proof.

udam-f2, line 5:
```suggestion
This column determines whether the commitment is resource-based or usage-based.
```

specification/dimensions/commitmentdiscounttype.md
```diff
@@ -0,0 +1,41 @@
...
+| Data type       | String           |
+| Allows nulls    | True             |
+| Value format    | List of values   |
+
+Allowed values:
+
+| Value      | Description                                                                                                                                                                   |
+|:---------------|:--------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
+| Spend-Based    | Commitments are purchased and measured in terms of the dollars per hour of equivalent on-demand spend.                                                                    |
+| Usage-Based | Commitments are purchased and measured in terms of the underlying resource usage.                                                                                              |
```
udam-f2, line 37:
```suggestion
| Value      | Description                                                                       |
|:---------------|:-----------------------------------------------|
| Spend-Based   | Commitments requiring an amount of spend.    |
| Usage-Based   | Commitments requiring an amount of usage     |
```

specification/dimensions/commitmentdiscounttype.md
flanakin:
```suggestion
This column determines whether the commitment is based on usage quantity or cost (aka "spend").
```

specification/dimensions/commitmentdiscounttype.md
```diff
@@ -0,0 +1,41 @@
+# Commitment Discount Type
+
+A commitment-based discount is a contractual commitment in exchange for a discounted unit price. These discounts may be available for spend-based commitments or resource-based commitments, depending on the Provider's offerings.
+
+This column determines whether the commitment is based on a minimum spend or minimum usage.
+
+The CommitmentDiscountType column MUST be present in the billing data. This column must be of type String, MUST be null when CommitmentDiscountId is null, and MUST NOT be null when CommitmentDiscountId is not null.
```
flanakin, line 7:
```suggestion
The CommitmentDiscountType column MUST be present in the billing data. This column MUST be of type String, MUST be null when CommitmentDiscountId is null, and MUST NOT be null when CommitmentDiscountId is not null.
```

specification/dimensions/commitmentdiscounttype.md
```diff
@@ -0,0 +1,41 @@
...
+
+The CommitmentDiscountType MUST be one of the allowed values.
```
flanakin, line 9:
```suggestion
The CommitmentDiscountType column MUST be present in the billing data. This column MUST be of type String, MUST be null when CommitmentDiscountId is null, and MUST NOT be null when CommitmentDiscountId is not null. The CommitmentDiscountType MUST be one of the allowed values.
```

specification/dimensions/commitmentdiscounttype.md
flanakin: Covered in the other comment.

specification/dimensions/commitmentdiscounttype.md
```diff
@@ -0,0 +1,41 @@
...
+
+CommitmentDiscountType
+
+## Display name
+
+Commitment Discount Type
+
+## Description
+
+The type of commitment agreed to in exchange for the discounted unit price.
```
flanakin, line 21:
```suggestion
Indicates whether the commitment-based discount identified in the CommitmentDiscountId column is based on usage quantity or cost (aka "spend").
```

specification/dimensions/commitmentdiscounttype.md
flanakin:
```suggestion
| Value      | Description                                                                                                                                                                   |
|:---------------|:--------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
| Spend-Based   | Commitment-based discounts that require a predetermined amount of spend.    |
	| Usage-Based   | Commitment-based discounts that require a predetermined amount of usage.     |
```

specification/dimensions/commitmentdiscounttype.md
```diff
@@ -0,0 +1,39 @@
...
+| Data type       | String           |
+| Allows nulls    | True             |
+| Value format    | List of values   |
+
+Allowed values:
+
+| Value      | Description                                                                                                                                                                   |
+|:---------------|:--------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
+| Spend-Based   | Commitment-based discounts that require a predetermined amount of spend.    |
+	| Usage-Based   | Commitment-based discounts that require a predetermined amount of usage.     |
```
flanakin, line 35:
```suggestion
| Spend-Based   | Commitment-based discounts that require a predetermined amount of spend.    |
| Usage-Based   | Commitment-based discounts that require a predetermined amount of usage.     |
```

specification/dimensions/commitmentdiscounttype.md
```diff
@@ -0,0 +1,39 @@
...
+| Data type       | String           |
+| Allows nulls    | True             |
+| Value format    | List of values   |
+
+Allowed values:
+
+| Value      | Description                                                                                                                                                                   |
+|:---------------|:--------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
+| Spend-Based   | Commitment-based discounts that require a predetermined amount of spend.    |
+| Usage-Based   | Commitment-based discounts that require a predetermined amount of usage.     |
```
flanakin, line 35: nit: Does having "-Based" in both (and potentially all future) values feel unnecessary to anyone else? Should we just use `Spend` and `Usage`?

marc-perreaut: I agree "-based" is needed in the description (plain text) but it is superfluous in the values, as it's present for all values.

ljadvey: +1

ljadvey (approved):

specification/dimensions/commitmentdiscounttype.md
ljadvey, line 35:
```suggestion
| Usage         | Commitment-based discounts that require a predetermined amount of usage.     |
```

ljadvey (changes requested):

salpaysenturus:
@flanakin can you confirm that this section of the spec would be the incorrect place to explicitly tell the novice FinOps practitioner that Reservation (or Reserved Instance) = Usage and Savings Plan = Spend?  I ask because, per our conversation earlier today, I increasingly feel that there will need to be a cheat sheet either within or alongside the spec that helps people understand these myriad translations from provider-specific to provider-agnostic...

udam-f2:
> @flanakin can you confirm that this section of the spec would be the incorrect place to explicitly tell the novice FinOps practitioner that Reservation (or Reserved Instance) = Usage and Savings Plan = Spend? I ask because, per our conversation earlier today, I increasingly feel that there will need to be a cheat sheet either within or alongside the spec that helps people understand these myriad translations from provider-specific to provider-agnostic...

The thinking is that we're referring to these as generic 'commitment based discounts' and that we would have a commitment program name that can provide the provider-specified name. I would be against us generically saying Reservations are usage based and SPs are spend-based. That may be true based on the way two providers have named things but it's not a generic concept.

specification/dimensions/commitmentdiscounttype.md
```diff
@@ -0,0 +1,39 @@
...
+Indicates whether the commitment-based discount identified in the CommitmentDiscountId column is based on usage quantity or cost (aka "spend").
+
+## Content constraints
+
+|    Constraint   |      Value       |
+|:----------------|:-----------------|
+| Column required | True             |
+| Data type       | String           |
+| Allows nulls    | True             |
+| Value format    | List of values   |
```
udam-f2, line 28: We have been using list-of-values
```suggestion
| Value format    | list-of-values   |
```

udam-f2 (approved):
@rodneyjoyce @flanakin what do we need to do to wrap this up and get checked in?

flanakin:
> @rodneyjoyce @flanakin what do we need to do to wrap this up and get checked in?

Just wanted to get maintainers inputs. Will share it with members to review...

specification/dimensions/commitmentdiscounttype.md
```diff
@@ -0,0 +1,39 @@
+# Commitment Discount Type
+
+A commitment-based discount is a contractual commitment for an amount of usage or spend throughout a specified term, in exchange for discounted unit pricing on that amount. The commitment may be based on quantities of resource units or monetary value, with various payment options and time frames.
```
cnharris10, line 3: Ideally, it'd be great to clearly distance this definition from a "capacity reservation" which absolutely isn't a discount.

You could call out that capacity guarantees MAY be coupled with a commitment discount type (but not always).

tobrien (approved):

ljadvey (approved):

specification/dimensions/commitmentdiscounttype.md
```diff
@@ -0,0 +1,39 @@
...
+| Data type       | String           |
+| Allows nulls    | True             |
+| Value format    | list-of-values   |
+
+Allowed values:
+
+| Value      | Description                                                                                                                                                                   |
+|:---------------|:--------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
+| Spend   | Commitment-based discounts that require a predetermined amount of spend.    |
+| Usage   | Commitment-based discounts that require a predetermined amount of usage.     |
```
udam-f2, line 35:
```suggestion
|:--------|:--------------------------------------------------------------------------------|
| Spend   | Commitment-based discounts that require a predetermined amount of spend.    |
| Usage   | Commitment-based discounts that require a predetermined amount of usage.    |
```

udam-f2: Personally, I think Spend-commitment and Usage commitment are more descriptive terms than just Spend and Usage. Curious what others think. I think that may have gotten updated recently so I assume it was discussed.

specification/dimensions/commitmentdiscounttype.md
udam-f2: If making an update here, I assume we would have to make it elsewhere. I like "Commitment Discount *" not being mixed with on-demand capacity reservations in AWS (and the equivalents) since they don't apply here. Would we make a differentiation where "Commitment Discount *" columns only apply to cases where your commitment gets discounts and "Commitment *" columns getting used when a column must apply to either?

specification/dimensions/commitmentdiscounttype.md
marc-perreaut: Personally, I find "shorter is better" for column values, as the column name gives the context for the values.
Another example is the PricingModel PR: I would find cumbersome if every value of the Pricing Model column is suffixed with "Pricing Model".
For column names, I find that values being more explicit and descriptive is indeed better.

joshk0 (approved):

specification/dimensions/commitmentdiscounttype.md
TDubovchenko: I would vote for Usage and Spend as column name is descriptive

specification/dimensions/commitmentdiscounttype.md
cnharris10: I think that makes sense

flanakin: Do you think we need to explicitly call this out here?

supporting_content/dimensions/commitmentdiscounttype.md
```diff
@@ -0,0 +1,55 @@
...
+Current values observed in billing data for various scenarios:
+
+| Provider | Data set                 | CommitmentDiscountType     | CommitmentDiscountProgram (Name TBD)     |
+|----------|--------------------------|----------------------------|------------------------------------------|
+| AWS | CUR (PurchaseOption)                   | Usage-Based                | Reserved Instances              |
+| AWS | CUR (PurchaseOption)                   | Usage-Based                | Savings Plans                   |
+| Google Cloud | BigQuery Billing Export | Usage-Based        | Resource-based CUD                            |
+| Google Cloud | BigQuery Billing Export | Spend-Based        | Spend-based CUD                               |
+| Microsoft | Cost Details (PricingModel)| Usage-Based                     | Savings Plan                     |
+| Microsoft | Cost Details (PricingModel)| Usage-Based                     | Reservation                      |
```
TDubovchenko, line 24: Aren't AWS and Azure Savings plans examples for spend-based commitments?
Aren't AWS and Azure Savings plans examples for spend-based commitments? 
The commitment is to spend hourly a certain amount, Although how the discount is applied and resulting savings depend on what kind of resources were running, the commitment is to spend hourly a certain amount.

specification/dimensions/commitmentdiscounttype.md
yprikhodko: Hey @cnharris10,
Why do you think it can be confused with capacity reservation? Capacity reservation doesn't provide discount and can be cancelled at any time, so it is not really a commitment and doesn't fall into above definition. 
Another note, what is meant by word 'contractual' here? Is it really needed?

supporting_content/dimensions/commitmentdiscounttype.md
```diff
@@ -0,0 +1,55 @@
...
+| Microsoft | Cost Details             | Not available               | N/A |
+
+## Example usage scenarios
+
+Current values observed in billing data for various scenarios:
+
+| Provider | Data set                 | CommitmentDiscountType     | CommitmentDiscountProgram (Name TBD)     |
+|----------|--------------------------|----------------------------|------------------------------------------|
+| AWS | CUR (PurchaseOption)                   | Usage-Based                | Reserved Instances              |
+| AWS | CUR (PurchaseOption)                   | Usage-Based                | Savings Plans                   |
```
yprikhodko, line 20: In Savings Plans you're committing to particular spend per hour in $, shouldn't it be marked as Spend-Based?

supporting_content/dimensions/commitmentdiscounttype.md
```diff
@@ -0,0 +1,55 @@
...
+| Microsoft | Cost Details (PricingModel)| Usage-Based                     | Reservation                      |
+
+## Documentation
+- Microsoft
+  - Azure:  Understand usage details fields: https://learn.microsoft.com/en-us/azure/cost-management-billing/automate/understand-usage-details-fields
+- GCP
+  - Google Commitment Types:  https://cloud.google.com/docs/cuds#spend_versus_resource_commitments (This is the most similar to our selected implementation)
+  - https://cloud.google.com/billing/docs/how-to/export-data-bigquery-tables/standard-usage - see credit.type
+- AWS
+  - Amazon: Reservation details - https://docs.aws.amazon.com/cur/latest/userguide/reservation-columns.html
```
yprikhodko, line 33: + Savings Plans details: https://docs.aws.amazon.com/savingsplans/

supporting_content/dimensions/commitmentdiscounttype.md
```diff
@@ -0,0 +1,55 @@
+# Column: CommitmentDiscountType
+
+## Example provider mappings
+
+Current column mappings found in available data sets:
+
+| Provider | Data set                 | Column                    |  Example Values  |
+|----------|--------------------------|---------------------------|------------------|
+| AWS | CUR                      | Not available   | N/A |
```
yprikhodko, line 9: For AWS can be calculated: if reservation/ReservationARN <> '' then Usage, else if savingsPlan/SavingsPlanArn
<> '' then Spend, else null

specification/dimensions/commitmentdiscounttype.md
udam-f2: I can see your point @yprikhodko and @flanakin. Ok if we don't make any reference to capacity reservations. 

Conceptually the only similarities are the fact that you have a used and unused portion that would need to be charged to the capacity reservation itself and the fact that there's a 'reservation' not a 'commitment' per se.

Also ok without 'contractual' in the description. It's no more contractual than any other usage or purchase with a provider.

cnharris10: Simply, the fact that AWS and Azure uses "reservation" in "Capacity Reservation", GCP literally calls a capacity reservation a "Reservation", and AWS/Azure-like "reservations" are "commitments" (except BigQuery reservations are "reservations") in GCP makes differentiating these terms incredibly complex.

Just a line that says, ~"Commitments may be bundled with infrastructure guarantees but entirely disjoint from purchases that don't incur a fix term" would go a long way.

rodneyjoyce: I have removed the word "contractual" in the following files:

- commitmentdiscountname
- commitmentdiscounttype
- commitmentdiscountid

supporting_content/dimensions/commitmentdiscounttype.md
rodneyjoyce: Thanks - this was an oversight - updated

supporting_content/dimensions/commitmentdiscounttype.md
rodneyjoyce: Thanks - this was an oversight - updated

supporting_content/dimensions/commitmentdiscounttype.md
rodneyjoyce: Ok, added to text

PR #176, opened 2023-09-18
keyvalueformat

kk09v:
Initial commit for KeyValueFormat to get broader review.

### DIFF
--- a/specification/attributes/key_value_format.md
+++ b/specification/attributes/key_value_format.md
@@ -16,14 +16,14 @@ Rules and formatting requirements for columns appearing in billing data which co
 
 ## Requirements
 
-* Data MUST be provided as a string in a format consistent with ECMA-404, as a single Object
-    * When applied to the key-value pair, string refers to the key and value refers to the value
-* Keys MUST be unique within an object
-    * In the case of key collisions, multiple sets of key value pairs may be provided by prefixing tag keys to disambiguate sources
-* Keys SHOULD be case insensitive
-* Values SHOULD be case sensitive
-* Source data containing only a keyword instead of a key-value pair SHOULD be expressed as a key-value pair with keyword being treated as the key and TRUE being the value
-* Source data containing a key with a null value MAY be expressed as a key-value pair with key being the key and an empty string being the value 
+* Data MUST be provided as a string in a format consistent with ECMA-404, as a single object.
+  * When applied to the key-value pair, "string" refers to the key and "value" refers to the value.
+* Keys MUST be unique within an object.
+  * In the case of key collisions, multiple sets of key-value pairs can be provided by prefixing tag keys to disambiguate sources.
+* Keys SHOULD be case-insensitive.
+* Values SHOULD be case-sensitive.
+* Source data containing only a keyword instead of a key-value pair SHOULD be expressed as a key-value pair with keyword being treated as the key and TRUE being the value.
+* Source data containing a key with a null value MAY be expressed as a key-value pair with key being the key and an empty string being the value.
 
 ## Exceptions
 
--- a/specification/attributes/key_value_format.md
+++ b/specification/attributes/key_value_format.md
@@ -22,8 +22,8 @@ Rules and formatting requirements for columns appearing in billing data which co
   * In the case of key collisions, multiple sets of key-value pairs can be provided by prefixing tag keys to disambiguate sources.
 * Keys SHOULD be case-insensitive.
 * Values SHOULD be case-sensitive.
-* Source data containing only a keyword instead of a key-value pair SHOULD be expressed as a key-value pair with keyword being treated as the key and TRUE being the value.
-* Source data containing a key with a null value MAY be expressed as a key-value pair with key being the key and an empty string being the value.
+* Source data containing only a keyword instead of a key-value pair SHOULD be expressed as a key-value pair with keyword being treated as the key and `true` being the value.
+* Source data containing a key with a null value SHOULD be expressed as a key-value pair with key being the key and `null` being the value.
 
 ## Exceptions
 
--- a/specification/attributes/key_value_format.md
+++ b/specification/attributes/key_value_format.md
@@ -1,0 +1,34 @@
+# Key-Value Format
+
+This is the specified format for providing data in the form of key-value pairs. In addition, this can support keyword data which has a key but not value. This format is used for data such as tags, labels, and resource metadata. 
+
+## Attribute ID
+
+KeyValueFormat
+
+## Attribute Name
+
+Key-Value Format
+
+## Description
+
+Rules and formatting requirements for columns appearing in billing data which convey key-value pairs and/or keyword attributes.
+
+## Requirements
+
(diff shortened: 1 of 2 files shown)


specification/attributes/key_value_format.md
```diff
@@ -0,0 +1,34 @@
...
+
+## Requirements
+
+* Data MUST be provided as a string in a format consistent with ECMA-404, as a single Object
+    * When applied to the key-value pair, string refers to the key and value refers to the value
+* Keys MUST be unique within an object
+    * In the case of key collisions, multiple sets of key value pairs may be provided by prefixing tag keys to disambiguate sources
+* Keys SHOULD be case insensitive
+* Values SHOULD be case sensitive
+* Source data containing only a keyword instead of a key-value pair SHOULD be expressed as a key-value pair with keyword being treated as the key and TRUE being the value
```
cnharris10, line 25: Should this be `"true"` instead of `TRUE` to fit within this definition? https://www.crockford.com/mckeeman.html

specification/attributes/key_value_format.md
```diff
@@ -0,0 +1,34 @@
...
+
+Key-Value Format
+
+## Description
+
+Rules and formatting requirements for columns appearing in billing data which convey key-value pairs and/or keyword attributes.
+
+## Requirements
+
+* Data MUST be provided as a string in a format consistent with ECMA-404, as a single Object
```
kk09v, line 19: Often referred to as serialized JSON

specification/attributes/key_value_format.md
```diff
@@ -0,0 +1,34 @@
...
+    * When applied to the key-value pair, string refers to the key and value refers to the value
+* Keys MUST be unique within an object
+    * In the case of key collisions, multiple sets of key value pairs may be provided by prefixing tag keys to disambiguate sources
+* Keys SHOULD be case insensitive
+* Values SHOULD be case sensitive
```
kk09v, line 24: When applied to string values

specification/attributes/key_value_format.md
```diff
@@ -0,0 +1,34 @@
...
+* Source data containing only a keyword instead of a key-value pair SHOULD be expressed as a key-value pair with keyword being treated as the key and TRUE being the value
+* Source data containing a key with a null value MAY be expressed as a key-value pair with key being the key and an empty string being the value
```
flanakin, line 26: Our null handling doc says to not use empty string. We may want to also update that to include this as an exception to that rule.

udam-f2:
@kk09v @cnharris10 Can we format this PR with the proper github issue ID etc

rileyjenk (approved):

specification/attributes/key_value_format.md
```diff
@@ -0,0 +1,34 @@
...
+
+Rules and formatting requirements for columns appearing in billing data which convey key-value pairs and/or keyword attributes.
+
+## Requirements
+
+* Data MUST be provided as a string in a format consistent with ECMA-404, as a single object.
+  * When applied to the key-value pair, "string" refers to the key and "value" refers to the value.
+* Keys MUST be unique within an object.
+  * In the case of key collisions, multiple sets of key-value pairs can be provided by prefixing tag keys to disambiguate sources.
+* Keys SHOULD be case-insensitive.
```
rileyjenk, line 23: I have concerns about this.....many data systems that deal in JSON, keys are case sensitive.    Databse like Redshift, Mongo, MySQL all support JSON fields and are all case sensitive. The JSON spec itself is case sensitive on keys (in that you can do it), therefore python, java, R, etc....they use case sesntive keys when dealing with JSON. I have concerns if we are adding this requirement...it makes our JSON special...I don't think it should be.

specification/attributes/key_value_format.md
```diff
@@ -0,0 +1,34 @@
...
+* Values SHOULD be case-sensitive.
+* Source data containing only a keyword instead of a key-value pair SHOULD be expressed as a key-value pair with keyword being treated as the key and TRUE being the value.
```
rileyjenk, line 25: This is not JSON. I don't believe the spec should have logic like this. Use JSON and use it only, don't mix other structures.

specification/attributes/key_value_format.md
kk09v: Updating to `true` to align with how it's expressed in ECMA 404, section 4

specification/attributes/key_value_format.md
```diff
@@ -0,0 +1,34 @@
...
+* Source data containing a key with a null value MAY be expressed as a key-value pair with key being the key and an empty string being the value.
```
rileyjenk, line 26: Bear with me but I spend all day on database systems...  Are we saying that as far as the spec is concerned that {"thing":null }, and {"thing":""} are the same. I know visually they are the same but I don't think we should be in the business wading into. Especially when both are valid entries depending on the use case.

rileyjenk (changes requested):

specification/attributes/key_value_format.md
kk09v: Updating this to reflect `null` as the desired value when a value is not provided. 
Honestly, it depends on how you think about it; I was considering the empty text field to be a blank string rather than a null since the value field's existence in the UI instantiates the value as a string. However, I do believe that treating these as null would improve the usability of this field.

kk09v:
> @kk09v @cnharris10 Can we format this PR with the proper github issue ID etc

There isn't an issue ID yet. Here's the task, if you're able to convert it to an issue.
https://github.com/orgs/FinOps-Open-Cost-and-Usage-Spec/projects/4/views/1?filterQuery=tag&pane=issue&itemId=37058323

specification/attributes/key_value_format.md
kk09v: Changing this to a null.
https://github.com/FinOps-Open-Cost-and-Usage-Spec/FOCUS_Spec/pull/176#discussion_r1336605635

specification/attributes/key_value_format.md
kk09v: Absolutely fair and up for discussion. The JSON spec explicitly does not have these restrictions but does allow for specifications using JSON to apply these restrictions.

Per ECMA 404, section 6:
> The JSON syntax does not impose any restrictions on the strings used as names, does not require that name strings be unique, and does not assign any significance to the ordering of name/value pairs. These are all semantic considerations that may be defined by JSON processors or in specifications defining specific uses of JSON for data interchange.

specification/attributes/key_value_format.md
flanakin: @rileyjenk I believe these rules were more about tags. What do you think about moving them into the tags spec?

specification/attributes/key_value_format.md
flanakin: I think this is only an issue for the usability tags. I'd move it there. Fwiw, it's highly unlikely that any vendor would use 2 attributes with the same name, but different casing. (I'm going to regret writing this, aren't I? 🤦‍♂️)

udam-f2:
Karl asked that we close this and that he will put out a new one given that he believes there are significant differences based on new discussions @kk09v

Issue #146, opened 2023-08-15
Create a Billed Unit Price metric


ijurica:
### Type
> Type of issue: Metric
Normalized: No

### Description
> Define how Billed Unit Price (actual after-discount rate/unit price) should appear in cost data under metric.

### Definition of done

- [ ] Rationalize vendor-neutral, cross-cloud naming
- [ ] Complete spec template and include naming (code name, display name), constraints, guidelines, compatibility with major providers etc.

udam-f2:
@ijurica  If still relevant, could we please re-create this with the new feedback template? Feels like we should close this out?

ijurica:
As it is an outcome of various savings activities and can be derived from BilledCost/PricingQuantity (cake), I **suggest closing this one for now** and **focusing on Negotiated Unit Price or Negotiated Discount** (setting aside the naming aspect), which seems more crucial.
If necessary, we can revisit the BilledUnitPrice later (using the feedback template) 

@flanakin, as you expressed interest in this metric, I would like to confirm if it is acceptable to you?

flanakin:
Why do we have to choose one over the other? Shouldn't we do both?

We've never closed on the ingredients vs. cake discussion, so I don't think that's a reason to exclude anything at this point.

jpradocueva:
Task Force 2 agreed to close the issue on the May 15 call.

PR #178, opened 2023-09-21
FOCUS #146: Billed Unit Price metric spec and supporting content

ijurica:


rileyjenk:
@ijurica I put together a tab with your column on the standardized sample sheet.  see #146 BilledUnitPrice

specification/metrics/billedunitprice.md
```diff
@@ -0,0 +1,31 @@
+# Billed Unit Price
+
+The Billed Unit Price represents the unit price for a single pricing measurement unit, which incorporates various discounts like volume/tier-based rates, negotiated discounts and commitment-based discounts. Billed Unit Price does not amortize upfront charges (one-time or recurring). This unit price can match the List Unit Price, if there are no discounts in place. This pricing is denominated in the  [Billing Currency](#billingcurrency). It is used to calculate Billed Cost and savings based on financial optimization activities.
```
udam-f2, line 3: Should we mention that this is a generated unit price added for convenience - since there's no published price like this and the providers outside of AWS may not even have this consistently. 

Should we also say billed unit prices may change over time as the benefits of negotiated discounts and commitment based discounts application priority changes to make it clear why this rate may jump around drastically (e.g. might be $0.00 one hour and be something much more expensive in another hour if someone else got the RI)? 

Thinking about if we would avoid adding this similar to effective unit price rate. Both seem to be cake not ingredient. 

Also, should we avoid implying this is used to calculate Billed Cost (and instead say the other way around if we're keeping this).

(As I make these suggestions, I fully recognize I wrote some of this.. :( so, sorry for not bringing this up earlier)

supporting_content/metrics/billedunitprice.md
```diff
@@ -0,0 +1,67 @@
...
+
+| Provider  | Data set                     | Column                   |
+|:----------|:-----------------------------|:-------------------------|
+| AWS       | CUR;                         | Use lineItem/NetUnblendedRate if available, otherwise ​​lineItem/UnblendedRate. |
+| GCP       | BigQuery Billing Export;<br>BigQuery Pricing Data Export;<br>Cloud Billing API | Not available in BigQuery Billing Export<br>**price.effective_price** (available only in detail export) is the closest match for a subset of resources/services; doesn't include all relevant discounts/credits.<br>*see Discussion / Scratch space - GCP column mappings related issues for more details* |
+| Microsoft | Cost details;<br>*TODO: Add reference to Custom Azure Prices REST API* | **unitPrice** is probably the best match, but according to  documentation it's available only for EA and PAYG. <br>*see Discussion / Scratch space - Microsoft column mappings related issues for more details* |
+| OCI       | Cost and Usage Report        | **cost/unitPrice** is the best match<br>*see Discussion / Scratch space - OCI column mappings related issues for more details* |
```
udam-f2, line 12: In these cases, are we sure these are the unit prices inclusive of all the discounts (mainly including commitment discounts?)

specification/metrics/billedunitprice.md
ljadvey, line 3:
```suggestion
The Billed Unit Price represents the unit price for a single pricing measurement unit, which incorporates various discounts like volume/tier-based rates, negotiated discounts and commitment-based discounts. Billed Unit Price does not amortize upfront charges (one-time or recurring). This unit price can match the List Unit Price, if there are no discounts in place. This pricing is denominated in the [Billing Currency](#billingcurrency). It is used to calculate Billed Cost and savings based on financial optimization activities.
```

ljadvey: Worth a discussion!

ljadvey: Minor double space (OCD)

supporting_content/metrics/billedunitprice.md
ljadvey: Inclusive of discounts but not amortized

flanakin (changes requested):

specification/metrics/billedunitprice.md
```diff
@@ -0,0 +1,31 @@
+# Billed Unit Price
+
+The Billed Unit Price represents the unit price for a single pricing measurement unit, which incorporates various discounts like volume/tier-based rates, negotiated discounts and commitment-based discounts. Billed Unit Price does not amortize upfront charges (one-time or recurring). This unit price can match the List Unit Price, if there are no discounts in place. This pricing is denominated in the [Billing Currency](#billingcurrency). It is used to calculate Billed Cost and savings based on financial optimization activities.
+
+The BilledUnitPrice column MUST be present in the billing data. This column MUST be a numeric value of type Decimal within the range of non-negative decimal values. It MUST NOT contain null values in cases where the [ChargeType](#chargetype) is 'usage' or 'purchase'. When BilledUnitPrice is not null, multiplying BilledUnitPrice by [QuantityInPricingUnit](#quantityinpricingunit) MUST equal [BilledCost](#billedcost).
```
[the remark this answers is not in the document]
ijurica: We are currently reviewing/fine-tuning normative paragraphs for all Pricing supporting columns.
I'll make sure these comments are copied in that GDoc so we don't forget them.

A few notes/questions regarding the BilledUnitPrice and pricesheet/invoice:
- It MUST match a price in a custom price sheet if there are no commitment-based discounts?
- Could someone verify the content of their invoices? The ones I've reviewed or have access to either contain aggregated data or don't provide the BilledUnitPrice even though they could (based on the aggregation level in that particular case).
- Just to confirm, FOCUS doesn't/can not interfere with invoice content, right?

PR #179, opened 2023-09-23
FOCUS #169: Pricing Unit Dimension

### LINKED ISSUE #169: Create a Pricing Unit dimension
Normalized: No

- Notes:
  - represents a specific measurement increment designated by the provider to determine the cost of utilizing a particular resource or service 
  - forms the basis upon which cost calculations are built

### Definition of done

- [ ] Rationalize vendor-neutral, cross-cloud naming
- [ ] Complete spec template and include naming (code name, display name), constraints, guidelines, compatibility with major providers etc.

JGrubb:


marcin-walkow:
Updated version to latest after weekly brainstorming session.
Ready for the review!

specification/dimensions/pricingunit.md
```diff
@@ -0,0 +1,31 @@
+# Pricing Unit
+
+The Pricing Unit represents a provider specified pricing measurement unit, which forms the foundation for determining unit prices of utilizing a particular resource and/or service. Common examples include the number of hours a compute appliance was running for a given period, the number of gigabytes * the number of hours (gigabyte-hours) for a storage appliance, or a cumulative count of requests for a network appliance or API service. For certain services providers can choose block pricing i.e., million requests as a base unit.
```
udam-f2, line 3:
```suggestion
The Pricing Unit is a measurement unit specified by the provider, serving as the basis for calculating unit prices associated with using a specific resource and/or service. Examples include the duration a compute appliance is active, the product of gigabytes and hours (gigabyte-hours) for a storage appliance, or the total count of requests for a network appliance or API service. In some cases, providers may opt for block pricing, such as pricing per million requests as the pricing unit. The Pricing Unit, in conjunction with [Quantity In Pricing Unit](#quantityinpricingunit), defines and describes the measurement method for determining costs.
```

specification/dimensions/pricingunit.md
```diff
@@ -0,0 +1,31 @@
...
+
+PricingUnit
+
+## Display name
+
+Pricing Unit
+
+## Description
+
+Represents a provider specified pricing measurement unit, which forms the foundation for determining unit prices of utilizing a particular resource and/or service.
```
udam-f2, line 18:
```suggestion
A measurement unit specified by the provider, serving as the basis for calculating unit prices associated with using a specific resource and/or service.
```

supporting_content/dimensions/pricingunit.md
```diff
@@ -0,0 +1,46 @@
...
+- [Understand usage details fields - Microsoft Cost Management](https://learn.microsoft.com/en-us/azure/cost-management-billing/automate/understand-usage-details-fields)
+
+### OCI
+
+- [Cost and Usage Reports Overview](https://docs.oracle.com/en-us/iaas/Content/Billing/Concepts/usagereportsoverview.htm)
+
+## Discussion / Scratch space
+
+### Add into an appendix that describes pricing units, tiers, strategies
+The most basic math for calculating costs of cloud or SaaS services is Unit Price * Quantity = Cost.  Because different services require measuring different fundamental usage units while using a common billing format, the quantity column’s meaning or scale is unclear without specifying the unit of measure.
```
udam-f2, line 46: Add notes about what we discussed about normalizing / recommending (reasons for/against) and any other discussion surrounding this topic. Probably also worth mentioning how we view 'usage unit' vs pricing unit.

specification/dimensions/pricingunit.md
```diff
@@ -0,0 +1,31 @@
+# Pricing Unit
+
+The Pricing Unit represents a provider specified pricing measurement unit, which forms the foundation for determining unit prices of utilizing a particular resource and/or service. Common examples include the number of hours a compute appliance was running for a given period, the number of gigabytes * the number of hours (gigabyte-hours) for a storage appliance, or a cumulative count of requests for a network appliance or API service. For certain services providers can choose block pricing i.e., million requests as a base unit.
+PricingUnit is complementary to and describes how [QuantityInPricingUnit](#quantityinpricingunit) is measured.
+
+The PricingUnit column MUST be present in the billing data. This column MUST be of type String. It MUST NOT contain null if [QuantityInPricingUnit](#quantityinpricingunit) is not null. The PricingUnit value MUST match what is shown on the invoice if a pricing unit is present on the invoice.
```
udam-f2, line 6: Is there not a standard set of units (or a combination) we're recommending like the usage unit team is doing? 

cc @amitw-g @ijurica

udam-f2, line 6: Should we mention the difference between usage unit and pricing? Im not suggesting we do but want to make sure that was discussed and agreed on.

specification/dimensions/pricingunit.md
udam-f2: My suggestion only.. use all/parts if it makes sense to do so.

specification/dimensions/pricingunit.md
ijurica: I believe UsageQuantity + UsageUnit pair can also be described as a semi-normalized version of QuantityinPricingUnit + PricingUnit (not-normalized) pair.  I agree, we should mention that in the supporting content. As for the definition, phrasing this is a challenge :(.  I suggest we get back to that at a later stage, once names and definitions of UsageUnit and Quantity are finalized.

Few more notes about the PricingUnit...

**For AWS, Microsoft, GCP:**

PricingUnit comes down to mapped/renamed Provider's column

**For OCI:**

There is a column in billing data which we could use but it might be better to fetch the corresponding value from Public Pricing API (we already plan to use it for ListUnitPrice). 
Example:

- values in billing data: `ONE GiB HOURS STORAGE_SIZE`, `ONE GiB HOURS MEMORY`
- corresponding value in Public Pricing API: `Gigabyte Per Hour`

flanakin: Would that be an appendix? (Asking not stating 🙂)

specification/dimensions/pricingunit.md
flanakin: I would love to avoid the term "basis". It's jargon that most people don't understand based on my conversations with customers outside of Finance.

flanakin (approved):

udam-f2:
@JGrubb @marcin-walkow What are the next steps here? Can we get this closed out by end of week?

specification/dimensions/pricingunit.md
JGrubb: I like this.  I'm wondering how serious @flanakin is about the word "basis" vs "foundation", but I'm inclined to accept this as is and then we can discuss that part.  Thoughts?

specification/dimensions/pricingunit.md
marcin-walkow: As a non-finance+non-native I would say that _basis_ is as understandable as _foundation_ in this context. 
@flanakin / @udam-f2 - maybe we should ship version 1.0 and eventually rethink such wordings after wider community feedback?

specification/dimensions/pricingunit.md
marcin-walkow: Sounds very robust, the wording is clear and the main message stays the same 👍

specification/dimensions/pricingunit.md
marcin-walkow: TL;DR
If we decide to keep comments regarding PricingUnit in UsageUnit spec then we should do the other way around here as well.

I think as PricingUnit is complementary we literally had to put this information there. For considerations and relations to non-complementary units we may want to stick to appendix BUT as the UsageUnit contributors are commenting on PricingUnit we can add those comments.

specification/dimensions/pricingunit.md
marcin-walkow: @udam-f2 I had a quick chat with @flanakin and Michael mentions that basis can be widely unknown in a non-finance world based on his past experience.
What do you think about such small tweak?
**_serving as the foundation (aka basis) for calculating unit prices_**

udam-f2: I'm ok with foundation.

specification/dimensions/pricingunit.md
marcin-walkow: We have discussed that it may be very hard to ship 1.0 with standard set of units, my believe is that we should keep examples as in the definition to get the idea of what is this column referring to. Later on we will see what is the value of those being standardised

specification/dimensions/pricingunit.md
```diff
@@ -0,0 +1,31 @@
+# Pricing Unit
+
+The Pricing Unit is a measurement unit specified by the provider, serving as the foundation (a.k.a. basis) for calculating unit prices associated with using a specific resource and/or service. Common examples include the number of hours a compute appliance was running for a given period, the number of gigabytes * the number of hours (gigabyte-hours) for a storage appliance, or a cumulative count of requests for a network appliance or API service. For certain services providers can choose block pricing i.e., million requests as a base unit.
+PricingUnit is complementary to and describes how [QuantityInPricingUnit](#quantityinpricingunit) is measured.
+
```
marcin-walkow, line 5:
```suggestion
The Pricing Unit for a particular line item may differ from the published Usage Unit (and complementary Quantity in Usage Unit) when providers use different units, unit increments, or columns for calculating cost. 
```

specification/dimensions/pricingunit.md
ijurica: I'd like to challenge the term `calculating` :). I feel like `determining` is a better fit in this context.

specification/dimensions/pricingunit.md
ijurica: I've added a remark regarding the UsageUnit/PricingUnit within Usage Unit PR as well. This seems a bit unclear and implies that there are UCs in which providers use different columns for cost calculation?!

Suggestions:
- focus on differing purposes (primary distinction?!) 
- avoid mentioning that the values may differ and when that may be the case as covering those UCs seams a bit challenging

We might consider something like this: 
_While PricingUnit and QuantityInPricing may appear to relate to resource and/or service consumption, their primary focus is on pricing and cost, so it's essential not to confuse them with UsageUnit and QuantityInUsageUnit, which are dedicated to tracking actual resource and service consumption._

Rephrase? Simplify? Definitely!

specification/dimensions/pricingunit.md
marcin-walkow: @ijurica - you are referring to the fact that `Pricing Unit` is more of an information than a `p * q = c` dimension?

ijurica: Costs are usually calculated and unit prices determined/specified therefore, in this particular sentence, I would prefer the term **determining** over **calculating**. 

_The Pricing Unit is a measurement unit specified by the provider, serving as the basis for **calculating** unit prices associated with using a specific resource and/or service._

specification/dimensions/pricingunit.md
tobrien: We have a set of recommended units, but the UsageUnit was transitioned from a "MUST use Allowed Values" to "SHOULD use recommended values" - maybe we should move that part into an appendix and refer to it from both?

specification/dimensions/pricingunit.md
marcin-walkow: Aligned during the call that we should determining 👍

specification/dimensions/pricingunit.md
ijurica: Discussed at a meeting and agreed to emphasize the difference in purpose rather than mentioning that UsageUnit and PricingUnit may differ and in which cases. A similar note or remark will be included in the definition of all four columns - PricingUnit, QuantityInPricingUnit, UsageUnit, QuantityInPricingUnit.

specification/dimensions/pricingunit.md
```diff
@@ -0,0 +1,30 @@
+# Pricing Unit
+
+The Pricing Unit is a measurement unit specified by the provider for determining unit prices associated with the usage or the purchase of a specific resource and/or service. Common examples include the number of hours a compute appliance was running for a given period, the number of gigabytes * the number of hours (gigabyte-hours) for a storage appliance, or a cumulative count of requests for a network appliance or API service. For certain services providers can choose block pricing i.e., million requests as a base unit. PricingUnit is complementary to and describes how [QuantityInPricingUnit](#quantityinpricingunit) is measured. While PricingUnit may appear to relate to resource and/or service consumption, PricingUnit is primarily focused on pricing and cost. It is essential not to confuse PricingUnit with [UsageUnit](#usageunit), which is dedicated to tracking resource and/or service consumption and is often at a lower granularity than PricingUnit.
```
ijurica, line 3:
```suggestion
The Pricing Unit is a measurement unit specified by the provider for determining unit prices associated with the usage or the purchase of a specific resource and/or service. Common examples include the number of hours a compute appliance was running for a given period, the number of gigabytes * the number of hours (gigabyte-hours) for a storage appliance, or a cumulative count of requests for a network appliance or API service. For certain services providers can choose block pricing i.e., million requests as a base unit. PricingUnit is complementary to and describes how [QuantityInPricingUnit](#quantityinpricingunit) is measured. The PricingUnit is primarily focused on pricing and cost. It should not be confused with [UsageUnit](#usageunit), which is dedicated to tracking resource and/or service consumption and is often at a lower granularity than the PricingUnit.
```

specification/dimensions/pricingunit.md
```diff
@@ -0,0 +1,30 @@
...
+
+The PricingUnit column MUST be present in the billing data. This column MUST be of type String. It MUST NOT contain null if [QuantityInPricingUnit](#quantityinpricingunit) is not null. The PricingUnit value MUST match what is shown on the invoice if a pricing unit is present on the invoice.
```
ijurica, line 5:
```suggestion
The PricingUnit column MUST be present in the billing data. This column MUST be of type String. It MUST NOT contain null if [QuantityInPricingUnit](#quantityinpricingunit) is not null. The PricingUnit MUST match the corresponding value published in the price list, and if present on the invoice, it MUST also match the value on the invoice.
```

ijurica, line 5:
```suggestion
The PricingUnit column MUST be present in the billing data. This column MUST be of type String. It MUST NOT contain null if [QuantityInPricingUnit](#quantityinpricingunit) is not null. The PricingUnit MUST match the corresponding value published in the price list. If the level of detail provided on invoice includes pricing units, PricingUnit values must be consistent in both the billing data and the invoice for the same billing period.
```

supporting_content/dimensions/pricingunit.md
ijurica: - notes regarding non-normalized vs normalized and future plans provided within supporting content
- inclusion of PricingUnit - UsageUnit cross-references in progress

specification/dimensions/pricingunit.md
ijurica: - notes regarding non-normalized vs normalized and future plans provided within supporting content
- inclusion of PricingUnit - UsageUnit cross-references in progress
- current values observed in billing data for OCI provided

specification/dimensions/pricingunit.md
ijurica: moving suggestion to another conversation which encapsulates additional suggestions for improvements

specification/dimensions/pricingunit.md
```diff
@@ -0,0 +1,33 @@
+# Pricing Unit
+
+The Pricing Unit is a measurement unit specified by the provider for determining unit prices associated with the usage or the purchase of a specific resource and/or service. Common examples include the number of hours a compute appliance was running for a given period, the number of gigabytes * the number of hours (gigabyte-hours) for a storage appliance, or a cumulative count of requests for a network appliance or API service. For certain services providers can choose block pricing i.e., million requests as a base unit. Pricing Unit is complementary to and describes how [Quantity In Pricing Unit](#quantityinpricingunit) is measured. The Pricing Unit is primarily focused on pricing and cost and it should not be confused with the [Usage Unit](#usageunit), which is dedicated to tracking resource and/or service consumption and is often at a lower granularity than the Pricing Unit.
```
ijurica, line 3: - Improve the PricingUnit and UsageUnit definitions, especially the PricingUnit - UsageUnit cross-references by incorporating the following suggestions: 
  - _The Pricing Unit indicates how a provider rates the measured usage and purchase quantities after pricing rules are applied. It may differ from the physical measurement unit or Usage Unit due to pricing rules, such as block pricing. The Pricing Unit is commonly used for scenarios like verifying unit prices against the price sheet and should not be directly correlated with the physical resource or service usage._
  - _The Usage Unit indicates how service usage and purchases are measured by the provider. The Usage Unit is a standard measure of physical quantity before the pricing has been applied. The Usage Unit describes and is used in conjunction with Quantity In Usage Unit to gain an understanding of how much of a given product or service was used or purchased. Usage Unit may be a lower-level granularity than Pricing Unit, which applies provider-specific pricing rules. The Usage Unit is commonly used for scenarios like verifying how much of a resource or service was used with external metrics and logging systems to optimize usage patterns and should not be used to correlate directly to the unit price._
- Ensure that the definitions of related columns (PricingUnit, QuantityInPricingUnit, UsageUnit, and QuantityInUsageUnit) are aligned.
- Discuss these changes with both VS1 and VS3 team members.

udam-f2, line 3:
```suggestion
The Pricing Unit is the provider-specified measurement unit for determining the unit prices associated with the use or purchase of a specific resource or service. Examples include measuring the hours a compute appliance runs, calculating gigabyte-hours for storage appliances, or tallying requests for network appliances or API services. In some cases, providers may opt for block pricing, such as using a million requests as a foundational unit. The Pricing Unit complements and describes how the measurement of [Quantity In Pricing Unit](#quantityinpricingunit) is conducted. The Pricing Unit pertains to pricing and cost determination and should not be confused with the [Usage Unit](#usageunit), which tracks resource and service consumption, typically at a finer granularity than the Pricing Unit.
```

udam-f2: Use if more concise. Credit ChatGPT :)

udam-f2 (approved):

specification/dimensions/pricingunit.md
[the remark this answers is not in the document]
ijurica: Taking into account the original version and the suggestions, this is what I'm currently voting for

marcin-walkow: We took some bits of those and commited pre-maintainer version to the repo

marcin-walkow: First 2 points covered, discussion was taken and maintainers reviews should cover the "discussion" part

marcin-walkow:
For later, not a blocker for maintainers review:
_Discuss these changes with both VS1 and VS3 team members._

ijurica (approved):

ghost (approved):

ljadvey (approved):

rileyjenk (approved):

AndersonOliveira2022 (approved):

specification/dimensions/pricingunit.md
```diff
@@ -0,0 +1,33 @@
+# Pricing Unit
+
+The Pricing Unit is the provider-specified measurement unit for determining the unit prices associated with the use or purchase of a specific resource and/or service. The Pricing Unit indicates how a provider rates the measured usage and purchase quantities after pricing rules, such as block pricing, are applied. Common examples include the number of hours for compute appliance runtime, gigabyte-hours for a storage appliance, or an accumulated count of requests for a network appliance or API service. In some cases, providers may opt for block pricing, such as using a million requests as the Pricing Unit instead of pricing individual requests. The Pricing Unit complements and describes how the measurement of [Quantity In Pricing Unit](#quantityinpricingunit) is conducted. The Pricing Unit pertains to pricing and cost determination and should not be confused with the [Usage Unit](#usageunit), which tracks resource and service consumption, typically at a finer granularity than the Pricing Unit.
```
udam-f2, line 3: Looking to see if there's a way to make the description more concise. The second sentence seems unnecessary since we describe the block pricing provision a little further down anyway. 

```suggestion
The Pricing Unit is the provider-specified measurement unit for determining the unit prices associated with the use or purchase of a specific resource and/or service. Common examples include the number of hours for compute appliance runtime, gigabyte-hours for a storage appliance, or an accumulated count of requests for a network appliance or API service. In some cases, providers may use block pricing, like a million requests as the Pricing Unit. Pricing Unit complements the [Quantity In Pricing Unit](#quantityinpricingunit) measurement. Distinct from the [Usage Unit](#usageunit), it focuses on pricing, not resource and service consumption, often at a coarser granularity.
```

ijurica, line 3:
```suggestion
The Pricing Unit is the provider-specified measurement unit for determining unit prices, indicating how a provider rates measured usage and purchase quantities after applying pricing rules like block pricing. Common examples include the number of hours for compute appliance runtime, gigabyte-hours for a storage appliance, or an accumulated count of requests for a network appliance or API service. Pricing Unit complements the [Quantity In Pricing Unit](#quantityinpricingunit) measurement. Distinct from the [Usage Unit](#usageunit), it focuses on pricing, not resource and service consumption, often at a coarser granularity.
```

ijurica: Similar to the [previous suggestion](https://github.com/FinOps-Open-Cost-and-Usage-Spec/FOCUS_Spec/pull/179#discussion_r1382305221). The main/only difference: referring to block pricing as a pricing rule rather than mentioning it via an example.

ijurica: Addressed in [a subsequent conversation](https://github.com/FinOps-Open-Cost-and-Usage-Spec/FOCUS_Spec/pull/179#discussion_r1382309783)

ijurica: introduced a minor modification to align with the definition of QuantityInPricingUnit

ljadvey (approved):

udam-f2 (approved):

specification/dimensions/pricingunit.md
```diff
@@ -0,0 +1,33 @@
+# Pricing Unit
```
flanakin, line 1: There have been conversations about having a semi/normalized pricing unit post-1.0 with a separate block size. If we do that, I'm assuming we'll want to keep PricingUnit and UsageUnit aligned in naming and values. Can we rename this to `PricingMeasurement` to keep the provider-specific value separate?

ijurica: Discussed at the meeting and decided to rename the **PricingUnit** column to **PricingMeasure** (non-normalized).

marcin-walkow:
As @ijurica earlier mentioned:
_Discussed at the meeting and decided to rename the PricingUnit column to PricingMeasure (non-normalized). Reasoning will be provided in the supporting content._

specification/dimensions/pricingmeasure.md
```diff
@@ -0,0 +1,33 @@
+# Pricing Measure
+
+The Pricing Measure is a provider-specified measurement unit for determining unit prices, indicating how a provider rates measured usage and purchase quantities after applying pricing rules like block pricing. Common examples include the number of hours for compute appliance runtime, gigabyte-hours for a storage appliance, or an accumulated count of requests for a network appliance or API service. Pricing Measure complements the [Pricing Quantity](#pricingquantity) measurement. Distinct from the [Usage Unit](#usageunit), it focuses on pricing and cost, not resource and service consumption, often at a coarser granularity.
+
+The PricingMeasure column MUST be present in the billing data. This column MUST be of type String. It MUST NOT contain null if PricingQuantity is not null. The PricingMeasure value MUST be aligned with the corresponding pricing measure value provided in:
```
aqu-erp, line 5:
```suggestion
The PricingMeasure column MUST be present in the billing data. This column MUST be of type String. It MUST NOT be null if PricingQuantity is not null. The PricingMeasure value MUST be aligned with the corresponding pricing measure value provided in:
```

supporting_content/dimensions/pricingmeasure.md
```diff
@@ -0,0 +1,62 @@
...
+
+### OCI
+
+* [Cost and Usage Reports Overview](https://docs.oracle.com/en-us/iaas/Content/Billing/Concepts/usagereportsoverview.htm)
+
+## Discussion / Scratch space
+
+### PricingMeasure - non-normalized vs normalized
+
+* After several discussions the team has decided to use **PricingMeasure (non-normalized)** instead of earlier proposed PricingUnit (semi-normalized).
```
salpaysenturus, line 48: I won't hold up the approval of this dimension with this comment, but I do want to be on record recommending against this recent name change.

In reporting and semantic models, "Measure" means something quite different than "Measurement" or "Unit of Measure".  A measure is an aggregation (e.g. sum, average, first, last) of something, typically a number, whereas a measurement or UoM is an attribute which we might slice a measure.

The leading semantic model platforms are quite aligned on the use of the term.  See the following:

[Power BI](https://learn.microsoft.com/en-us/power-bi/transform-model/desktop-measures)
[Tableau](https://help.tableau.com/current/pro/desktop/en-us/datafields_typesandroles.htm)
[Cognos](https://www.ibm.com/docs/en/cognos-analytics/11.1.0?topic=dimensions-creating-measure)
[Qlik](https://help.qlik.com/en-US/cloud-services/Subsystems/Hub/Content/Sense_Hub/Measures/measures.htm)

The name "Pricing Measure" implies that the values contained therein will include something about measures associated with pricing.  The end user will have to remember that this is not the case, and I assume that a non-trivial portion of FOCUS consumers will end up changing this to "Measurement" or "Unit" or "UoM" or "Unit of Measure" in their data models for the purposes of self-documentation anyway.  As a FinOps data modeler, I can assure you that I personally will.

Obviously this name was discussed and agreed, so I won't stop the train from leaving the station.  But I am certain that this name will cause some amount of confusion down the road.

ijurica: For the similar reason, it took me a while to get accustomed to the way we use the term 'Attribute'. :)

salpaysenturus: I absolutely plan to advocate for changing the use of the terms "dimension" and "attribute" post 1.0 to align more with common data modeling parlance... but that's a debate for another day!

supporting_content/dimensions/pricingmeasure.md
```diff
@@ -0,0 +1,62 @@
...
+|-----------|-------------------------|--------------------------------------------------------------------------------------------------------------------------------------|
+| AWS       | CUR                     | hours; Hrs; Queries; GB; Secrets; vCPU-Hours; API Requests; Keys; Alarms etc.                                                        |
+| GCP       | BigQuery Billing Export | tebibyte; count; gibibyte; tebibyte; hour; gibibyte hour; gibibyte month; gibibyte hour; gibibyte; hour; count; gibibyte month; etc. |
+| Microsoft | Cost details            | 1 Hour; 10 Hours; 1/Month; 1 GB; 10K; 1 GB-Month; etc.                                                                               |
+| OCI       | Cost and Usage Report   | ONE GiB HOURS STORAGE_SIZE; ONE GiB HOURS MEMORY; etc.                                                                               |
+| OCI       | List Pricing REST API   | Gigabyte Per Hour; etc.<br>*Note: Gigabyte Per Hour corresponds to both ONE GiB HOURS STORAGE_SIZE and ONE GiB HOURS MEMORY*         |
```
salpaysenturus, line 24: Both Cost and Usage Report and List Pricing REST API are listed on line 12... Is only the CUR here on line 24 on purpose...?

supporting_content/dimensions/pricingmeasure.md
mike-finopsorg: Just putting it out there as an option to consider. If we wanted to avoid conflict of terms (and I believe we should) `PricingMeasurementUnit` is an option that while is a long name, it feels like a good descriptor for the content and avoids this conflict. Also happy if everyone wants to proceed with `PricingMeasure` as the internet (Chat-GPT) also offers `“measure” refers to a standard unit used to express size, quantity, or degree`

udam-f2: Agree that Measure is confusing (used interchangeably with Metric for data analysis). Measurement is even more confusing. 

ChatGPT says: 
"Measure" and "measurement" are related terms, but they have distinct meanings:

Measure:

"Measure" can function as both a verb and a noun.
As a verb, it means to ascertain the size, dimensions, quantity, or degree of something. For example, you might measure the length of a table.
As a noun, "measure" refers to a standard unit used to express size, quantity, or degree. In this sense, it is a reference to a specific dimension or extent. For example, meters or feet are measures of length.
Measurement:

"Measurement" is the process or result of quantifying or determining the extent, size, quantity, or capacity of something using a standardized scale or unit.
It is the act of assigning a number or value to a characteristic or property of an object or phenomenon.
Measurements are typically expressed using specific units, such as inches, kilograms, seconds, etc.
In summary, "measure" is the action of determining the size or quantity, while "measurement" refers to the result or the process of quantifying, often involving the use of standardized units. "Measure" is more action-oriented, while "measurement" is the outcome or the concept of quantification.

Disagree with assessment of dimensions. While in data modeling dimension might mean that, in data analysis, its commonly used to describe qualitative columns that enhance your data and can be used for grouping/filtering etc. Let's revisit post v1.0 :)

supporting_content/dimensions/pricingmeasure.md
ijurica: The idea was to highlight that in the case of **OCI**, the values in **billing data** and the **Pricing API** are **aligned but not identical**. (The values in the Pricing API (an alternative OCI data source) represent the exact match.) 

We can use `cost/billingUnitReadable` as a **close match**, but considering that we already need the same API for ListUnitPrice values, we can also retrieve the **exact match** values for PricingMeasure from the **Pricing API**.

supporting_content/dimensions/pricingmeasure.md
salpaysenturus: I respect the fact that ChatGPT thinks "measure" is a possible means of describing a unit in a vacuum, but I will reiterate that every major BI tool that will inevitably be used to house and present FOCUS data has a precise, non-unit-related definition of that specific term.

I prefer `PricingMeasurementUnit` to `PricingMeasure`.  (I do have a separate concern about reserving the obvious choice of `PricingUnit` for a more normative column that will sit alongside this extant column in the future, where the names will not self-document the difference.)  That said, I would prefer any name that sidesteps the specific word "measure".

supporting_content/dimensions/pricingmeasure.md
salpaysenturus: Okay!  Given the verbiage in line 12, I would expect to see an example of said "close match" in line 24, but it's not a showstopper.  Feel free to resolve if it works for you as-is.

ijurica: - In line 23 there are two 'close match' values (values from the OCI cost and usage report) 
- In line 24 there is the exact match, i.e. what we would expect to get (value from the Pricing API).

supporting_content/dimensions/pricingmeasure.md
ijurica: Something to consider...

- NON-NORMALIZED
  - **PricingUnit** (currently named PricingMeasure)
  
- SEMI-NORMALIZED
  - **UsageUnitCategory** (currently named UsageUnit)
  - **PricingUnitCategory** (to be introduced in 1,0+)

supporting_content/dimensions/pricingmeasure.md
salpaysenturus: Geez.  I'm sorry Irena... I don't know how I missed that.  Sorry to waste your time... amateur hour over here.

supporting_content/dimensions/pricingmeasure.md
salpaysenturus: For me, that (or something like it) would work well, assuming that the list of values in the semi-normalized columns group the values in the non-normalized equivalents.

specification/dimensions/pricingmeasure.md
aqu-erp: Edit: 

- "be null" for consistency 
- if we impose a `MUST` condition, I believe the action should also be concrete. Align is fuzzy, hence "match" , which should be more clear as an enforceable condition.

supporting_content/dimensions/pricingmeasure.md
flanakin: In an effort to get something in, let's go with `PricingMeasurementUnit` for now and continue the discussion in Slack. I suspect there's a better name, but also feel it's best to get something in while we brainstorm.

The current thinking is to create `PricingUnit` as a semi-normalized column like `UsageUnit` post-1.0 and the chances are it would be used more often than the provider-specific string (`PricingMeasurementUnit`), hence trying to stay away from that one.

kk09v (approved):

specification/dimensions/pricingmeasure.md
ijurica: - Replaced `contain` with `be` -> _It MUST NOT be null if PricingQuantity is not null._
- Originally used `match` but replaced it with `be aligned with` on purpose. 
  - The reasoning behind this is that it seems those values are sometimes aligned but don't exactly match (Pricing APIs, billing data, invoices). 
  - Note that Pricing APIs and Invoices are out of our scope.

@aqu-erp Please let me know if I can close this conversation. Thanks!

aqu-erp: > An implementation of the [specification][] is not compliant if it fails to satisfy one or more of the "MUST"...

An alternative would be to use `SHOULD` and `match`. This way, it still prescribes "alignment" but overall spec compliance does not hinge on a fuzzy interpretation.

I'll defer to what maintainers or @udam-f2 thinks, we can resolve the conversation after.

ijurica: Here are the options:

- `MUST match`: Pricing APIs and Invoices are out of scope.
- `MUST be aligned with`: Fuzzy interpretation, verifiable by humans?
- `SHOULD match`: Compliant even if those are not aligned.
- `SHOULD match, and at least, MUST be aligned`: Sounds awkward

@udam-f2 @flanakin Need an executive decision 😄

supporting_content/dimensions/pricingmeasure.md
rileyjenk: @salpaysenturus  I do not agree that there is ubiquitous agreement on the use and application of "Measure" among the BI platforms out there. There are as many differences as there are similarities. Also this is introducing a term that is used less frequently than unit. Units of Measure sure. Unit Price is a very common term.

rileyjenk (changes requested):
The Rename of this dimension is going to lead to confusion.

specification/dimensions/pricingmeasure.md
flanakin: Hmm... I'm definitely in the "MUST" camp since "SHOULD" doesn't drive accountability. Beyond that, I'm nit-picking words. 

I suppose "be aligned with" might be subjective. "Yes, one is per 1GB and the other is 10GB, but they're both GB, so that makes them aligned." Not sure if that's a sellable argument or not 😜

"Match" is an equality statement, which is strong if there's a known text difference but if the issue is the string and not the meaning, would it make sense to say, "SHOULD be the same as and MUST semantically match"? I also thought about "quantitatively match", but I'm not sure if that's too much. There's probably a better adjective.

ijurica:
> The Rename of this dimension is going to lead to confusion.

@rileyjenk are you voting for **PricingMeasurementUnit** or?

supporting_content/dimensions/pricingmeasure.md
marc-perreaut: I am old-school so I prefer to use Wikipedia rather than ChatGPT:

From https://en.wikipedia.org/wiki/Unit_of_measurement:

> A unit of measurement is a definite magnitude of a quantity, defined and adopted by convention or by law, that is used as a standard for measurement of the same quantity.

So `PricingMeasurementUnit `or `PricingUnitOfMeasurement `would be better than `PricingMeasure`.

specification/dimensions/pricingmeasure.md
aqu-erp: > The PricingMeasure value MUST `semantically match` the corresponding pricing measure value provided in...

Probably sufficiently tackles the below scenario while retaining accountability, and sufficiently restricting definition of equivalence. 

### S3 charge, monthly invoice
| charge | price list | compliant |
-|-|-
1 GB | 1 GiB | TRUE |
1 GB | 1000 MB | FALSE
1 GB | 1024 MB | TRUE
1 GB | 953 MB | FALSE
1 GB | 1 GB-month | TRUE
1 GB | 1 GB-day | FALSE

ijurica: going with ` ...MUST be equal to the...`

@aqu-erp Please let me know if I can close this conversation. Thanks!

specification/dimensions/pricingmeasure.md
```diff
@@ -0,0 +1,33 @@
+# Pricing Measure
+
+The Pricing Measure is a provider-specified measurement unit for determining unit prices, indicating how a provider rates measured usage and purchase quantities after applying pricing rules like block pricing. Common examples include the number of hours for compute appliance runtime, gigabyte-hours for a storage appliance, or an accumulated count of requests for a network appliance or API service. Pricing Measure complements the [Pricing Quantity](#pricingquantity) measurement. Distinct from the [Usage Unit](#usageunit), it focuses on pricing and cost, not resource and service consumption, often at a coarser granularity.
+
+The PricingMeasure column MUST be present in the billing data. This column MUST be of type String. It MUST NOT be null if PricingQuantity is not null. The PricingMeasure value MUST be equal to the corresponding pricing measure value provided in:
```
ijurica, line 5:
```suggestion
The PricingMeasure column MUST be present in the billing data. This column MUST be of type String. It MUST NOT be null if PricingQuantity is not null. The PricingMeasure value MUST be equal to the corresponding pricing measurement unit value provided in:
```

ijurica:
After deliberation, the group decided to opt for semi-normalized version of the dimension, adhering to the values and format requirements specified in the Unit Format attribute, similar to Usage Unit.

TODO:

- Rename and redefine the column
- Edit the PR title
- Etc.

aqu-erp (approved):

supporting_content/dimensions/pricingunit.md
```diff
@@ -0,0 +1,62 @@
...
+
+### OCI
+
+* [Cost and Usage Reports Overview](https://docs.oracle.com/en-us/iaas/Content/Billing/Concepts/usagereportsoverview.htm)
+
+## Discussion / Scratch space
+
+### PricingUnit - non-normalized vs normalized
+
+* After several discussions the team has decided to use **PricingMeasure (non-normalized)** instead of earlier proposed PricingUnit (semi-normalized).
```
aqu-erp, line 48:
```suggestion
* After several discussions the team has decided to use **PricingUnit (semi-normalized)** instead of earlier proposed PricingMeasure (non-normalized).
```

Some changes needed below as well to follow new decision.

aqu-erp (approved):

supporting_content/dimensions/pricingunit.md
ijurica: Of course, additional adjustments to the supporting content are coming soon. We tried to communicate that through the commit message. I will let you know when it's ready for review.

aqu-erp: Doh, that's on me for not expanding the commit message. Thanks @ijurica

ijurica: No problem, I actually don't read them either :). 
I thought you are all resting and wouldn't even notice that I've published a partial change :).

Now, it should be ready for the review.

udam-f2 (approved):

specification/dimensions/pricingunit.md
```diff
@@ -0,0 +1,35 @@
...
+
+The Pricing Unit represents a provider-specified measurement unit for determining unit prices, indicating how a provider rates measured usage and purchase quantities after applying pricing rules like block pricing. Common examples include the number of hours for compute appliance runtime (e.g. `Hours`), gigabyte-hours for a storage appliance (e.g., `GB-Hours`), or an accumulated count of requests for a network appliance or API service (e.g., `1000 Requests`). Pricing Unit complements the [Pricing Quantity](#pricingquantity) metric. Distinct from the [Usage Unit](#usageunit), it focuses on pricing and cost, not resource and service consumption, often at a coarser granularity.
```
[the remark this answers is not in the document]
ijurica: Already taken care of

specification/dimensions/pricingunit.md
```diff
@@ -0,0 +1,35 @@
+# Pricing Unit
+
+The Pricing Unit represents a provider-specified measurement unit for determining unit prices, indicating how a provider rates measured usage and purchase quantities considering pricing rules like block pricing. Common examples include the number of hours for compute appliance runtime (e.g. `Hours`), gigabyte-hours for a storage appliance (e.g., `GB-Hours`), or an accumulated count of requests for a network appliance or API service (e.g., `1000 Requests`). Pricing Unit complements the [Pricing Quantity](#pricingquantity) metric. Distinct from the [Usage Unit](#usageunit), it focuses on pricing and cost, not resource and service consumption, often at a coarser granularity.
```
flanakin, line 3: This uses an example with a block size, but I thought the Unit Format doesn't not want numbers in the value, right? Should we remove that example?

specification/dimensions/pricingunit.md
```diff
@@ -0,0 +1,35 @@
+# Pricing Unit
```
flanakin, line 1: I really like the alignment to Unit Format for consistency. My only request would be to also add PricingBlockSize as the unit without the block size isn't enough information to understand how pricing works.

flanakin (approved):

specification/dimensions/pricingunit.md
aqu-erp: Is a quantifier dimension eg `pricingBlockSize` in 1.0 or post-1.0?

udam-f2: Pricing Unit is logically the same as what providers have today with a recommendation to use consistent formatting specified in Unit Format. Block size can be a fast follow in v1.0+ since it's only necessary in a small number of (still important) use cases. Primary need is for providing a way for users to make sense of the rates/quantities/cost and that's accomplished by this column.

specification/dimensions/pricingunit.md
udam-f2: Block size can be a fast follow in v1.x since it's only necessary in a small number of (still important) use cases. Added here: [FOCUS Data Model](https://docs.google.com/spreadsheets/d/148Hdg1ws1Co3MDI65T_KtGCAzbtmFbzOU0pGvRnmbq8/edit#gid=1318248972) as 1.x

udam-f2:
> The Rename of this dimension is going to lead to confusion.

@rileyjenk we're back to Pricing Unit. What's the confusing part? It's now Pricing Unit as you suggested. Are you good? If so, can we resolve and push?

aqu-erp (approved):

rileyjenk (approved):

specification/dimensions/pricingunit.md
ijurica: One of the notes in the Discussion/Scratch space refers to some suggestions for 1.0+

specification/dimensions/pricingunit.md
ijurica: One of the notes in the Discussion/Scratch space refers to some suggestions for 1.0+

supporting_content/dimensions/pricingmeasure.md
ijurica: opted for semi-normalized PricingUnit

Issue #152, opened 2023-08-18
CommitmentUtilization column


flanakin:
Determine how unused commitments will be represented within the cost data. This may be in an existing column or a new column.

This column will be normalized so it is consistent across providers.

> ℹ️ _This column name is a placeholder and is not final._

udam-f2:
@flanakin If still relevant, could we please re-create this with the new feedback template?

flanakin:
This is completed with ChargeSubcategory for usage.

Issue #161, opened 2023-08-18
PricingModel column


flanakin:
Define a column to identify the type of rate discount applied to a charge. For instance, this would be used to differentiate between on-demand usage and committed usage/cost.

This column will be normalized so it is consistent across providers.

> ℹ️ _This column name is a placeholder and is not final._

udam-f2:
@flanakin Should we capture this in the new issue format and close this out?

flanakin:
Completed with PricingCategory column

PR #180, opened 2023-09-24
FOCUS #152: Commitment Utilization

chandra-cloudtrakr:
This is the first version of defining CommitmentUtilization to identify used and unused commitments. Please review and let me know your feedback.


specification/dimensions/usagetype.md
```diff
@@ -0,0 +1,42 @@
...
+| On-Demand         | This usage type describes the consumption of resources or services that are billed at the standard on-demand unit price.                                                |
+| Interruptible     | The interruptible usage type pertains to resources that are provisioned using spot availability. These resources are billed at a fluctuating spot price, which is typically lower than on-demand prices.                                                                                  |
+| Committed         | Committed usage refers to the consumption of resources that benefit from a discounted rate due to long-term commitments, such as reserved instances or savings plans. These resources are billed at this reduced rate. |
+| Unused Commitment | This usage type indicates a scenario where a customer has committed to a specific amount or capacity (hourly, daily, etc.) but hasn't fully utilized it. The line item reflects the capacity or monetary value that went unused.                                                                        |
```
rileyjenk, line 38: Could we not have spaces in the names?

specification/dimensions/usagetype.md
```diff
@@ -0,0 +1,42 @@
+# Usage Type
+
+UsageType is a categorization parameter that defines how a resource or service was consumed during a specified period. The UsageType can influence the price point and also indicate whether purchased commitments are used or unused. This classification offers a nuanced view into the resource or service's consumption patterns and helps in understanding how you're billed, especially when considered alongside attributes like ChargeType and PurchaseType.
```
marc-perreaut, line 3: > defines how a resource or service was consumed 
In case of unused commitment, there is no resource or service attached to this record (unless one considers the commitment is a resource, which is arguable), so referring a resource or service can be misleading. Maybe using the plural form works better: "defines how resources or services were used".

> during a specific period
Maybe "during the charge period" is more precise

> especially when considered alongside attributes like ChargeType and PurchaseType.
I understood that UsageType would be used to drilldown the type of the record in the case ChargeType=Usage. So UsageType would be populated (not empty, not null) if and only if ChargeType=Usage. In the same way, PurchaseType would be populated if and only if ChargeType=Purchase. Or would ChargeType, UsageType and PurchaseType work in another way?

specification/dimensions/usagetype.md
```diff
@@ -0,0 +1,42 @@
...
+| Column required | True                                     |
+| Data type       | String                                   |
+| Allows nulls    | False                                    |
+| Value format    | list-of-values                           |
+
+Allowed values:
+
+| Value      | Description                                                                                                                                                                   |
+|:-----------|:------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
+| On-Demand         | This usage type describes the consumption of resources or services that are billed at the standard on-demand unit price.                                                |
```
marc-perreaut, line 35: Maybe OnDemand (without -) is simpler?

How would we deal with Google Sustained Usage Discount?
Would it be another value for UsageType, or a particular case for UsageType=OnDemand. In the latter case, which dimension would bear this information?

specification/dimensions/usagetype.md
```diff
@@ -0,0 +1,42 @@
...
+| Interruptible     | The interruptible usage type pertains to resources that are provisioned using spot availability. These resources are billed at a fluctuating spot price, which is typically lower than on-demand prices.                                                                                  |
```
flanakin, line 36: Do we want to avoid the word "spot", since it's a product name?

specification/dimensions/usagetype.md
```diff
@@ -0,0 +1,42 @@
...
+
+Indicates whether the record pertains to an on-demand or recurring charge, or relates to the cost or capacity using a committed discount rate or from unutilized commitments.
+
+## Content Constraints
+
+| Constraint      | Value                                    |
+| :-------------- | :--------------------------------------- |
+| Column required | True                                     |
+| Data type       | String                                   |
+| Allows nulls    | False                                    |
```
flanakin, line 28: Will be null for purchases and adjustments
```suggestion
| Allows nulls    | True                                     |
```

flanakin (changes requested):
Can you also add supporting content with some high-level takeaways from our discussion last week?

specification/dimensions/usagetype.md
udam-f2: +1 on this being a spec-wide decision. Where it would run into issues is things like Service Category values where without spaces, its not going to be very user friendly.

specification/dimensions/usagetype.md
udam-f2: +1 on SUDs being volume pricing and not a discount.

specification/dimensions/usagetype.md
```diff
@@ -0,0 +1,42 @@
+# Usage Type
```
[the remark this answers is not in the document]
udam-f2: Strongly oppose this being called UsageType. Pricing model makes sense other than the fact that we have put a sub-property of 'unused commitment' as a first-level value for this column IMO. 

I would use a value like 'Committed' or 'Committed Usage' and then have a separate property to say its used vs unused. If we're against creating an explicit column, maybe this can be figured out based on the fact that the resource id is a commitment OR the resource type for the line is a commitment discount (as opposed to some other actual resource you would use - e.g. lambda, vm, etc.) to know that the purchased discount didn't apply to actual resources.

specification/dimensions/usagetype.md
nfonrose: I'm not so sure about that because it's not really volume. You can consume the same "744 core-hours" of compute by maximizing SUD or not depending on how your organize your workloads.

And Practionners work on how to optimize their SUD  in a similar way as they work on optimizing their commitment discounts :
 - Avoid spikes in instance numbers (if you spread your compute usage, you maximize your SUD)
 - Use consistent instance families (if you use N2 all around instead of "N1 here and N2 there", you maximize your SUD)
 
So i understand the point about considering it a "volume pricing" but there are also aspects which favor considering it as a discount like CUD or RIs.

specification/dimensions/usagetype.md
nfonrose: One of the advantages of "no space inside names" is that it makes searching for the terms easiers. And it turn these names into "keywords" rather than usual language, reducing the risk of ambiguities.

So it might be more user-friendly for non tech people to have spaces, but it makes life easier in several cases.

I think hashtags have popularized names with no spaces in the "general population".

specification/dimensions/usagetype.md
flanakin: @nfonrose, what value would you want to see for that? It's not a commitment. 

/cc @rodneyjoyce for `PricingModel` 🙂

specification/dimensions/usagetype.md
flanakin: I would move this to supporting content.

specification/dimensions/usagetype.md
marc-perreaut: Indeed, SUD could be managed with another dimension, but for simplicity of the FOCUS model, this information could be put in the same dimension. The value could be simply 'Sustained'.

specification/dimensions/usagetype.md
ljadvey: Agreed

specification/dimensions/usagetype.md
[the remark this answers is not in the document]
marc-perreaut: I would propose ChargeSubType:
- The 'Sub' pattern already exists with SubAccountId/Name
- The ChargeSubType dimension can be used to subcategorize other types of charges
- If find the pattern ChargeType/ChargeSubType simple to handle in visual reporting (for example PowerBI) for both filtering and presentation. For example, I love the columns the MeterCategory and MeterSubCategory in Azure billing data for this.

I would propose the same pattern for Service: ServiceName, SubServiceName.

specification/dimensions/usagetype.md
aqu-erp:
```suggestion
| Interruptible     | Interruptible usage type pertains to consumption of resources provisioned with a cloud provider's ability to stop or interrupt workloads at any time.                                                                       |
```

aqu-erp: Removed language referring to dynamic pricing, as it's not guaranteed to be true moving forward.

specification/dimensions/usagetype.md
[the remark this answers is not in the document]
flanakin: FYI - After some discussion, we decided to split this column into 2: `PricingModel` and `CommitmentUtilization` (name TBD), where this column will cover `Used` and `Not Used` capacity, spend, and usage commitments.

specification/dimensions/usagetype.md
flanakin: I'm going to resolve this here and propose that we bring this up outside of this column and discuss it holistically. We can then make a sweeping change to apply whatever we decide.

specification/dimensions/usagetype.md
flanakin: Keeping this comment active until the PricingModel PR comes out. Will update everyone here when that's ready.

specification/dimensions/usagetype.md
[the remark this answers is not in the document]
flanakin: @BenShy, can you make sure this is covered in the amortization handling attribute? 

@marc-perreaut, we generally only put the rules for the current column in the column specs. Ben will make sure this is covered in the amortization handling attribute. Do you think we should explicitly add it to the BilledCost column? My only concern would be that spreading the amortization logic in many columns might lead to confusion. Something to consider 🤔

specification/dimensions/usagetype.md
flanakin: Keeping this comment active until the PricingModel PR comes out. Will update here.
/cc @rodneyjoyce for PricingModel

specification/dimensions/usagetype.md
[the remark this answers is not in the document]
marc-perreaut: I am fine with the principle of putting rules only for the current column.
Still I find important that the specification says how commitments are handled overall, ie over different columns and over different rows. It is useful for both FOCUS data producers and consumers.
Maybe there could be a dedicated appendix about commitments?

specification/dimensions/usagetype.md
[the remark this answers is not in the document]
marc-perreaut: There might be a risk on dispatching the information over different columns for both user understanding and data filtering & presentation. I do not know the rationale of the split, but I actually miss the holistic view / while picture on how charges and commitments are organized / structured.

specification/dimensions/commitmentutilization.md
```diff
@@ -0,0 +1,37 @@
+# Commitment Utilization
+
+Commitment Utilization indicates whether a charge utilized a commitment during the charge period. Commitment Utilization applies to all commitments and commitment types applied to the current charge. This includes, but is not limited to capacity commitments and commitment-based disocunts represented by the Commitment Disocunt ID. Commitment Utilization is commonly used to identify wastage when a commitment has not been used and to calculate commitment utilization.
```
salpaysenturus, line 3: small typo: disocunts

salpaysenturus:
Does the structure as proposed here imply that every Commitment Discount will generate two rows per timeframe?  Alternatively, was there a discussion of instead adding two columns (something like Commitment Discount Used and Commitment Discount Unused), and keep each CD to one row per timeframe?  I recognize that that adds two columns of overhead to many line items that can't have Commitment Discounts -- but then makes it significantly easier to natively calculate both Utilization and Coverage across the entire dataset (when aggregating either by CD or Resource, respectively).  Depending on the organization, it may also result in a smaller dataset size overall.

If the proposed structure is already too far downstream to alter in so drastically a fashion, then I understand.  But this is how our organization elected to structure the data, to good effect.

specification/dimensions/commitmentutilization.md
```diff
@@ -0,0 +1,37 @@
+# Commitment Utilization
```
salpaysenturus, line 1: I am personally not in favor of the use of the term "Utilization" in this fashion.  To my mind, the term "Utilization" implies a crucial FinOps KPI that, as we all know, is a percentage measure that returns the result of Used / (Used + Not Used).  However, here, it is being used to describe a attribute that one could use in order to derive that calculation.  This introduces a semantic conflict later, if and when the FOCUS spec add "Utilization" to the list of Metrics (which I hope it does!).

I feel that the term "Utilized" or "Utilized Flag" would not only help delineate from that, but it would also be more self-documenting and imply that this is an attribute with two (and only two) possible values.

I also prefer the members of a flag attribute to follow the same pattern as the name of the attribute.  To that end, I propose the use of "Utilized" and "Not Utilized" instead of "Used" and "Not Used".

rodneyjoyce (commented):
Is there any Supporting Content for this PR @chandra-cloudtrakr? Would be good to see some examples across different clouds and links to resources

specification/dimensions/usagetype.md
rodneyjoyce: Yes, we do not refer to Spot as it is a Vendor term. We use interruptible (not a Proper Noun) in Pricing Model

specification/dimensions/usagetype.md
flanakin: Can everyone please go review the PricingModel PR and leave comments there for anything in this area?
See #214

specification/dimensions/commitmentutilization.md
flanakin:
```suggestion
Commitment Utilization indicates whether a charge utilized a commitment during the charge period. Commitment Utilization applies to all commitments and commitment types applied to the current charge. This includes, but is not limited to capacity commitments and commitment-based discounts represented by the Commitment Discount ID. Commitment Utilization is commonly used to identify wastage when a commitment has not been used and to calculate commitment utilization.
```

specification/dimensions/commitmentutilization.md
flanakin: This is the primary thing I was expecting to hear from people on this column 🙂 I love that "Utilization" binds it to the KPI, but it is a good point that there could possibly be confusion around the column vs. KPI.

A few other thoughts on names:
1. `CommitmentStatus` was our original name. We thought about "Status" for coverage calculation, but that would require separate columns per commitment type, so we scoped it out.
2. `CommitmentUtilized` sounds like a boolean (true/false). I'm personally not a huge fan of that word given it's not something we use often in English, but it is at least aligned to the KPI.
3. `CommitmentUsage` is another option I suppose. "Usage" is a more common term than "Utilization".

"UtilizationType" feels unnecessarily long.

I'll keep this open to get more feedback.

specification/dimensions/usagetype.md
flanakin: FYI - Here's the PricingModel PR. Please move any comments over there!
See #161

flanakin:
> Does the structure as proposed here imply that every Commitment Discount will generate two rows per timeframe? 
Not exactly. The discount will apply to the usage, so there are no extra rows to track usage. The only time there's an extra row not associated with usage is when commitments are not used.

> Alternatively, was there a discussion of instead adding two columns (something like Commitment Discount Used and Commitment Discount Unused) in order to keep each CD to one row per timeframe? 
Hopefully my comment above explains this. Commitment usage is tracked on the resource usage rows. This column will indicate whether a commitment was used or not. @BenShy is working on an Amortization handling attribute that will explain this in detail.

> I recognize that that adds two columns of overhead to many line items that can't have Commitment Discounts -- but then makes it significantly easier to natively calculate both Utilization and Coverage across the entire dataset (when aggregating either by CD or Resource, respectively).
FYI - Coverage cannot be calculated from a single column since not all services are available to all commitment types. We'd need separate columns to indicate whether certain usage is eligible for spend commitments separate from usage commitments separate from capacity commitments.

specification/dimensions/commitmentutilization.md
marc-perreaut: I agree CommitmentUtilization refers more to a Commitment KPI, and my preference goes to 3. CommitmentUsage.

Another possibility may be "UnusedCommitment" with values "Yes" and "No".

flanakin: Added your suggestion above so they're all together. 
Tracking 2 votes against CommitmentUtilization, so that sounds good.
What do others think? Should we do a poll or can we get consensus here?

specification/dimensions/usagetype.md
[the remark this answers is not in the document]
flanakin: FYI - Amortization Handling is here: #229

@marc-perreaut would love your eyes on that to see if we captured this adequately.

rodneyjoyce (commented):
lgtm- I am voting on the name in Slack

specification/dimensions/commitmentutilization.md
```diff
@@ -0,0 +1,37 @@
...
+
+## Description
+
+Indicates whether a charge utilized a commitment during the charge period.
+
+## Content Constraints
+
+| Constraint      | Value          |
+| :-------------- | :------------- |
+| Column required | True           |
```
joe-ferrero, line 23: @BenShy @flanakin Regarding Line 23

When an unused value is rolled over from one contract to a subsequent contract - is that an exception to the "MUST" criteria? I'm familiar with a pre-paid commitment with a 10% rollover clause; in practice - if the customer uses 92% of what they prepaid - the remaining 8% is added to the subsequent contract - but does NOT have to be purchased again.

This would get complicated in the scenario where a customer consumes 85% of what they pre-paid, rolls over 10% to a subsequent contract and has 5% that they have to add into their cost data.

Happy to work through scenarios if you believe it needs to be part of V1.0
```suggestion
| Column required | False |

**Review by joe-ferrero — APPROVED** [review]

File: specification/dimensions/commitmentutilization.md [hunk 10]
```diff
@@ -0,0 +1,37 @@
+# Commitment Utilization
+
+Commitment Utilization indicates whether a charge utilized a commitment during the charge period. Commitment Utilization applies to all commitments and commitment types applied to the current charge. This includes but is not limited to capacity commitments and commitment-based discounts represented by the Commitment Discount ID. Commitment Utilization is commonly used to identify wastage when a commitment has not been used and to calculate commitment utilization.
```
yprikhodko, line 3: Do we have definition of charge period? Maybe we should say billing period instead?

specification/dimensions/commitmentutilization.md
```diff
@@ -0,0 +1,37 @@
...
+## Content Constraints
+
+| Constraint      | Value          |
+| :-------------- | :------------- |
+| Column required | True           |
+| Data type       | String         |
+| Allows nulls    | True           |
+| Value format    | list-of-values |
+
+Allowed values:
```
yprikhodko, line 28: What is the real value of this column for practitioners especially when it's partially used? Wouldn't be more helpful to display in this column the unused portion of commitment so practitioners could easily calculate the 'waste' across all the commitments. And to display 0 if it's fully used/utilized.
What is suggested value in case of commitment is partially used?

specification/dimensions/commitmentutilization.md
TDubovchenko: Maybe `CommittmentUsed` or `CommitmentUtilized`?

specification/dimensions/commitmentutilization.md
TDubovchenko, line 3: **Commitment Utilization** is commonly used to identify wastage when a commitment has not been used and to calculate commitment utilization.
Commitment Utilization is used twice referring to different meaning

specification/dimensions/commitmentutilization.md
flanakin: Vote here: https://finopsfoundation.slack.com/archives/C053CNCRVCL/p1698191641650719

specification/dimensions/commitmentutilization.md
flanakin: See the ChargePeriodStart/End columns. The Charge Period is the granularity of the data, so a day or hour of usage in this case. Whatever the row is covering.

specification/dimensions/commitmentutilization.md
flanakin: The intent isn't to put a number here. It's to identify whether this row was for `Used` commitment or `Not Used` commitment. So if you want to calculate utilization, you can group by this column to get what was used vs. not used. You can also quantify waste by filtering on `Not Used`.

If that's not clear, we should tweak the text to make it clearer. We'll also cover it in the Discount Handling spec, but let's make sure it's clear here first. Any thoughts on how we could make it more obvious?

specification/dimensions/commitmentutilization.md
flanakin: Great catch! We're going to change the column name, so it's probably fine... Let's see how it looks when we update the column name.

specification/dimensions/commitmentutilization.md
yprikhodko: After getting into more details it's clear. Probably the name is a bit misleading which I believe is a subject for discussion anyway.

specification/dimensions/commitmentutilization.md
yprikhodko: Clear now, thanks

flowerinthenight:
Merged with [ChargeSubcategory PR](https://github.com/FinOps-Open-Cost-and-Usage-Spec/FOCUS_Spec/pull/218). It's probably better to add these information to that PR's supporting content for additional context.

udam-f2:
This is closed in favor of ChargeSubcategory

PR #187, opened 2023-10-03
Expanded the language to be a bit more precise on format, added language to recommended unit set

tobrien:
Couple of things to noticed in this commit:

1. Start in the Recommended Unit Set - Time Units are sorted from largest scope to smallest, everything has an abbreviation so I added one to count, I've also added two paragraphs explaining that capitalization is important (everything lowercase except for abbreviations where specified.)

2. Next I expanded and revised the language up at the topic of the section.  In particular, I added language to clarify what we mean by "It must, if possible."


tobrien:
Ugh, I have to close this pull requests - I used the wrong email address in this particular commit.

PR #189, opened 2023-10-04
Adding a paragraph to get explicit about numeric value and format acceptable in the usage quantity dimension

tobrien:
Adding language to clarify exactly what the number format is for the usage quantity dimension and making sure to specify that the value is non-negative.  Also, make sure that we specify that fractional values are not allowed.

### DIFF
--- a/specification/dimensions/usagequantity.md
+++ b/specification/dimensions/usagequantity.md
@@ -2,8 +2,12 @@
 
 Usage Quantity is the amount of resources or services consumed. The Usage Quantity is measured in units specified by the Usage Unit. Usage Quantity may differ from a Pricing Quantity when providers use a different unit of measure for calculating cost. Usage Quantity is often used when analyzing environments to gain understanding of usage or consumption of a given product/service in an environment.
 
-
 Usage Quantity MUST exist in billing data. The column MUST be a numeric value and MUST NOT contain nulls.
+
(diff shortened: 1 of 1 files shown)


specification/dimensions/usagequantity.md
```diff
@@ -2,9 +2,13 @@
...
 
-
 Usage Quantity MUST exist in billing data. The column MUST be a numeric value and MUST NOT contain nulls.
 
+The value of UsageQuantity MUST be a non-negative, real number with a single value greater than or equal to zero. The single numeric value in this column can be expressed as integer, decimal value, or scientific notation. Fractional notation or a value expressed with an arithmetic operator (e.g. 1/2) is not permitted.  Multiple values, arrays, or ranges are not permitted in the UsageUnit dimension. Numerical values expressed with mathematical symbol, operators, or exponent values not required for scientific notation are not permitted.
+
+* Examples of valid values for Usage Quantity are "0", "1", "1.5", "1.234", "1.23342 x 10^7"
+* Examples of invalid values for Usage Quantity are "-100", "1 1/2", "[3,5,8]", "[4:5]", "5i + 4", "sqrt(2)", "2.3^3"
```
amitw-g, line 10: I think we allowed negative usage values for corrections. Also the valid/invalid values should be covered by the Type constraints I think.

tobrien: Perfect, I'll take some of the numeric language and drop it in the numeric section of the appendix.

Issue #190, opened 2023-10-04
Numeric Format Attribute


tobrien:
### Type
Attribute

Normalized? Yes | No

### Description
Define format requirements for metrics that define Numeric type.

### Definition of done

- [X] Rationalize vendor-neutral, cross-cloud naming
- [ ] Complete spec template and include naming (code name, display name), constraints, guidelines, compatibility with major providers etc.

PR #192, opened 2023-10-04
FOCUS #190: Numeric Format Attribute

tobrien:
This Pull Request contains proposed language for the numeric format attribute.   Note that this also proposes the ability to define more granular numeric types.    

Numeric
Integer
Decimal

There is also an option to add a "Data Precision" constraint to a metric to define precision over a defined set or precision specifiers.


specification/attributes/numeric_format.md
```diff
@@ -0,0 +1,104 @@
+# Date/Time Format
```
udam-f2, line 1:
```suggestion
# Numeric Format
```

specification/attributes/numeric_format.md
```diff
@@ -0,0 +1,104 @@
...
+
+Columns that provide numeric values conforming to specified rules and formatting requirements ensure clarity, accuracy, and ease of interpretation for both humans and systems.
+
+All columns capturing a numeric value, defined in the FOCUS specification, MUST be a real number represented as an integer, a decimal value, or a value represented in scientific notation as defined in this section.
+
+## Attribute ID
```
udam-f2, line 7: I would consider putting Numeric Format, Numeric Precision and Examples above this line (maybe under sub headings (H3 or lower) so we don't mix up the descriptive stuff with the spec items that start with this Attribute ID header. Feels like Numeric Format being the first section and then precision and then examples make sense to me).

I think there is a normative sentence in the Precision section that we can probably move into the above normative section (line 5)

udam-f2 (commented):
Awesome to see this.. Thanks for putting the time and effort on this!

specification/attributes/numeric_format.md
```diff
@@ -0,0 +1,104 @@
...
+Numeric Format
+
+## Description
+
+Rules and formatting requirements for numeric columns appearing in billing data.
+
+## Requirements
+
+* Metrics with a Numeric format MUST contain a single value.
+* The single value captured in metrics with a Numeric format MUST be a real, rational.
```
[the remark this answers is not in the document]
ljadvey: Agree with this

specification/attributes/numeric_format.md
```diff
@@ -0,0 +1,104 @@
...
+* Numeric values MUST NOT be used to represent binary values (e.g. 0 or 1).
+
+## Examples of Numeric Values Meeting Requirements
+
+* Values Meeting Numeric Requirements:
+  * "-100.2"
+  * "0"
+  * "1"
+  * "1.234"
+  * "1.23342 x 10^7"
```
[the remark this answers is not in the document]
ljadvey: I read it as such. Double quotes shouldn’t be used.

specification/attributes/numeric_format.md
```diff
@@ -0,0 +1,104 @@
...
+  * "sqrt(2)" - contains a mathematical symbol or operations
+  * "2.3^3" - contains an exponent not related to scientific notation
+  * "32 GiB" - contains a unit of measure
+  * "$32" - contains a currency symbol
+  * "3,432,342" - contains a comma
+  * "+333" - contains a positive sign
+
+## Numeric Precision
+
+The FOCUS specification does not require a specific level of precision for numeric values.  The level of precision required for a given metric is determined by the provider unless specified in the metric specification.  The level of precision required for a given metric SHOULD be documented in the metric definition.  The following table outlines the options for specifying the level of precision for a given metric.
```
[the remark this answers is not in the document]
ljadvey: Difficult…. Prefer not to define our precision but following the above on REAL number.

specification/attributes/numeric_format.md
```diff
@@ -0,0 +1,104 @@
+# Numeric Format
+
+Columns that provide numeric values conforming to specified rules and formatting requirements ensure clarity, accuracy, and ease of interpretation for both humans and systems.
+
+All columns capturing a numeric value, defined in the FOCUS specification, MUST be a real number represented as an integer, a decimal value, or a value represented in scientific notation as defined in this section.
```
ljadvey, line 5: Scientific notation is not compute friendly…. I would prefer we don’t allow this. Thoughts?

specification/attributes/numeric_format.md
```diff
@@ -0,0 +1,104 @@
...
+| :--------------| :------------------- | :------------------ | :-------------------- |
+| (unspecified)  | (unspecified)        | 64-bit binary format IEEE 754-2008 floating-point |
+| Integer        | (unspecified)        | 32-bit signed two's complement integer | -2,147,483,647 to +2,147,483,647 |
+| Integer        | Short                | 16-bit signed two's complement integer | -32,767 to +32,767 |
+| Integer        | Long                 | 32-bit signed two's complement integer | -2,147,483,647 to +2,147,483,647 |
+| Integer        | Extended             | 64-bit signed two's complement integer *or higher* | extended |
+| Float          | (unspecified)        | 64-bit binary format IEEE 754-2008 floating-point | 16 |
+| Float          | Single               | 32-bit binary format IEEE 754-2008 floating-point | 9 |
+| Float          | Double               | 64-bit binary format IEEE 754-2008 floating-point | 16 |
+| Float          | Extended             | 128-bit binary format IEEE 754-2008 floating-point *or higher* | extended |
```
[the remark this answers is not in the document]
ljadvey: Quantity would probably have integer….? But, we need to look at it case by case.

specification/attributes/numeric_format.md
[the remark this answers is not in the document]
tobrien: I changed it to:

* Columns with a Numeric format MUST contain a single, numeric value.
* Numeric values MUST be expressed as integer or decimal values. Fractional notation is NOT allowed.

The "real, rational number" requirement doesn't need to be there, and I think it was just distracting.  If we don't allow fractions, and we specify integer or decimal values then we're only talking about real and rational numbers.

Also, I took the line item out about arrays.  If the first line item states "single, numeric value" then we don't have to explicitly disallow arrays or ranges.

specification/attributes/numeric_format.md
tobrien: Yeah, I could go both ways.  I think I'm leaning towards us not stating anything about scientific notation.   I just added this "just in case," but I don't see a need for it.

@ljadvey I'm going to take it out of this request - it's a distraction.

tobrien: Just to make it clear, I don't know if I'm going to expressly prohibit it, but I think we just leave it out.    From a practical perspective, if I have floating point numbers, every language I know of is going to know how to parse them, but I just don't think we need to state this one way or the other.

The only place I can think of where something like this is useful, is when we have several underutilized storage  - if you are being charged in PBs, but you are only consuming a fraction of a percent of that unit, you end up with numbers like 0.000000342.   But, I don't think it's worth diving into this level of specificity on the format

specification/attributes/numeric_format.md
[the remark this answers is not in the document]
tobrien: Decimal is the same as float from a standards perspective, I'm getting more specific here and just getting rid of the word float all together.   I'll reference the decimal32,64,128 to get more specific - look for the next commit to fix this.

specification/attributes/numeric_format.md
tobrien: I rearranged the sections.

tobrien:
@ljadvey  @udam-f2  - I've updated the content based on feedback from the PR.

* Removed Quotes from Examples
* Simplified the requirements
* Got rid of any mention of scientific notation (also removed the example)
* Added language that stated that providers should define data precision, but removed any language related to precision being set in the standard.
* Identifying a ISO/IEC standard to root the integer precisions values in ISE/IEC 9899:2018
* Changed all mention of the term "float" to "decimal" and referencing the decimal32,64,128 formats in IEEE 754

In general, this revision should be more concise.

specification/attributes/numeric_format.md
```diff
@@ -0,0 +1,91 @@
...
+
+## Description
+
+Rules and formatting requirements for numeric columns appearing in billing data.
+
+## Requirements
+
+* Columns with a Numeric format MUST contain a single, numeric value.
+* Numeric values MUST be expressed as integer or decimal values. Fractional notation is NOT allowed.
+* Numerical values expressed with mathematical symbols, operators, or exponent values are not permitted.
```
udam-f2, line 23: I assume this is intentionally written as Numerical and not Numeric?

specification/attributes/numeric_format.md
```diff
@@ -0,0 +1,91 @@
...
+Numeric Format
+
+## Description
+
+Rules and formatting requirements for numeric columns appearing in billing data.
+
+## Requirements
+
+* Columns with a Numeric format MUST contain a single, numeric value.
+* Numeric values MUST be expressed as integer or decimal values. Fractional notation is NOT allowed.
```
udam-f2, line 22:
```suggestion
* Numeric values MUST be expressed as integer or decimal values. Fractional notation MUST NOT be used.
```

udam-f2: NOT is not a part of https://www.rfc-editor.org/rfc/rfc2119.txt. We can use SHALL NOT if that's preferable here.

specification/attributes/numeric_format.md
udam-f2, line 23:
```suggestion
* Numerical values MUST NOT be expressed with mathematical symbols, operators, or exponent values.
```

specification/attributes/numeric_format.md
```diff
@@ -0,0 +1,91 @@
...
+* Numerical values expressed with mathematical symbols, operators, or exponent values are not permitted.
+* Numeric values MUST NOT contain qualifiers or additional characters (e.g. currency symbols, units of measure, etc.).
+* Numeric values MUST NOT contain commas or punctuation marks except for a single decimal point if required to express a decimal value.
+* Numeric values MUST NOT include a character to represent sign for a positive value.  The only sign permitted is a negative sign (-) to indicate a negative value.
```
udam-f2, line 26: Not a 100% sure if one of the BCP-14 values should be used for the second sentence. 'MAY' is optional but in the case of negative values, that's not optional So would we say "Negative sign (-) MUST be used to indicate a negative value."?

specification/attributes/numeric_format.md
```diff
@@ -0,0 +1,91 @@
...
+
+The numeric Data type is specified in the metric definition as the "Data type" constraint as shown in the following example table.
+
+|    Constraint   |      Value      |
+|:----------------|:----------------|
+| Column required | True            |
+| Data type       | Decimal         |
+| Allows nulls    | False           |
+| Value format    | Numeric value   |
+
```
udam-f2, line 64: I'm a little confused by this section. Do we need this? What are the possible choices for Data type and Value format in the above table? Are both necessary or should Value format just be Numeric value and Data type just say Decimal | Integer?

Also feels like this should go above the precision section

specification/attributes/numeric_format.md
```diff
@@ -0,0 +1,91 @@
...
+## Examples of Numeric Values Meeting Requirements
+
+* Values Meeting Numeric Requirements:
+  * -100.2
+  * 0
+  * 1
```
udam-f2, line 70: since we say above not to use it for binary values, should we just suggest a different value for either line 69 or 70? e.g. 103

specification/attributes/numeric_format.md
```diff
@@ -0,0 +1,91 @@
...
+
+* Columns with a Numeric format MUST contain a single, numeric value.
+* Numeric values MUST be expressed as integer or decimal values. Fractional notation is NOT allowed.
+* Numerical values expressed with mathematical symbols, operators, or exponent values are not permitted.
+* Numeric values MUST NOT contain qualifiers or additional characters (e.g. currency symbols, units of measure, etc.).
+* Numeric values MUST NOT contain commas or punctuation marks except for a single decimal point if required to express a decimal value.
+* Numeric values MUST NOT include a character to represent sign for a positive value.  The only sign permitted is a negative sign (-) to indicate a negative value.
+* Numeric values MUST NOT be used to represent binary values (e.g. 0 or 1).
+
+## Numeric Precision
```
udam-f2, line 29: Thoughts on this and the two sections below being a sub-heading and not the same level as the standard headers we use for the sepc content? For example we could park these as H3s after the normative content at the very top (before ## Attribute ID).

specification/attributes/numeric_format.md
```diff
@@ -0,0 +1,91 @@
...
+
+The FOCUS specification does not require a specific level of precision for numeric values.  The level of precision required for a given column is determined by the provider and SHOULD be documented as a part of a data definition published by the provider.  The following table outlines the options for specifying the level of precision for a given metric in a data definition document.
```
udam-f2, line 31: If we have to use BCP-14 normative content, it would be best if we can limit it to the requirements section so everything is in one place.

specification/attributes/numeric_format.md
tobrien: Here's what I'm trying to say in plain-speak - "You can list Numeric, Integer, or Decimal.   If you list Numeric, you are basically saying - could be either. The spec has no opinion."

Do we want to just make it require to specify either Integer or Decimal?  I'm good with that if it makes it more straightforward.

specification/attributes/numeric_format.md
tobrien: Yep, that was a mistake.  Fixed in the latest push to this branch.  thanks for catching

specification/attributes/numeric_format.md
tobrien: ok, I rearranged content in the latest push to the branch.

specification/attributes/numeric_format.md
tobrien: Ok, the precision section is an interesting case because it's normative that providers SHOULD publish this, but I still think it's not an enforceable requirement.   I rearranged everything that is a MUST up top and left this at the end.   WDYT?

specification/attributes/numeric_format.md
tobrien: Done. Updated

specification/attributes/numeric_format.md
tobrien: I reworded things a bit - see if it makes sense in the latest push.

specification/attributes/numeric_format.md
```diff
@@ -0,0 +1,87 @@
...
+  * sqrt(2) - contains a mathematical symbol or operation
+  * 2.3^3 - contains an exponent
+  * 32 GiB - contains a unit of measure
+  * $32 - contains a currency symbol
+  * 3,432,342 - contains a comma
+  * +333 - contains a positive sign
+
+## Numeric Precision
+
+The FOCUS specification does not require a specific level of precision for numeric values.  The level of precision required for a given column is determined by the provider and SHOULD be documented as a part of a data definition published by the provider.  The following table outlines the options for specifying the level of precision for a given column in a data definition document.
```
flanakin, line 66: nit
```suggestion
FOCUS does not require a specific level of precision for numeric values.  The level of precision required for a given column is determined by the provider and SHOULD be documented as a part of a data definition published by the provider.  The following table outlines the options for specifying the level of precision for a given column in a data definition document.
```

flanakin (approved):

ljadvey (approved):

specification/attributes/numeric_format.md
```diff
@@ -0,0 +1,72 @@
+# Numeric Format
+
+Columns that provide numeric values conforming to specified rules and formatting requirements ensure clarity, accuracy, and ease of interpretation for humans and systems. The FOCUS specification does not require a specific level of precision for numeric values. The level of precision required for a given column is determined by the provider and should be part of a data definition published by the provider.
+
+All columns capturing a numeric value, defined in the FOCUS specification, MUST follow the formatting requirements listed below. Provider generated date/time related columns SHOULD adopt the same format requirements over time.
```
ijurica, line 5:
```suggestion
All columns capturing a numeric value, defined in the FOCUS specification, MUST follow the formatting requirements listed below. Provider generated numeric value capturing columns SHOULD adopt the same format requirements over time.
```

specification/attributes/numeric_format.md
```diff
@@ -0,0 +1,72 @@
...
+  * [3,5,8] - contains an array
+  * [4:5] - contains a range
+  * 5i + 4 - contains a complex number
+  * sqrt(2) - contains a mathematical symbol or operation
+  * 2.3^3 - contains an exponent
+  * 32 GiB - contains a unit of measure
+  * $32 - contains a currency symbol
+  * 3,432,342 - contains a comma
+  * +333 - contains a positive sign
+
```
ijurica, line 65: E notation is prohibited or? I suggest including it as one of the examples.

specification/attributes/numeric_format.md
ijurica: except for this small oversight, great job!

ijurica (approved):

specification/attributes/numeric_format.md
```diff
@@ -0,0 +1,73 @@
...
+Rules and formatting requirements for numeric columns appearing in billing data.
+
+## Requirements
+
+* Columns with a Numeric value format MUST contain a single numeric value.
+* Numeric values MUST be expressed as integer or decimal values. Fractional notation MUST NOT be used.
+* Numeric values MUST NOT be expressed with mathematical symbols, operators, or exponent values.
+* Numeric values MUST NOT contain qualifiers or additional characters (e.g., currency symbols, units of measure, etc.).
+* Numeric values MUST NOT contain commas or punctuation marks except for a single decimal point if required to express a decimal value.
```
flanakin, line 25: Should we be explicit about the "decimal point" being a period given regional variations?
```suggestion
* Numeric values MUST NOT contain commas or punctuation marks except for a single decimal point (`.`) if required to express a decimal value.
```

flanakin (approved):

specification/attributes/numeric_format.md
marc-perreaut: I understand from the requirements that the E/e exponential notation is disallowed.
This notation is supported by JSON, Python and many languages.
This notation is used in Azure billing data.
What is the rationale for disallowing it?

specification/attributes/numeric_format.md
```diff
@@ -0,0 +1,73 @@
+# Numeric Format
+
+Columns that provide numeric values conforming to specified rules and formatting requirements ensure clarity, accuracy, and ease of interpretation for humans and systems. The FOCUS specification does not require a specific level of precision for numeric values. The level of precision required for a given column is determined by the provider and should be part of a data definition published by the provider.
+
+All columns capturing a numeric value, defined in the FOCUS specification, MUST follow the formatting requirements listed below. Provider generated numeric value capturing columns SHOULD adopt the same format requirements over time.
```
marc-perreaut, line 5: >  Provider generated numeric value capturing columns SHOULD adopt the same format requirements over time.

I would propose a less strong statement:

>  Provider generated numeric value capturing columns SHOULD adopt *backward-compatible* format requirements over time.

joshk0 (changes requested):
looks good.
I wonder if we should prescribe wire formats when advanced formats (parquet, avro etc.) provide numbers?
For example, we've cooked ourselves as a data platform using FLOAT64 in BigQuery (IEEE 754 style), when we instead should have used NUMERIC. This is especially noticeable when numbers are very small; such as the unit cost of 1 byte of RAM for one hour in a compute service ($0.00000001 type number) If you take the float64 and multiply it over many hours, you might get a wrong result.

It may be out of scope for this spec though which I totally get. Just always want to find ways for people to avoid shooting themselves in the foot where I have before.

specification/attributes/numeric_format.md
tobrien: I wasn't aware that Azure allowed this.  @marc-perreaut - I just committed language to allow it, and to specify that values must adhere to the E notation format.  Thanks for catching this.

tobrien:
> looks good. I wonder if we should prescribe wire formats when advanced formats (parquet, avro etc.) provide numbers? For example, we've cooked ourselves as a data platform using FLOAT64 in BigQuery (IEEE 754 style), when we instead should have used NUMERIC. This is especially noticeable when numbers are very small; such as the unit cost of 1 byte of RAM for one hour in a compute service ($0.00000001 type number) If you take the float64 and multiply it over many hours, you might get a wrong result.
> 
> It may be out of scope for this spec though which I totally get. Just always want to find ways for people to avoid shooting themselves in the foot where I have before.

@joshk0 - Good evening Mr. Kwan.  I was thinking about this, and how to put it into the column specification, I don't know if it makes sense to talk about the wire format in the attribute, but I do think something needs to be said somewhere.   Can we leave it out of this format specifier and figure out another place to put it.

Example, I think we need to think about maybe another document that is focused on general tips for implementation that can go into arbitrary precision data types in different languages and databases.   Does that make sense?   So, I'm marking it as resolved for this PR, but I do think a real resolution would be to drop it into a discussion section.

specification/attributes/numeric_format.md
tobrien: I implemented this suggestion.  Thanks

specification/attributes/numeric_format.md
```diff
@@ -0,0 +1,76 @@
+# Numeric Format
+
+Columns that provide numeric values conforming to specified rules and formatting requirements ensure clarity, accuracy, and ease of interpretation for humans and systems. The FOCUS specification does not require a specific level of precision for numeric values. The level of precision required for a given column is determined by the provider and should be part of a data definition published by the provider.
+
+All columns capturing a numeric value, defined in the FOCUS specification, MUST follow the formatting requirements listed below. Provider-generated numeric value capturing columns SHOULD adopt backward-compatible format requirements over time.
```
udam-f2, line 5: In other attributes, we're going with language like this: 
"All columns capturing a date/time value, defined in the FOCUS specification, MUST follow the formatting requirements listed below. Provider-generated, date/time-related columns SHOULD adopt the same format requirements over time."

Should we make the others look like this OR make this consistent with the others? Fine either way. I don't feel like we need to explicitly say backward-compatible FWIW. How providers solve these issues for their customers is up to them. They might create new data sets for all we know.

udam-f2 (commented):
LGTM other than a minor comment.

joshk0 (approved):
Let's figure out how to make a mention of precision somewhere later, but that's an improvement over everything here which is already good and should just go in.

specification/attributes/numeric_format.md
marc-perreaut: I think I proposed the backward-compatibility when reviewing the PR, because the original statement looked too strong to me. Sorry, I had not cross checked the language used in other language attributes. I do not have a strong opinion about stating explicitly backward compatibility and I am fine to make this consistent with others.

ljadvey (approved):

udam-f2 (approved):
LGTM

PR #193, opened 2023-10-04
FOCUS #180.2 - Adding language to make the 'Mandatory Alignment with Recommended Units" Clear

tobrien:
Ok, three changes:

This "You must align with recommended units" language was unclear.      It can be paraphrased as this:

If your unit is one of the dimensions we have in the recommended unit sets table.  You must use those units. If you have to report another dimension, like distance, go ahead and introduce a new unit."

I also split the numeric prefixes out of the recommended unit names table.

### DIFF
--- a/specification/dimensions/usageunit.md
+++ b/specification/dimensions/usageunit.md
@@ -2,9 +2,19 @@
 
 Usage Unit refers to a unit of measurement for the consumption or usage of resources or services. The Usage Unit for a particular line item may differ from the published Pricing Unit when providers use different units, unit increments, or dimensions for calculating cost.  This unit is commonly used when auditing or reconciling consumption or billing data.
 
-The UsageUnit column MUST be present and MUST NOT be null or empty. UsageUnit is of type String, and UsageUnit MUST be composed of the list of recommended units or unit abbreviations if it is possible to be expressed with the units listed in "Recommended Unit Set" below. If the UsageUnit value is a composite value made from combinations of one or more units, each component MUST also align with the set of recommended units if possible.  For example, a composite unit of "GB/Month" is allowed.
+The UsageUnit column MUST be present and MUST NOT be null or empty. UsageUnit is of type String.
 
 Instead of "per" or "-" to denote a Composite Unit, slash ("/") SHOULD be used as a common convention.  Count based units like requests, instances, tokens SHOULD be expressed as "count".  For example, if a usage unit is measured as a rate of requests or instances over a period of time, the unit should be listed as "count/day" to signify the number of requests per day.
+
+## Mandatory Alignment with Recommended Unit Set
+
+UsageUnit MUST be composed of the list of recommended units or unit abbreviations listed in "Recommended Unit Set" below unless the UsageUnit value covers a dimension not listed in the recommended unit set.  If the UsageUnit value is a composite value made from combinations of one or more units, each component MUST also align with the set of recommended units or unit abbreviations.
+
+This specification requires alignment with recommended units or unit abbreviations to ensure consistency and compatibility across providers, but it allows providers to adapt to emerging units or dimensions as they are introduced.
+
+For example, if a UsageUnit value is "requests/day", the unit "count/day" MUST be used instead as "requests" is not a recommended unit.  In another example, if a UsageUnit must cover distance or length, a new unit term may be introduced as the recommended unit set listed in this specification does not cover a distance of the length dimension.  
+
+Another example of a non-compliant value illustrates that count-based units must align with the recommended unit set.  If a UsageUnit captures "tokens/sec", the unit "count/sec" MUST be used instead as "tokens" is not a recommended unit.  Any unit that consists of a number of events, occurrences, or instances MUST be expressed as "count" to maintain compatibility with this specification.
 
 ## Column ID
 
@@ -33,18 +43,37 @@ Units listed in the following table are presented in the following format: "abbr
 
 If the unit is not listed in the table, it is to be used over a functional equivalent with similar meaning or incompatible capitalization.  An example of unit now allowed is a unit such as "calls" or "number" to signify the number of times an event occurred in this case, the unit "count" MUST be used to maintain compatibility with this specification.
 
-| TIME_UNIT        | DATA_UNIT | COUNT_UNIT   | NUMERIC PREFIXES        |
-|------------------|-----------|--------------|-------------------------|
-| yr: year         | bit: bit  | cnt: count   | Ki: kibi (2^10)         |
-| mo: month        | B: byte   |              | Mi: mebi (2^20)         |
-| wk: week         |           |              | Gi: gibi (2^30)         |
-| d: day           |           |              | Ti: tebi (2^40)         |
-| h: hour          |           |              | Pi: pebi (2^50)         |
-| min: minute      |           |              | K: kilo (10^3)          |
-| s: second        |           |              | M: mega (10^6)          |
-| ms: millisecond  |           |              | G: giga (10^9)          |
-| us: microsecond  |           |              | T: tera (10^12)         |
-| ns: nanosecond   |           |              | P: peta (10^15)         |
+The following table lists the valid units for 3 dimensions of measurement: time, data, and count.  
+
+| Time             | Data      | Count        |
+|------------------|-----------|--------------|
+| yr: year         | bit: bit  | cnt: count   |
+| mo: month        | B: byte   |              |
+| wk: week         |           |              |
+| d: day           |           |              |
+| h: hour          |           |              |
+| min: minute      |           |              |
+| s: second        |           |              |
+| ms: millisecond  |           |              |
+| us: microsecond  |           |              |
+| ns: nanosecond   |           |              |
+
+Each unit can be associated with a numeric prefix to multiply the value by a defined power of ten.  The following table lists the valid numeric prefixes.
+
+| NUMERIC PREFIXES        |
+|-------------------------|
+| Ki: kibi (2^10)         |
+| Mi: mebi (2^20)         |
+| Gi: gibi (2^30)         |
+| Ti: tebi (2^40)         |
+| Pi: pebi (2^50)         |
+| Ei: exbi (2^60)         |
+| K: kilo (10^3)          |
+| M: mega (10^6)          |
+| G: giga (10^9)          |
+| T: tera (10^12)         |
+| P: peta (10^15)         |
+| E: exa (10^15)          |
 
 ## Introduced (version)
 
--- a/specification/dimensions/usageunit.md
+++ b/specification/dimensions/usageunit.md
@@ -1,20 +1,16 @@
 # Usage Unit
 
-Usage Unit refers to a unit of measurement for the consumption or usage of resources or services. The Usage Unit for a particular line item may differ from the published Pricing Unit when providers use different units, unit increments, or dimensions for calculating cost.  This unit is commonly used when auditing or reconciling consumption or billing data.
+Usage Unit refers to a unit of measurement for the purchase or consumption of resources or services. The Usage Unit for a particular line item may differ from the published Pricing Unit when providers use different units, unit increments, or dimensions for calculating cost.  This unit is commonly used when auditing or reconciling consumption or billing data.
 
-The UsageUnit column MUST be present and MUST NOT be null or empty. UsageUnit is of type String.
+The UsageUnit column MUST be present and MUST NOT be null or empty. UsageUnit is of type String. UsageUnit should be expressed as a single unit of measure adhering to the following format options:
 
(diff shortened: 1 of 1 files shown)


specification/dimensions/usageunit.md
```diff
@@ -2,10 +2,20 @@
...
 
-The UsageUnit column MUST be present and MUST NOT be null or empty. UsageUnit is of type String, and UsageUnit MUST be composed of the list of recommended units or unit abbreviations if it is possible to be expressed with the units listed in "Recommended Unit Set" below. If the UsageUnit value is a composite value made from combinations of one or more units, each component MUST also align with the set of recommended units if possible.  For example, a composite unit of "GB/Month" is allowed.
+The UsageUnit column MUST be present and MUST NOT be null or empty. UsageUnit is of type String.
 
 Instead of "per" or "-" to denote a Composite Unit, slash ("/") SHOULD be used as a common convention.  Count based units like requests, instances, tokens SHOULD be expressed as "count".  For example, if a usage unit is measured as a rate of requests or instances over a period of time, the unit should be listed as "count/day" to signify the number of requests per day.
```
flanakin, line 7: This is written as if there's an existing expectation about how the value should be formatted. Can we phrase it so it stands on its own and just define what we want? I'm hoping we can keep it as simple as `<plural-units>` or `<singular-unit> <plural-time-units>` (e.g., "GB Hours") or `<plural-units>/<singular-time-unit>`. (I think those are the only three 🤔)

The values need to be title cased, like everything else. We don't want this to be the only column that's all lowercase. Also, instead of "count", can we use "Units" as the generic identifier? That sounds better in a sentence.

flanakin: On second thought, move this down to an **Allowed values** section, like the other normalized columns.

specification/dimensions/usageunit.md
```diff
@@ -2,10 +2,20 @@
...
 
+## Mandatory Alignment with Recommended Unit Set
```
flanakin, line 9: Can we just merge this with the other section below?

specification/dimensions/usageunit.md
```diff
@@ -2,10 +2,20 @@
...
+
+UsageUnit MUST be composed of the list of recommended units or unit abbreviations listed in "Recommended Unit Set" below unless the UsageUnit value covers a dimension not listed in the recommended unit set.  If the UsageUnit value is a composite value made from combinations of one or more units, each component MUST also align with the set of recommended units or unit abbreviations.
+
+This specification requires alignment with recommended units or unit abbreviations to ensure consistency and compatibility across providers, but it allows providers to adapt to emerging units or dimensions as they are introduced.
```
flanakin, line 13: You can probably replace these last 3 paragraphs with a simple table of the options with examples.

specification/dimensions/usageunit.md
```diff
@@ -33,18 +43,37 @@ Units listed in the following table are presented in the following format: "abbr
...
-| mo: month        | B: byte   |              | Mi: mebi (2^20)         |
-| wk: week         |           |              | Gi: gibi (2^30)         |
-| d: day           |           |              | Ti: tebi (2^40)         |
-| h: hour          |           |              | Pi: pebi (2^50)         |
-| min: minute      |           |              | K: kilo (10^3)          |
-| s: second        |           |              | M: mega (10^6)          |
-| ms: millisecond  |           |              | G: giga (10^9)          |
-| us: microsecond  |           |              | T: tera (10^12)         |
-| ns: nanosecond   |           |              | P: peta (10^15)         |
+The following table lists the valid units for 3 dimensions of measurement: time, data, and count.
```
flanakin, line 46: This feels like I need a secret decoder ring 😕 Can we use plain English, like our other columns?

specification/dimensions/usageunit.md
```diff
@@ -33,18 +43,37 @@ Units listed in the following table are presented in the following format: "abbr
 
 If the unit is not listed in the table, it is to be used over a functional equivalent with similar meaning or incompatible capitalization.  An example of unit now allowed is a unit such as "calls" or "number" to signify the number of times an event occurred in this case, the unit "count" MUST be used to maintain compatibility with this specification.
```
flanakin, line 44: GitHub won't let me comment on the other section since it's from a previous commit 😒

Can we label this section "Allowed values" and make it an H3, like the other normalized columns?

specification/dimensions/usageunit.md
```diff
@@ -33,18 +43,37 @@ Units listed in the following table are presented in the following format: "abbr
...
-| min: minute      |           |              | K: kilo (10^3)          |
-| s: second        |           |              | M: mega (10^6)          |
-| ms: millisecond  |           |              | G: giga (10^9)          |
-| us: microsecond  |           |              | T: tera (10^12)         |
-| ns: nanosecond   |           |              | P: peta (10^15)         |
+The following table lists the valid units for 3 dimensions of measurement: time, data, and count.  
+
+| Time             | Data      | Count        |
+|------------------|-----------|--------------|
+| yr: year         | bit: bit  | cnt: count   |
```
flanakin, line 50: Can we use "Units" as the generic count? It sounds better and aligns more with the other values. 
"24 Hours" 
"15 Units"

Also, why are we not allowing other common words? I'm fine if we don't have everything, but there are a core set of common words we should be able to use.

FYI - Here are Microsoft's unique values: https://github.com/microsoft/finops-toolkit/blob/dev/src/open-data/PricingUnits.csv
```suggestion
| yr: year         | bit: bit  | Units   |
```

specification/dimensions/usageunit.md
```diff
@@ -33,18 +43,37 @@ Units listed in the following table are presented in the following format: "abbr
...
+| mo: month        | B: byte   |              |
```
flanakin, line 51: These all need to be title cased like the rest of the spec.
```suggestion
| Months        | B: byte   |              |
```

specification/dimensions/usageunit.md
```diff
@@ -33,18 +43,37 @@ Units listed in the following table are presented in the following format: "abbr
...
+| Gi: gibi (2^30)         |
+| Ti: tebi (2^40)         |
+| Pi: pebi (2^50)         |
+| Ei: exbi (2^60)         |
+| K: kilo (10^3)          |
+| M: mega (10^6)          |
+| G: giga (10^9)          |
+| T: tera (10^12)         |
+| P: peta (10^15)         |
+| E: exa (10^15)          |
```
flanakin, line 76: While I understand the desire to mix and match, it would arguably be more straightforward to just list out the allowed values. It would be much more precise without people having to decipher what's allowed. Technically, this sounds like we're allowing kibi-minutes or something 😜
```suggestion
Data size MUST be abbreviated using the following standard abbreviations.

| Data size in bits    | Data size in bytes    |
| -------------------- | --------------------- |
| Kib (kibibit = 2^10) | KiB (kibibyte = 2^10) |
| Mib (mebibit = 2^20) | MiB (mebibyte = 2^20) |
| Gib (gibibit = 2^30) | GiB (gibibyte = 2^30) |
| Tib (tebibit = 2^40) | TiB (tebibyte = 2^40) |
| Pib (pebibit = 2^50) | PiB (pebibyte = 2^50) |
| Eib (exbibit = 2^60) | EiB (exbibyte = 2^60) |
| Kb (kilobit = 10^3)  | KB (kilobyte = 10^3)  |
| Mb (megabit = 10^6)  | MB (megabyte = 10^6)  |
| Gb (gigabit = 10^9)  | GB (gigabyte = 10^9)  |
| Tb (terabit = 10^12) | TB (terabyte = 10^12) |
| Pb (petabit = 10^15) | PB (petabyte = 10^15) |
| Eb (exabit = 10^15)  | EB (exabyte = 10^15)  |
```

specification/dimensions/usageunit.md
```diff
@@ -33,18 +43,37 @@ Units listed in the following table are presented in the following format: "abbr
...
+| yr: year         | bit: bit  | cnt: count   |
+| mo: month        | B: byte   |              |
+| wk: week         |           |              |
+| d: day           |           |              |
+| h: hour          |           |              |
+| min: minute      |           |              |
+| s: second        |           |              |
+| ms: millisecond  |           |              |
+| us: microsecond  |           |              |
+| ns: nanosecond   |           |              |
```
flanakin, line 59: Do we really want to allow this? It feels entirely too low-level. We can represent all of these using fractions of hours or minutes. When you're talking about something this small at a large scale, it becomes entirely unreadable by humans, which breaks one of our tenants of being human readable/consumable.

flanakin (changes requested):

specification/dimensions/usageunit.md
JGrubb: The only problem is that multiple of the vendors use the generic "count".

specification/dimensions/usageunit.md
JGrubb: In practice, I can't think of anything that comes across in weeks or days. It's either months or hours.

specification/dimensions/usageunit.md
JGrubb: Does this help?  I've been working w cloud carbon footprint a bit lately and they boil it down to 4 primary usage units.

- Compute: Hours
- Storage: Terabyte-hours
- Memory: Gigabyte-hours
- Networking: Gigabytes

This multiplicative table doesn't simplify the missions of improving understanding and normalizing data.

https://www.cloudcarbonfootprint.org/docs/creating-a-lookup-table/

specification/dimensions/usageunit.md
tobrien: +1, I'll revise based on all this feedback, and then resubmit the PR against the feature branch.

specification/dimensions/usageunit.md
tobrien: @flanakin - look at the latest PR revision, I incorporated the "Data size MUST by abbreviated using the following.

@JGrubb  - I am seeing some charges for PBs from one provider, and I have a few charges showing up in TB.   I agree that [M,G,T,B] is a smaller, more common set, but I think we list Exabytes for forward compatibility.   Mega and kila?  Not sure.  Let me take a look - for bandwidth, I think we need to keep both.

specification/dimensions/usageunit.md
tobrien: Mostly done - I dropped everything but "Day, Hour, Minute, Second" - why did I keep second?    I would rather parse 90 seconds than 1.5 minutes.  I think we should leave some flexibility on level up and one level down.     Also, from a licensing perspective, days is something I see as relevant.

Check the latest revision of the PR.

specification/dimensions/usageunit.md
tobrien: Done - look at the latest rev

specification/dimensions/usageunit.md
tobrien: @JGrubb @flanakin @amitw-g - I've gone back and forth here, but I'm aligned with what Michael's saying here, why would we not just allow a provider to use a unit term like "Token", "Domain", "Certificate"?  There is a large list here.

What I ended up proposing is this "Use the Allowed Values, but if you have a term for count-based unit that isn't present.  Anything unrecognized will be considered a new count-based unit."     I really don't want to get a charge from a provider for 400000 count on a Gen AI service, I'd like to know exactly what the count unit was.

specification/dimensions/usageunit.md
tobrien: Merged

specification/dimensions/usageunit.md
```diff
@@ -2,10 +2,20 @@
 
 Usage Unit refers to a unit of measurement for the consumption or usage of resources or services. The Usage Unit for a particular line item may differ from the published Pricing Unit when providers use different units, unit increments, or dimensions for calculating cost.  This unit is commonly used when auditing or reconciling consumption or billing data.
```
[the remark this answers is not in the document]
tobrien: Done

specification/dimensions/usageunit.md
```diff
@@ -2,10 +2,20 @@
...
+The UsageUnit column MUST be present and MUST NOT be null or empty. UsageUnit is of type String.
 
 Instead of "per" or "-" to denote a Composite Unit, slash ("/") SHOULD be used as a common convention.  Count based units like requests, instances, tokens SHOULD be expressed as "count".  For example, if a usage unit is measured as a rate of requests or instances over a period of time, the unit should be listed as "count/day" to signify the number of requests per day.
 
+## Mandatory Alignment with Recommended Unit Set
+
+UsageUnit MUST be composed of the list of recommended units or unit abbreviations listed in "Recommended Unit Set" below unless the UsageUnit value covers a dimension not listed in the recommended unit set.  If the UsageUnit value is a composite value made from combinations of one or more units, each component MUST also align with the set of recommended units or unit abbreviations.
```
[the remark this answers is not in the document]
tobrien: Done, moved all the MUST to a previous paragraph.   Look at latest rev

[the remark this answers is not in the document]
tobrien: Done, I removed abbreviations from this entire column def

PR #196, opened 2023-10-05
Update supporting content to add OCI mappings to billingcurrency

arrarama:
Added OCI reference and sample for currency.


udam-f2 (approved):
LGTM

PR #197, opened 2023-10-05
Update supporting content to add OCI mappings to availabilityzone

arrarama:
Corrected Availability Zone examples.


udam-f2 (approved):
LGTM

PR #214, opened 2023-10-10
FOCUS #161: Pricing Category column

rodneyjoyce:
First draft from 0.5


specification/dimensions/pricingmodel.md
```diff
@@ -0,0 +1,39 @@
+# Pricing Model
+
+Pricing Model indicates whether a charge is a fixed, on-demand or discounted.
```
flanakin, line 3: First draft... 
```suggestion
Pricing Model indicates the type of unit price applied to a charge at the time it was used or purchased. Pricing Model is commonly used to differentiate between charges at on-demand vs. reduced prices when investigating optimization opportunities, like for commitment-based discount coverage.
```

specification/dimensions/pricingmodel.md
```diff
@@ -0,0 +1,39 @@
...
+
+The PricingModel column MUST be present. This column MUST be of type String and MUST be one of the Supported Values.
```
flanakin, line 5: I wonder if the last sentence takes it too far, given the values are defined 🤔
```suggestion
The PricingModel column MUST be present and MUST NOT be null or empty. This column is of type String and MUST be one of the allowed values. PricingModel MUST be "Commitment-Based Discount" when CommitmentDiscountId is not null. PricingModel MUST be "Dynamic" when pricing is determined by the provider and may change over time, regardless of predetermined agreement pricing.
```

specification/dimensions/pricingmodel.md
```diff
@@ -0,0 +1,39 @@
...
+
+PricingModel
+
+## Display name
+
+Pricing Model
+
+## Description
+
+TBD once definition is confirmed
```
flanakin, line 17:
```suggestion
Type of unit price applied to a charge at the time it was used or purchased.
```

specification/dimensions/pricingmodel.md
```diff
@@ -0,0 +1,39 @@
...
+| Value format    | List of values   |
+
+Allowed values:
+
+| Value      | Description                                                                                                                                                                   |
+|:---------------|:--------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
+| Fixed    | TBC                                                                 |
+| OnDemand | TBC                                                                                             |
+| Commitment Based Discount    | TBC                                                                 |
+| Interruptible | TBC                                                                                             |
```
flanakin, line 35:
```suggestion
| Value                     | Description                                                                                                                                                                                                  |
| :------------------------ | :----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
| Fixed                     | Charges priced at a constant rate, regardless of quantity used or purchased.                                                                                                                                 |
| On-Demand                 | Charges priced at the standard rate for the billing account. This includes any applicable negotiated discounts and volume/tiered pricing, but does not include dynamic or commitment-based discount pricing. |
| Dynamic                   | Charges priced at a variable rate determined by the provider. This includes any product or service with a unit price the provider can change without notice, like interruptible or low priority resources.   |
| Commitment-Based Discount | Charges with reduced prices due to a commitment-based discount specified by the Commitment Discount ID.                                                                                                      |
```

supporting_content/dimensions/pricingmodel.md
```diff
@@ -0,0 +1,36 @@
+# Column: PricingModel - WIP
+
+## Example provider mappings
+
+Current column mappings found in available data sets:
+
+| Provider | Data set                 | Column                    |  Example Values  |
+|----------|--------------------------|---------------------------|------------------|
+| AWS | CUR                      |  | |
```
flanakin, line 9:
```suggestion
| AWS | CUR                      |  product/PurchaseOption | On-Demand, Reserved Instances, Spot Instances, Dedicated Hosts |
```

specification/dimensions/pricingmodel.md
```diff
@@ -0,0 +1,39 @@
...
+## Content constraints
+
+|    Constraint   |      Value       |
+|:----------------|:-----------------|
+| Column required | True             |
+| Data type       | String           |
+| Allows nulls    | False             |
+| Value format    | List of values   |
+
+Allowed values:
```
flanakin, line 28: Should we add a "Shared" option to account for dedicated host for AWS?

supporting_content/dimensions/pricingmodel.md
```diff
@@ -0,0 +1,36 @@
...
+
+## Example provider mappings
+
+Current column mappings found in available data sets:
+
+| Provider | Data set                 | Column                    |  Example Values  |
+|----------|--------------------------|---------------------------|------------------|
+| AWS | CUR                      |  product/PurchaseOption | On-Demand, Reserved Instances, Spot Instances, Dedicated Hosts |
+| Google Cloud | BigQuery Billing Export |              |  |
+| Microsoft | Cost Details             | PricingModel             | OnDemand, Spot, Reservation |
```
flanakin, line 11:
```suggestion
| Microsoft | Cost Details             | PricingModel             | OnDemand, Spot, Reservation, SavingsPlan |
```

supporting_content/dimensions/pricingmodel.md
```diff
@@ -0,0 +1,36 @@
...
+| Microsoft | Cost Details (PricingModel)| Spot                    | Reservation                      |
+| Microsoft | Cost Details (PricingModel)| OnDemand                    | Reservation                      |
+
+## Documentation
+- Microsoft
+  - Azure:  Understand usage details fields: https://learn.microsoft.com/en-us/azure/cost-management-billing/automate/understand-usage-details-fields
+- GCP
+  -
+- AWS
+  - Amazon:
```
flanakin, line 34:
```suggestion
- AWS: https://docs.aws.amazon.com/cur/latest/userguide/product-columns.html
- GCP
- Microsoft:  Understand usage details fields: https://learn.microsoft.com/azure/cost-management-billing/automate/understand-usage-details-fields
```

specification/dimensions/pricingmodel.md
```diff
@@ -0,0 +1,39 @@
...
+
+| Value                     | Description                                                                                                                                                                                                  |
+| :------------------------ | :----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| Fixed                     | Charges priced at a constant rate, regardless of quantity used or purchased.                                                                                                                                 |
+| On-Demand                 | Charges priced at the standard rate for the billing account. This includes any applicable negotiated discounts and volume/tiered pricing, but does not include dynamic or commitment-based discount pricing. |
```
flanakin, line 33: With this definition of "on-demand", I would argue purchases apply to that. I'm thinking we get rid of "Fixed" unless we know of a special need for it. As is, this seems unclear as to which one to use for purchases based on negotiated rates.

specification/dimensions/pricingmodel.md
```diff
@@ -0,0 +1,39 @@
+# Pricing Model
+
+Pricing Model indicates the type of unit price applied to a charge at the time it was used or purchased. Pricing Model is commonly used to differentiate between charges at on-demand vs. reduced prices when investigating optimization opportunities, like for commitment-based discount coverage.
```
dwbrown2, line 3: Should `tiers` be a part of this concept/definition or do we think that's better covered elsewhere?

specification/dimensions/pricingmodel.md
```diff
@@ -0,0 +1,39 @@
...
+
+| Value                     | Description                                                                                                                                                                                                  |
+| :------------------------ | :----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| Fixed                     | Charges priced at a constant rate, regardless of quantity used or purchased.                                                                                                                                 |
+| On-Demand                 | Charges priced at the standard rate for the billing account. This includes any applicable negotiated discounts and volume/tiered pricing, but does not include dynamic or commitment-based discount pricing. |
+| Dynamic                   | Charges priced at a variable rate determined by the provider. This includes any product or service with a unit price the provider can change without notice, like interruptible or low priority resources.   |
```
rodneyjoyce, line 34: So is the key phrase here "can change without notice"? 

Because the Provider can change the cost of On-Demand as prices increase (like Azure did recently as they moved to peg it to USD) however, they gave notice for this increase.

So the difference here is that you sign up to the service without knowing the EXACT cost in advance and when it will changed?

specification/dimensions/pricingmodel.md
```diff
@@ -0,0 +1,39 @@
...
```
flanakin, line 34: @ijurica FYI - We should be explicit about how we want the unit price columns to work given these different pricing models (especially dynamic).

specification/dimensions/pricingmodel.md
flanakin: Are you talking about tiered/graduated pricing?

specification/dimensions/pricingmodel.md
flanakin: Is the difference that the price changes happen within a billing period here rather than across different billing periods for list price changes? Thoughts on this:
```suggestion
| Dynamic                   | Charges priced at a variable rate that is designed to change throughout the day. This includes any product or service with a unit price the provider can change at any moment without notice, like interruptible or low priority resources.   |
```

Fwiw, you do know the price you'll have. Part of creating a Spot VM (at least in Azure) is that you define a max price you're willing to pay. Not sure about other clouds.

flanakin (approved):

specification/dimensions/pricingmodel.md
dwbrown2: Yep, one I've seen recently is IBM Cloud's implementation -- https://cloud.ibm.com/docs/billing-usage?topic=billing-usage-charges

specification/dimensions/pricingmodel.md
```diff
@@ -0,0 +1,39 @@
...
+| Column required | True             |
+| Data type       | String           |
+| Allows nulls    | False             |
+| Value format    | list-of-values   |
+
+Allowed values:
+
+| Value                     | Description                                                                                                                                                                                                  |
+| :------------------------ | :----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| Fixed                     | Charges priced at a constant rate, regardless of quantity used or purchased.                                                                                                                                 |
```
marc-perreaut, line 32: How is "Fixed" different from "On-Demand"?
What would be an example?

specification/dimensions/pricingmodel.md
marc-perreaut: I find the naming Dynamic is not clear enough.

For example, I would think that GCP SUD is included in Dynamic.
I believe we had a discussion to include GCP SUD within OnDemand, and to address GCP SUD with further dimensions.

Why not just use the value "Interruptible"?

marc-perreaut: Actually, I am not clear why Pricing Model would differentiate Spot (Dynamic/Interruptible) but not GCP SUD.

In Azure, Spot instances have no SLA: so it's not only a Pricing Model, it's also a different Product/Sku/UsageType (we are currently discussing on this naming :-)). Conversely, GCP CUD and SUD can apply to the same Product/Sku/UsageType, so it relates to a Pricing Model.

aqu-erp: Because SUDs are computed as a separate line item combining all discounts of all machine types for all projects, and at the end of a billing period, I'm more inclined to push it into an adjustment dimension. Something more akin to a perma-conditional credit or pricing adjustment.

specification/dimensions/pricingmodel.md
flanakin: I love this! Is that in their cost dataset? Do you know what they call the column? I can't tell if this is a logical explanation or the definition of a value in their data.

There are a few interesting things here:
1. They use `Fixed` for purchases or monthly fees. That matches our intent. Might be good to borrow from their definition if it adds any additional clarity to what we have, if we agree the differentiation is necessary here. (I'm still on the fence given how close it is with On-Demand. But I want to know what others think.)
2. They use `Metered` instead of `On-Demand`. Logically, I can understand why they have that. I personally prefer On-Demand because of it is more explicit about the fact that there aren't any reduced prices in play. But I'm good with whatever the group thinks.
3. Technically, `Tiered` is still an on-demand rate. It's just that they decided to offer lower rates based on how much you use. From that perspective, do we expect practitioners to filter both values in when they want to see everything that wasn't a reduced, on-demand rate? That's the main value of `On-Demand` here. This might be a good, simple poll for the member group.
4. They use `Reserved`. We could use `Committed` if we want to align to the past-tense form for all words. But we do need to explicitly say this is a commitment-based discount. Should we use `Committed Usage` / `Committed Spend`? This should probably align with how we handle `Tiered`, since both splits are provided to share granularity on the levels of savings (Dynamic > Committed Usage > Committed Spend > Tiered > Metered/On-Demand). (Although "Dynamic" (interruptible/spot) is the only one that isn't past-tense 😣)

All thought-provoking things 🤔 Would love to hear others' thoughts on this!

specification/dimensions/pricingmodel.md
flanakin: Here's how IBM Cloud defines in:
> _Fixed-rate pricing is based on an agreed upon monthly charge. If you buy an additional bare metal server, virtual server, or storage resource after the start of the monthly billing period, the cost of the resource for the first month is prorated based on the number of days remaining in the billing period. The prorated cost is billed in a separate invoice._
Thanks @dwbrown2 for sharing that!

I would vote we either make this more specific, like what IBM has, or just get rid of it and use `On-Demand` for the same purpose. Would love to know what others think.

specification/dimensions/pricingmodel.md
udam-f2: I'm still struggling with this (after this many months). I feel like there's two different things we might be mixing up here. One is how the usage/purchase is priced - e.g. Fixed, Metered: Fixed, Metered:Dynamic, Metered:Tiered (seem like the types of values though it could also just be Fixed and Metered and Metered could have a sub categorization for Fixed, Dynamic, and Tiered)
The second is how I ended up paying for it on a particular line - on-demand/full-price OR committed/discounted.

Not sure if that makes sense.. but if we try to merge these two concepts together, it feels like we will have many overlaps.

flanakin: I looked at this at one point and realized that, even if they did overlap, you're ultimately getting one benefit the overrides the rest. If we look at cost, `On-Demand` > `Tiered` > `Dynamic` (spot) > `Committed Spend` > `Committed Usage`. Even if we had tiered spot usage that was covered by commitments, there's ultimately something that's driving the cost for that line item. So if a spend commitment is responsible for zero-ing out the charge, then that's what the Pricing Model value should be set to.

I ran across [this article](https://www.exoscale.com/syslog/cloud-pricing-models/) which covers it as well. I wonder if this is just standard operating procedure for cloud, where it may not make sense to go against the grain 🤔 The more I look at this, the more important I think it is that we differentiate each potential layer of price reduction, like IBM did. I'm a fan of splitting these out to be more detailed.

I propose we keep this as defined right now so we can close the PR, then start a poll to see what everyone thinks about having each of these either together or separate:
1. `Fixed` and `On-Demand`
2. `On-Demand` and `Tiered`
3. `Committed Usage` and `Committed Spend`

Based on the outcome of that, we can either update the closed spec or leave it as-is. Unless anyone sees a major issue, of course. If so, let's get that fixed now.

flanakin: We discussed this in VS2 and decided to keep Pricing Model focused on quantifying the explicit actions you take to generate savings. With that tiered pricing wouldn't fit. The good news is we've talked about a separate column (`SKU Tier`) that would identify the pricing tier. That can be used to identify the tiered pricing in a way that will actually convey the level of savings as well. We also have `Commitment Discount Type/Category` to differentiate committed usage and spend, so splitting that here may not be necessary.

specification/dimensions/pricingmodel.md
flanakin: I'll resolve this in favor of Marc's comment which is similarly questioning Fixed.

flanakin (approved):

specification/dimensions/pricingmodel.md
marc-perreaut: Thanks a lot, that clarifies.
BTW, IBM does not say what happens for the last month.

Thus, to support IBM Cloud, the value Fixed would make sense.
Maybe the description could be made more explicit, like "charges priced at a constant rate *over the billing period*".

yprikhodko: I'd vote to use On-Demand for that purpose

specification/dimensions/pricingmodel.md
yprikhodko: Another way to call Dynamic or spot is Spare Capacity Discount. Interesting to hear opinions about this name

specification/dimensions/pricingmodel.md
rodneyjoyce: IMO we should not add it now and we can always add it in a point release once we have feedback from it's use in the wild. It's easier to add then to remove/update. Will close this for now

specification/dimensions/pricingmodel.md
rodneyjoyce: Vote here to keep or remove: https://finopsfoundation.slack.com/archives/C053CNCRVCL/p1698650850102459

specification/dimensions/pricingmodel.md
```diff
@@ -0,0 +1,39 @@
...
+| On-Demand                 | Charges priced at the standard rate for the billing account. This includes any applicable negotiated discounts and volume/tiered pricing, but does not include dynamic or commitment-based discount pricing. |
```
salpaysenturus, line 33:
```suggestion
| Standard                | Charges priced at the standard rate for the billing account. This includes any applicable negotiated discounts and volume/tiered pricing, but does not include dynamic or commitment-based discount pricing. |
```

"On-Demand" here is a tricky term.  To my mind, the primary definition of that term is “I ran this resource whenever I wanted” instead of “I paid the non-discounted rate”.  (I can respect the perspective that “On Demand” might sometimes mean one or the other — but our own [current glossary draft](https://docs.google.com/document/d/1PvJrjXSeqtrbVAHW-aht1gzLM47EJnzX3jmjWQ8ZxkE/edit) makes no mention of rate when defining the term.)  A resource can be both run on demand _and_ covered by a CBD, and as such, these proposed groupings become potentially confusing for the practitioner.

In terms of FOCUS values in a Pricing Model attribute, how about sidestepping that semantic conflict and using a term like “**Standard**” (or “Full”) instead of “On Demand”?

specification/dimensions/pricingmodel.md
rodneyjoyce: It is removed

specification/dimensions/pricingmodel.md
rodneyjoyce: VS2 Squad considered this, and we have decided that On-Demand is a more generic, understood term than Standard, which would invite discussions as it is broader. We will leave it as On-Demand for now and move this to Maintainers for review. This conversation will not be closed though to invite further comments on this.

specification/dimensions/pricingmodel.md
```diff
@@ -0,0 +1,38 @@
+# Pricing Model
+
+Pricing Model indicates the type of unit price applied to a charge at the time it was used or purchased. Pricing Model is commonly used to differentiate between charges at on-demand vs. reduced prices when investigating optimization opportunities, like for commitment-based discount coverage.
+
+The PricingModel column MUST be present and MUST NOT be null or empty. This column is of type String and MUST be one of the allowed values. PricingModel MUST be "Commitment-Based Discount" when CommitmentDiscountId is not null. PricingModel MUST be "Dynamic" when pricing is determined by the provider and may change over time, regardless of predetermined agreement pricing.
```
thecloudman, line 5: Capitalise not for CommitmentDiscountId is not null

thecloudman (commented):
I think this is looking good. Just one comment, should we cahnge "not" to NOT" in the description where you say when CommitmentDiscountId is not...

specification/dimensions/pricingmodel.md
flowerinthenight: Looking at the merged items in our `working_draft`, there's no need to capitalize when it's inside the condition part.

[Statement] if [condition], or
[Statement] when [condition].

flowerinthenight (approved):

ljadvey (approved):

ljadvey (approved):

specification/dimensions/pricingmodel.md
udam-f2: To me, from a pricing / purchasing model perspective, on-demand means it was purchased without any prior commitment. It has nothing to do with if a negotiated rate was applied or not. The same row has metrics that have metrics that are with negotiated discounts and some without. I'm good with on-demand instead of standard/fixed IF we can make this nullable and not have everything default to on-demand which reduces the value of this column (tax, support, and to a lesser degree, purchases aren't on-demand IMO). If it has to be non-nullable, then it feels like we need another value that can cover some of the cases like tax, support etc.

supporting_content/dimensions/pricingmodel.md
```diff
@@ -0,0 +1,33 @@
...
+## Example usage scenarios
+
+Current values observed in billing data for various scenarios:
+
+| Provider | Data set                 | PricingModel     |      |
+|----------|--------------------------|----------------------------|------------------------------------------|
+| AWS | CUR ()                   |                 |              |
+| AWS | CUR ()                   |                |                    |
+| Google Cloud | BigQuery Billing Export |         |                            |
+| Google Cloud | BigQuery Billing Export |         |                               |
```
udam-f2, line 22: Should this be populated with values you've listed above as example values?

supporting_content/dimensions/pricingmodel.md
```diff
@@ -0,0 +1,33 @@
...
+| Microsoft | Cost Details (PricingModel)| Reservation                    | Reservation                      |
+| Microsoft | Cost Details (PricingModel)| Spot                    | Reservation                      |
+| Microsoft | Cost Details (PricingModel)| OnDemand                    | Reservation                      |
+
+## Documentation
+- AWS: https://docs.aws.amazon.com/cur/latest/userguide/product-columns.html
+- GCP
+- Microsoft:  Understand usage details fields: https://learn.microsoft.com/azure/cost-management-billing/automate/understand-usage-details-fields
+
+## Discussion Topics
```
udam-f2, line 33: Seems like we've had a lot of valuable conversations on this that should be captured here some how? I know there's been lots of twists and turns here but at least what we can recall should be captured so anyone wants to pick this up in the future can figure out why we went the direction we did.

specification/dimensions/pricingmodel.md
marc-perreaut: I agree with Udam's proposition to make Pricing Model nullable, because it's about price. For example a tax has arguably a price. Support has a price: maybe support could be charged based on actual hourly support, so on-demand, or on a fixed monthly amount?

> Pricing Model indicates the type of unit price applied to a charge at the time it was used or purchased.

rodneyjoyce: Ok, we have now added NULLABILTY (to be used when SkuPriceID is NULL and an OTHER option (for when a Pricing Model is present, but does not fit in to any of the other options (we can then expand on this in a future version once we capture more data). Please review and tell me if your comments above have been addressed or if more changes are needed...

supporting_content/dimensions/pricingmodel.md
rodneyjoyce: It SHOULD BE use, but it COULD NOT be due to time. I will do it now

supporting_content/dimensions/pricingmodel.md
rodneyjoyce: We need a Note taking bot to give us the transcripts and summary!

specification/dimensions/pricingmodel.md
```diff
@@ -0,0 +1,39 @@
...
+| Data type       | String           |
+| Allows nulls    | False             |
+| Value format    | list-of-values   |
+
+Allowed values:
+
+| Value                     | Description                                                                                                                                                                                                  |
+| :------------------------ | :----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| On-Demand                 | Charges priced at the standard rate for the billing account. This includes any applicable negotiated discounts and volume/tiered pricing, but does not include dynamic or commitment-based discount pricing. |
+| Dynamic                   | Charges priced at a variable rate determined by the provider. This includes any product or service with a unit price the provider can change without notice, like interruptible or low priority resources.   |
```
sarahcloudpm, line 33: this seems to overlap with on-demand -- you can have dynamic on-demand pricing

specification/dimensions/pricingmodel.md
```diff
@@ -0,0 +1,39 @@
...
+| Commitment-Based Discount | Charges with reduced prices due to a commitment-based discount specified by the Commitment Discount ID.                                                                                                      |
+| Other | Use when a PricingModel is present but none of the other allowed values apply. |
```
sarahcloudpm, line 35: what value does other vs. null have?

specification/dimensions/pricingmodel.md
```diff
@@ -0,0 +1,39 @@
...
+
+Type of unit price applied to a charge at the time it was used or purchased.
+
+## Content constraints
+
+|    Constraint   |      Value       |
+|:----------------|:-----------------|
+| Column required | True             |
+| Data type       | String           |
+| Allows nulls    | False             |
```
TDubovchenko, line 25: On line 5 it states: "**MUST be null** if SkuPriceId is null", however the column is defined as not allowing nulls.

specification/dimensions/pricingmodel.md
mike-finopsorg: Other is intentional, if you have a SKU that does not use the other pricing models the options would be use NULL against the definition of the column or use one of the other options anyhow reducing the value of those other values. Other allows the practitioner to identify how much of their spend is covered by other pricing models.

specification/dimensions/pricingmodel.md
mike-finopsorg, line 25:
```suggestion
| Allows nulls    | True             |
```

rodneyjoyce:
> I think this is looking good. Just one comment, should we cahnge "not" to NOT" in the description where you say when CommitmentDiscountId is not...

I did a search across all PRs and will leave it as lowercase to be consistent

specification/dimensions/pricingmodel.md
rodneyjoyce: Good catch, that slipped past me

rodneyjoyce: Thanks, fixed

supporting_content/dimensions/pricingmodel.md
flanakin: Ya know, Teams has that built-in 😉

specification/dimensions/pricingmodel.md
flanakin: We discussed that and determined "On-Demand" was the clearest term for the non-committed, non-spot pricing. That said, we also talked about a second level of detail here. We decided to push that to post-1.0 due to time.

What are you defining as "dynamic on-demand pricing"?

specification/dimensions/pricingmodel.md
```diff
@@ -0,0 +1,39 @@
...
+| :------------------------ | :----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| On-Demand                 | Charges priced at the standard rate for the billing account. This includes any applicable negotiated discounts and volume/tiered pricing, but does not include dynamic or commitment-based discount pricing. |
+| Dynamic                   | Charges priced at a variable rate determined by the provider. This includes any product or service with a unit price the provider can change without notice, like interruptible or low priority resources.   |
+| Commitment-Based Discount | Charges with reduced prices due to a commitment-based discount specified by the Commitment Discount ID.                                                                                                      |
+| Other | Use when a PricingModel is present but none of the other allowed values apply. |
```
mike-finopsorg, line 35: After looking at this probably for way too long I have a proposal I would like the community to consider here. There is questions about the use of dynamic vs on-demand for Spot/Interruptible/Preemptive instances. There are examples of how this differs across CSPs. I am proposing we rename this dimension to Pricing Category with a future Pricing Subcategory to come.

The definition of values already here works ok when paired with some possible subcategories. E.g.

Description | Pricing Category | Pricing Subcategory | Charge Frequency | Commitment Discount ID
-- | -- | -- | -- | --
Purchase something as a one-time buy (cert, domain) | On-Demand | Standard | One-Time | NULL
Purchase something on a recurring basis (RI with partial upfront) | On-Demand | Standard | Recurring | NULL
Use something based on on-demand prices with no commitments (EBS volume usage) | On-Demand | Standard | Usage-Based | NULL
Use something based on tiered prices with no commitments (S3 storage) | On-Demand | Tiered | Usage-Based | NULL
Use something based on a commitment (EC2 with commitment) | Commitment-Based | Spend | Usage-Based | ID:123
Use a Spot instance | Dynamic | Spot | Usage-Based | NULL

I like the consistency this builds with category and subcategory. It also avoids the conflict in naming for PricingModels and PurchaseModels as this is just putting pricing into different categories that the FinOps practitioner is looking for. 

Some examples of how this can be used by the practitioner:

- In order to find usage that is priced at list price `ChargeCategory=Usage,PricingCategory=On-Demand`
- In order to find usage that is spot `ChargeCategory=Usage,PricingSubcategory=Spot`
- To work out how much of your usage has tiering `ChargeCategory=Usage,PricingCategory=On-Demand,PricingSubcategory=Tiered`

specification/dimensions/pricingmodel.md
```diff
@@ -0,0 +1,39 @@
...
+
+| Value                     | Description                                                                                                                                                                                                  |
+| :------------------------ | :----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| On-Demand                 | Charges priced at the standard rate for the billing account. This includes any applicable negotiated discounts and volume/tiered pricing, but does not include dynamic or commitment-based discount pricing. |
+| Dynamic                   | Charges priced at a variable rate determined by the provider. This includes any product or service with a unit price the provider can change without notice, like interruptible or low priority resources.   |
+| Commitment-Based Discount | Charges with reduced prices due to a commitment-based discount specified by the Commitment Discount ID.                                                                                                      |
```
flanakin, line 34: nit: Could we simpilify this to "Commitment-Based" pricing? I'm not sure the "Discount" adds much. Thoughts?
```suggestion
| Commitment-Based | Charges with reduced prices due to a commitment-based discount specified by the Commitment Discount ID.                                                                                                      |
```

specification/dimensions/pricingmodel.md
flanakin: I love this! I can imagine a lot of possibilities with PricingSubcategory to cover more scenarios that we weren't able to get in. I was debating using ChargeSubcategory for that (hence the null over there), but I do feel like this makes _much_ more sense, given these things are truly about "pricing" and not a generic explanation of the type of "usage".

specification/dimensions/pricingmodel.md
ljadvey: Agreed

specification/dimensions/pricingmodel.md
ljadvey: Simply: Yes!

flanakin: Tiering (aka volume-based discounts) is a pricing model where you get a built-in discount the more you use. You don't "select" a tier, but you do select a service that leverages tiering. You may not have a choice for that specific service, but there are likely other services that can be used that don't have tiering. More importantly, this also applies vice-versa. Let's say you use a DB that isn't tiered, then you see some DB charges are tiered and others aren't. As an optimization, you might move that untiered workload over to the tiered service to get more out of what you're spending. (This assumes all other things being equal, of course.)

flanakin: Turns out spot isn't a commercial thing at all. It's a general financial/stock term that refers to the price something can be immediately sold for "right now", noting that this price is always changing based on supply and demand. Sound familiar? 🙂 We should be fine without PricingType, but it's nice to have that option, if we do need it. I suspect anything like that would end up in a SKU column.

flanakin: > Side question about "Purchase something on a recurring basis (RI with partial upfront)" having PricingCategory=On-Demand. If this is about regular payment for reservations, I would see PricingCategory=Commitment-Based, so as that I can identify the Purchase of commitment with ChargeType=Purchase and PricingModel=CommitmentBased. Or should such identification be done in another way?

I like the idea of making commitment purchases easily identifiable, but if the purpose of this column is to identify how that purchase was "priced", then it wouldn't be commitment-based pricing. Yes, you're purchasing a disocunt, but that purchase isn't based on a commitment. (Unless it is, then it's a different story 😜) You can use CommitmentDiscountCategory for that, but you'd need to filter out nulls rather than filtering to a specific value.

supporting_content/dimensions/pricingmodel.md
rodneyjoyce: +1 for Teams ;)

specification/dimensions/pricingmodel.md
rodneyjoyce: I like the concept - the fact that we always were coming back to discussions on PricingModel suggested that there was an issue with overloading it, so I think the sub category that further defines it makes sense to me. 

@sarahcloudpm Thoughts on this?

specification/dimensions/pricingmodel.md
rodneyjoyce: See Mike's new proposal below and let me know if that meets this requirement @sarahcloudpm ?

flanakin: According to this definition, dynamic is for variable rates that change and on-demand is a standard, predetermined rate, so we should be good. I'll go ahead and resolve this one, but please reopen it if we're missing anything.

specification/dimensions/pricingmodel.md
flanakin: Updated. We'll keep PricingSubcategory separate given the time. If anyone feels it's critical for 1.0 and we have enough people willing to close it quickly, we can bring it in. Please speak up in Slack 🙂

specification/dimensions/pricingcategory.md
```diff
@@ -0,0 +1,47 @@
+# Pricing Category
+
+Pricing Category describes the kind of pricing model used for a charge at the time it was used or purchased. Pricing Category is commonly used for scenarios like differentiating between charges at on-demand vs. reduced prices when investigating optimization opportunities, like for commitment-based discount coverage.
```
aqu-erp, line 3:
```suggestion
Pricing Category describes the pricing model used for a charge at the time of use or purchase. It can be useful for distinguishing between charges with [List Unit Prices] #ListUnitPrice or reduced prices, exposing optimization opportunities such as increasing commitment-based discount coverage
```

aqu-erp (approved):

specification/dimensions/pricingcategory.md
```diff
@@ -0,0 +1,47 @@
...
+| Column required | True           |
+| Data type       | String         |
+| Allows nulls    | True           |
+| Value format    | list-of-values |
+
+Allowed values:
+
+| Value            | Description                                                                                                                                                                                                  |
+| :--------------- | :----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| On-Demand        | Charges priced at the standard rate for the billing account. This includes any applicable negotiated discounts and volume/tiered pricing, but does not include dynamic or commitment-based discount pricing. |
```
udam-f2, line 40: I think On-demand isn't a good value here if I'm holistically looking to define this highest-level categorization. I say that knowing the term on-demand has been used to describe a few different things by cloud providers including the pricing model. The issue with on-demand is, as Shawn called out in a previous iteration, it means you're able to purchase something without purchasing discounts or committing to something. Any regular Joe/Jane can walk up and get this price with no commitments. To me, dynamic priced offerings can be described the same way. While they offer reduced rates, a consumer is getting the same exact pricing whenever the SKU is made avaialble at the published price - It's consumed in an on-demand manner.

For me, Fixed or Standard pricing seems like a better option here. If we introduce Fixed|Standard here, Tiered and Dynamic also naturally fit in as a pricing category. On-demand and commitment-based move to the purchase-model (or pricing subcategory in post v1.0).

Given it's late, I'm ok for us to push forward with this for 1.0 preview but curious to see what others think. 


| Description                                                                    | Pricing Model / Pricing Category | Purchasing Model / Pricing Subcategory |
|--------------------------------------------------------------------------------|----------------------------------|----------------------------------------|
| Purchase something as a one-time buy (cert, domain)                            | Fixed                            | On-demand                              |
| Purchase something on a recurring basis (RI with partial upfront)              | Fixed                            | On-demand                              |
| Use something based on on-demand prices with no commitments (EBS volume usage) | Fixed                            | On-demand                              |
| Use something based on tiered prices with no commitments (S3 storage)          | Tiered                           | On-demand                              |
| Use something based on a commitment (EC2 with RI commitment)                   | Discounted                       | Commitment-based                       |
| Unused commitment for a RI                                                     | Discounted                       | Commitment-based                       |
| Use something based on a commitment (EC2 with RI commitment) negation          | Discounted                       | Commitment-based                       |
| Use a Spot instance                                                            | Dynamic                          | Spot                                   |
| I Purchase something that is part of a bundle                                  | Fixed                            | Bundled                                |
| I use a service offered as preview pricing                                     | Fixed                            | Preview                                |
| Use a Interuptable VM (OCI)                                                    | Fixed                            | Interuptable                           |

specification/dimensions/pricingmodel.md
marc-perreaut: Thanks Michael for all the explanations!

About Tiering, I would optimize the cost of a DB workload by moving it from one DB technology (SKU) to another if it makes the cost decrease. It may happen that the cost decreases when moving from a Tiered DB technology to a non-Tiered DB technology. Moreover, there are other Tiering pricing dimensions like for example Standard/Premium Tier in Azure DataBricks. A cost optimization is indeed to switch from Premium to Standard when it's not needed. In the end, that's why I have a feeling that it will be difficult for Pricing Model / Category & Sub Category to reflect different pricing dimensions like Tier-ing and it makes more sense to me to keep it only high-level, i.e. typically on-demand / commitment-based / interruptible.

Thanks about the explanation of spot! But then spot would also apply for AWS reservation marketplace? Moreover, does spot imply that the SKU can be interrupted (different SLA from non-spot)?

I am not very comfortable with identifying purchases of commitments with CommitmentDiscountCategory, because it would indeed apply in the theoretical (funny) case where the purchase is based on a commitment. BTW, in this case we would have two commit IDs: the one used for the purchase, and the one resulting from the purchase. Will provider ever introduce commitment-based purchases of commitments? In that case, I would introduce a PurchasingPricingCategory and a PurchasingCommitmentDiscountId. Indeed, I would really prefer to identify purchases and usages of commitments in the same way, based on PricingCategory.

Maybe I miss some background: feel free to ignore these comments, as the V1.0 deadline is coming soon.

specification/dimensions/pricingcategory.md
aqu-erp: I think directionally, the feedback on "levels" of categorization make sense for post v1.0 enhancements. Current looks good to go. 

Based on the example, sounds like "Fixed" would be constrained to charges where `SkuId` does not have many  `SkuPriceIds`, in order for `Tiered`, `Discounted`, `Dynamic` to make sense as peer high-level categorizations. 

I don't think we currently have any collisions, but could there be a scenario in the future where providers include `Tiered` priced offerings as eligible for 'Discounted' mechanisms as well? Particularly with pricing models like GCP Bigquery (editions, capacity vs on-demand, omni, etc.) we will likely need to be prescriptive with how the data is mapped to columns. 

In a way `PricingCategory` is also describing publicly available opportunities for rate-optimization. 

#### Local SSD charge pricingCategory based on its VM
Fixed: Standard VM
Dynamic: Spot VM
Tiered: n/a
Discounted: 1 or 3 yr CUD with local SSD reservation

BenShy (approved):

specification/dimensions/pricingcategory.md
flanakin: We've discussed this a few times and keep coming back to On-Demand being the clearest, most obvious term that practitioners are already familiar with.
```suggestion
| On-Demand        | Charges priced at the standard rate for the billing account. This includes any flat rate and volume/tiered pricing but does not include dynamic or commitment-based discount pricing. |
```

specification/dimensions/pricingcategory.md
```diff
@@ -0,0 +1,47 @@
...
+| Dynamic          | Charges priced at a variable rate determined by the provider. This includes any product or service with a unit price the provider can change without notice, like interruptible or low priority resources.   |
+| Commitment-Based | Charges with reduced prices due to a commitment-based discount specified by the Commitment Discount ID.                                                                                                      |
```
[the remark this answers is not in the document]
flanakin: I covered this in the supporting content, but the concern with Discounted is that it groups too many different types of pricing models together. There's a big difference between a commitment-based discount and a flat rate discount and other discounts that will come in the future. If we group them all together, then it dilutes the goal of differentiating the categories of discounted pricing approaches that the column is seeking.

jasonkellyAP (approved):
LGTM

specification/dimensions/pricingcategory.md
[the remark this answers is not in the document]
rodneyjoyce: I was not quite sure how to put my answer into words, but I think I agree with the way @flanakin described it - in this column we are interested in the _manner_ in which it was priced (and in this case it was a discount based on a commitment).

cnharris10 (approved):

specification/dimensions/pricingcategory.md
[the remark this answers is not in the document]
flanakin: Resolving this so we can close this out.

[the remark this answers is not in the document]
lbernalposada: Hi team, if the value is 'Commitment-Based' are we making sure in other column there is a arn to id the commitment, I'm saying this because I have an open conversation with our AWS TAM, since the purchase option data from AWS CUR is not enough so we need to add a CASE to validate that if it has a commitment arn it should be mark as 'Commitment-Based'

mike-finopsorg (approved):

PR #217, opened 2023-10-12
Remove broken ChargeType link

flanakin:
Removing a link that was accidentally left in 0.5 when we moved the allowed values into the column spec.

### DIFF
--- a/specification/dimensions/chargetype.md
+++ b/specification/dimensions/chargetype.md
@@ -8,8 +8,6 @@ allocation logic to purchases or adjustments.
 The ChargeType column MUST be present and MUST NOT be null or empty. This column is of type String and MUST be one of the
 allowed values.
 
-See [Appendix: Charge Type](#chargetype-1) for details about allowed values and governance criteria for this dimension.
-
 ## Column ID
 
 ChargeType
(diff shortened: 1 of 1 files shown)


salpaysenturus (approved):
Awesome, looks good!  I'll admit I didn't even see the values had been embedded in the column definition because I couldn't get past the broken link.  That might sound dumb, but I'm sure other people have done the same!

specification/dimensions/chargetype.md
```diff
@@ -25,17 +23,17 @@ after-the-fact adjustment (e.g., credits), or taxes.
...
+| Data type       | String         |
+| Allows nulls    | False          |
+| Value format    | list-of-values |
 
 Allowed values:
 
 | Value      | Description                                                                                                                                                                   |
-|:-----------|:------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
+| :--------- | :---------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
 | Adjustment | Any adjustments that are applied after the original usage or purchase record. Adjustments may be related to multiple charges.                                                 |
```
SanjnaSrivatsaVMware, line 37: Do out-of-cycle bills, refunds and credits apply as an Adjustment?

flanakin: Refunds and credits on those out-of-cycle bills would be adjustments. Other line items would show as they normally would. We're more focused on the cost details than invoice generation today.

flanakin: Let's discuss in Slack if you think there's a gap here we should account for in 1.0.

ljadvey (approved):

udam-f2 (approved):

PR #218, opened 2023-10-13
FOCUS #223: Charge Subcategory column

thecloudman:


specification/dimensions/adjustmenttype.md
```diff
@@ -0,0 +1,54 @@
+# adjustment type
+
+Adjustment Type indicates the record represents a change to your monthly invoice through after-the-fact adustements (e.g. refunds, pricing changes). An adjustment can be positive or negative.
+Adjustment Type can be used by practitioners to assess any impacts to spend from positive or negative adjustments.
```
flanakin, line 4: Not all organizations have monthly invoices. But given this happens before the invoice, I'm wondering if we need that explicit reference at all.

Should we start with an intro definition to what an "adjustment" is? I pulled in the definition from ChargeType.

Can we be a little more precise with the use case? I feel like there's something better but will suggest something as a starting point. (I also aligned it to how we usually phrase the scenarios.)
```suggestion
An adjustment is a positive or negative change in cost applied after the original usage or purchase record(s). Adjustments may be related to one or more charges and are identified by the ChargeType column.

Adjustment Type indicates what kind of after-the-fact adjustment the record represents. Adjustment Type is commonly used to identify changes like credits and refunds.
```

specification/dimensions/adjustmenttype.md
```diff
@@ -0,0 +1,54 @@
...
+
+See [Appendix: adjustment type) for details about allowed values and governance criteria for this dimension.
+
+## Column ID
+
+AdjustmentType
+
+## Display Name
+
+Adjustments
```
flanakin, line 30:
```suggestion
Adjustment Type
```

specification/dimensions/adjustmenttype.md
```diff
@@ -0,0 +1,54 @@
+# adjustment type
+
+Adjustment Type indicates the record represents a change to your monthly invoice through after-the-fact adustements (e.g. refunds, pricing changes). An adjustment can be positive or negative.
+Adjustment Type can be used by practitioners to assess any impacts to spend from positive or negative adjustments.
+
+The AdjustmentType column MUST be present and MUST NOT be null or empty. This column is of type String and MUST be one of the allowed values.
```
flanakin, line 6: We need to qualify when it's null and when it's not.
```suggestion
The AdjustmentType column MUST be present and MUST NOT be null or empty when ChargeType is "Adjustment". AdjustmentType MUST be null when ChargeType is not "Adjustment". This column is of type String and MUST be one of the allowed values.
```

specification/dimensions/adjustmenttype.md
```diff
@@ -0,0 +1,54 @@
...
+- General Adjustment
+- Promo Credit
+- Refund - Reservations
+- Refund - Savings Plan
+- Refund - General usage
+- Tax error
+- Rounding Error
+- Account Closure
+
+See [Appendix: adjustment type) for details about allowed values and governance criteria for this dimension.
```
flanakin, line 22: We don't need this.
```suggestion
```

specification/dimensions/adjustmenttype.md
```diff
@@ -0,0 +1,54 @@
+# adjustment type
+
+Adjustment Type indicates the record represents a change to your monthly invoice through after-the-fact adustements (e.g. refunds, pricing changes). An adjustment can be positive or negative.
+Adjustment Type can be used by practitioners to assess any impacts to spend from positive or negative adjustments.
+
+The AdjustmentType column MUST be present and MUST NOT be null or empty. This column is of type String and MUST be one of the allowed values.
+
+- Usage Correction
```
flanakin, line 8: Move all of these to the table below and add definitions.

Did you get these from somewhere? This is a lot more detailed than I was expecting 🙂 

We should add an "Other" option since we likely won't catch everything.

specification/dimensions/adjustmenttype.md
```diff
@@ -0,0 +1,54 @@
...
+- Price Correction
+- GoodWill
+- SLA Violation
+- Balance Transfer
+- General Adjustment
```
flanakin, line 13: Should we have something so generic? I worry a provider might only use this value to abide by the spec without giving practitioners what they're actually looking for.

specification/dimensions/adjustmenttype.md
```diff
@@ -0,0 +1,54 @@
...
+- Promo Credit
+- Refund - Reservations
+- Refund - Savings Plan
+- Refund - General usage
```
flanakin, line 17: We should only have one of these. We have other columns to indicate whether it's related to a commitment. Also "reservation" and "savings plan" are provider-specific terms.

flanakin (changes requested):
This is a good start! Don't forget to add supporting content.

specification/dimensions/adjustmenttype.md
thecloudman: I got these from invoices i have had from Google, AWS and Azure.

specification/dimensions/adjustmenttype.md
thecloudman: I think we should specify these but also allow providers to put in what they want. These are what i have seen in invoices and examples from vendors.

specification/dimensions/adjustmenttype.md
thecloudman: This was to show a refund could be multiple types of refund. In Azure, reservation refunds show up in reservations transactions and not usage details. So thought we should maybe call out different refund types.

specification/dimensions/adjustmenttype.md
flanakin: Can you add the explicit things you saw and how they're tracked to the supporting content?

specification/dimensions/adjustmenttype.md
flanakin: Ah, so was this your suggestion instead of "Other"? If that's the case, I wonder if we should put a statement like "MUST/SHOULD NOT use "General Adjustment" unless a more specific adjustment type does not exist." (Or maybe a better wording, since not logic is difficult to follow.)

specification/dimensions/adjustmenttype.md
flanakin: Understood. My point is that when a refund is for a reservation, the provider should indicate that in the CommitmentDiscountId/Name/Type/Plan columns.

flanakin: We have an underlying goal to make single-purpose columns that are easy to use for summarization and filtering. In this case, we'd use AdjustmentType to identify refunds and then other columns to identify what the refund was for.

specification/dimensions/adjustmenttype.md
thecloudman: I have updated the whole thing and used SHOULD for allowed values examples

thecloudman (commented):
Updated with new information and requested changes

specification/dimensions/adjustmenttype.md
```diff
@@ -0,0 +1,78 @@
...
+- Price Correction
+- GoodWill
+- SLA Violation
+- Balance Transfer
+- General Adjustment
+- Promo Credit
+- Refund
+- Tax error
+- Rounding Error
+- Account Closure
```
flanakin, line 59: Sharing this based on our discussion. Not intending to change what we have here...
```
- Refund
  - Usage Correction (quantity?)
  - Price Correction
  - Tax error
- Credit
  - Promo Credit
  - SLA Violation
  - Account Closure
  - GoodWill
- General Adjustment
  - Price Correction
  - Rounding Error (should this be separate)
  - Balance Transfer
```

specification/dimensions/adjustmenttype.md
```diff
@@ -0,0 +1,78 @@
...
+
+AdjustmentType
+
+## Display Name
+
+Adjustment Type
+
+## Description
+
+Indicates whether the record represents an after-the-fact change to an invoice or cost of usage that already occurred (e.g., refunds, credits).
```
aqu-erp, line 32: Might just be a style thing, but I don't think the positioning of "whether" between two alternatives is necessary here. This column indicates charges from retrospective changes, not charges related to discrete usage.

aqu-erp (approved):

specification/dimensions/adjustmenttype.md
```diff
@@ -0,0 +1,78 @@
...
+- Refund
+- Tax error
+- Rounding Error
+- Account Closure
+|
+
+Allowed Value Type Definition:
+|
+- Usage Correction | Applied when vendor has incorrectly billed usage - example, Virual machine ran for 700 hours but was billed for 730 hours, the adjustment would show usage correction, with comments or notes stating a correction was made in billing for usage of virtual machines due to billing calculation errors.
+- Price Correction | Applied as an adjustment where a vendor has incorrecly applied the price to a service, this could be a negative or positive value.
```
mattray, line 65: "incorrectly"

specification/dimensions/adjustmenttype.md
```diff
@@ -0,0 +1,78 @@
...
+- Promo Credit
+- Refund
+- Tax error
+- Rounding Error
+- Account Closure
+|
+
+Allowed Value Type Definition:
+|
+- Usage Correction | Applied when vendor has incorrectly billed usage - example, Virual machine ran for 700 hours but was billed for 730 hours, the adjustment would show usage correction, with comments or notes stating a correction was made in billing for usage of virtual machines due to billing calculation errors.
```
mattray, line 64: "virtual"

specification/dimensions/adjustmenttype.md
```diff
@@ -0,0 +1,78 @@
...
+- Price Correction | Applied as an adjustment where a vendor has incorrecly applied the price to a service, this could be a negative or positive value.
+- GoodWill | Similar to General Adjustment
+- SLA Violation | Applied after customers request refunds for SLA violations. SLA Violations are where a consumed service did not meet the vendor specified SLO.
+- Balance Transfer | Applied to an account where a customer migrates an account from one account to another where the migrated account had a positive balance | Applied where an account had a positive balance that was not consumed in the previous month
+- General Adjustment | Any adjustment the vendor applies to a customer account that does not have a specific type or is an adjusment negotiated by the customer with the vendor
```
mattray, line 69: "adjustment" not "adjusment"

specification/dimensions/adjustmenttype.md
thecloudman: I have chosen to leave the examples as they where just reordered a bit

specification/dimensions/adjustmenttype.md
thecloudman: removed "whether"

specification/dimensions/adjustmenttype.md
```diff
@@ -0,0 +1,53 @@
+# Adjustment Type
+
+An adjustment is a positive or negative change in cost applied after the original usage or purchase record(s). Adjustments may be related to one or more charges and are identified by the ChargeType column.
+
+Adjustment Type indicates what kind of after-the-fact adjustment the record represents. Adjustment Type is commonly used to identify changes like credits and refunds.
+
+The AdjustmentType column MUST be present and MUST NOT be null or empty when ChargeType is "Adjustment". AdjustmentType MUST be null when ChargeType is not "Adjustment". This column is of type String and SHOULD be one of the example Allowed values or a value of choosing from the vendor.
```
flanakin, line 7:
```suggestion
The AdjustmentType column MUST be present and MUST NOT be null or empty when ChargeType is "Adjustment". AdjustmentType MUST be null when ChargeType is not "Adjustment". This column is of type String and SHOULD be one of the example allowed values or a value of choosing from the vendor.
```

specification/dimensions/adjustmenttype.md
```diff
@@ -0,0 +1,53 @@
...
+
+AdjustmentType
+
+## Display Name
+
+Adjustment Type
+
+## Description
+
+Indicates the record represents an after-the-fact change to an invoice or cost of usage that already occurred (e.g., refunds, credits).
```
flanakin, line 19: We usually take the first sentence and remove the column name for descriptions...
```suggestion
A positive or negative change in cost applied after the original usage or purchase record(s).
```

specification/dimensions/adjustmenttype.md
```diff
@@ -0,0 +1,53 @@
...
+
+Allowed values:
+
+| Value      | Description                                                                                                                                                                   |
+|:-----------|:------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
+| Adjustment | Any adjustments that are applied after the original usage or purchase record. Adjustments may be related to multiple charges.  Adjustment types can be any of the example values or a value the vendor chooses, but MUST be added if ChargeType is of type Adjustment.
+
+
+Allowed Value Type Definition:
+| Value | Description |
```
flanakin, line 38: We don't need this extra table
```suggestion
```

thecloudman:
Updates have been made, ready for review and approval

specification/dimensions/adjustmenttype.md
```diff
@@ -0,0 +1,43 @@
...
+|:----------------|:-----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
+| Refund | Applied by the vendor where a refund is given to the vendor, refund can be multiple types, the refund type should be noted in the column ChargeType, this could be for example a refund for unused reservations, or unused pre-commit amounts.
+| Usage Correction | Applied when vendor has incorrectly billed usage - example, a virtual machine ran for 700 hours but was billed for 730 hours, the adjustment would show usage correction, with comments or notes stating a correction was made in billing for usage of virtual machines due to billing calculation errors.
+| SLA Violation | Applied after customers request refunds for SLA violations. SLA Violations are where a consumed service did not meet the vendor specified SLO.
+| Tax error | Applied where a vendor has applied the wrong tax calculations
```
flanakin, line 37: Aren't these all refunds?

specification/dimensions/adjustmenttype.md
```diff
@@ -0,0 +1,43 @@
...
+| Constraint      | Value                                    |
+| :-------------- | :--------------------------------------- |
+| Column required | True                                     |
+| Data type       | String                                   |
+| Allows nulls    | True                                     |
+| Value format    | list-of-values                           |
+
+Allowed values:
+
+| Value      | Description                                                                                                                                                                   |
```
flanakin, line 32: What about credits?

specification/dimensions/adjustmenttype.md
```diff
@@ -0,0 +1,43 @@
...
+## Content Constraints
+
+| Constraint      | Value                                    |
+| :-------------- | :--------------------------------------- |
+| Column required | True                                     |
+| Data type       | String                                   |
+| Allows nulls    | True                                     |
+| Value format    | list-of-values                           |
+
+Allowed values:
```
flanakin, line 30: I believe we show this as a level 3 header to match others...
```suggestion
### Allowed values
```

specification/dimensions/adjustmenttype.md
thecloudman: Technically yes, but it would be good to have these values to describe the refund. Simply having "Refund" would not provide enough detail to categorise the refund.

specification/dimensions/adjustmenttype.md
```diff
@@ -0,0 +1,40 @@
...
+## Content Constraints
+
+| Constraint      | Value                                    |
+| :-------------- | :--------------------------------------- |
+| Column required | True                                     |
+| Data type       | String                                   |
+| Allows nulls    | True                                     |
+| Value format    | list-of-values                           |
+
+### Allowed values
```
flanakin, line 30:
```suggestion
Allowed values:
```

specification/dimensions/adjustmenttype.md
```diff
@@ -0,0 +1,40 @@
...
+|:----------------|:-----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
+| Refund | Applied by the vendor where a refund is given to the vendor, refund can be multiple types, the refund type should be noted in the column ChargeType, this could be for example a refund for unused reservations, or unused pre-commit amounts.
+|Credits  | Applied when vendor has incorrectly billed usage or applied a credit for usage such as a promo credit, tax error, SLA violation, rounding error, usage correction or pricing correction.
+| Balance Transfer | Applied to an account where a customer migrates an account from one account to another where the migrated account had a positive balance | Applied where an account had a positive balance that was not consumed in the previous month
+| General Adjustment | Any adjustment the vendor applies to a customer account that does not have a specific type
```
flanakin, line 37:
```suggestion
| General Adjustment | Any adjustment the vendor applies to a customer account that does not have a specific type. |

```

jasonkellyAP (approved):
LGTM

specification/dimensions/chargesubcategory.md
```diff
@@ -0,0 +1,62 @@
...
+
+- The ChargeSubcategory MUST be present in the billing data.
+- ChargeSubcategory is of type String and MUST be one of the allowed values.
+- ChargeSubcategory MUST NOT be null or empty when ChargeType is "Usage" and the charge is covered by a commitment.
+  - When a usage charge is covered by a commitment, ChargeSubcategory MUST be "Commitment Used".
+  - When a commitment is not used within the committed period, ChargeSubcategory MUST be "Commitment Unused" for the unused usage charge.
+- ChargeSubcategory MUST NOT be null or empty when ChargeType is "Adjustment".
+  - When ChargeSubcategory is "Refund" or "Credit", the charge MUST be negative.
+  - When an adjustment applies to a specific item, the corresponding FOCUS columns that identify that item MUST NOT be null and MUST match the applicable item details the adjustment pertains to.
+- ChargeSubcategory MUST be null when ChargeType is "Tax" or "Usage" and not covered by a commitment.
```
thecloudman, line 21: ChargeSubcategory MUST be null when ChargeType is "Tax" or "Usage" and not covered by a commitment or adjustment.

specification/dimensions/chargesubcategory.md
```diff
@@ -0,0 +1,62 @@
...
+| :---------------- | :--------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| Commitment Used   | This Charge Subcategory value shows that the usage is covered by and benefiting from discounted rates due to pre-existing commitment agreements, such as Reserved Instances or Savings Plans.                                                                                                                                                              |
+| Commitment Unused | This value would represent the portion of a commitment that has not been used. For example, if an organization has reserved instances but is not utilizing all of them, the unused portion falls under this category. It highlights an area where the organization is not fully leveraging its commitments, which could be a lost cost-saving opportunity. |
```
thecloudman, line 49: Do we need this line: It highlights an area where the organization is not fully leveraging its commitments, which could be a lost cost-saving opportunity. |

thecloudman (commented):
Two comments to be reviewed

specification/dimensions/chargesubcategory.md
```diff
@@ -0,0 +1,62 @@
...
+
+When Charge Type is "Usage", the Charge Subcategory specifies if the usage is accounted for by pre-purchased commitments or if the usage has not benefited from these discounts.
+
+When Charge Type is "Adjustment", the Charge Subcategory indicates what kind of after-the-fact adjustment the record represents. Adjustment Category is commonly used to identify changes like credits and refunds.
+
+ChargeSubcategory MUST abide by the following requirements:
+
+- The ChargeSubcategory MUST be present in the billing data.
+- ChargeSubcategory is of type String and MUST be one of the allowed values.
+- ChargeSubcategory MUST NOT be null or empty when ChargeType is "Usage" and the charge is covered by a commitment.
```
thecloudman, line 15: ChargeSubcategory MUST NOT be null and MUST NOT be empty when ChargeType is "Usage" and the charge is covered by a commitment.

thecloudman (commented):
Adding more MUST NOT

specification/dimensions/chargesubcategory.md
rodneyjoyce, line 49:
```suggestion
| Commitment Unused | This value would represent the portion of a commitment that has not been used. For example, if an organization has Reserved Instances but is not utilizing all of them, the unused portion falls under this category. It highlights an area where the organization is not fully leveraging its commitments, which could be a lost cost-saving opportunity. |
```

rodneyjoyce (commented):
Minor change as to be consistent with previous sentence

specification/dimensions/chargesubcategory.md
```diff
@@ -0,0 +1,62 @@
+# Charge Subcategory
+
+Charge Subcategory is a detailed descriptor within the billing and usage reports that acts as a supplementary detail to the main ChargeType column. It essentially dissects the primary category of charge to indicate more specific information about how each cost item relates to predefined pricing commitments.
+
+This linkage to the parent ChargeType means that for every entry under ChargeType, there is a corresponding Charge Subcategory that further refines the nature of the charge. It's a nested level of detail that allows users to see not just what type of charge was incurred, but also how that charge interacts with their purchasing agreements and discounts.
+
+When Charge Type is "Usage", the Charge Subcategory specifies if the usage is accounted for by pre-purchased commitments or if the usage has not benefited from these discounts.
+
+When Charge Type is "Adjustment", the Charge Subcategory indicates what kind of after-the-fact adjustment the record represents. Adjustment Category is commonly used to identify changes like credits and refunds.
```
flowerinthenight, line 9:
```suggestion
When Charge Type is "Adjustment", the Charge Subcategory indicates what kind of after-the-fact adjustment the record represents. Adjustment type is commonly used to identify changes like credits and refunds.
```

specification/dimensions/chargesubcategory.md
```diff
@@ -0,0 +1,62 @@
...
+
+ChargeSubcategory MUST abide by the following requirements:
+
+- The ChargeSubcategory MUST be present in the billing data.
+- ChargeSubcategory is of type String and MUST be one of the allowed values.
+- ChargeSubcategory MUST NOT be null or empty when ChargeType is "Usage" and the charge is covered by a commitment.
+  - When a usage charge is covered by a commitment, ChargeSubcategory MUST be "Commitment Used".
+  - When a commitment is not used within the committed period, ChargeSubcategory MUST be "Commitment Unused" for the unused usage charge.
```
flowerinthenight, line 17:
```suggestion
  - When a commitment is not used within the committed period, ChargeSubcategory MUST be "Commitment Unused" for the unused usage charge.
- ChargeSubcategory MUST be null when ChargeType is "Usage" and is not covered by a commitment.
```

specification/dimensions/chargesubcategory.md
flowerinthenight, line 21:
```suggestion
- ChargeSubcategory MUST be null when ChargeType is "Tax".
```

specification/dimensions/chargesubcategory.md
```diff
@@ -0,0 +1,63 @@
+# Charge Subcategory
+
+Charge Subcategory is a detailed descriptor within the billing and usage reports that acts as a supplementary detail to the main ChargeType column. It essentially dissects the primary category of charge to indicate more specific information about how each cost item relates to predefined pricing commitments.
```
flanakin, line 3: Does "detailed descriptor" make it sound like it'd be a textual description? Suggesting some tweaks to simplify and align with ChargeType...
```suggestion
Charge Subcategory indicates what kind of usage or adjustment the row represents. Charge Subcategory is a supplemental detail to the Charge Type. It provide additional context to describe the primary category of charge.
```

specification/dimensions/chargesubcategory.md
```diff
@@ -0,0 +1,63 @@
...
+- The ChargeSubcategory MUST be present in the billing data.
+- ChargeSubcategory is of type String and MUST be one of the allowed values.
+- ChargeSubcategory MUST NOT be null or empty when ChargeType is "Usage" and the charge is covered by a commitment.
+  - When a usage charge is covered by a commitment, ChargeSubcategory MUST be "Commitment Used".
+  - When a commitment is not used within the committed period, ChargeSubcategory MUST be "Commitment Unused" for the unused usage charge.
+- ChargeSubcategory MUST be null when ChargeType is "Usage" and is not covered by a commitment.
+- ChargeSubcategory MUST NOT be null or empty when ChargeType is "Adjustment".
+  - When ChargeSubcategory is "Refund" or "Credit", the charge MUST be negative.
+  - When an adjustment applies to a specific item, the corresponding FOCUS columns that identify that item MUST NOT be null and MUST match the applicable item details the adjustment pertains to.
+- ChargeSubcategory MUST be null when ChargeType is "Tax".
```
flanakin, line 22: Looks like "Purchase" was missed...
```suggestion
- ChargeSubcategory MUST be null when ChargeType is "Purchase" or "Tax".
```

specification/dimensions/chargesubcategory.md
```diff
@@ -0,0 +1,63 @@
...
+
+ChargeSubcategory
+
+## Display Name
+
+Charge Subcategory
+
+## Description
+
+Charge Subcategory is a detailed descriptor within the billing and usage reports that acts as a supplementary detail to the main ChargeType column.
```
flanakin, line 34:
```suggestion
Indicates what kind of usage or adjustment the row represents.
```

specification/dimensions/chargesubcategory.md
```diff
@@ -0,0 +1,63 @@
...
+| :---------------- | :--------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| Commitment Used   | This Charge Subcategory value shows that the usage is covered by and benefiting from discounted rates due to pre-existing commitment agreements, such as Reserved Instances or Savings Plans.                                                                                                                                                              |
+| Commitment Unused | This value would represent the portion of a commitment that has not been used. For example, if an organization has Reserved Instances but is not utilizing all of them, the unused portion falls under this category. It highlights an area where the organization is not fully leveraging its commitments, which could be a lost cost-saving opportunity. |
```
flanakin, line 50: Every time I see "Used" and "Unused" next to each other, I worry they're so close, they might get glossed over. What do others think? Should we use "Not Used" to distinguish them more?
```suggestion
| Commitment Not Used | This value would represent the portion of a commitment that has not been used. For example, if an organization has Reserved Instances but is not utilizing all of them, the unused portion falls under this category. It highlights an area where the organization is not fully leveraging its commitments, which could be a lost cost-saving opportunity. |
```

flanakin (approved):

flanakin (approved):

specification/dimensions/chargesubcategory.md
```diff
@@ -0,0 +1,63 @@
...
+
+When Charge Type is "Adjustment", the Charge Subcategory indicates what kind of after-the-fact adjustment the record represents. Adjustment type is commonly used to identify changes like credits and refunds.
+
+ChargeSubcategory MUST abide by the following requirements:
+
+- The ChargeSubcategory MUST be present in the billing data.
+- ChargeSubcategory is of type String and MUST be one of the allowed values.
+- ChargeSubcategory MUST NOT be null or empty when ChargeType is "Usage" and the charge is covered by a commitment.
+  - When a usage charge is covered by a commitment, ChargeSubcategory MUST be "Commitment Used".
+  - When a commitment is not used within the committed period, ChargeSubcategory MUST be "Commitment Unused" for the unused usage charge.
```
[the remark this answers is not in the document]
ljadvey: Yes - I like Not Used vs. Unused

specification/dimensions/chargesubcategory.md
```diff
@@ -0,0 +1,63 @@
+# Charge Subcategory
+
+Charge Subcategory indicates what kind of usage or adjustment the row represents. Charge Subcategory is a supplemental detail to the Charge Type. It provide additional context to describe the primary category of charge.
+
+This linkage to the parent ChargeType means that for every entry under ChargeType, there is a corresponding Charge Subcategory that further refines the nature of the charge. It's a nested level of detail that allows users to see not just what type of charge was incurred, but also how that charge interacts with their purchasing agreements and discounts.
+
+When Charge Type is "Usage", the Charge Subcategory specifies if the usage is accounted for by pre-purchased commitments or if the usage has not benefited from these discounts.
```
marc-perreaut, line 7: Isn't it the role of the Pricing Model dimension to specify whether the charge is covered by a commitment or not?

specification/dimensions/chargesubcategory.md
```diff
@@ -0,0 +1,63 @@
...
+
+When Charge Type is "Adjustment", the Charge Subcategory indicates what kind of after-the-fact adjustment the record represents. Adjustment type is commonly used to identify changes like credits and refunds.
```
marc-perreaut, line 9: I understood adjustments will be used also for GCP SUD.
It may be worth mentioning, as GCP SUDs are important discounts.

specification/dimensions/chargesubcategory.md
```diff
@@ -0,0 +1,63 @@
...
+| :---------------- | :--------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| Commitment Used   | Usage charges that are covered by pre-existing commitments, such as capacity commitments or commitment-based discounts.                                                                                                                                                              |
+| Commitment Not Used | Amortized usage charges for the portion of a commitment that has not been used. For example, if an organization has a commitment-based discount that is not fully utilized, the unused portion falls under this category. It highlights an area where the organization is not fully leveraging its commitments, which could be a lost cost-saving opportunity. |
```
marc-perreaut, line 50: I understand there are 2 cases for charges with Commitment Not Used:
- The charge is for the unused portion of a commitment. The charge must be present for every commitment. The value MUST be 0 if the commitment is fully used. I would not elaborate on the potentially lost cost-saving opportunity, because a commitment is beneficial as long as the effective cost is lower than the negotiated cost (on-demand equivalent cost) over the commitment period.
- The charge is for a SKU that is billed at the on-demand rate. This is currently not documented?

Is it correct?

specification/dimensions/chargesubcategory.md
```diff
@@ -0,0 +1,63 @@
+# Charge Subcategory
+
+Charge Subcategory indicates what kind of usage or adjustment the row represents. Charge Subcategory is a supplemental detail to the Charge Type. It provide additional context to describe the primary category of charge.
```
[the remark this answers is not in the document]
flanakin: Yes, we're waiting to do that after all the columns are in because several columns have references to ChargeType and we want to wait for everything to be in before we rename it to ensure nothing gets missed.

specification/dimensions/chargesubcategory.md
flanakin: This is about whether the commitment was used or not. But I see your point about that not being clear... Thoughts on this...?
```suggestion
When Charge Type is "Usage" and the charge is covered by a commitment, the Charge Subcategory indicates whether the commitment was used or not.
```

specification/dimensions/chargesubcategory.md
flanakin: We discussed this and landed on treating SUDs as credits.

specification/dimensions/chargesubcategory.md
marc-perreaut: A usage charge may not be related to a commitment (case where on-demand rate is used). For a commitment, the charge may relate to the used or unused portion. So there are 3 cases. How are they handled?

specification/dimensions/chargesubcategory.md
marc-perreaut: Maybe the long form is OK:

> Credits can be used for certain types of discounts that cover multiple usage charges, like discounts for sustained usage of resources.

specification/dimensions/chargesubcategory.md
marc-perreaut: Maybe it would be clearer to have 3 values for Usage:
- No commitment: usage billed at on-demand rate
- Commitment Used: usage covered by a commitment
- Commitment Not Used: unused commitment

chandra-cloudtrakr: I think we have a blank or null value for "no commitment" that represents the usage is billed at on-demand rate.

specification/dimensions/chargesubcategory.md
```diff
@@ -0,0 +1,63 @@
...
+
+This linkage to the parent ChargeType means that for every entry under ChargeType, there is a corresponding Charge Subcategory that further refines the nature of the charge. It's a nested level of detail that allows users to see not just what type of charge was incurred, but also how that charge interacts with their purchasing agreements and discounts.
```
udam-f2, line 5: Seems like lines 5-9 could be summarized? feels like currently it's repeating a lot of stuff that's already said above. Below is a suggestion only that could be merged with the first line. It feels like listing sub category values for each parent value pattern won't be scalable when we look at upcoming subcategorizations for ServiceCategory etc.

```suggestion
This linkage to the parent Charge Type means that for every entry under Charge Type, there can be a corresponding Charge Subcategory that further refines the nature of the charge. It's a nested level of detail that allows users to see not just what type of charge was incurred. Current sub-categorization currently applies to Charge Type values 'Usage' and 'Adjustment'. Support for others will be added as needed.
```

specification/dimensions/chargesubcategory.md
```diff
@@ -0,0 +1,63 @@
...
+
+When Charge Type is "Usage", the Charge Subcategory specifies if the usage is accounted for by pre-purchased commitments or if the usage has not benefited from these discounts.
+
+When Charge Type is "Adjustment", the Charge Subcategory indicates what kind of after-the-fact adjustment the record represents. Adjustment type is commonly used to identify changes like credits and refunds.
+
+ChargeSubcategory MUST abide by the following requirements:
```
udam-f2, line 11: I think in other places we've said something like "adheres to " can we check and make consistent.

specification/dimensions/chargesubcategory.md
```diff
@@ -0,0 +1,63 @@
...
+
+- The ChargeSubcategory MUST be present in the billing data.
+- ChargeSubcategory is of type String and MUST be one of the allowed values.
+- ChargeSubcategory MUST NOT be null or empty when ChargeType is "Usage" and the charge is covered by a commitment.
+  - When a usage charge is covered by a commitment, ChargeSubcategory MUST be "Commitment Used".
+  - When a commitment is not used within the committed period, ChargeSubcategory MUST be "Commitment Not Used" for the unused usage charge.
+- ChargeSubcategory MUST be null when ChargeType is "Usage" and is not covered by a commitment.
+- ChargeSubcategory MUST NOT be null or empty when ChargeType is "Adjustment".
+  - When ChargeSubcategory is "Refund" or "Credit", the charge MUST be negative.
```
udam-f2, line 20: Seems like this is not something to enforce here? We have different metrics where its possible not all of them are going to be guaranteed to have this dynamic?

specification/dimensions/chargesubcategory.md
marc-perreaut: IMHO, it would be clearer and better for filtering and for reporting to introduce an explicit value for Charge Sub Category for Usage when it relates to no commitment.

specification/dimensions/chargesubcategory.md
```diff
@@ -0,0 +1,63 @@
...
+## Content Constraints
+
+| Constraint      | Value          |
+| :-------------- | :------------- |
+| Column required | True           |
+| Data type       | String         |
+| Allows nulls    | True           |
+| Value format    | list-of-values |
+
+Allowed values for usage:
```
udam-f2, line 45: Do we want these to be more specific?

```suggestion
Allowed values for Charge Type `Usage`:
```

specification/dimensions/chargesubcategory.md
```diff
@@ -0,0 +1,63 @@
...
+
+| Value             | Description                                                                                                                                                                                                                                                                                                                                                |
+| :---------------- | :--------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| Commitment Used   | Usage charges that are covered by pre-existing commitments, such as capacity commitments or commitment-based discounts.                                                                                                                                                              |
```
udam-f2, line 49: Why do we want to mix the two here? Wouldn't that not allow people to sum the cost covered by (or not covered by) commitment discounts? or are we saying they'll have to use another column to differentiate this (CommitmentDiscountId?)

specification/dimensions/chargesubcategory.md
joshk0, line 45:
```suggestion
Allowed values when ChargeType is `Usage`:
```

specification/dimensions/chargesubcategory.md
```diff
@@ -0,0 +1,63 @@
...
+| Commitment Not Used | Amortized usage charges for the portion of a commitment that has not been used. For example, if an organization has a commitment-based discount that is not fully utilized, the unused portion falls under this category. It highlights an area where the organization is not fully leveraging its commitments, which could be a lost cost-saving opportunity. |
+
+Allowed values for adjustments:
```
joshk0, line 52:
```suggestion
Allowed values when ChargeType is `Adjustment`:
```

specification/dimensions/chargesubcategory.md
```diff
@@ -0,0 +1,63 @@
...
+| Refund             | Negative charges that were previously billed and are being returned by the provider. Providers can have multiple types of refunds such as resolving a tax error or for returned or exchanged commitment-based discounts. |
+| Credit             | Negative charges granted by the provider for various scenarios, like negotiated benefits, usage discounts, or promotional credits.                                                                                       |
+| Rounding Error     | Positive or negative charges that are needed to ensure raw cost and usage data aggregations match the total cost on the invoice, which may be rounded.                                                                   |
+| General Adjustment | Positive or negative charges the provider applies that do not fall into other adjustment category values.                                                                                                                |
```
udam-f2, line 59: What's an example of when someone would use General Adjustment?

joshk0 (changes requested):

specification/dimensions/chargesubcategory.md
aqu-erp: This may be a carryover from the previous unreviewed CommitmentUtilization draft. 

Agreed the distinction here creates more confusion than clarity. 

1. By distinguishing capacity commitments from commitment-based discounts (which can have a capacity commitment component ie: GCP BigQuery), it implies the capacity commitment does not provide a discount. 
2. AFAIK, usage covered a capacity reservation produces billing data that could be sufficiently captured with on-demand usage charges `chargeCategory == Usage` and `chargeSubcategory == null`
3. The implication is that capacity commitments (with no discounts) could also create charges with `chargeCategory == Usage` and `chargeSubcategory == Commitment Not Used` and I'm not sure that's true

```suggestion
| Commitment Used   | Usage charges associated with the consumption of a commitment-based discount's underlying basis.                                                                                                                              |
```

specification/dimensions/chargesubcategory.md
aqu-erp: If a CSP under-charges over a period of time, and posts a positive charge for an adjusting amount in the subsequent period, I believe it would fall under `General Adjustment`. Not sure what negative charges could be created that are not `Refund` or `Credit`

Ex: Correcting previous bills where a SKU was given a lower price than intended, or applying a breach of private pricing spend commitment (true-up)

```suggestion
| General Adjustment | Positive or negative charges the provider applies to support contractual terms or to correct billing errors that do not fall into other adjustment category values                                                                                                               |
```

specification/dimensions/chargesubcategory.md
flanakin: In that case, we're keeping this column null for now. We do think there should be a value but felt it's too late to lock it by EOW. We cover that on line 18 below.

specification/dimensions/chargesubcategory.md
flanakin: The purpose of the original CommitmentUtilization column was to cover both, so that's why we did it here. Practitioners can use PricingModel == "Commitment-Based Discount" for filtering.

specification/dimensions/chargesubcategory.md
flanakin: > The charge is for the unused portion of a commitment. The charge must be present for every commitment. The value MUST be 0 if the commitment is fully used.

The Discount Handling attribute might be a better place to convey the intent of the "Not Used" row. That row is only there _if_ there's an unused amount. Commitments that are fully utilized will not have an extra row.

> The charge is for a SKU that is billed at the on-demand rate. This is currently not documented?

Not exactly. This is where showing data would be helpful. The "Not Used" row is tied to the commitment, but not any usage (because there is no usage).

> IMHO, it would be clearer and better for filtering and for reporting to introduce an explicit value for Charge Sub Category for Usage when it relates to no commitment.

Agreed. We were keeping it null because we think there could be multiple values. If you only care about on-demand, then you can get that out of PricingModel.

specification/dimensions/chargesubcategory.md
flanakin: "charges the provider applies that do not fall into other adjustment category values"

specification/dimensions/chargesubcategory.md
```diff
@@ -0,0 +1,63 @@
+# Charge Subcategory
+
+Charge Subcategory indicates what kind of usage or adjustment the row represents. Charge Subcategory is a supplemental detail to the [Charge Type](#chargetype). It provide additional context to describe the primary category of charge.
+
+This linkage to the parent Charge Type means that for every entry under Charge Type, there can be a corresponding Charge Subcategory that further refines the nature of the charge. It's a nested level of detail that allows users to see not just what type of charge was incurred. Current sub-categorization currently applies to Charge Type values 'Usage' and 'Adjustment'. Support for others will be added as needed.
+
+When Charge Type is "Usage", the Charge Subcategory specifies if the usage is accounted for by pre-purchased commitments or if the usage has not benefited from these discounts.
+
+When Charge Type is "Adjustment", the Charge Subcategory indicates what kind of after-the-fact adjustment the record represents. Adjustment type is commonly used to identify changes like credits and refunds.
```
flanakin, line 9: Just realized we have an old "adjustment type" reference here...
```suggestion
When Charge Type is "Adjustment", the Charge Subcategory indicates what kind of after-the-fact adjustment the record represents.
```

specification/dimensions/chargesubcategory.md
flanakin: This is what we use in attributes:
```suggestion
ChargeSubcategory MUST follow the requirements listed below:
```

specification/dimensions/chargesubcategory.md
flanakin: @aqu-erp, does that also cover anything we haven't accounted for? That was the original intent.

specification/dimensions/chargesubcategory.md
marc-perreaut: > Commitments that are fully utilized will not have an extra row.

As a practitioner, I would find very convenient that are all commitments have an unused commitment charge, so that I can easily calculate every commitment utilization from FOCUS data and I find it's also better for transparency ("zero is a number like others"). The absence of unused commitment charge may also denote a data error.

specification/dimensions/chargesubcategory.md
chandra-cloudtrakr: Can we use the "Commitment Type" whether "Discount-based" or "Capacity-based". I think we have a dimension for this, the "CommitmentDiscountType", right?

chandra-cloudtrakr: Can we use the "Commitment Type" whether "Discount-based" or "Capacity-based". I think we have a dimension for this, the "CommitmentDiscountType", right?

"Charge Type" is "Usage"
"Charge Discount Type" is "Discount-based"
"Charge Subcategory" is "Commitment Used" or "Commitment Unused" or null

"Charge Type" is "Usage"
"Charge Discount Type" is "Capacity-based"
"Charge Subcategory" is "Commitment Used" or "Commitment Unused" or null

specification/dimensions/chargesubcategory.md
aqu-erp: 😆 I think `contractual terms` and `billing errors` also add color in a generic way that doesn't exclude anything, given everything on the invoice should be related to billing mechanics or described in a service agreement. I added the examples since the other descriptions had them as well, potentially seems less jarring this way.

specification/dimensions/chargesubcategory.md
aqu-erp: After further review, my presumption may have been incomplete, I changed the suggestion to cover both discounted and discount-less commitments, but retained vocabulary to use less implied terms such as "covered".  @flanakin @chandra-cloudtrakr 

```suggestion
| Commitment Used   | Usage charges that are associated with consumption of a commitment's underlying basis.                                                                                                                                              |
```

joshk0 (approved):
Modulo pending discussions, I am good to go on this one

specification/dimensions/chargesubcategory.md
thecloudman: Null works here because you can then use the null value with commitment charges value to determine unused value.

thecloudman: if all agree @marc-perreaut @flanakin  i will resolve this as we keep null

specification/dimensions/chargesubcategory.md
thecloudman: @udam-f2  not sure what you are inferring here? Do you want to suggest a change?

specification/dimensions/chargesubcategory.md
chandra-cloudtrakr: @thecloudman  - I'm resolving this conversation in the VS2 meeting

specification/dimensions/chargesubcategory.md
flanakin: Instead of having this line here, we should add a "MUST be negative" statement to the columns that we're requiring to be negative. For now, let's remove this from here. We can discuss adding it to other columns if we feel it's required post-1.0.
```suggestion
```

flanakin (approved):

specification/dimensions/chargesubcategory.md
thecloudman: keeping the original and can be addressed post 1.0

specification/dimensions/chargesubcategory.md
```diff
@@ -0,0 +1,62 @@
...
+| :---------------- | :--------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| Commitment Used   | Usage charges that are associated with consumption of a commitment's underlying basis.                                                                                                                                              |
+| Commitment Not Used | Amortized usage charges for the portion of a commitment that has not been used. For example, if an organization has a commitment-based discount that is not fully utilized, the unused portion falls under this category. It highlights an area where the organization is not fully leveraging its commitments, which could be a lost cost-saving opportunity. |
```
mike-finopsorg, line 49:
```suggestion
| Used Commitment   | Usage charges that are associated with consumption of a commitment's underlying basis.                                                                                                                                              |
| Unused Commitment | Amortized usage charges for the portion of a commitment that has not been used. For example, if an organization has a commitment-based discount that is not fully utilized, the unused portion falls under this category. It highlights an area where the organization is not fully leveraging its commitments, which could be a lost cost-saving opportunity. |
| On-Demand |  Usage charges that are associated with a commitment |
```

I think `Commitment Not Used` is very unclear if this means a commitment wasn't applied to this row or there was commitment that wasn't used in the row. I think while the `Unused` vs `Used` doesn't offer a large visual difference, the clarity is probably worth it. Adding an On-Demand here also means its easy to group usage costs into course groups.

specification/dimensions/chargesubcategory.md
```diff
@@ -0,0 +1,62 @@
+# Charge Subcategory
+
+Charge Subcategory indicates what kind of usage or adjustment the row represents. Charge Subcategory is a supplemental detail to the [Charge Type](#chargetype). It provide additional context to describe the primary category of charge.
```
flanakin, line 3:
```suggestion
Charge Subcategory indicates what kind of usage or adjustment the row represents. Charge Subcategory is a supplemental detail to the [Charge Type](#chargetype). It provides additional context to describe the primary category of charge.
```

specification/dimensions/chargesubcategory.md
thecloudman: @mike-finopsorg I dont think adding On-Demand should be in the ChargeSubcategory as it is already used in chargetype (chargecategory) If we added on-demand then all chargetypes would be considered subcategories

mike-finopsorg: I do not see OnDemand in the Chargetype? https://github.com/FinOps-Open-Cost-and-Usage-Spec/FOCUS_Spec/blob/working_draft/specification/dimensions/chargetype.md I think I am missing something that helps me understand your comment @thecloudman

specification/dimensions/chargesubcategory.md
```diff
@@ -0,0 +1,63 @@
+# Charge Subcategory
+
+Charge Subcategory indicates what kind of usage or adjustment the row represents. Charge Subcategory is a supplemental detail to the [Charge Type](#chargetype). It provides additional context to describe the primary category of charge.
+
+This linkage to the parent Charge Type means that for every entry under Charge Type, there can be a corresponding Charge Subcategory that further refines the nature of the charge. It's a nested level of detail that allows users to see not just what type of charge was incurred. Current sub-categorization currently applies to Charge Type values 'Usage' and 'Adjustment'. Support for others will be added as needed.
+
+When Charge Type is "Usage" and the charge is related to a commitment, the Charge Subcategory indicates whether the row represents committed usage or is an amortized charge for the unused portion of the commitment. Charge Subcategory is commonly used for scenarios like calculating commitment utilization when ChargeType is "Usage".
```
aqu-erp, line 7:
```suggestion
When Charge Type is "Usage" and the charge is related to a commitment, the Charge Subcategory indicates whether the row represents committed usage or is an amortized charge for the unused portion of the commitment. Charge Subcategory is commonly used for scenarios like calculating commitment utilization when Charge Type is "Usage".
```

specification/dimensions/chargesubcategory.md
```diff
@@ -0,0 +1,63 @@
...
+
+When Charge Type is "Adjustment", the Charge Subcategory indicates what kind of after-the-fact adjustment the record represents and is commonly used to identify changes like credits and refunds.
+
+ChargeSubcategory MUST follow the requirements listed below:
+
+- The ChargeSubcategory MUST be present in the billing data.
+- ChargeSubcategory is of type String and MUST be one of the allowed values.
+- ChargeSubcategory MUST NOT be null or empty when ChargeType is "Usage" and the charge is covered by a commitment.
+  - When a usage charge is covered by a commitment, ChargeSubcategory MUST be "Used Commitment".
+  - When a commitment is not used within the committed period, ChargeSubcategory MUST be "Unused Commitment" for the unused usage charge.
```
aqu-erp, line 17:
```suggestion
  - When a commitment is not completely used within the committed period, ChargeSubcategory MUST be "Unused Commitment" for the unused usage charge.
```

thecloudman: @aqu-erp can we say instead of not completely used, say when a commitment is not used (Fully or partially)....

udam-f2 (approved):
LGTM

Issue #185, opened 2023-10-02
Create a Tag Attribute definition


udam-f2:

kk09v:
This attribute has been renamed to Key-Value as it will be applicable to fields other than tags

PR #219, opened 2023-10-16
FOCUS #185: Key-Value Format

kk09v:
Second iteration of the Key-Value Format after much discussion and lots of help from the people mentioned below

Co-authored-by: @flanakin
Co-authored-by: @rileyjenk
Co-authored-by: @cnharris10
Co-authored-by: @udam-f2

### DIFF
--- a/specification/attributes/key_value_format.md
+++ b/specification/attributes/key_value_format.md
@@ -21,7 +21,7 @@ Rules and formatting requirements for columns appearing in billing data which co
 * Columns subject to Key-Value Format MUST be a serialized JSON string, consistent with the ECMA-404 definition of an object.
 * Keys MUST be unique within an object.
 * Values MAY be a number, string, `true`, `false`, or `null`.
-  * Values MAY NOT be an object nor an array.
+  * Values MUST NOT be an object nor an array.
 
 ## Exceptions
 
--- a/specification/attributes/key_value_format.md
+++ b/specification/attributes/key_value_format.md
@@ -1,0 +1,32 @@
+# Key-Value Format
+
+This is the specified format for providing data in the form of key-value pairs, which are also known as name-value pairs, attribute-value pairs, and field-value pairs in other contexts. 
+
+This format is often used in place of enumerating data which would be inherently sparse and/or without predetermined keys (attributes) into separate columns. To faciliate parsing, this format does not support nesting.
+
+## Attribute ID
+
+KeyValueFormat
+
+## Attribute Name
+
+Key-Value Format
+
+## Description
+
+Rules and formatting requirements for columns appearing in billing data which convey data as key-value pairs.
+
+## Requirements
+
+* Columns subject to Key-Value Format MUST be a serialized JSON string, consistent with the ECMA-404 definition of an object.
+* Keys MUST be unique within an object.
+* Values MAY be a number, string, `true`, `false`, or `null`.
+  * Values MAY NOT be an object nor an array.
+
+## Exceptions
+
+None
+
+## Introduced (version)
+
+1.0
--- a/supporting_content/attributes/key_value_format.md
+++ b/supporting_content/attributes/key_value_format.md
@@ -1,0 +1,10 @@
+# Key-Value Format
+
+## Examples
+
+- Valid key-value format demonstrating all 5 valid values: `{"KeyString":"Value1", "KeyNumber":42, "KeyBool":true, "KeyBool2":false, "KeyEmpty":"", "KeyNull":null}`
+- Disambiguating keys using a prefix: `{"parent/key1":"Value1", "child/key1":"Value2"}`
+
+## References:
+
+- [ECMA 404](https://www.ecma-international.org/wp-content/uploads/ECMA-404_2nd_edition_december_2017.pdf)
--- a/specification/attributes/attributes.mdpp
+++ b/specification/attributes/attributes.mdpp
@@ -8,4 +8,5 @@ for data granularity, recency, frequency, etc. Requirements defined in attribute
 !INCLUDE "column_naming_convention.md",1
 !INCLUDE "currency_code_format.md",1
 !INCLUDE "datetime_format.md",1
+!INCLUDE "key_value_format.md",1
 !INCLUDE "null_handling.md",1


specification/attributes/key_value_format.md
```diff
@@ -0,0 +1,32 @@
...
+## Description
+
+Rules and formatting requirements for columns appearing in billing data which convey data as key-value pairs.
+
+## Requirements
+
+* Columns subject to Key-Value Format MUST be a serialized JSON string, consistent with the ECMA-404 definition of an object.
+* Keys MUST be unique within an object.
+* Values MAY be a number, string, `true`, `false`, or `null`.
+  * Values MUST NOT be an object nor an array.
```
cnharris10, line 24:
```suggestion
 * Values MUST NOT be an object nor an array.
```

specification/attributes/key_value_format.md
```diff
@@ -0,0 +1,32 @@
+# Key-Value Format
+
+This is the specified format for providing data in the form of key-value pairs, which are also known as name-value pairs, attribute-value pairs, and field-value pairs in other contexts. 
+
+This format is often used in place of enumerating data which would be inherently sparse and/or without predetermined keys (attributes) into separate columns. To faciliate parsing, this format does not support nesting.
+
```
cnharris10, line 6: From your initial paragraphs, tried to a bit less technical.  Feel free to judge as you wish ...

The Key-Value Format specifies the format for providing data as key-value pairs, also known as name-value pairs, attribute-value pairs, and field-value pairs in other contexts.

The Key-Value Format mirrors JSON's ECMA-404 with a handful or extra requirements and in most cases, more efficiently consolidates column-based data than the CSV format.

cnharris10: Also removed this line "To faciliate parsing, this format does not support nesting." because it seemed duplicated in the requirements below

cnharris10 (changes requested):

specification/attributes/key_value_format.md
```diff
@@ -0,0 +1,32 @@
...
+## Description
+
+Rules and formatting requirements for columns appearing in billing data which convey data as key-value pairs.
+
+## Requirements
+
+* Columns subject to Key-Value Format MUST be a serialized JSON string, consistent with the ECMA-404 definition of an object.
+* Keys MUST be unique within an object.
+* Values MAY be a number, string, `true`, `false`, or `null`.
```
cnharris10, line 23: Another option: `Values MUST be either a ...`

specification/attributes/key_value_format.md
```diff
@@ -0,0 +1,32 @@
...
+
+Key-Value Format
+
+## Description
+
+Rules and formatting requirements for columns appearing in billing data which convey data as key-value pairs.
+
+## Requirements
+
+* Columns subject to Key-Value Format MUST be a serialized JSON string, consistent with the ECMA-404 definition of an object.
```
udam-f2, line 21:
```suggestion
* Columns of type Key-Value Format MUST contain serialized JSON string, consistent with the ECMA-404 definition of an object.
```

specification/attributes/key_value_format.md
udam-f2, line 23: Is it clear that this is referring to the Key and Value of each KV pair defined in this column? Line 23 for example could get confused with the overall value of the cell? (I could be misreading that)

specification/attributes/key_value_format.md
udam-f2: @cnharris10 Not seeing where this is mentioned in requirements

specification/attributes/key_value_format.md
```diff
@@ -0,0 +1,32 @@
+# Key-Value Format
+
+This is the specified format for providing data in the form of key-value pairs, which are also known as name-value pairs, attribute-value pairs, and field-value pairs in other contexts.
```
udam-f2, line 3: For consistency, you might consider adding a sentence like we do for other format attributes where we explain why this would be helpful to consumers.

specification/attributes/key_value_format.md
udam-f2, line 6: To be consistent with the other attributes, we should add a normative line here that says MUST meet all requirements listed below (in the requirements section)

cnharris10: Referring to this line, `* Values MUST NOT be an object nor an array.`.  If this is true, nesting is not possible.

cnharris10 (approved):

specification/attributes/key_value_format.md
```diff
@@ -0,0 +1,32 @@
...
+
+Key-Value Format
+
+## Description
+
+Rules and formatting requirements for columns appearing in billing data which convey data as key-value pairs.
+
+## Requirements
+
+* Key-Value Format columns MUST contain a serialized JSON string, consistent with the ECMA-404 definition of an object.
```
joshk0, line 21: Do we want to have an opinion on case sensitivity? As it stands this is permitting "keynull" and "KeYNuLl". Some CSP's have case sensitive tags, some CSP's don't.

joshk0 (commented):
Nothing to stop shipping it, but some light edge cases I was able to think of

salpaysenturus (approved):

specification/attributes/key_value_format.md
joshk0: never mind. it is probably use case dependent (e.g. this is a discussion better left for `Tags`)

supporting_content/attributes/key_value_format.md
```diff
@@ -0,0 +1,10 @@
+# Key-Value Format
+
+## Examples
+
+- Valid key-value format demonstrating all 5 valid values: `{"KeyString":"Value1", "KeyNumber":42, "KeyBool":true, "KeyBool2":false, "KeyEmpty":"", "KeyNull":null}`
```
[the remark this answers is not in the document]
joshk0: same thing, this is really about `Tags`

joshk0 (approved):

specification/attributes/key_value_format.md
```diff
@@ -0,0 +1,32 @@
+# Key-Value Format
+
+Columns that provide Key-Value information are often used in place of separate columns for enumerating data which would be inherently sparse and/or without predetermined keys. This consolidates related information and provides more consistency in the schema. Key-value pairs are also referred to as name-value pairs, attribute-value pairs, or field-value pairs.
+
+All key-value related columns defined in the FOCUS specification MUST follow the key-value formatting requirements listed below. These columns SHALL NOT contain nested elements.
```
udam-f2, line 5: Is the last sentence needed here? It's one of the requirements listed in the requirements already.

specification/attributes/key_value_format.md
```diff
@@ -0,0 +1,32 @@
...
+
+## Attribute ID
+
+KeyValueFormat
+
+## Attribute Name
+
+Key-Value Format
```
udam-f2, line 13: Should it be KeyValue Format? 

I ask because we also have Date/Time Format and it feels like it would be better to not have - or /. Otherwise we should pick one Applies here and for the top of the page header (see [here](https://github.com/FinOps-Open-Cost-and-Usage-Spec/FOCUS_Spec/blob/working_draft/specification/attributes/datetime_format.md))

ljadvey: I like Keyvalue, DateTime....

specification/attributes/key_value_format.md
ljadvey: No reason to duplicate

specification/attributes/key_value_format.md
flanakin: Wikipedia uses [key-value](https://en.wikipedia.org/wiki/Key%E2%80%93value_database) and [date/time](https://en.wikipedia.org/wiki/ISO_8601#EDTF) just to keep us on our toes 🙃 I believe "DateTime" is slightly different in that it refers to a datatype in a programming language. I vaguely recall looking that up when we went thru that one.

flanakin (approved):

specification/attributes/key_value_format.md
kk09v: To me, the Attribute Name is written in plain English as you would when using it in an email and Attribute ID is written in the way you'd use it in a system or reference it in a slack message with technical teams.

In English, it's very atypical to have a compound word written in Pascal case.

For now, we can leave it as key-value since we have it written as date/time in the working draft. If we want to change this, we should do both (and any similar examples we find) in the same PR.

specification/attributes/key_value_format.md
kk09v: It was intended to make this more approachable, but the common opinion is that this redundant comment doesn't add value.

joshk0 (approved):

udam-f2 (approved):
LGTM.

PR #221, opened 2023-10-17
Add Charge Description dimension

cnharris10:
Closes https://github.com/FinOps-Open-Cost-and-Usage-Spec/FOCUS_Spec/issues/220
[Doc](https://docs.google.com/document/d/1kMjXpEIJhpKq4dkBblg9Lk4iq8iAGJ92uWQKlXuxOF4/edit)

# Preview

image

### DIFF
--- a/specification/dimensions/chargedescription.md
+++ b/specification/dimensions/chargedescription.md
@@ -2,7 +2,7 @@
 
 A Charge Description provides a high-level context of a line item without requiring additional discovery.  This column is typically a brief, human-readable summary of a select group of corresponding details across a cost and usage dataset or provides information not otherwise available.
 
-The Charge Description column MUST be present within billing data, MUST be of type String, and SHOULD not be null.
+The Charge Description column MUST be present within billing data, MUST be of type String, and SHOULD NOT be null.
 
 ## Column ID
 
--- a/specification/dimensions/chargedescription.md
+++ b/specification/dimensions/chargedescription.md
@@ -1,6 +1,6 @@
 # Charge Description
 
-A Charge Description provides a high-level context of a line item without requiring additional discovery.  This column is typically a brief, human-readable summary of a select group of corresponding details across a cost and usage dataset or provides information not otherwise available.
+A Charge Description provides a high-level context of a line item without requiring additional discovery.  This column is a self-contained, summary explanation of what the charge is for and how it was priced.  It typically covers a select group of corresponding details across a cost and usage dataset or provides information not otherwise available.
 
 The Charge Description column MUST be present within the dataset, MUST be of type String, and SHOULD NOT be null.  Providers SHOULD specify the length of this column in their publicly available FOCUS documentation.
 
--- a/specification/dimensions/chargedescription.md
+++ b/specification/dimensions/chargedescription.md
@@ -1,6 +1,6 @@
 # Charge Description
 
-A Charge Description provides a high-level context of a line item without requiring additional discovery.  This column is typically a brief, human-readable summary of a select group of corresponding details or provides information not otherwise available.  
+A Charge Description provides a high-level context of a line item without requiring additional discovery.  This column is typically a brief, human-readable summary of a select group of corresponding details across a cost and usage dataset or provides information not otherwise available.
 
 The Charge Description column MUST be present within billing data, MUST be of type String, and SHOULD not be null.
 
--- a/specification/dimensions/chargedescription.md
+++ b/specification/dimensions/chargedescription.md
@@ -2,7 +2,7 @@
 
 A Charge Description provides a high-level context of a line item without requiring additional discovery.  This column is typically a brief, human-readable summary of a select group of corresponding details across a cost and usage dataset or provides information not otherwise available.
 
-The Charge Description column MUST be present within billing data, MUST be of type String, and SHOULD NOT be null.
+The Charge Description column MUST be present within billing data, MUST be of type String, and SHOULD NOT be null.  Providers SHOULD specify the length of this column in their publicly available FOCUS documentation. 
 
 ## Column ID
 
--- a/specification/dimensions/chargedescription.md
+++ b/specification/dimensions/chargedescription.md
@@ -2,7 +2,7 @@
 
 A Charge Description provides a high-level context of a line item without requiring additional discovery.  This column is typically a brief, human-readable summary of a select group of corresponding details across a cost and usage dataset or provides information not otherwise available.
 
-The Charge Description column MUST be present within billing data, MUST be of type String, and SHOULD NOT be null.  Providers SHOULD specify the length of this column in their publicly available FOCUS documentation.
+The Charge Description column MUST be present within the dataset, MUST be of type String, and SHOULD NOT be null.  Providers SHOULD specify the length of this column in their publicly available FOCUS documentation.
 
 ## Column ID
 
--- a/specification/dimensions/chargedescription.md
+++ b/specification/dimensions/chargedescription.md
@@ -1,0 +1,30 @@
+# Charge Description
+
+A Charge Description provides a high-level context of a line item without requiring additional discovery.  This column is typically a brief, human-readable summary of a select group of corresponding details or provides information not otherwise available.  
+
+The Charge Description column MUST be present within billing data, MUST be of type String, and SHOULD not be null.
+
+## Column ID
+
+ChargeDescription
+
+## Display Name
+
+Charge Description
+
+## Description
+
+A brief, human-readable summary of a line item.
+
+## Content Constraints
+
+|    Constraint   |      Value       |
+|:----------------|:-----------------|
+| Column required | True             |
+| Data type       | String           |
+| Allows nulls    | True             |
+| Value format    | \<not specified> |
+
+## Introduced (version)
+
+1.0
--- a/specification/dimensions/chargedescription.md
+++ b/specification/dimensions/chargedescription.md
@@ -2,7 +2,7 @@
 
 A Charge Description provides a high-level context of a line item without requiring additional discovery.  This column is typically a brief, human-readable summary of a select group of corresponding details across a cost and usage dataset or provides information not otherwise available.
 
-The Charge Description column MUST be present within billing data, MUST be of type String, and SHOULD NOT be null.  Providers SHOULD specify the length of this column in their publicly available FOCUS documentation. 
+The Charge Description column MUST be present within billing data, MUST be of type String, and SHOULD NOT be null.  Providers SHOULD specify the length of this column in their publicly available FOCUS documentation.
 
 ## Column ID
 
--- a/supporting_content/dimensions/chargedescription.md
+++ b/supporting_content/dimensions/chargedescription.md
@@ -8,7 +8,7 @@ Current resource types found or extracted from available data sets:
 | :-------- | :---------------------- | :-------------------------------------------------------|
 | AWS       | CUR                     | lineItem/LineItemDescription 				|
 | GCP       | BigQuery Billing Export | sku.description                                         |
-| Microsoft | Cost details            | Not available      			                |
+| Microsoft | Cost details            | ProductName      			                |
 
 ## Example usage scenarios
 
--- a/supporting_content/dimensions/chargedescription.md
+++ b/supporting_content/dimensions/chargedescription.md
@@ -20,8 +20,14 @@ Current resource types found or extracted from available data sets:
 
 ## Discussion Topics
 
-- Should we explicitly define the minimum required information for this column? A list of FOCUS (Irena)
+### Minimum Information Requirements vs Open-Ended
+
+We deliberated if we should be more specific and define the minimum required information for this column?
+
+- A list of FOCUS columns which we believe provide this high-level context without the need for additional discovery and thus must be included (concatenated) as part of the Charge description? (Irena)
 - Are these enough? Region/location? Qty? (Larry)
-	- What: lineItem/LineItemDescription
-	- Where: Region
-	- How (many): Qty
+  - What: lineItem/LineItemDescription
+  - Where: Region
+  - How (many): Qty
+
+Considering a variable landscape across cloud, SaaS, etc. we decided to keep it open-ended.
--- a/supporting_content/dimensions/chargedescription.md
+++ b/supporting_content/dimensions/chargedescription.md
@@ -4,21 +4,11 @@
 
 Current resource types found or extracted from available data sets:
 
-| Provider  | Data set                | Column                                            		|
+| Provider  | Data set                | Column                                            	|
 | :-------- | :---------------------- | :-------------------------------------------------------|
-| AWS       | CUR                     | lineItem/LineItemDescription 							|
-| GCP       | BigQuery Billing Export | Related to the combination of ProductName and MeterName |
-| Microsoft | Cost details            | Related to sku.description      						|
-
-## Example provider mappings
-
-Current resource types found or extracted from available data sets:
-
-| Provider  | Data set                | Column                                            		|
-| :-------- | :---------------------- | :-------------------------------------------------------|
-| AWS       | CUR                     | lineItem/LineItemDescription 							|
-| GCP       | BigQuery Billing Export | Related to the combination of ProductName and MeterName |
-| Microsoft | Cost details            | Related to sku.description      						|
+| AWS       | CUR                     | lineItem/LineItemDescription 				|
+| GCP       | BigQuery Billing Export | sku.description                                         |
+| Microsoft | Cost details            | Not available      			                |
 
 ## Example usage scenarios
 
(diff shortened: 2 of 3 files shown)


ljadvey (changes requested):
Let's be specific with the Columns here like with AWS since this is a required field

cnharris10:
> Let's be specific with the Columns here like with AWS since this is a required field

Thanks Larry, I'm assuming that you're saying that it's not enough to just add `lineItem/LineItemDescription` to the supporting content?  If that's not right, can you give a little more detail?

ljadvey:
> > Let's be specific with the Columns here like with AWS since this is a required field
> 
> Thanks Larry, I'm assuming that you're saying that it's not enough to just add `lineItem/LineItemDescription` to the supporting content? If that's not right, can you give a little more detail?

Sorry @cnharris10 :  lineItem/LineItemDescription is GOOD, the other columns had "Related to the " language in it that I am proposing we remove to be more specific like AWS has simply "lineItem/LineItemDescription"

ljadvey (approved):

specification/dimensions/chargedescription.md
```diff
@@ -0,0 +1,30 @@
+# Charge Description
+
+A Charge Description provides a high-level context of a line item without requiring additional discovery.  This column is typically a brief, human-readable summary of a select group of corresponding details across a cost and usage dataset or provides information not otherwise available.
+
+The Charge Description column MUST be present within billing data, MUST be of type String, and SHOULD not be null.
```
mattray, line 5: [RFC 2119](https://www.rfc-editor.org/rfc/rfc2119.html) says this should be "SHOULD NOT"

supporting_content/dimensions/chargedescription.md
```diff
@@ -0,0 +1,27 @@
...
+| GCP       | BigQuery Billing Export | Not available                      | Not Available            |
+| Microsoft | Cost details            | via Cost export file		   | Not Available            |
+
+## Discussion Topics
+
+- Should we explicitly define the minimum required information for this column? A list of FOCUS (Irena)
+- Are these enough? Region/location? Qty? (Larry)
+	- What: lineItem/LineItemDescription
+	- Where: Region
+	- How (many): Qty
```
ijurica, line 27:
```suggestion
### Minimum Information Requirements vs Open-Ended

We deliberated if we should be more specific and define the minimum required information for this column? 
- A list of FOCUS columns which we believe provide this high-level context without the need for additional discovery and thus must be included (concatenated) as part of the Charge description? (Irena)

- Are these enough? Region/location? Qty? (Larry)
  - What: lineItem/LineItemDescription
  - Where: Region
  - How (many): Qty
Considering a variable landscape across cloud, SaaS, etc. we decided to keep it open-ended.
```

ljadvey (approved):

ijurica (approved):

specification/dimensions/chargedescription.md
```diff
@@ -0,0 +1,30 @@
+# Charge Description
+
+A Charge Description provides a high-level context of a line item without requiring additional discovery.  This column is typically a brief, human-readable summary of a select group of corresponding details across a cost and usage dataset or provides information not otherwise available.
+
+The Charge Description column MUST be present within billing data, MUST be of type String, and SHOULD NOT be null.
```
salpaysenturus, line 5:
```suggestion
The Charge Description column MUST be present within billing data, MUST be of type String, and SHOULD NOT be null.  Providers SHOULD specify the length of this column in their publicly available FOCUS documentation. 
```

salpaysenturus: Whether we specify this per column or generally in a separate section of the spec is up for debate.

salpaysenturus (changes requested):

anatri1:
Are we sure "Charge" description is the right terminology? Is it generic enough to describe all types of line items in the FOCUS spec? Looking at some of the AWS ([AWS Cost and Usage Reports tracks your AWS usage and provides estimated charges associated with your account](https://docs.aws.amazon.com/cur/latest/userguide/what-is-cur.html)) and Azure's public documentation [,](https://learn.microsoft.com/en-us/azure/cost-management-billing/understand/understand-usage) it seems like the term "charge" is being used as a synonym of "cost".

cnharris10:
> Are we sure "Charge" description is the right terminology? Is it broad enough to describe all types of line items in the FOCUS spec? Looking at some of the AWS ([AWS Cost and Usage Reports tracks your AWS usage and provides estimated charges associated with your account](https://docs.aws.amazon.com/cur/latest/userguide/what-is-cur.html)) and Azure's public documentation https://learn.microsoft.com/en-us/azure/cost-management-billing/understand/understand-usage it seems like the term "charge" is being used as a synonym of "cost".

This is a good debate which we should have and might take some time.  The result of it though is quite is to apply here (either "ChargeDescription" or "<new_name>Description").  Given this, is there anything else in the spec that you would question?

anatri1:
No. I think the rest looks good. Thanks, Chris!

specification/dimensions/chargedescription.md
```diff
@@ -0,0 +1,30 @@
+# Charge Description
+
+A Charge Description provides a high-level context of a line item without requiring additional discovery.  This column is typically a brief, human-readable summary of a select group of corresponding details across a cost and usage dataset or provides information not otherwise available.
+
+The Charge Description column MUST be present within billing data, MUST be of type String, and SHOULD NOT be null.  Providers SHOULD specify the length of this column in their publicly available FOCUS documentation.
```
joshk0, line 5: The length thing is a bit interesting, perhaps it would be good as a spec to just set an upper bound. I've never seen anything above 200 chars in the field.

supporting_content/dimensions/chargedescription.md
```diff
@@ -0,0 +1,33 @@
...
+
+## Example provider mappings
+
+Current resource types found or extracted from available data sets:
+
+| Provider  | Data set                | Column                                            	|
+| :-------- | :---------------------- | :-------------------------------------------------------|
+| AWS       | CUR                     | lineItem/LineItemDescription 				|
+| GCP       | BigQuery Billing Export | sku.description                                         |
+| Microsoft | Cost details            | Not available      			                |
```
joshk0, line 11: I think `MeterName` is as close as it gets in Azure: [link](https://azure.microsoft.com/en-us/updates/update-reports-for-improved-meter-names/)

We can call it out as not a precise match because it requires other columns' values to have sufficient context, but this is what Ternary uses for this column.

supporting_content/dimensions/chargedescription.md
```diff
@@ -0,0 +1,33 @@
...
+| GCP       | BigQuery Billing Export | Not available                      | Not Available            |
+| Microsoft | Cost details            | via Cost export file		   | Not Available            |
+
+## Discussion Topics
+
+### Minimum Information Requirements vs Open-Ended
+
+We deliberated if we should be more specific and define the minimum required information for this column?
+
+- A list of FOCUS columns which we believe provide this high-level context without the need for additional discovery and thus must be included (concatenated) as part of the Charge description? (Irena)
```
joshk0, line 27: I would venture that the charge description should be unambiguous once combined with the  product name / service description information.

joshk0 (approved):
nothing to stop a ship it, but providing additional info for Microsoft Azure

salpaysenturus (approved):

specification/dimensions/chargedescription.md
cnharris10: SaaS descriptions are MUCH longer given that the description contains details that are not otherwise part of the an invoice, so I'd be hesitant to define a static length, and instead let providers do so.

specification/dimensions/chargedescription.md
```diff
@@ -0,0 +1,30 @@
+# Charge Description
+
+A Charge Description provides a high-level context of a line item without requiring additional discovery.  This column is typically a brief, human-readable summary of a select group of corresponding details across a cost and usage dataset or provides information not otherwise available.
```
flanakin, line 3: This isn't very explicit and will likely result in wildly different implementations making it useless for practitioners. Can we be more specific in what we want to be here and what it's for? "High-level context" could be anything.

I'd like to be explicit about the (at least conceptual) length so we don't end up with paragraphs. I believe the intent is for this to be a one-liner, right? Suggesting "single sentence" since "one-liner" is also a bit ambiguous, but there are at least rules around sentences. I'd be happy with either.
```suggestion
A Charge Description provides a self-contained, summary explanation of what the charge is for and how it was priced in a single sentence. The Charge Description depicts the SKU that was used or purchased, applicable discounts, and any attributes that impact the price relative to other variations of the same SKU (e.g., different regions). The column should be able to stand on its own as a clear description of what is being charged for without any additional columns. The exact price, quantity, and cost do not need to be included.
```

specification/dimensions/chargedescription.md
```diff
@@ -0,0 +1,30 @@
...
+
+ChargeDescription
+
+## Display Name
+
+Charge Description
+
+## Description
+
+A brief, human-readable summary of a line item.
```
flanakin, line 17: Pairing with my other suggestion above...
```suggestion
Self-contained, summary explanation of what the charge is for and how it was priced in a single sentence.
```

supporting_content/dimensions/chargedescription.md
flanakin: Actually, `ProductName` is the same thing as `sku.description`, but with a little more context. 
```suggestion
| Microsoft | Cost details            | ProductName      			                |
```

supporting_content/dimensions/chargedescription.md
flanakin: I think we need to be explicit here at least at a conceptual level. I'm fine if we don't give the algorithm for concatenation, but we need to be clear for providers who don't do this today. If we aren't, it's going to be wildly inconsistent and ultimately useless. What happens when we get a paragraph description of marketing content here? Is that okay? I'm assuming not.

flanakin (commented):
I'm on the fence about explicitly approving. I don't have a fundamental problem with the column as-is, but it's not clear what the purpose of this column is or how to fill it out. I left suggestions that would be much more explicit. At the same time, I don't want to block moving forward.

supporting_content/dimensions/chargedescription.md
cnharris10: Given that this dimension is meant to be a bit subjective, I don't think it makes sense to try to concoct a list of columns that tries to mimic AWS' or GCP's corresponding columns.  While MeterName and ProductName may contain important aspects of the entire line item, they likely won't encompass summary.  Thoughts?

supporting_content/dimensions/chargedescription.md
cnharris10: But, we can't think of this from just a cloud perspective.  From the SaaS angle, this column is meant to act as the part of the charge that gives the most context around what is being billed.  Most of the time, from an invoice or billing/usage summary from SaaS products, these details don't exist as columns to concatenate.

specification/dimensions/chargedescription.md
joshk0: I really like this.
Speaking to the difference between ProductName/MeterName vs sku.description etc etc, I think we have a responsibility to set examples for what good SKU descriptions look like.

Not to throw barbs, but AWS is a biggest violator of the standard you have put forth here. A single line item description can cover 3 usage tiers of the exact same thing. Free tier, level 1 usage, and level 2 usage. So I think it is important for us to word it like you have, but also keep in mind that AWS will want to do their tiered pricing and it should be possible with our schema to still do that in a better way than it is being done today in CUR.

cnharris10: I like and will incorporate this portion, "... self-contained, summary explanation of what the charge is for and how it was priced ...".  I think the rest is too oriented around cloud-based cost/usage though and don't apply broadly to the SaaS intention.

joshk0: So I think I would add to your paragraph to call it out explicitly - If there are multiple pricing tiers for the same type of service usage, then, that should all fold into one charge description, and XXX column (what would it be?) should be used to distinguish between the pricing tiers.

joshk0:
I'm not able to comment on the last review in the timeline above me, but, it would be good for there to be an examples section of idiomatic charge descriptions vs non idiomatic ones, so we can ground out all of our explanations with working examples.

* "X3 Compute Core in USA"
* "Class B API Requests" (perhaps only makes sense when combined with service description")
* "Foobarsoft OS Licensing Fee"

And not

* "$0.005 for the first 1,000 API requests - Free tier" plus "$0.005 for the next 10,000 API requests" plus "$0.0025 per 1,000 requests thereafter" (includes pricing and has multiple skus for the same fundamental usage - sorry, guess i have an axe to grind 👼🏼 )
* "Megacloud Compute Platform X3 instances with Intel Core i25. Compute Engine is an awesome service and way better than the others. Use it today and try our Y5 as well!" (marketing copy)
* "R3_USAGE" (opaque sku-id/internal type identifier that is not human readable)
* "Usage" (not specific enough)

supporting_content/dimensions/chargedescription.md
joshk0: I know why you would say that! 

But I want to fight a little harder for CSP applicability. Remember we're getting articles written about us because of the applicability to public cloud. We cannot and should not walk away from that..

I am happy to address the SaaS case additionally in this copy, but I think it's super important to make sure things are super clear for current and future cloud-provider-type services.

flanakin:
> I'm not able to comment on the last review in the timeline above me

@joshk0 I'm assuming you're referring to my review comment outside the markdown. It'd probably be better to reply to the comments in the markdown where you think these things can fit in. I left comments there as well. I just wanted to explain why I wasn't explicitly signing off or requesting changes. (I try to avoid non-committal reviews 🙂)

> it would be good for there to be an examples section of idiomatic charge descriptions vs non idiomatic ones, so we can ground out all of our explanations with working examples.

I love this! I would just ask that we codify that into the normative requirements (MUST/SHOULD statements). If it's not there, then we can't hold anyone accountable.

specification/dimensions/chargedescription.md
cnharris10: @joshk0 @flanakin 

How about ...

Charge Description provides a self-contained, summary explanation of what the charge is and how it was priced, without requiring additional discovery.  This column is typically a brief, human-readable summary of a select group of corresponding details across a cost and usage dataset or provides information not otherwise available.

1. This keeps the description at a high-level (not just a CSP column)
2. It better defines "high-level" with "self-contained, summary explanation of what the charge is and how it was priced"
3. "how it was priced" implicitly encapsulate Josh's point around multiple tiers without diving into these details.  For v1.0 and our current timeline, I don't want to delineate between cloud, SaaS, and potentially other sources but we can do this for a post-v1.0.

specification/dimensions/chargedescription.md
cnharris10: "Self-contained, summary explanation of what the charge is for and how it was priced".  We have a clause in the description saying that the provider should define the length of this column

supporting_content/dimensions/chargedescription.md
```diff
@@ -0,0 +1,33 @@
...
+
+## Example usage scenarios
+
+| Provider  | Data set                | Scenario                           | Value                    |
+|:----------|:------------------------|:-----------------------------------|:-------------------------|
+| AWS       | CUR                     | Not available                      | $0.00 per GB - US West (Oregon) data transfer from US West (Northern California) |
+| GCP       | BigQuery Billing Export | Not available                      | Not Available            |
+| Microsoft | Cost details            | via Cost export file		   | Not Available            |
+
+## Discussion Topics
```
joshk0, line 21: Took a crack at this. Not sure how it works with the suggestion system, so be careful if accepting.

```suggestion
## Guiding Examples

Given that this is a free-form text field, it may be difficult for providers to understand how they should create ChargeDescriptions for their services. The following table grounds out the recommendations above with guiding examples of ChargeDescription values, whether they are recommended, and why or why not.

| Example Value | Recommended | Reason |
| ----------    | ----------- | ------ |
| X3 Compute Core in USA | Yes | Specific, concise, and may be comprehensible even without the  `ServiceName` |
| Class B API Requests | Yes | Specific and concise, even though it might only make sense when combined with the `ServiceName` |
| Foobarsoft OS Licensing Fee | Yes | Specific, concise, and may be comprehensible even without the  `ServiceName` |
| $0.005 for the first 1,000 API requests - Free tier<br/>$0.005 for the next 10,000 API requests<br/>$0.0025 per 1,000 requests thereafter | *No* | Includes pricing which is redundant with another column; has multiple descriptions for the same fundamental usage and has multiple skus for the same fundamental usage |
| Megacloud Compute Platform X3 instances with Intel Core i25. Compute Engine is an awesome service and way better than the others. Use it today and try our Y5 as well! | **No** | Not concise; is essentially marketing copy |
| R3_USAGE | *No* | Opaque SKU ID-type or provider-internal identifier that is not necessarily human readable |
| Usage | *No* | Not specific enough |

## Discussion Topics
```

supporting_content/dimensions/chargedescription.md
joshk0: added suggestion above. thanks for prodding

supporting_content/dimensions/chargedescription.md
joshk0: I think it is ok to call out that they all feel like a good fit sometimes, and there is not one exact fit in the current data format. That would be representative of my experience as a data platform creator.

joshk0: Just bumping here, I would be in favor of either setting this row to ProductName (accept Michael's suggestion), setting it to "ProductName and MeterName provide similar information", but i feel strongly it shouldn't be marked as 'unavailable'

udam-f2 (approved):
LGTM

PR #224, opened 2023-10-17
service-category-dimension update

rodneyjoyce:
All other Dimensions use this standard (checked all 0.5 and 1.0)

### DIFF
--- a/specification/dimensions/servicecategory.md
+++ b/specification/dimensions/servicecategory.md
@@ -23,7 +23,7 @@ Highest-level classification of a service based on the core function of the serv
 | Column required | True           |
 | Data type       | String         |
 | Allows nulls    | False          |
-| Value format    | List of values |
+| Value format    | list-of-values |
 
 Allowed values:
 


flanakin (approved):
This is interesting. This format doesn't lend well to the goal of "user friendly" strings. Signing off for consistency, but I'd argue we should probably go the other way 😜

@udam-f2, curious about your thoughts.

rodneyjoyce:
> This is interesting. This format doesn't lend well to the goal of "user friendly" strings. Signing off for consistency, but I'd argue we should probably go the other way 😜
> 
> @udam-f2, curious about your thoughts.



> This is interesting. This format doesn't lend well to the goal of "user friendly" strings. Signing off for consistency, but I'd argue we should probably go the other way 😜
> 
> @udam-f2, curious about your thoughts.

@udam-f2  suggested this change - at least now 0.5 and 1.0 Dimensions are all the same. I personally prefer the user-friendly/English version.

flanakin:
Let's get it consistent first and then we can handle the broad change separately, if we want.

udam-f2:
> Let's get it consistent first and then we can handle the broad change separately, if we want.

Yep, When I said it, I added it to my list of on-going spec-wide changes that we need to discuss and make to make more consistent. If you find any, let me know and I can add to my list. This one is #10 on the list :D

udam-f2 (approved):

PR #227, opened 2023-10-18
Add Tags dimension

### LINKED ISSUE #226: Create a Tags dimension

cnharris10:
Closes #226 
[Doc](https://docs.google.com/document/d/1PmVpxtGDewBfthgsJjKjGCrftH1kMvfoj4t5qu9TMtY/edit)

# Preview (updated 10/31/23)

image
image


specification/dimensions/tags.md
```diff
@@ -0,0 +1,62 @@
+# Tags
+
+Tags are the set of finalized User-Defined and/or Provider-Defined Tags assigned to Tag Sources.  Tags commonly add business context to billing data to identify and accurately allocate charges.
+
+If a Tag is subject to a set of prioritized rules for determining its value, it is finalized after a single value is chosen from a set of corresponding values.  Otherwise, a Tag and a finalized Tag are equal.
```
ljadvey, line 5: I am not understanding what "finalized tag" means...   Needs more context please.

specification/dimensions/tags.md
```diff
@@ -0,0 +1,64 @@
...
+* Providers MUST NOT alter User-Defined Tag keys.
+
+Provider-defined Tags additionally adhere to the following requirements:
+
+* Provider-defined Tags MUST be prefixed with a provider-defined prefix.
+* Providers SHOULD publish all provider-defined prefixes within their respective documentation.
+
+## Provider-Defined vs. User-Defined Tags
+
+The following is an example of one User-Defined Tag and one Provider-Defined Tag, respectively, with tag key, foo.  The 1st property is not prefixed, and the Provider has a pre-selected prefix, `marketplace/`, applied to denote that it is a different type of tag.
```
mattray, line 21: I'd suggest replacing "tag key, foo" with "tag key `foo`" and replace "The 1st property" with "The first property"

specification/dimensions/tags.md
```diff
@@ -0,0 +1,64 @@
...
+## Finalized Tags
+
+The following example shows a simplified cost and usage data with one metered Account that contains one metered Virtual Machine.  The Provider supports tag inheritance where an Account's User-Defined Tags are applied to all Resources' User-Defined Tags, where applicable.
+
+| ResourceType    | ResourceId | Column                             |
+| :---------------| :----------| :----------------------------------|
+| Account         | my-account | { "team": "web", "env": "prod" }   |
+| Virtual Machine | my-vm      | { "team": "web", *"env": "prod"* } |
+
+After all tag inheritance rules are processed, the finalized cost and usage dataset shows that the Virtual Machine Resource did inherit tag, `env:prod`, from its corresponding Account because no tag with the same key exists.  Conversely, the Virtual Machine Resource did not inherit tag, `team:web`, because the Provider's tag inheritance rules determine that aResource already containing a tag with the same key maintains that same tag.
```
mattray, line 39: "that a Resource" instead of "that aResource"

specification/dimensions/tags.md
cnharris10: Added a few options here.  Any of these clearer or not?

Option 1 (original):
```
If a Tag is subject to a set of prioritized rules for determining its value, it is finalized after a single value is chosen from a set of corresponding values.  Otherwise, a Tag and a finalized Tag are equal.
```

Option 2:
A Tag can either be static or dynamic.  If a Tag is static, its value is immutable.  If a Tag is dynamic, its value is determined by a set of predefined user or provider rules.  A finalized Tag is the final result of any static or dynamic Tag.

Option 3:
A Tag is typically unaltered after being attached to a Tag Source.  In special circumstances, predefined rules configured by a user or provider, can dynamically determine a Tag's value.  In any case, a finalized Tag is the Tag's key and resulting value.

mattray: I like option #2

mattray (approved):
LGTM

specification/dimensions/tags.md
cnharris10: @ljadvey any thoughts here?

ljadvey: Yes, #2 I agree with as well!

cnharris10: Ok, settling on Option 2 and will see how this lands with others.

specification/dimensions/tags.md
```diff
@@ -0,0 +1,64 @@
...
+```
+
+## Finalized Tags
+
+A Tag can either be static or dynamic. If a Tag is static, its value is immutable. If a Tag is dynamic, its value is determined by a set of predefined user or provider rules. A finalized Tag is the final result of any static or dynamic Tag.
+
+| ResourceType    | ResourceId | Column                             |
+| :---------------| :----------| :----------------------------------|
+| Account         | my-account | { "team": "web", "env": "prod" }   |
+| Virtual Machine | my-vm      | { "team": "web", *"env": "prod"* } |
```
udam-f2, line 37: did we mean to have different values here at the different levels? in order to show the inheritance?

specification/dimensions/tags.md
```diff
@@ -0,0 +1,64 @@
...
+```json
+    {
+        "foo":"bar",
+        "marketplace/foo": "bar"
+    }
+```
+
+## Finalized Tags
+
+A Tag can either be static or dynamic. If a Tag is static, its value is immutable. If a Tag is dynamic, its value is determined by a set of predefined user or provider rules. A finalized Tag is the final result of any static or dynamic Tag.
```
udam-f2, line 32: I think we need to figure out if we need to keep capitalizing terms we are putting in the glossary. Maybe the first one gets capitalized and hyperlinked? And subsequent ones are just in lower case? This is a spec-wide decision but maybe clearest here because of the column and concepts Tag, Tags etc.

specification/dimensions/tags.md
```diff
@@ -0,0 +1,64 @@
...
+## Description
+
+The set of finalized tags assigned to Tag Sources.
+
+## Content Constraints
+
+|    Constraint   |      Value       |
+|:----------------|:-----------------|
+| Column required | True             |
+| Data type       | Object           |
```
udam-f2, line 58: Is the Data type JSON? 
I know its confusing.. but if value format is KeyValue, maybe its clear?

specification/dimensions/tags.md
udam-f2: Wonder if we should have something above the table to say how the example works.. e.g. provider has x rules and the following tags are set:

specification/dimensions/tags.md
ljadvey: Let's be clear with JSON

specification/dimensions/tags.md
ljadvey: +1 Capitalized and hyperlinked then... then column camelcased would be helpful to pick it out.  I have seen other columns defined like this.

specification/dimensions/tags.md
```diff
@@ -0,0 +1,64 @@
+# Tags
+
+Tags are the set of finalized User-Defined and/or Provider-Defined Tags assigned to Tag Sources.  Tags commonly add business context to billing data to identify and accurately allocate charges.
+
+If a Tag is subject to a set of prioritized rules for determining its value, it is finalized after a single value is chosen from a set of corresponding values.  Otherwise, a Tag and a finalized Tag are equal.
```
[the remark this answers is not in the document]
ljadvey: Agreed, do not need.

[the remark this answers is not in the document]
cnharris10: Will remove this paragraph.  The one below is sufficient

specification/dimensions/tags.md
cnharris10: Everything capitalized is meant to be part of the Glossary

specification/dimensions/tags.md
cnharris10: @udam-f2 @ljadvey 
I assume you both mean `JSON Object` instead of just `JSON` as that is ambiguous?

specification/dimensions/tags.md
cnharris10: `*"env": "prod"*` is italicized to indicate that the value was added.  I can make this more obvious though.

cnharris10: Take a look at the new edit that first presents a list of resources and then the finalized cost and usage dataset

supporting_content/dimensions/tags.md
```diff
@@ -0,0 +1,44 @@
...
+
+## Example provider mappings
+
+Current resource types found or extracted from available data sets:
+
+| Provider  | Dataset                 | Column                                      | Hierarchical Resources | Supports Inheritance?
+| :-------- | :---------------------- | :-------------------------------------------| :-------------------------------------------| :----------
+| AWS       | CUR                     | resourceTags/user:\*, costCategories/\*       | Organization, Organizational Unit(s), Account | No
+| GCP       | BigQuery Billing Export | Tags 									 	| Resource Group, Subscription | Yes
+| Microsoft | Cost details            | tags, labels, system_labels, project.labels	| Folder(s), Project | Yes
```
TDubovchenko, line 11:
```suggestion
| GCP       | BigQuery Billing Export | tags, labels, system_labels, project.labels 									 	|  Folder(s), Project | Yes
| Microsoft | Cost details            | Tags | Subscription, Resource Group | Yes
```

TDubovchenko (dismissed):

specification/dimensions/tags.md
```diff
@@ -0,0 +1,77 @@
...
+```json
+    {
+        "foo":"bar",
+        "marketplace/foo": "bar"
+    }
+```
+
+## Finalized Tags
+
+A Tag can either be static or dynamic. If a Tag is static, its value is immutable. If a Tag is dynamic, its value is determined by a set of predefined user or provider rules. A finalized Tag is the final result of any static or dynamic Tag.
```
joshk0, line 31: It seems like this is a choice the CSP / generator of FOCUS data has to make, but, it seems sort of implicit. This looks more like a review of how CSPs do it, which doesn't seem spec-appropriate.

Perhaps something more like

> Providers MAY offer multi-tiered tagging taxonomies as part of their management capabilities. In this case, the value of Tags could be the result of applying the taxonomy and precedence rules set by the provider, potentially suppressing visibility of identical values of a tag key that were defined at a lower priority level.

ultimately the Tags end up being just a pile of JSON, and this is just a small window into how a provider ought to set it.

The following section (example) could be framed as a reference implementation for new providers who may not be sure how to implement the tag merging / precedence functionality.

specification/dimensions/tags.md
```diff
@@ -0,0 +1,77 @@
+# Tags
+
+The Tags column represents the set of finalized user-defined and/or provider-defined tags assigned to Tag Sources.  Tags commonly add business context to billing data to identify and accurately allocate charges.
+
+The Tags column adheres to the following requirements:
+
+* The Tags column MUST contain user-defined and provider-defined tags.
```
mike-finopsorg, line 7: user-defined and/or provider-defined

specification/dimensions/tags.md
```diff
@@ -0,0 +1,77 @@
+# Tags
+
+The Tags column represents the set of finalized user-defined and/or provider-defined tags assigned to Tag Sources.  Tags commonly add business context to billing data to identify and accurately allocate charges.
```
mike-finopsorg, line 3: We should describe what `finalized` means here when its first used.

specification/dimensions/tags.md
cnharris10: Yes, it is up to the CSP/generator on how to they want to or if they want to configure tag processing guidelines, but I think the purpose of this column is not it's not a pile of JSON but a specific subset of that JSON pile and however the process upstream, we expect EXACTLY 1 key and 1 value downstream, or a finalized tag.

joshk0: ok, we can revisit if people end up being confused about the static / dynamic classification, but it's definitely not a sticking point for me.

joshk0 (approved):

specification/dimensions/tags.md
```diff
@@ -0,0 +1,79 @@
...
+The set of finalized tags assigned to Tag Sources.
+
+## Content Constraints
+
+|    Constraint   |      Value       |
+|:----------------|:-----------------|
+| Column required | True             |
+| Data type       | JSON Object      |
+| Allows nulls    | True             |
+| Value format    | MUST adhere to KeyValueFormat Attribute |
```
flanakin, line 75: nit: Shouldn't this just be a simple reference to the attribute? The MUST statement is above already, right? (I'm fine either way. Just thinking we can simplify a tad. Good with whatever direction you want to go.)
```suggestion
| Value format    | Key-Value Format |
```

specification/dimensions/tags.md
```diff
@@ -0,0 +1,79 @@
...
+
+A tag is 'finalized' after a single tag value is chosen from a set of corresponding tag values for a single tag key.  Whether a tag key can have multiple, possible values is determined by the Provider.
+
+The Tags column adheres to the following requirements:
+
+* The Tags column MUST contain user-defined and/or provider-defined tags.
+* The Tags column MUST only contain finalized tags.
+* The Tags column MUST be in Key-Value Format.
+* A Tag without a specified value MUST have its value set to null.
+* Providers MUST NOT alter user-defined Tag keys or values.
```
flanakin, line 13: Should we have something like, "If a provider allows a resource or resource grouping construct to be tagged, the provider MUST include that tag in the billing data"? We've never used the term "scope", but not sure how to generically reference things that we don't talk about in FOCUS.

supporting_content/dimensions/tags.md
```diff
@@ -0,0 +1,44 @@
...
+
+## Example provider mappings
+
+Current resource types found or extracted from available data sets:
+
+| Provider  | Dataset                 | Column                                      | Hierarchical Resources | Supports Inheritance?
+| :-------- | :---------------------- | :-------------------------------------------| :-------------------------------------------| :----------
+| AWS       | CUR                     | resourceTags/user:\*, costCategories/\*       | Organization, Organizational Unit(s), Account | No
+| GCP       | BigQuery Billing Export | tags, labels, system_labels, project.labels 									 	|  Folder(s), Project | Yes
+| Microsoft | Cost details            | Tags | Subscription, Resource Group | Yes
```
flanakin, line 11: Not sure if the hierarchy is about the cloud or the billing data. If it's what's in billing data, then management group should not be in this list. Including it for completeness just in case.
```suggestion
| Microsoft | Cost details            | Tags | Billing Account, Billing Profile, Invoice Section, Department, Enrollment Account, Management Group, Subscription, Resource Group | Yes
```

flanakin (approved):

specification/dimensions/tags.md
cnharris10: Let's leave for now and revisit after 1.0

specification/dimensions/tags.md
ljadvey: @flanakin With your latest comment in mind then it'd be the following?

```suggestion
A tag can either be static or dynamic. If a tag is static, its value is immutable. If a tag is dynamic, its value is determined by a set of predefined user or provider rules. A finalized tag is the final result of any static or dynamic Tag.
```

vs. when we state the Tag column (above) we use Tag.

ljadvey (dismissed):

udam-f2 (approved):
LGTM. Couple of minor things to consider.

specification/dimensions/tags.md
flanakin: Good catch. I wasn't trying to make a statement about that. I was suggesting this because I know cases where tags are supposed to be there but aren't. I was hoping to get that covered 😅 We could tweak this to account for that...

> All tags applied to purchases, resources, or resource grouping constructs MUST be included in the billing data. If a provider supports selecting a subset of tags for inclusion in the billing data, the provider MAY only include the configured tags.

specification/dimensions/tags.md
```diff
@@ -0,0 +1,76 @@
...
+
+Tags
+
+## Display Name
+
+Tags
+
+## Description
+
+The set of finalized tags assigned to Tag Sources.
```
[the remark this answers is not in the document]
joshk0: Yeah I'm pro removing mentions of the following terms:

* static
* dynamic
* finalized

Put it another way, what is the change we want to effect, resulting attribute of data we are trying to gain, or problem we want to solve by taxonomizing tags in this fashion?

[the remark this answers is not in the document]
udam-f2: I misread this earlier. @joshk0 @flanakin What's the issue with finalized? We want to explain that this is the reduced set of values once inheritance rules have been taken into consideration.

Static/dynamic I agree with and have felt should be described differently as well.

specification/dimensions/tags.md
```diff
@@ -0,0 +1,76 @@
+# Tags
+
+The Tags column represents the set of finalized user-defined and/or provider-defined tags assigned to Tag Sources.  Tags are commonly used for scenarios like adding business context to billing data to identify and accurately allocate charges.
```
cnharris10, line 3: Option 1:
The Tags column represents the set of tags assigned to Tag Sources with respect to potential provider-defined or user-defined tag evaluations.

Option 2:
The Tags column represents the set of tags assigned to Tag Sources that respect potential provider-defined or user-defined tag evaluations.

Option 3:
The Tags column represents the set of tags assigned to Tag Sources that also account for potential provider-defined or user-defined tag evaluations

cnharris10: @flanakin @udam-f2 @mike-finopsorg Thoughts

specification/dimensions/tags.md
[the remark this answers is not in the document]
cnharris10: Leaving current text as is and we can revist post-1.0

specification/dimensions/tags.md
```diff
@@ -0,0 +1,76 @@
...
+A tag can either be static or dynamic. If a tag is static, its value is immutable. If a tag is dynamic, its value is determined by a set of predefined user or provider rules. A finalized tag is the final result of any static or dynamic Tag.
+
+As a example, let's assume 1 sub account exists with 1 virtual machine with the following details:
+
+* Sub Account
+  * id: *my-sub-account*
+  * user-defined tags: *team:ops*, *env:prod*
+* Virtual Machine
+  * id: *my-vm*
+  * user-defined tags: *team:web*
```
salpaysenturus, line 42: I think it would be worth mentioning somewhere here that, in this particular example, the finalization rules are such that the lower tier of the hierarchy trumps the higher tier, and inherits the higher tier otherwise.

specification/dimensions/tags.md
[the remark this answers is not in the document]
salpaysenturus: I agree with Michael that "finalized" in a short-form definition here is problematic because it's not a Googleable term.  Perhaps a link to the glossary with an entry for "finalized" could help solve that.  Don't delay on resolution on my part -- but I did want to weigh in.

specification/dimensions/tags.md
salpaysenturus: I think any of those options are an improvement.  The first time we see the term "finalized" then follows in the second paragraph, and we are immediately told what it means.

marc-perreaut: I am not educated enough, so "finalized" and "tag sources" look mysterious too me in first sentence, and I believe I understand actually the first sentence  when reading the following sentences.

Here is a proposition. Feel free to ignore if irrelevant

> The Tags column represents key-value pairs that are either user-defined or provided-defined and that result from an evaluation process, knowing that original tags can be defined at resource level or at different levels of a resource group hierarchy (if supported by the Provider).

specification/dimensions/tags.md
```diff
@@ -0,0 +1,76 @@
...
+The Tags column represents the set of finalized user-defined and/or provider-defined tags assigned to Tag Sources.  Tags are commonly used for scenarios like adding business context to billing data to identify and accurately allocate charges.
+
+A tag becomes 'finalized' when a single value is selected from a set of possible tag values assigned to the tag key at different levels of a resource grouping hierarchy (if supported by the Provider).
+
+The Tags column adheres to the following requirements:
+
+* The Tags column MUST contain user-defined and/or provider-defined tags.
+* The Tags column MUST only contain finalized tags.
+* The Tags column MUST be in Key-Value Format.
+* A Tag without a specified value MUST have its value set to null.
```
marc-perreaut, line 12: In my experience, I have never seen a tag without value, and I would not not know what to do with it.
I would be interested to have an example or link if possible.
In any case, feel free to resolve this comment.

specification/dimensions/tags.md
```diff
@@ -0,0 +1,76 @@
...
+```json
+    {
+        "foo":"bar",
+        "marketplace/foo": "bar"
+    }
+```
+
+## Finalized Tags
+
+A tag can either be static or dynamic. If a tag is static, its value is immutable. If a tag is dynamic, its value is determined by a set of predefined user or provider rules. A finalized tag is the final result of any static or dynamic Tag.
```
marc-perreaut, line 33: Do we need to elaborate about static or dynamic tags?

I believe the management of tags is specific for every provider.
For example in Azure, tags from subscriptions and resource groups are not automatically inherited by resources: such behavior has to be implemented by cloud users, typically by creating Azure policies.

In the end, tags are assigned by providers to charges and FOCUS only needs to specify the format for them.

specification/dimensions/tags.md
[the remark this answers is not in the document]
marc-perreaut: I believe static/dynamic/finalized are not concepts that are common to all providers.
Every provider has its own tag processing and FOCUS just needs to specify the format for the resulting tags without any assumption on the upstream processing.

specification/dimensions/tags.md
cnharris10: https://cloud.google.com/vpc/docs/add-remove-network-tags

specification/dimensions/tags.md
cnharris10: In this column, we're not exposing the universe of tags set by the provider and/or user, just the subset that tags that  are resulting, or "finalized".  If a tag, "team" has 3 different values across an Organization (`team:a`), Account (`team:b`), and Resource (`team:c`) and inheritance favors the resource, only `team:c` appears in this column.

In your comment above, we're proposing a post-1.0 column, currently called `TagDetails`, that contains every possible tag set.

specification/dimensions/tags.md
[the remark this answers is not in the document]
cnharris10: Agree with Shawn/Michael.  Will add `Finalized Tag` to the glossary

specification/dimensions/tags.md
cnharris10: Going to go with Option 3

specification/dimensions/tags.md
joshk0: I'm still quite unhappy with the static, dynamic and final terminology. Could I make a proposal that I think keeps the details you've put in but eliminates this terminology?

```suggestion
Within a provider, tag keys may be associated with multiple values, and potentially defined at different levels within the provider, such as accounts, folders, resources and other organizational constructs. Providers MUST reduce these multiple levels of definition to a single value of Tags where one key is associated with exactly one value. The method by which this is done is up to the provider's discretion, but MUST be documented for end-users of the FOCUS-formatted data.
```

If we agree to this kind of direction, it would precipitate other changes throughout the document, but let's start here.

specification/dimensions/tags.md
```diff
@@ -0,0 +1,76 @@
+# Tags
+
+The Tags column represents the set of tags assigned to Tag Sources that also account for potential provider-defined or user-defined tag evaluations.  Tags are commonly used for scenarios like adding business context to billing data to identify and accurately allocate charges.
+
+A tag becomes *finalized* when a single value is selected from a set of possible tag values assigned to the tag key.  This can occur when a tag is set at different levels of a resource grouping hierarchy (if supported by the Provider), or a tag is value is determined from a set of user-defined rules.
```
udam-f2, line 5:
```suggestion
A tag becomes *finalized* when a single value is selected from a set of possible tag values assigned to the same tag key at different levels of a resource grouping hierarchy (if supported by the Provider).
```

specification/dimensions/tags.md
```diff
@@ -0,0 +1,76 @@
...
+
+The Tags column adheres to the following requirements:
+
+* The Tags column MUST contain user-defined and/or provider-defined tags.
+* The Tags column MUST only contain finalized tags.
+* The Tags column MUST be in Key-Value Format.
+* A Tag without a specified value MUST have its value set to null.
```
udam-f2, line 12: If clearer, lets go with this. Otherwise, ignore.
```suggestion
* A Tag key without a specified value MUST have its tag value set to null.
* If Tag finalization is supported, providers MUST publish tag finalization methods and semantics within their respective documentation.
```

specification/dimensions/tags.md
marc-perreaut: I would agree to postpone the formalization of tag processing by providers and the terminology (static/dynamic/final) when the TagDetails dimension is actually introduced after FOCUS v1.0.

Meanwhile the proposition done by Joshua could be enough for v1.0.

specification/dimensions/tags.md
```diff
@@ -0,0 +1,76 @@
...
+```json
+    {
+        "foo":"bar",
+        "marketplace/foo": "bar"
+    }
+```
+
+## Finalized Tags
+
+A tag can either be static or dynamic. If a tag is static, its value is immutable. If a tag is dynamic, its value is determined by a set of predefined user or provider rules. A finalized tag is the final result of any static or dynamic Tag.
```
udam-f2, line 33: Few minor modifications to @joshk0 's description. Ignore if they don't add value.
```suggestion
Within a provider, tag keys may be associated with multiple values, and potentially defined at different levels within the provider, such as accounts, folders, resources and other resource grouping constructs. When finalizing, providers must reduce these multiple levels of definition to a single value where each key is associated with exactly one value. The method by which this is done and the semantics are up to each provider, but must be documented within their respective documentation.
```

joshk0:
Since we're going this direction I will do a follow up change to tweak, but happy for this to go in right now.

specification/dimensions/tags.md
salpaysenturus: I really really like that suggestion, @joshk0.  Really helps lay out what the deal is.

salpaysenturus: Ah -- I see now @cnharris10 committed a revised version of this per @udam-f2.  That looks good.

salpaysenturus (approved):
Let's put you to bed, grandpa.

joshk0 (approved):

Issue #157, opened 2023-08-18
Amortization handling attribute


flanakin:
Document how amortization should be handled. This should include the EffectiveUnitPrice, EffectiveCost, and how unused commitments are handled,

udam-f2:
@flanakin Could we please re-create this with the new feedback template.

flanakin:
We covered the original goals here in discount handling appendix.

PR #229, opened 2023-10-19
FOCUS #157: Discount Handling attribute

BenShy:
Adding discount handling attribute


specification/attributes/amortization_handling.md
```diff
@@ -0,0 +1,31 @@
+# Amortization Handling
```
flanakin, line 1:
```suggestion
# Amortization Handling

```

specification/attributes/amortization_handling.md
```diff
@@ -0,0 +1,31 @@
...
+Amortization is a process used to break down and spread purchase costs over the period of time or term of use. 
+
+Amortization is primarily used in purchasable commitment-based discounts. When a purchase is applicable to resources, like commitment-based discounts, the amortized cost of a resource takes the initial payment and term into account and distributes it out based on resource usage, attributing the prorated cost for each hour of billing. Amortization provides a method for FinOps practitioners to reallocate significant purchase charges to the appropriate audience in support of their cost allocation efforts. 
+
+## Attribute ID
```
flanakin, line 6:
```suggestion
## Attribute ID

```

specification/attributes/amortization_handling.md
```diff
@@ -0,0 +1,31 @@
...
+AmortizationHandling
+
+## Attribute name
```
flanakin, line 9:
```suggestion
## Attribute name

```

specification/attributes/amortization_handling.md
flanakin, line 1: Don't forget to update the `attributes.mdpp` file!

flanakin (approved):

joe-ferrero (commented):
@BenShy @flanakin Regarding Line 23 

When an unused value is rolled over from one contract to a subsequent contract - is that an exception to the "MUST" criteria? I'm familiar with a pre-paid commitment with a 10% rollover clause;  in practice - if the customer uses 92% of what they prepaid - the remaining 8% is added to the subsequent contract - but does NOT have to be purchased again.

This would get complicated in the scenario where a customer consumes 85% of what they pre-paid, rolls over 10% to a subsequent contract and has 5% that they have to add into their cost data.

Happy to work through scenarios if you believe it needs to be part of V1.0

specification/attributes/amortization_handling.md
```diff
@@ -0,0 +1,34 @@
+# Amortization Handling
+
+Amortization is a process used to break down and spread purchase costs over the period of time or term of use. 
+
+Amortization is primarily used in purchasable commitment-based discounts. When a purchase is applicable to resources, like commitment-based discounts, the amortized cost of a resource takes the initial payment and term into account and distributes it out based on resource usage, attributing the prorated cost for each hour of billing. Amortization provides a method for FinOps practitioners to reallocate significant purchase charges to the appropriate audience in support of their cost allocation efforts.
```
marc-perreaut, line 5: > When a purchase is applicable to resources, like commitment-based discounts, the amortized cost of a resource

I wonder if it's better to say that a purchase is applicable to the usage of a resource rather than to a resource.

Taking example of Azure, reservations can be made for VM usage and for Azure DataBricks (software). So for a single resource, multiple purchases can apply for the same resource: VM reservation and Azure DataBricks reservation.

marc-perreaut, line 5: >  the amortized cost of a resource 

I believe it was renamed effective cost

marc-perreaut, line 5: > and distributes it out based on resource usage,

The distribution is done to the resource usages for which the purchase applied.
Maybe it's implicit and it's fine as is.

marc-perreaut, line 5: >  to reallocate significant purchase charges to the appropriate audience

Why "significant", and for example not "full"?
Does the "appropriate audience" refers the resource usages?

To me, the main benefit of amortization is to allocate a purchase cost transparently through the resource usages where it applies.
Sorry, I am not English native speaker: maybe this is what is actually stated here.

specification/attributes/amortization_handling.md
```diff
@@ -0,0 +1,34 @@
...
+
+Amortization Handling
+
+## Description
+
+Rules and requirements defining how amortization is shown in the billing data.
+
+## Requirements
+
+* Every charge MUST include all applicable discounts. (e.g., negotiated and commitment-based discounts)
```
marc-perreaut, line 21: May negotiated costs imply a purchase, which is then amortized?

In my mind, but maybe I am not educated enough, costs are negotiated with providers based on a financial commitment, without any purchase.

Indeed, credits can be negotiated. To my knowledge, credits are not amortized.

specification/attributes/amortization_handling.md
```diff
@@ -0,0 +1,34 @@
...
+* Discounts SHOULD be applied directly to the charges they are discounting.
+* All purchases SHOULD be amortized.
+* MUST NOT add extra rows to subtract the costs due to an applied discount.
```
marc-perreaut, line 24: Maybe "discount" should be refined with "discount related to a purchase".

Indeed, GCP SUD is a discount, but it's not a related to a purchase

specification/attributes/amortization_handling.md
```diff
@@ -0,0 +1,34 @@
...
+
+## Description
+
+Rules and requirements defining how amortization is shown in the billing data.
+
+## Requirements
+
+* Every charge MUST include all applicable discounts. (e.g., negotiated and commitment-based discounts)
+* Discounts SHOULD be applied directly to the charges they are discounting.
+* All purchases SHOULD be amortized.
```
joe-ferrero, line 23: When an unused value is rolled over from one contract to a subsequent contract - is that an exception to the "MUST" criteria? I'm familiar with a pre-paid commitment with a 10% rollover clause; in practice - if the customer uses 92% of what they prepaid - the remaining 8% is added to the subsequent contract - but does NOT have to be purchased again.
This would get complicated in the scenario where a customer consumes 85% of what they pre-paid, rolls over 10% to a subsequent contract and has 5% that they have to add into their cost data.
Happy to work through scenarios if you believe it needs to be part of V1.0

specification/attributes/amortization_handling.md
```diff
@@ -0,0 +1,34 @@
...
+* MUST NOT add extra rows to subtract the costs due to an applied discount.
+* When multiple discounts apply to a single resource/product/service, but are not applicable to the entire charge, those charges should be split into separate rows.
```
mattray, line 25: "SHOULD"

specification/attributes/amortization_handling.md
BenShy: I just removed 'significant' which should read better.

specification/attributes/discount_handling.md
```diff
@@ -0,0 +1,44 @@
...
+
+While providers may use different terms to describe discounts, FOCUS identifies a discount as being a reduced price applied directly to a row or charge. Any price or cost reductions that are awarded after the fact are identified as a "Credit" Adjustment Type. One example might be when a provider offers a reduced rate after passing a certain threshold of usage or spend.
+
```
flanakin, line 8: We need a normative paragraph with the core MUST/SHOULD statements...
```suggestion

All rows defined in FOCUS MUST follow the discount handling requirements listed below.

```

flanakin (approved):

BenShy (commented):
Redoing to cover discount handling

specification/attributes/amortization_handling.md
thecloudman: Negotiated discounts can also be discounts applied to an agreement for example retail price minus negotiated discount applies to on demand charges and is not necessarily an amortized cost. So do negotiated discounts need to be called out in amortization handling?

specification/attributes/discount_handling.md
```diff
@@ -0,0 +1,44 @@
...
+* All discounts applied to a row MUST apply to the entire charge. 
+  * If a discount only applies to a subset of the charge, then that portion of the charge MUST be split into a separate row. One with the discount and one without.
+  * If multiple discounts apply, there MUST be separate rows for each unique combination of applied discounts or undiscounted amount.
+  * Each discount MUST be identifiable using existing FOCUS columns. 
+    * Discounts applied to charges/rows from a commitment-based discount MUST include a CommitmentDiscountId.
+    * If a provider applies a discount that cannot be represented by a FOCUS column, they SHOULD include additional columns to identify the source of the discount.
+* Purchased discounts (e.g., commitment-based discounts) MUST specify CommitmentUtilization on the usage rows that received the reduced rates from the discount. 
+* Purchased discounts (e.g., commitment-based discounts) MUST be amortized.
+  * The BilledCost MUST be 0 for the discounted portion of the charge and the EffectiveCost MUST be the portion of the purchase cost that applies to this row.
```
thecloudman, line 32: EffectiveCost MUST be the portion of the purchase
Change to
EffectiveCost MUST be the portion of the amortized

specification/attributes/discount_handling.md
```diff
@@ -0,0 +1,44 @@
...
+  * CommitmentUtilization MUST be "Used" for rows that received a reduced price from that commitment.
+  * If a commitment is not fully utilized, the provider MUST include a row that represents the unused portion of the commitment for that charge period. CommitmentUtilization MUST be "Not Used".
```
thecloudman, line 34: If a commitment is not fully utilized, the provider MUST include a row that represents the unused portion of the commitment for that charge period. CommitmentUtilization MUST be "Not Used".
Should this be worded to align to other columns where we say Commitment Unused

specification/attributes/discount_handling.md
```diff
@@ -0,0 +1,44 @@
...
+  * The sum of the EffectiveCost for all "Used" and "Not Used" rows for each CommitmentDiscountId MUST be the same as the BilledCost of the commitment-based discount purchase.
+* Credits that are applied after the fact MUST use a ChargeType of "Adjustment" and AdjustmentCategory of "Credit".
```
flowerinthenight, line 36:
```suggestion
* Credits that are applied after the fact MUST use a ChargeType of "Adjustment" and ChargeSubcategory of "Credit".
```

specification/attributes/discount_handling.md
```diff
@@ -0,0 +1,44 @@
...
+
+While providers may use different terms to describe discounts, FOCUS identifies a discount as being a reduced price applied directly to a row or charge. Any price or cost reductions that are awarded after the fact are identified as a "Credit" Adjustment Type. One example might be when a provider offers a reduced rate after passing a certain threshold of usage or spend.
```
flowerinthenight, line 7:
```suggestion
While providers may use different terms to describe discounts, FOCUS identifies a discount as being a reduced price applied directly to a row or charge. Any price or cost reductions that are awarded after the fact are identified as a "Credit" Charge Subcategory. One example might be when a provider offers a reduced rate after passing a certain threshold of usage or spend.
```

specification/attributes/discount_handling.md
```diff
@@ -0,0 +1,44 @@
+# Discount Handling
+
+A discount is a type of pricing model in which cloud providers offer their customers a reduced price that is applied to the usage of the cloud provider’s services. There are many types of discounts that a cloud provider can support, including but not limited to commercially negotiated discounts, commitment-based discounts that offer lower prices when you agree to a certain amount of their services or spend, and bundled discounts where you receive free or discounted usage of one product or service based on the usage of another product or service. Discount handling is commonly used in scenarios for verifying discounts were applied and calculating cost savings.
```
flanakin, line 3: Should we avoid "pricing model" given there's another column or did you want to reference that intentionally? Also generalizing to support discounts for purchases and suggesting a few ways to shorten this a bit...
```suggestion
A discount is a pricing construct where providers offer a reduced price for services. Providers may have many types of discounts, including but not limited to commercially negotiated discounts, commitment-based discounts when you agree to a certain amount of usage or spend, and bundled discounts where you receive free or discounted usage of one product or service based on the usage of another. Discount Handling is commonly used in scenarios like verifying discounts were applied and calculating cost savings.
```

specification/attributes/discount_handling.md
```diff
@@ -0,0 +1,44 @@
...
...ed prices. The most common example of this is commitment-based discounts, where you "purchase" a commitment to use or spend a specific amount within a period. There are instances where a portion of the purchased discount may go unused. This can result in effectively lost savings and needs to be clearly identifiable at a granular level to enable FinOps scenarios. In order to facilitate this, the amount of a purchased discount is amortized over the term that the discount is applied to (e.g., 1 year). Amortization is a process used to break down and spread the purchase costs over a period of time or term of use. When a purchase is applicable to resources, like commitment-based discounts, the amortized cost of a resource takes the initial payment and term into account and distributes it out based on the resource's usage, attributing the prorated cost for each hour of billing. Amortization provides a method for FinOps practitioners to reallocate purchase charges to the appropriate audience in support of their cost allocation efforts. Discount Handling for purchased commitments is commonly used for scenarios like calculating utilization and implementing chargeback of the purchase amount.
```
flanakin, line 5: Simplifying a tad to shorten it...
```suggestion
Some discount offers can be purchased from a provider to get reduced prices. The most common example is commitment-based discounts, where you "purchase" a commitment to use or spend a specific amount within a period. There are instances where a portion of the purchased discount may go unused, which results in lost savings and needs to be clearly identifiable at a granular level. To facilitate this, the purchase amount is amortized over the term the discount is applied to (e.g., 1 year). Amortization is a process used to break down and spread purchase costs over a period of time or term of use. When a purchase is applicable to resources, like commitment-based discounts, the amortized cost of a resource takes the initial payment and term into account and distributes it out based on the resource's usage, attributing the prorated cost for each hour of billing. Amortization enables FinOps practitioners to distribute purchase charges to the appropriate audience in support of cost allocation efforts. Discount Handling for purchased commitments is commonly used for scenarios like calculating utilization and implementing chargeback for the purchase amount.
```

specification/attributes/discount_handling.md
flanakin: Also adding a caveat in case the commitment only discounts a part of the cost and not the full cost.
```suggestion
  * The BilledCost MUST be 0 for any row where the commitment covers the entire cost for the charge period.
  * The EffectiveCost MUST be the portion of the amortized purchase cost that applies to this row.
```

specification/attributes/discount_handling.md
flanakin: Good catch! Applying suggestion...
```suggestion
  * If a commitment is not fully utilized, the provider MUST include a row that represents the unused portion of the commitment for that charge period. Charge Subcategory MUST be "Commitment Unused".
```

specification/attributes/discount_handling.md
```diff
@@ -0,0 +1,44 @@
...
+    * Discounts applied to charges/rows from a commitment-based discount MUST include a CommitmentDiscountId.
+    * If a provider applies a discount that cannot be represented by a FOCUS column, they SHOULD include additional columns to identify the source of the discount.
+* Purchased discounts (e.g., commitment-based discounts) MUST specify CommitmentUtilization on the usage rows that received the reduced rates from the discount. 
+* Purchased discounts (e.g., commitment-based discounts) MUST be amortized.
+  * The BilledCost MUST be 0 for the discounted portion of the charge and the EffectiveCost MUST be the portion of the purchase cost that applies to this row.
+  * CommitmentUtilization MUST be "Used" for rows that received a reduced price from that commitment.
+  * If a commitment is not fully utilized, the provider MUST include a row that represents the unused portion of the commitment for that charge period. CommitmentUtilization MUST be "Not Used".
+  * The sum of the EffectiveCost for all "Used" and "Not Used" rows for each CommitmentDiscountId MUST be the same as the BilledCost of the commitment-based discount purchase.
```
flanakin, line 37:
```suggestion
  * The sum of the EffectiveCost for all "Commitment Used" and "Commitment Unused" rows for each ChargeSubcategory MUST be the same as the BilledCost of the commitment-based discount purchase.
```

flanakin (approved):

rodneyjoyce (commented):
I agree with the changes above and have some other minor comments

specification/attributes/amortization_handling.md
flanakin: Resolving this since there was a major refactor of the attribute. Can you take another look?

specification/attributes/amortization_handling.md
flanakin: I believe this is covered in the latest version. Please take a look and leave any comments. Resolving this one, since this reference is gone now.

specification/attributes/amortization_handling.md
flanakin: This has been refactored in the latest version. Can you take a look? Resolving this for now. Please leave any comments if you notice anything.

specification/attributes/discount_handling.md
flanakin: Updated my suggestion above to merge in @rodneyjoyce's comment about using "unit" instead of "hour".

specification/attributes/amortization_handling.md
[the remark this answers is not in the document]
flanakin: Looks like this comment got flagged as outdated due to other changes. I copied the change to a suggestion I made. I'll close this one since GitHub is hiding it anyway.

specification/attributes/amortization_handling.md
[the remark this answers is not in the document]
flanakin: This was removed in the latest iteration. It now only addresses purchased discounts.

supporting_content/attributes/discount_handling.md
```diff
@@ -0,0 +1,88 @@
...
+* Partnership-based discounts
+
+## Commitment-based discounts
+
+Notes:
+
+* Usage-based commitment discounts
+* Spend-based commitment discounts
+
+## Tier-based discounts
```
marc-perreaut, line 18: Is tier-based pricing (pricing user tiers) considered as type of discount? Indeed, one can see tier-based pricing as the implementation of a volume-based discount.

Otherwise, is "tier-based discount" referring to discounts applying to specific tiers? For example, a discount could be applied on the 1st Tier to implement a trial period.

specification/attributes/discount_handling.md
```diff
@@ -0,0 +1,46 @@
...
+Discount Handling
+
+## Description
+
+Indicates how to include and apply discounts to usage charges or rows. 
+
+## Requirements
+
+* All applicable discounts MUST be applied to each row they pertain to and MUST NOT be negated in a separate row.
+* All discounts applied to a row MUST apply to the entire charge.
```
marc-perreaut, line 26: I believe this relates to the SKU Price ID, which applies to the entire charge.
Maybe we could state that the applied discount, if any, must be reflected in the SKU price, which is identified by the SKU Price ID.

specification/attributes/discount_handling.md
```diff
@@ -0,0 +1,46 @@
...
+
+## Requirements
+
+* All applicable discounts MUST be applied to each row they pertain to and MUST NOT be negated in a separate row.
+* All discounts applied to a row MUST apply to the entire charge. 
+  * If a discount only applies to a subset of the charge, then that portion of the charge MUST be split into a separate row. One with the discount and one without.
+  * If multiple discounts apply, there MUST be separate rows for each unique combination of applied discounts or undiscounted amount.
+  * Each discount MUST be identifiable using existing FOCUS columns. 
+    * Discounts applied to charges/rows from a commitment-based discount MUST include a CommitmentDiscountId.
+    * If a provider applies a discount that cannot be represented by a FOCUS column, they SHOULD include additional columns to identify the source of the discount.
```
marc-perreaut, line 31: I was asking myself how the organization discount is represented in FOCUS.
I understand it's represented by the Billing Account ID/Name: the Negotiated Cost SHOULD represents a discount of the List Cost that is related to the Billing Account ID/Name or a parent entity (eg SCE in Azure). Correct?

supporting_content/attributes/discount_handling.md
flowerinthenight: The thinking behind this tier-based discount is on the partnership level i.e. premier tier has more enterprise (or volume, etc) discounts than, say, advanced tier. But then again, we could also just consider this as under the partnership discounts. I wasn't really sure. Do you think it's better to remove? I'll add a note under Discussion / Scratch Space.

specification/attributes/discount_handling.md
flowerinthenight, line 31: Do we consider spot instance usage as a form of discount? For GCP SUD, we could consider the discount row as a General Adjustment -> Credit type but spot usage is associated with a resource but with a discounted price; and no CommitmentDiscountId. Or we just consider it as a normal Usage row but with a different (lower) price list, meaning, not discounted?

> If a provider applies a discount that cannot be represented by a FOCUS column, they SHOULD include additional columns to identify the source of the discount.
I was thinking that spot instance usages would fall under this provision but this line seems a bit open-ended. Specifically the "additional columns" part, which could potentially be handled post v1.0?

marc-perreaut: IMHO, spot is not strictly a discount, because spot instances have a different SLA than regular instances: their usage can be interrupted. I see more spot as "low-cost" than "discount".

specification/attributes/discount_handling.md
```diff
@@ -0,0 +1,46 @@
...
...ed prices. The most common example of this is commitment-based discounts, where you "purchase" a commitment to use or spend a specific amount within a period. There are instances where a portion of the purchased discount may go unused. This can result in effectively lost savings and needs to be clearly identifiable at a granular level to enable FinOps scenarios. In order to facilitate this, the amount of a purchased discount is amortized over the term that the discount is applied to (e.g., 1 year). Amortization is a process used to break down and spread the purchase costs over a period of time or term of use. When a purchase is applicable to resources, like commitment-based discounts, the amortized cost of a resource takes the initial payment and term into account and distributes it out based on the resource's usage, attributing the prorated cost for each hour of billing. Amortization provides a method for FinOps practitioners to reallocate purchase charges to the appropriate audience in support of their cost allocation efforts. Discount Handling for purchased commitments is commonly used for scenarios like calculating utilization and implementing chargeback of the purchase amount.
```
marc-perreaut, line 5: > When a purchase is applicable to resources, like commitment-based discounts, the amortized cost of a resource takes the initial payment and term into account and distributes it out based on the resource's usage, attributing the prorated cost for each hour of billing.

I would propose to refine "resource" into "resource usage", everywhere it's not done already.

Original comment: https://github.com/FinOps-Open-Cost-and-Usage-Spec/FOCUS_Spec/pull/229#discussion_r1368252300

specification/attributes/discount_handling.md
```diff
@@ -0,0 +1,46 @@
...
+
+DiscountHandling
+
+## Attribute name
+
+Discount Handling
+
+## Description
+
+Indicates how to include and apply discounts to usage charges or rows.
```
flowerinthenight, line 21:
```suggestion
Indicates how to include and apply discounts to usage charges or rows.
```
For build errors.

specification/attributes/discount_handling.md
flowerinthenight, line 26:
```suggestion
* All discounts applied to a row MUST apply to the entire charge.
```
For build errors.

specification/attributes/discount_handling.md
```diff
@@ -0,0 +1,46 @@
...
+
+Indicates how to include and apply discounts to usage charges or rows. 
+
+## Requirements
+
+* All applicable discounts MUST be applied to each row they pertain to and MUST NOT be negated in a separate row.
+* All discounts applied to a row MUST apply to the entire charge. 
+  * If a discount only applies to a subset of the charge, then that portion of the charge MUST be split into a separate row. One with the discount and one without.
+  * If multiple discounts apply, there MUST be separate rows for each unique combination of applied discounts or undiscounted amount.
+  * Each discount MUST be identifiable using existing FOCUS columns.
```
flowerinthenight, line 29:
```suggestion
  * Each discount MUST be identifiable using existing FOCUS columns.
```
For build errors.

specification/attributes/discount_handling.md
```diff
@@ -0,0 +1,46 @@
...
+## Requirements
+
+* All applicable discounts MUST be applied to each row they pertain to and MUST NOT be negated in a separate row.
+* All discounts applied to a row MUST apply to the entire charge. 
+  * If a discount only applies to a subset of the charge, then that portion of the charge MUST be split into a separate row. One with the discount and one without.
+  * If multiple discounts apply, there MUST be separate rows for each unique combination of applied discounts or undiscounted amount.
+  * Each discount MUST be identifiable using existing FOCUS columns. 
+    * Discounts applied to charges/rows from a commitment-based discount MUST include a CommitmentDiscountId.
+    * If a provider applies a discount that cannot be represented by a FOCUS column, they SHOULD include additional columns to identify the source of the discount.
+* Purchased discounts (e.g., commitment-based discounts) MUST specify CommitmentUtilization on the usage rows that received the reduced rates from the discount.
```
flowerinthenight, line 32:
```suggestion
* Purchased discounts (e.g., commitment-based discounts) MUST specify CommitmentUtilization on the usage rows that received the reduced rates from the discount.
```
For build errors.

specification/attributes/discount_handling.md
BenShy: resolving as this is for the initial pr

specification/attributes/discount_handling.md
```diff
@@ -0,0 +1,47 @@
...
+
+Some discount offers can be purchased from a provider to get reduced prices. The most common example is commitment-based discounts, where you "purchase" a commitment to use or spend a specific amount within a period. There are instances where a portion of the purchased discount may go unused, which results in lost savings and needs to be clearly identifiable at a granular level. To facilitate this, the purchase amount is amortized over the term the discount is applied to (e.g., 1 year). Amortization is a process used to break down and spread purchase costs over a period of time or term of use. When a purchase is applicable to resources, like commitment-based discounts, the amortized cost of a resource takes the initial payment and term into account and distributes it out based on the resource's usage, attributing the prorated cost for each unit of billing. Amortization enables FinOps practitioners to distribute purchase charges to the appropriate audience in support of cost allocation efforts. Discount Handling for purchased commitments is commonly used for scenarios like calculating utilization and implementing chargeback for the purchase amount.
```
joshk0, line 5: softened FinOps practitioner to user of billing data, is that spicy?

FOCUS is for everyone, not just people who are finops certified or whose orgs use finops.

```suggestion
Some discount offers can be purchased from a provider to get reduced prices. The most common example is a commitment-based discount, where you "purchase" a commitment to use or spend a specific amount within a period. There are instances where a portion of the purchased discount may go unused, which results in lost savings and needs to be clearly identifiable at a granular level. To facilitate this, the purchase amount is amortized over the term the discount is applied to (e.g., 1 year).

Amortization is a process used to break down and spread purchase costs over a period of time or term of use. When a purchase is applicable to resources, like commitment-based discounts, the amortized cost of a resource takes the initial payment and term into account and distributes it out based on the resource's usage, attributing the prorated cost for each unit of billing. Amortization enables users of billing data to distribute purchase charges to the appropriate audience in support of cost allocation efforts. Discount Handling for purchased commitments is commonly used for scenarios like calculating utilization and implementing chargeback for the purchase amount.
```

joshk0 (commented):
Lazyweb question: Does an actual FOCUS dimension allow the provider to notate when a special kind of discount (contractual being my biggest example) has applied to a row?

If there's no dimension coverage, I worry that one of the most frustrating things about CSP billing data (non-representation of EDP's and whatnot in core data sources like CUR) will go unaddressed in FOCUS (whether action is actually taken by CSPs or not on this matter, I would want to take a stand.)

In my handling of a contractual discount with a client and CSP I will not name, here is how the discounts were handled:

* List prices and costs were given in the data file;
* An excel sheet was shared with the client containing information about what types of usage would be discounted (%) or set to a fixed $ rate per usage;
* Every month, a massive Refund line item would arrive in the data file containing the total amount of discount that applied to the billing account, without any breakdown based on service, discount amount, or anything useful.
* Much work has to be done to calculate whether the CSP got the refund right, and by this time we have already hand calculated a per-line-item context for whether a discount did apply to this row or not, based on the other information in the row (which in all likelihood is how the CSP themselves issues the discount in the first place.)

My motivation as part of FOCUS would be to advocate for elimination of this toil for other users in a similar boat.

specification/attributes/discount_handling.md
```diff
@@ -0,0 +1,49 @@
...
+* Purchased discounts (e.g., commitment-based discounts) MUST specify CommitmentUtilization on the usage rows that received the reduced rates from the discount.
+* Purchased discounts (e.g., commitment-based discounts) MUST be amortized.
+  * The BilledCost MUST be 0 for any row where the commitment covers the entire cost for the charge period.
+  * The EffectiveCost MUST be the portion of the amortized purchase cost that applies to this row.
+  * CommitmentUtilization MUST be "Used" for rows that received a reduced price from that commitment.
+  * If a commitment is not fully utilized, the provider MUST include a row that represents the unused portion of the commitment for that charge period. Charge Subcategory MUST be "Commitment Unused".
+  * The sum of the EffectiveCost for all "Commitment Used" and "Commitment Unused" rows for each ChargeSubcategory MUST be the same as the BilledCost of the commitment-based discount purchase.
+* Credits that are applied after the fact MUST use a ChargeType of "Adjustment" and ChargeSubcategory of "Credit".
```
joshk0, line 41: should negotiated cost be mentioned or link to around here?

supporting_content/attributes/discount_handling.md
BenShy: wouldn't tier pricing be covered by usage based?

specification/attributes/discount_handling.md
BenShy: yeah, I'd view spot as a lower cost vm.

specification/attributes/amortization_handling.md
BenShy: resolving as amortization specific information was removed.

specification/attributes/discount_handling.md
BenShy: I couldn't find the NegotiatedCost doc. Happy to link to it, but couldn't seem to find it. Resolving this PR, and I'll add the link to it at a later date.

specification/attributes/discount_handling.md
BenShy: I was planning on linking to NegotiatedCost doc, once I get the link to it.

specification/attributes/discount_handling.md
```diff
@@ -0,0 +1,49 @@
...
+
+Amortization is a process used to break down and spread purchase costs over a period of time or term of use. When a purchase is applicable to resources, like commitment-based discounts, the amortized cost of a resource takes the initial payment and term into account and distributes it out based on the resource's usage, attributing the prorated cost for each unit of billing. Amortization enables users of billing data to distribute purchase charges to the appropriate audience in support of cost allocation efforts. Discount Handling for purchased commitments is commonly used for scenarios like calculating utilization and implementing chargeback for the purchase amount.
+
+While providers may use different terms to describe discounts, FOCUS identifies a discount as being a reduced price applied directly to a row or charge. Any price or cost reductions that are awarded after the fact are identified as a "Credit" Charge Subcategory. One example might be when a provider offers a reduced rate after passing a certain threshold of usage or spend.
```
flanakin, line 9:
```suggestion
While providers may use different terms to describe discounts, FOCUS identifies a discount as being a reduced price applied directly to a row. Any price or cost reductions that are awarded after the fact are identified as a "Credit" Charge Subcategory. One example might be when a provider offers a reduced rate after passing a certain threshold of usage or spend.
```

specification/attributes/discount_handling.md
```diff
@@ -0,0 +1,49 @@
...
+Indicates how to include and apply discounts to usage charges or rows.
+
+## Requirements
+
+* All applicable discounts MUST be applied to each row they pertain to and MUST NOT be negated in a separate row.
+* All discounts applied to a row MUST apply to the entire charge.
+  * If a discount only applies to a subset of the charge, then that portion of the charge MUST be split into a separate row. One with the discount and one without.
+  * If multiple discounts apply, there MUST be separate rows for each unique combination of applied discounts or undiscounted amount.
+  * Each discount MUST be identifiable using existing FOCUS columns.
+    * Discounts applied to charges/rows from a commitment-based discount MUST include a CommitmentDiscountId.
```
flanakin, line 32:
```suggestion
    * Discounts applied to rows from a commitment-based discount MUST include a CommitmentDiscountId.
```

flowerinthenight (approved):

aqu-erp (approved):

specification/attributes/discount_handling.md
flanakin: There's a risk this may not land in time. Let's see which one lands first and we can easily add the link in either PR.

specification/attributes/discount_handling.md
```diff
@@ -0,0 +1,49 @@
...
+    * If a provider applies a discount that cannot be represented by a FOCUS column, they SHOULD include additional columns to identify the source of the discount.
+* Purchased discounts (e.g., commitment-based discounts) MUST specify CommitmentUtilization on the usage rows that received the reduced rates from the discount.
+* Purchased discounts (e.g., commitment-based discounts) MUST be amortized.
+  * The BilledCost MUST be 0 for any row where the commitment covers the entire cost for the charge period.
+  * The EffectiveCost MUST be the portion of the amortized purchase cost that applies to this row.
+  * CommitmentUtilization MUST be "Used" for rows that received a reduced price from that commitment.
+  * If a commitment is not fully utilized, the provider MUST include a row that represents the unused portion of the commitment for that charge period. Charge Subcategory MUST be "Commitment Unused".
+  * The sum of the EffectiveCost for all "Commitment Used" and "Commitment Unused" rows for each ChargeSubcategory MUST be the same as the BilledCost of the commitment-based discount purchase.
```
flanakin, line 40:
```suggestion
  * If a commitment is not fully utilized, the provider MUST include a row that represents the unused portion of the commitment for that charge period. Charge Subcategory MUST be "Commitment Not Used".
  * The sum of the EffectiveCost for all "Commitment Used" and "Commitment Not Used" rows for each ChargeSubcategory MUST be the same as the BilledCost of the commitment-based discount purchase.
```

joshk0 (approved):
Merge first but return quickly to crosslink to NegotiatedCost once it's ready.

could also just mention NegotiatedCost without cross linking and come back to crosslink after.

supporting_content/attributes/discount_handling.md
```diff
@@ -0,0 +1,88 @@
+# Discount Types
+
+Providers offer various discounting schemes for their service offerings. These discounts typically fall into common types such as:
+
+* Commitment-based discounts
+* Tier-based discounts
+* Negotiated discounts
+* Promotional discounts
+* Partnership-based discounts
```
cnharris10, line 9:
```suggestion
* Usage-based discounts
```

https://cloud.google.com/compute/docs/sustained-use-discounts

flanakin: @flowerinthenight is this the same as what you were thinking or separate? Might be good to add the link in the supporting content as well.

Thanks, @cnharris10!

flowerinthenight: Yes, the same. I'll add the link.

specification/attributes/discount_handling.md
```diff
@@ -0,0 +1,49 @@
...
+
+Discount Handling
+
+## Description
+
+Indicates how to include and apply discounts to usage charges or rows.
+
+## Requirements
+
+* All applicable discounts MUST be applied to each row they pertain to and MUST NOT be negated in a separate row.
```
mike-finopsorg, line 27:
```suggestion
* All applicable discounts SHOULD be applied to each row they pertain to and SHOULD NOT be negated in a separate row.
```

Just thinking through the adoption curve here, there are some CSPs that lean heavily on negation lines today, can we discourage the use of negation lines now and look towards moving this to a stronger stance in a later release?

cnharris10 (approved):

ljadvey (approved):

specification/attributes/discount_handling.md
```diff
@@ -0,0 +1,49 @@
+# Discount Handling
+
+A discount is a pricing construct where providers offer a reduced price for services. Providers may have many types of discounts, including but not limited to commercially negotiated discounts, commitment-based discounts when you agree to a certain amount of usage or spend, and bundled discounts where you receive free or discounted usage of one product or service based on the usage of another. Discount Handling is commonly used in scenarios like verifying discounts were applied and calculating cost savings.
+
+Some discount offers can be purchased from a provider to get reduced prices. The most common example is a commitment-based discount, where you "purchase" a commitment to use or spend a specific amount within a period. There are instances where a portion of the purchased discount may go unused, which results in lost savings and needs to be clearly identifiable at a granular level. To facilitate this, the purchase amount is amortized over the term the discount is applied to (e.g., 1 year).
```
udam-f2, line 5: I feel like this implies that the reason to amortize is to find the unused portion of a commitment discount which I've not seen in the past. It was more to get a view of cost without the periodic spikes so that the spend is spread based on the period the purchase applies for - which gives you a more accurate view of your spend regardless of the period you're analyzing spend for (allows you to do things like budgeting and forecasting more easily since it's accounting for purchases that were made a long time ago that you're still receiving the benefit of). 

I'm sure @kk09v would have a better accounting / finance explanation that might make sense to list here as well.

specification/attributes/discount_handling.md
```diff
@@ -0,0 +1,49 @@
...
+
+Amortization is a process used to break down and spread purchase costs over a period of time or term of use. When a purchase is applicable to resources, like commitment-based discounts, the amortized cost of a resource takes the initial payment and term into account and distributes it out based on the resource's usage, attributing the prorated cost for each unit of billing. Amortization enables users of billing data to distribute purchase charges to the appropriate audience in support of cost allocation efforts. Discount Handling for purchased commitments is commonly used for scenarios like calculating utilization and implementing chargeback for the purchase amount.
+
+While providers may use different terms to describe discounts, FOCUS identifies a discount as being a reduced price applied directly to a row. Any price or cost reductions that are awarded after the fact are identified as a "Credit" Charge Subcategory. One example might be when a provider offers a reduced rate after passing a certain threshold of usage or spend.
```
udam-f2, line 9: Not sure I understand what the Credit here is. What are some use cases for this? Are we talking about tiered pricing (last sentence sort of hints at that) or things like sustained use discounts in GCP?

specification/attributes/discount_handling.md
```diff
@@ -0,0 +1,49 @@
...
+
+## Requirements
+
+* All applicable discounts SHOULD be applied to each row they pertain to and SHOULD NOT be negated in a separate row.
+* All discounts applied to a row MUST apply to the entire charge.
+  * If a discount only applies to a subset of the charge, then that portion of the charge MUST be split into a separate row. One with the discount and one without.
+  * If multiple discounts apply, there MUST be separate rows for each unique combination of applied discounts or undiscounted amount.
+  * Each discount MUST be identifiable using existing FOCUS columns.
+    * Discounts applied to rows from a commitment-based discount MUST include a CommitmentDiscountId.
+    * If a provider applies a discount that cannot be represented by a FOCUS column, they SHOULD include additional columns to identify the source of the discount.
```
udam-f2, line 33: Not sure this is clear. What's an example? We don't today have a general discount type or id or anything right?

udam-f2 (commented):
Excited to see this get to the finished line..

specification/attributes/discount_handling.md
flanakin: This was about nudging them to make sure they explain discounts. We don't have it in FOCUS columns, but we're asking them to have it in theirs. Goal would be to cover this in FOCUS columns in a future release.

specification/attributes/discount_handling.md
flanakin: This is trying to cover SUDs without calling it out by name.

specification/attributes/discount_handling.md
```diff
@@ -0,0 +1,49 @@
...
+## Description
+
+Indicates how to include and apply discounts to usage charges or rows.
+
+## Requirements
+
+* All applicable discounts SHOULD be applied to each row they pertain to and SHOULD NOT be negated in a separate row.
+* All discounts applied to a row MUST apply to the entire charge.
+  * If a discount only applies to a subset of the charge, then that portion of the charge MUST be split into a separate row. One with the discount and one without.
+  * If multiple discounts apply, there MUST be separate rows for each unique combination of applied discounts or undiscounted amount.
```
rileyjenk, line 30: I am not sure how this would work? Say their is a commitment based discount and a general percentage discount.....how is bullet number one and bullet number 4 adhered to?

specification/attributes/discount_handling.md
marc-perreaut: I believe the point is GCP SUDs are applied globally, with no detail per resource.
That's why there is a need for a separate charge/row.
Maybe this could be added to the text to help for the udnerstanding.

specification/attributes/discount_handling.md
marc-perreaut: I believe:
- If the commitment-based discount and the general discount are combined together, they should apply to every row.
- If the general discount is applied after the fact based on some overall criteria like total cost exceeding a threshold, then the general discount should appear as a credit

specification/attributes/discount_handling.md
flanakin:
```suggestion
Some discount offers can be purchased from a provider to get reduced prices. The most common example is a commitment-based discount, where you "purchase" a commitment to use or spend a specific amount within a period. When a commitment isn't fully utilized, the unused amount reduces the potential savings from the discount and can even result in paying higher costs than without the discount. Due to this risk, unused commitment amounts need to be clearly identifiable at a granular level. To facilitate this, unused commitments are recorded with a separate row for each charge period where the commitment was not fully utilized. In order to show the impact of purchased discounts on each discounted row, discount purchases need the purchase amount the be amortized over the term the discount is applied to (e.g., 1 year) with each charge period split and applied to each row that received the discount.
```

supporting_content/attributes/discount_handling.md
```diff
@@ -0,0 +1,115 @@
...
+  - EdpDiscount - Enterprise Discount Program
+  - SppDiscount - Solutions Provider Program
+  - RiVolumeDiscount
+  - BundledDiscount - discounted usage on a specific product/service based on the usage of another product/service.
+
+- GCP discount types include:
+  - DISCOUNT - credits earned after a contractual spending threshold is reached.
+  - COMMITTED_USAGE_DISCOUNT - resource-based commitment discounts.
+  - COMMITTED_USAGE_DISCOUNT_DOLLAR_BASE - Spend-based commitment discounts.
+  - SUSTAINED_USAGE_DISCOUNT - automatic discount earned for prolonged usage.
```
flanakin, line 99:
```suggestion
  - SUSTAINED_USAGE_DISCOUNT - automatic discount earned for prolonged usage.
    > [!NOTE]
    > Sustained Use Discounts (SUDs) are handled as after-the-fact adjustments (aka credits) and do not apply to the Discount Handling attribute since this attribute requires that discounts be applied directly to the row they're discounting.
```

specification/attributes/discount_handling.md
flanakin: Added to supplemental data.

specification/attributes/discount_handling.md
flanakin: That's correct. Let me give a different example:
Let's say you have a 10% negotiated discount (A), a 10% bundled discount (B), and a $1/hour spend commitment purchased for $0.50/hour (C). For the sake of this example, let's say the negotiated and bundled discounts cap out and don't apply to all resources so I can show what happens with a partially undiscounted amount. All examples are $0.05/minute for hourly data...

| Scenario | Resource | Discounts | Quantity | ListCost | OnDemandCost | BilledCost | EffectiveCost |
|-----------|-------|------------|------|----|---------|------------|----------------|
| No discounts | Foo0 |  | 60 | $3.00 | $3.00 | $3.00 | $3.00 |
| Negotiated only | Foo1 | A | 60 | $3.00 | $2.70 | $2.70 | $2.70 |
| Commitment only | Foo2 | C | 20 | $1.00 | $1.00 | $0.00 | $0.50 |
| Uncommitted part | Foo2 |    | 40 | $2.00 | $2.00 | $2.00 | $2.00 |
| Commitment only | Foo3 | C | 10 | $0.50 | $0.50 | $0.00 | $0.25 |
| Unused commitment |     | C | 10 | $0.50 | $0.50 | $0.00 | $0,25 |
| Negotiated&nbsp;+&nbsp;committed | Foo4 | A+C | 22.222 | $1.11 | $1.00 | $0.00 | $0.50 |
| Negotiated&nbsp;+&nbsp;uncommitted | Foo4 | A | 37.778 | $1.89 | $1.70 | $1.70 | $1.70 |
| All discounts<sup>1</sup> | Foo5 | A+B+C | 24.691 | $1.11 | $1.00 | $0.00 | $0.50 |
| Negotiated&nbsp;+&nbsp;bundled | Foo5 | A+B | 5.309 | $0.24 | $2.15 | $2.15 | $2.15 |
| Negotiated only | Foo5 | A | 15 | $0.75 | $0.675 | $0.675 | $0.675 |
| All discounts capped out | Foo5 |  | 15 | $0.75 | $0.75 | $0.75 | $0.75 |

<sup>_* As we've defined SKUs, bundled discounts would be applied to the list price, so that's why it doesn't look like there's a discount applied compared to Foo4._</sup>

rileyjenk (approved):

supporting_content/attributes/discount_handling.md
```diff
@@ -0,0 +1,117 @@
...
+    > [!NOTE]
+    > Sustained Use Discounts (SUDs) are handled as after-the-fact adjustments (aka credits) and do not apply to the Discount Handling attribute since this attribute requires that discounts be applied directly to the row they're discounting.
+  - FREE_TIER - free resource usage up to specified limits.
+  - PROMOTION - includes free tial, marketing campaign credits, or other grants to use GCP.
+  - RESELLER_MARGIN - indicates the Reseller Program Discounts earned on every eligible line item.
+  - SUBSCRIPTION_BENEFIT - credits earned by purchasing long-term subscriptions to services in exchange for discounts.
+
+- Where does spot instance usage-based discount fall under? Should there be a **Usage-based discounts** type?
+
+- Where does GCP's sustained usage discount fall under? Maybe the same as spot instance, in which case, **Usage-based discounts**?
```
aqu-erp, line 109: Low prio: Are we going to remove these notes, now that we have clarity? Or should we answer the questions?

specification/attributes/discount_handling.md
```diff
@@ -0,0 +1,49 @@
...
+
+## Description
+
+Indicates how to include and apply discounts to usage charges or rows.
+
+## Requirements
+
+* All applicable discounts SHOULD be applied to each row they pertain to and SHOULD NOT be negated in a separate row.
+* All discounts applied to a row MUST apply to the entire charge.
+  * If a discount only applies to a subset of the charge, then that portion of the charge MUST be split into a separate row. One with the discount and one without.
```
udam-f2, line 29: Not sure if clearer. Ignore if not.
```suggestion
  * If a discount only applies to a portion of a charge, then that portion of the charge MUST be split into a separate row while the standard priced portion should be kept in it's own row.
```

specification/attributes/discount_handling.md
```diff
@@ -0,0 +1,49 @@
...
+  * If multiple discounts apply, there MUST be separate rows for each unique combination of applied discounts or undiscounted amount.
```
udam-f2, line 30: Technically, this is for non-negotiated discounts? Negotiated discounts aren't going to split out rows.

specification/attributes/discount_handling.md
```diff
@@ -0,0 +1,49 @@
...
+  * Each discount MUST be identifiable using existing FOCUS columns.
+    * Discounts applied to rows from a commitment-based discount MUST include a CommitmentDiscountId.
```
udam-f2, line 32: Not sure if clearer. Ignore if not.
```suggestion
    * Rows with a commitment-based discount applied to it MUST include a CommitmentDiscountId.
```

specification/attributes/discount_handling.md
udam-f2: I feel like it's a bit unclear what we're asking providers to do. Are we saying additional provider-defined columns should be included to identify the discount details?

specification/attributes/discount_handling.md
```diff
@@ -0,0 +1,49 @@
...
+    * If a provider applies a discount that cannot be represented by a FOCUS column, they SHOULD include additional columns to identify the source of the discount.
+* Purchased discounts (e.g., commitment-based discounts) MUST specify CommitmentUtilization on the usage rows that received the reduced rates from the discount.
```
udam-f2, line 34: I think VS2 decided to do away with CommitmentUtilization and instead has similar values in ChargeSubcategory

specification/attributes/discount_handling.md
```diff
@@ -0,0 +1,49 @@
...
+* Purchased discounts (e.g., commitment-based discounts) MUST be amortized.
+  * The BilledCost MUST be 0 for any row where the commitment covers the entire cost for the charge period.
```
udam-f2, line 36:
```suggestion
  * The BilledCost MUST be 0 for any ChargeType 'Usage' row where the commitment covers the entire cost for the charge period.
```

specification/attributes/discount_handling.md
```diff
@@ -0,0 +1,49 @@
...
+  * The EffectiveCost MUST be the portion of the amortized purchase cost that applies to this row.
+  * CommitmentUtilization MUST be "Used" for rows that received a reduced price from that commitment.
```
udam-f2, line 38: CommitmentUtilization doesn't exist. Use ChargeSubcategory instead.

specification/attributes/discount_handling.md
```diff
@@ -0,0 +1,49 @@
...
+  * If a commitment is not fully utilized, the provider MUST include a row that represents the unused portion of the commitment for that charge period. Charge Subcategory MUST be "Commitment Not Used".
```
udam-f2, line 39: I think 'that charge period' is vague. I think what you want to say is for each charge period at the lowest granularity supported by the dataset?
```suggestion
  * If a commitment is not fully utilized, the provider MUST include a row that represents the unused portion of the commitment for that charge period. ChargeSubcategory MUST be "Unused Commitment".
```

specification/attributes/discount_handling.md
```diff
@@ -0,0 +1,49 @@
...
+  * The sum of the EffectiveCost for all "Commitment Used" and "Commitment Not Used" rows for each ChargeSubcategory MUST be the same as the BilledCost of the commitment-based discount purchase.
```
udam-f2, line 40: This won't work. I see a few issues:
- First, I think you mean to say the sum of Used and Unused for the duration of the commitment should equal the billed cost of the purchase.
- The above statement would be true for all up-front paid RIs. In the case of partial and no up-front RIs, the effective cost would be the amortized portion + on-going rate you pay? So the sum will not equal the purchase
- In cases like partial or no up-front, are we suggesting there is a need for a single row that represents the 'purchase' record in each lowest-granularity charge period?

```suggestion
  * The sum of the EffectiveCost for all "Used Commitment" and "Unused Commitment" rows for each ChargeSubcategory MUST be the same as the BilledCost of the commitment-based discount purchase.
```

specification/attributes/discount_handling.md
marc-perreaut: That's great example to explain how commitment-related charges work.
If it's planned to have illustrative examples in FOCUS, it's a must-have there!

specification/attributes/discount_handling.md
aqu-erp: Made an edit for even more clarity^ 

> then the discounted

specification/attributes/discount_handling.md
aqu-erp, line 40: I think some of the confusion is coming from lines 36-40 not at low-enough level to indicate: 
> how to include and apply discounts to usage charges or rows


```suggestion
* Prepaid portions of purchased discounts (e.g., commitment-based discounts) MUST be amortized.
  * The BilledCost MUST be 0 for any row where ChargeType is "Usage" and the commitment applies to the entire charge for the charge period.
  * The EffectiveCost MUST include the portion of the amortized purchase cost proportional to the Pricing Unit and the time granularity of the charge, for any row that received a reduced price from a commitment.
  * ChargeSubcategory MUST be "Used Commitment" for rows that received a reduced price from that commitment.
  * The provider MUST include a row that represents unused portions of a commitment based on the time granularity of the dataset. ChargeSubcategory MUST be "Unused Commitment" for these rows.
```

I removed my draft of line 40 because ChargeSubCategory does not support "Purchase" yet.
>* The sum of BilledCost for all rows where ChargeSubcategory is "Used Commitment" or "Unused Commitment" MUST equal the value committed by all commitments within the time granularity of the dataset.

specification/attributes/discount_handling.md
flanakin: This also applies to negotiated discounts. I left an example in comments. I'll pull that into the supporting content. Specifically, for negotiated discounts, they generally apply to all charges or a specific set of charges. They're generally applicable to the entire charge which is all this is asking for. The only issue would be if a negotiated discount only applied to a portion of a charge. In that case, we'd need to split the charge into 2 rows. One with the negotiated discount and one without it. This isn't specific to negotiated discounts. This applies to all discounts.

supporting_content/attributes/discount_handling.md
flanakin: Will add notes rather than remove them...

specification/attributes/discount_handling.md
BenShy: yes, this is focused on non-negotiated discounts.

BenShy: closing this out, will open if need more data

supporting_content/attributes/discount_handling.md
flanakin, line 109:
```suggestion
- Where does spot pricing discount fall under? Should there be a **Usage-based discounts** type?
  - We decided to handle spot as PricingCategory = `Dynamic` and PricingSubcategory = `Spot`.
  - We have not yet discussed if there are any special nuances to how spot pricing works that should be explicitly incorporated into Discount Handling beyond identification of spot-priced charges.

- Where does GCP's sustained usage discount fall under? Maybe the same as spot instance, in which case, **Usage-based discounts**?
  - Since GCP SUDs are an after-the-fact discount that includes a negative charge, they're being tracked as ChargeCategory = `Adjustment` and ChargeSubcategory = `Credit`.
  - In the main spec content, SUDs are covered by the "Any price or cost reductions that are awarded after the fact are identified as a `Credit` Charge Subcategory" sentence.
```

specification/attributes/discount_handling.md
flanakin: I'll add this to supporting content for now. I like the idea of figuring out the best way to be more explicit with examples in the main spec. We should think about that after this release.

supporting_content/attributes/discount_handling.md
```diff
@@ -0,0 +1,117 @@
+# Discount Types
+
```
flanakin, line 2:
```suggestion

## Example implementation of the Discount Handling attribute

Let's say you have 3 discounts:

1. **10% negotiated** discount
2. **10% bundled** discount
3. **$1/hour spend commitment** purchased for **$0.50/hour**

For the sake of this example, let's say the negotiated and bundled discounts cap out and don't apply to all resources to demonstrate what happens with a partially undiscounted amount.

All examples are $0.05/minute for hourly data...

| Scenario | Resource | Qty | List | OnDemand | Billed | Effective |
|-----------|-----------|------|-----|--------------|-------|-----------|
| No discounts | Foo0 | 60 | $3.00 | $3.00 | $3.00 | $3.00 |
| Negotiated only | Foo1 |  60 | $3.00 | $2.70 | $2.70 | $2.70 |
| Commitment only | Foo2 |  20 | $1.00 | $1.00 | $0.00 | $0.50 |
| Uncommitted part | Foo2 | 40 | $2.00 | $2.00 | $2.00 | $2.00 |
| Commitment only | Foo3 | 10 | $0.50 | $0.50 | $0.00 | $0.25 |
| Unused commitment | | | $0.00 | $0.00 | $0.00 | $0.25 |
| Negotiated +&nbsp;committed | Foo4 | 22.22 | $1.11 | $1.00 | $0.00 | $0.50 |
| Negotiated +&nbsp;uncommitted | Foo4 | 37.78 | $1.89 | $1.70 | $1.70 | $1.70 |
| All discounts<sup>1</sup> | Foo5 | 24.69 | $1.11 | $1.00 | $0.00 | $0.50 |
| Negotiated +&nbsp;bundled | Foo5 | 5.31 | $0.24 | $2.15 | $2.15 | $2.15 |
| Negotiated only | Foo5 | 15 | $0.75 | $0.67 | $0.67 | $0.67 |
| All discounts capped out | Foo5 | 15 | $0.75 | $0.75 | $0.75 | $0.75 |

<sup>_1. As we've defined SKUs, bundled discounts would be applied to the list price, so that's why it doesn't look like there's a discount applied compared to Foo4._</sup>

Explanation of the unused commitment row:

1. The **Foo3** resource only used **10min @ $0.05/min**.
2. This is **$0.50** list _and_ on-demand since the negotiated discount doesn't apply on this row for the sake of the example (see Foo5 for stacked discounts).
3. The billed cost is nothing, since this $0.50 was covered by the spend commitment.
4. Since the spend commitment was for $1/hour, this means there's $0.50 left that was not used.
5. The "Unused commitment" row is for that unused amount.
   - List, billed, and on-demand costs are nothing since there was no usage and nothing was billed.
   - Effective cost is the unused amount of the spend commitment ($0.50 in this case).
   - Since spend commitments don't have a unit or a price, they won't have a quantity, so the unused amount doesn't have a quantity.
   - The only thing it has is the unused amount, which is the effective price per hour ($0.50), minus the used portion ($0.25).
   - **_Why $0.25?_**
     - The commitment was $1/hour with a 50% discount per hour.
     - This means the effective/amortized cost per hour is $0.50.
     - The on-demand cost was $0.50. After applying the amortized price, it's 50% off or $0.25.
     - Since the on-demand cost was only $0.50, there's still another $0.50 left unused.
     - And when you apply the 50% discount on the unused portion, you get $0.25.
     
```

flanakin (approved):

specification/attributes/discount_handling.md
```diff
@@ -0,0 +1,49 @@
...
+    * Rows with a commitment-based discount applied to it MUST include a CommitmentDiscountId.
+    * If a provider applies a discount that cannot be represented by a FOCUS column, they SHOULD include additional columns to identify the source of the discount.
+* Purchased discounts (e.g., commitment-based discounts) MUST specify ChargeSubcategory on the usage rows that received the reduced rates from the discount.
+* Purchased discounts (e.g., commitment-based discounts) MUST be amortized.
+  * The BilledCost MUST be 0 for any row where the commitment covers the entire cost for the charge period.
+  * The EffectiveCost MUST be the portion of the amortized purchase cost that applies to this row.
+  * ChargeSubcategory MUST be "Used" for rows that received a reduced price from that commitment.
+  * If a commitment is not fully utilized, the provider MUST include a row that represents the unused portion of the commitment for that charge period. ChargeSubcategory MUST be "Unused Commitment".
+  * The sum of the EffectiveCost for all "Used Commitment" and "Unused Commitment" rows for each ChargeSubcategory MUST be the same as the BilledCost of the commitment-based discount purchase.
```
rileyjenk, line 40: This is not true in the case of upfront or partial upfront commitment based discounts.

rileyjenk (changes requested):

specification/attributes/discount_handling.md
mike-finopsorg: I think this means the total `EffectiveCost` across the *whole term of the commitment* should match the SUM of the BilledCost.

specification/attributes/discount_handling.md
```diff
@@ -0,0 +1,49 @@
...
+## Description
+
+Indicates how to include and apply discounts to usage charges or rows.
+
+## Requirements
+
+* All applicable discounts SHOULD be applied to each row they pertain to and SHOULD NOT be negated in a separate row.
+* All discounts applied to a row MUST apply to the entire charge.
+  * If a discount only applies to a portion of a charge, then the discounted portion of the charge MUST be split into a separate row while the standard priced portion should be kept in it's own row.
+  * If multiple discounts apply, there MUST be separate rows for each unique combination of applied discounts or undiscounted amount.
```
mike-finopsorg, line 30:
```suggestion
* All discounts applied to a row MUST apply to the entire charge. 
  * Multiple discounts MAY apply to a row, but they MUST apply to the entire charge covered by that row.
  * If a discount only applies to a portion of a charge, then the discounted portion of the charge MUST be split into a separate row.
```

Suggestion to help clarify the splitting of lines isn't due to multiple discounts, but the differing portion of application of discounts. So multiple discounts can apply to a single row as long as they all apply to the total charge. If they differ (e.g. One discount (A) applies to the whole charge but the other discount (B) only to half then break into two rows, one with both A and B and the other with just A)

specification/attributes/discount_handling.md
mike-finopsorg, line 40:
```suggestion
  * The sum of the EffectiveCost for all "Used Commitment" and "Unused Commitment" rows for each CommitmentDiscountId over the entire duration of the commitment MUST be the same as the total BilledCost of the commitment-based discount.
```

Suggestion to clarify the total amount you pay in EffectiveCost where the discount is applied (or unused) should total the amount you are billed for the commitment purchase.

flanakin: @rileyjenk, take a look at the updates.

specification/attributes/discount_handling.md
```diff
@@ -0,0 +1,49 @@
...
+## Requirements
+
+* All applicable discounts SHOULD be applied to each row they pertain to and SHOULD NOT be negated in a separate row.
+* All discounts applied to a row MUST apply to the entire charge. 
+  * Multiple discounts MAY apply to a row, but they MUST apply to the entire charge covered by that row.
+  * If a discount only applies to a portion of a charge, then the discounted portion of the charge MUST be split into a separate row.
+  * Each discount MUST be identifiable using existing FOCUS columns.
+    * Rows with a commitment-based discount applied to it MUST include a CommitmentDiscountId.
+    * If a provider applies a discount that cannot be represented by a FOCUS column, they SHOULD include additional columns to identify the source of the discount.
+* Purchased discounts (e.g., commitment-based discounts) MUST specify ChargeSubcategory on the usage rows that received the reduced rates from the discount.
```
aqu-erp, line 34:
```suggestion
* ChargeSubCategory MUST NOT be null for rows where ChargeType is "Usage" and the row received reduced rates from a discount. 
```

aqu-erp: MUST condition on "Purchased discounts" sound a little funny

specification/attributes/discount_handling.md
```diff
@@ -0,0 +1,49 @@
...
+Discount Handling
+
+## Description
+
+Indicates how to include and apply discounts to usage charges or rows.
+
+## Requirements
+
+* All applicable discounts SHOULD be applied to each row they pertain to and SHOULD NOT be negated in a separate row.
+* All discounts applied to a row MUST apply to the entire charge.
```
flanakin, line 28:
```suggestion
* All discounts applied to a row MUST apply to the entire charge.
```

specification/attributes/discount_handling.md
```diff
@@ -0,0 +1,49 @@
...
+## Requirements
+
+* All applicable discounts SHOULD be applied to each row they pertain to and SHOULD NOT be negated in a separate row.
+* All discounts applied to a row MUST apply to the entire charge. 
+  * Multiple discounts MAY apply to a row, but they MUST apply to the entire charge covered by that row.
+  * If a discount only applies to a portion of a charge, then the discounted portion of the charge MUST be split into a separate row.
+  * Each discount MUST be identifiable using existing FOCUS columns.
+    * Rows with a commitment-based discount applied to it MUST include a CommitmentDiscountId.
+    * If a provider applies a discount that cannot be represented by a FOCUS column, they SHOULD include additional columns to identify the source of the discount.
+* ChargeSubCategory MUST NOT be null for rows where ChargeType is "Usage" and the row received reduced rates from a discount.
```
flanakin, line 34:
```suggestion
* ChargeSubCategory MUST NOT be null for rows where ChargeType is "Usage" and the row received reduced rates from a discount.
```

specification/attributes/discount_handling.md
```diff
@@ -0,0 +1,49 @@
...
+* All discounts applied to a row MUST apply to the entire charge.
+  * Multiple discounts MAY apply to a row, but they MUST apply to the entire charge covered by that row.
+  * If a discount only applies to a portion of a charge, then the discounted portion of the charge MUST be split into a separate row.
+  * Each discount MUST be identifiable using existing FOCUS columns.
+    * Rows with a commitment-based discount applied to it MUST include a CommitmentDiscountId.
+    * If a provider applies a discount that cannot be represented by a FOCUS column, they SHOULD include additional columns to identify the source of the discount.
+* ChargeSubCategory MUST NOT be null for rows where ChargeType is "Usage" and the row received reduced rates from a discount.
+* Purchased discounts (e.g., commitment-based discounts) MUST be amortized.
+  * The BilledCost MUST be 0 for any row where the commitment covers the entire cost for the charge period.
+  * The EffectiveCost MUST be the portion of the amortized purchase cost that applies to this row.
```
mike-finopsorg, line 37:
```suggestion
  * The EffectiveCost MUST include the portion of the amortized purchase cost that applies to this row.
```

Currently reads like EffectiveCost must equal the amortised amount, which is not correct if there includes charges on the row itself.

specification/attributes/discount_handling.md
```diff
@@ -0,0 +1,49 @@
...
+  * ChargeSubcategory MUST be "Used" for rows that received a reduced price from that commitment.
```
aqu-erp, line 38:
```suggestion
  * ChargeSubcategory MUST be "Used Commitment" for rows that received a reduced price from that commitment.
```

mike-finopsorg:
It feels like Discount Handling needs to answer the question of where costs are represented between purchases and usage. Billed and Effective. We need to answer this for 1.0-Final

mike-finopsorg (approved):

Issue #237, opened 2023-10-27
Create a Negotiated Cost metric


ijurica:
### Type
> Type of issue: Metric
Normalized: No

### Description
> Define how Negotiated Cost should appear in cost data under metrics.

### Definition of done

- [ ] Rationalize vendor-neutral, cross-cloud naming
- [ ] Complete spec template and include naming (code name, display name), constraints, guidelines, compatibility with major providers etc.

flanakin:
This is a must-have!

I'd like to see this aligned to the existing terminology we have for pricing that is not committed (PricingCategory = "On-Demand"). Not everyone has negotiated discounts, so something that is more inclusive and consistent with what we already have would be ideal.

ijurica:
This Cost would be used to **calculate savings based on negotiation activities**, by comparing it with the List Cost.
As @flanakin mentioned, we should also consider introducing the **corresponding unit price or discount**.

**Note:** 
Some time ago, we initiated the preparation of NegotiatedCost and NegotiatedUnitPrice metrics; however, we decided to put them on-hold at that point (refer to closed PRs #238 and #236).

flanakin:
We should be clear that the value we use here is the price/cost when commitment discounts are not applied. It's the on-demand price/cost, which applies even when a negotiated discount is not applicable. It's not about negotiated discounts exclusively. We need to make sure the name of the columns is inclusive of all account types and doesn't imply that it's only about negotiated discounts.

jpradocueva:
@ijurica will link the Issue to the PR and the group, TF-2,  agreed to close the Issue (on May 15)

ijurica:
Addressed in PR #364.

PR #238, opened 2023-10-27
FOCUS #237: NegotiatedCost metric

ijurica:
Negotiated Cost metric specification and supporting content

### DIFF
--- a/specification/metrics/negotiatedcost.md
+++ b/specification/metrics/negotiatedcost.md
@@ -1,0 +1,34 @@
+# Negotiated Cost
+
+The Negotiated Cost represents a charge inclusive of the impacts of volume/tier-based rates and negotiated discounts while excluding any commitment-based discounts or the amortization of upfront charges (one-time or recurring). This cost is denominated in the [Billing Currency](#billingcurrency). The Negotiated Cost equals the [List Cost](#listcost) when no negotiated discounts are present. It is used to calculate savings based on negotiation activities, by comparing it with List Cost.
+
+The NegotiatedCost column MUST be present in the billing data. This column MUST be a valid numeric value of type Decimal and MUST NOT contain null values. When a [NegotiatedUnitPrice](#negotiatedunitprice) is not null, multiplying the NegotiatedUnitPrice by [QuantityInPricingUnit](#quantityinpricingunit ) MUST produce the NegotiatedCost. In cases where the NegotiatedUnitPrice is null, the following applies:
+
+* If the line item is based on other charges (e.g. [ChargeType](#chargetype) is 'Tax'), the NegotiatedCost MUST be calculated based on NegotiatedCost of the related charges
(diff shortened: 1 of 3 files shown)


specification/metrics/negotiatedcost.md
```diff
@@ -0,0 +1,34 @@
...
+
+The NegotiatedCost column MUST be present in the billing data. This column MUST be a valid numeric value of type Decimal and MUST NOT contain null values. When a [NegotiatedUnitPrice](#negotiatedunitprice) is not null, multiplying the NegotiatedUnitPrice by [QuantityInPricingUnit](#quantityinpricingunit ) MUST produce the NegotiatedCost. In cases where the NegotiatedUnitPrice is null, the following applies:
+
+* If the line item is based on other charges (e.g. [ChargeType](#chargetype) is 'Tax'), the NegotiatedCost MUST be calculated based on NegotiatedCost of the related charges
+* If the line item is unrelated to other charges (e.g. [AdjustmentType](#adjustmenttype) is 'Credit'), the NegotiatedCost MUST match the BilledCost.
```
marc-perreaut, line 8: Unless the line item (=the charge, right?) relates to a commitment?

marc-perreaut, line 8: What about an unused commitment charge?

To me, the negotiated cost should be 0 in this case.

ijurica: If we expect to receive all Purchase and Usage related records, are we supposed to calculate unused commitments or do we also anticipate receiving records for unused commitments?

marc-perreaut: In the Azure amortized costs dataset, I have witness that an unused commitment record exists for every reservation, in which the amortized cost is 0 in case the reservation is used at 100%.
I do not know for GCP and AWS: I believe converters will need to calculate and to generate unused commitment records if they do not exist in GCP and AWS cost and usage data.

AndersonOliveira2022 (approved):

specification/metrics/negotiatedcost.md
jpickardgreen: The Negotiated Cost excludes the Commitments. The Billed Cost includes the Commitment.

marc-perreaut: Yes!

Sorry, I was not clear.

I believe there are several cases for a line item that is unrelated to other charges, where the Negotiated Cost may be equal or different from the Billed Cost:
- Purchase of a commitment (whatever upfront, partial upfront, no upfront): Negotiated Cost = Billed Cost
- Usage of a commitment: Negotiated Cost != Billed Cost. The Billed Cost must be actually 0, since the cost resides in the commitment purchase record.
- Unused Commitment:  Negotiated Cost = Billed Cost.

udam-f2:
Closing as out of scope for V1.0

PR #239, opened 2023-11-01
#242- Create a SKU ID Dimension

arrarama:
Add a definition for dimension SKUID in the 1.0 FOCUS specification.


salpaysenturus (approved):

specification/dimensions/skuid.md
```diff
@@ -0,0 +1,91 @@
+## Spec Content
+
+### SKUID
+
+A SKU ID is an identifier assigned to the SKU that’s associated with a charge. SKU ID can be referenced on a price list published by a provider to look up detailed information about the SKU including the different unit prices associated with the SKU.
+
+The SKUId column MUST be present in the billing data. This column MUST be of type String. The SKUId MUST NOT be null when [Charge Type](#ChargeType) is ‘Purchase’ or ‘Usage’. The SKUId SHOULD NOT be null when [Charge Type](#ChargeType) is ‘Adjustment’ or [Adjustment Type](#AdjustmentType) is ‘Refund’.
```
udam-f2, line 7:
```suggestion
The SKUId column MUST be present in the billing data. This column MUST be of type String. The SKUId MUST NOT be null when [Charge Type](#ChargeType) is ‘Purchase’ or ‘Usage’. The SKUId SHOULD NOT be null when [Adjustment Type](#AdjustmentType) is ‘Refund’. The SKUId MUST be null when [Adjustment Type](#AdjustmentType) not ‘Refund’.
```
I'm not sure I fully understand what the values listed for Adjustment type are tbh. But ultimately, what we want to suggest here is that if you're providing an adjustment where its related to a particular SKU then the SKU ID needs to be specified. If it's more of an ad-hoc adjustment, then it does not need to be set.

specification/dimensions/skuid.md
```diff
@@ -0,0 +1,91 @@
...
+
+
+#### Display name
+
+SKU ID
+
+
+#### Description
+
+A provider-assigned identifier that maps to a unique product offering that has a charge associated with it.
```
udam-f2, line 28: Identifier assigned to the SKU that’s associated with a charge.

specification/dimensions/skuid.md
```diff
@@ -0,0 +1,91 @@
...
+
+
+#### Content constraints
+
+|         Constraint             | Value                   |
+| -------------------- | ------------------ |
+| Column required | True          |
+|  Data type       |  String        |
+|  Allows nulls    |  True          |
+| Value format    | string-format |
```
udam-f2, line 38:
```suggestion
|    Constraint   |      Value      |
|:----------------|:----------------|
| Column required | True            |
| Data type       | String          |
| Allows nulls    | True            |
| Value format    | \<not specified> |
```

specification/dimensions/skuid.md
```diff
@@ -0,0 +1,91 @@
...
+
+
+#### Introduced (version)
+
+1.0
+
+
+## Supporting content
```
udam-f2, line 46: This all goes into a separate supporting content file under /supporting_content with the same file name

specification/dimensions/skuid.md
```diff
@@ -0,0 +1,91 @@
...
+
+#### Example provider mappings 
+
+Current column mappings found in available data sets:
+
+|              |                             |                 |
+| ------------ | --------------------------- | --------------- |
+| **Provider** | **Data set**                | **Column**      |
```
udam-f2, line 54: looks like the formatting is off here.

specification/dimensions/skuid.md
```diff
@@ -0,0 +1,91 @@
...
+|              |                             |                 |
+
+
+#### Example scenarios for current provider data
+
+Current values observed in billing data for various scenarios:
+
+|              |                            |                  |
+| ------------ | -------------------------- | ---------------- |
+| **Provider** | **Scenario**               | **Pattern**      |
```
udam-f2, line 68: looks like the formatting is off here.

specification/dimensions/skuid.md
salpaysenturus: @udam-f2 This section was directly lifted from your v1 draft... I'm sure @arrarama is happy to make the change per your suggestion.

salpaysenturus: Upon discussion, the suggestion works with the following revision:

> The SKUId column MUST be present in the billing data. This column MUST be of type String. The SKUId MUST NOT be null when [Charge Type](#ChargeType) is ‘Purchase’ or ‘Usage’. The SKUId SHOULD NOT be null when [Adjustment Type](#AdjustmentType) is ‘Refund’.

The stricture around forcing null for Adjustment Type != Refund feels too strong per Mike Fuller.

specification/dimensions/skuid.md
```diff
@@ -0,0 +1,91 @@
+## Spec Content
+
+### SKUID
```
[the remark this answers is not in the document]
salpaysenturus: Makes sense.  Will take as an action item.

specification/dimensions/skuid.md
```diff
@@ -0,0 +1,91 @@
...
+
+A SKU ID is an identifier assigned to the SKU that’s associated with a charge. SKU ID can be referenced on a price list published by a provider to look up detailed information about the SKU including the different unit prices associated with the SKU.
```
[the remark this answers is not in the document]
salpaysenturus: Makes sense.  Will partner with Trig to make this happen.

specification/dimensions/skuid.md
salpaysenturus: Discussed in November 2/3 call and agreed original verbiage is more approachable.

specification/dimensions/skuid.md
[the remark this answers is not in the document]
salpaysenturus: I have now included this change on the SKU Price PR.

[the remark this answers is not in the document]
arrarama: Raised PR https://github.com/FinOps-Open-Cost-and-Usage-Spec/FOCUS_Spec/pull/245.

specification/dimensions/skuid.md
arrarama: Fixed.

specification/dimensions/skuid.md
arrarama: Fixed.

specification/dimensions/skuid.md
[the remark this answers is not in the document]
udam-f2: Looks like both created the same update. Did you all discuss if in the column name we should preserve the pascal case even in the case of a acronym? I think SKUPriceID is still that and not SkuPriceId. In the case of Id, we decided to not use the upper case ID in the column name for better readability. If it wasn't discussed, we can bring it up with maintainers and the group async. Thanks

specification/dimensions/skuid.md
udam-f2: I think other groups have gone with the approach of saying when somehting MUST be or SHOULD BE null and also adding when it MUST NOT be null. This wasn't what we did in 0.5 so its on my list of inconsistencies to bring up with the group at the end.

What I meant above on AdjustmentType was I'm not sure I understand the values VS2 is proposing on AdjustmentType but the idea of the values per my understanding was that if it's a refund, then it will have SKU and other relevant info (unit rates, quantities etc.). If it doesn't have those, then it cant be a refund and is instead an adjustment. Now that 2-3 other adjustment types are defined, im not sure where the boundary lies. Would be good to follow up with VS2 since without some guidance, providers will have to chose when and when not to put the value here. (I understand this is even inconsistent with my own suggestion above :))

specification/dimensions/skuid.md
```diff
@@ -0,0 +1,91 @@
...
+
+The SkuId column MUST be present in the billing data. This column MUST be of type String. The SkuId MUST NOT be null when Charge Type is ‘Purchase’ or ‘Usage’. The SkuId SHOULD NOT be null when Adjustment Type is ‘Refund’.
+
+SkuId may have a collection of nullable attributes such as Service, Region, instance type, etc. Future versions of the FOCUS specification may include more precise definitions of those attributes.  In the interim, the SkuId can be used in concert with an externally-sourced price sheet to facilitate practitioner data modeling and analysis. 
+
+A SkuId doesn’t necessarily define the price as there may be additional attributes such as tier, region, etc. of the SkuId that map to a price.
```
marc-perreaut, line 11: Indeed, the SKUPriceId is defining the price, so maybe it could be cross-referenced here?

colinjacksnow (approved):
Approved

specification/dimensions/skuid.md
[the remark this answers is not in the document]
arrarama: Closed in favor of PR #244.

specification/dimensions/skuid.md
arrarama: PR #253 created for supporting_content update. Deleted items from file.

specification/dimensions/skuid.md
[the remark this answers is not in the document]
arrarama: Resolving as per Udam, VS4 already has a reference for SKU in the glossary.

udam-f2:
Closing in favor of https://github.com/FinOps-Open-Cost-and-Usage-Spec/FOCUS_Spec/pull/255

PR #244, opened 2023-11-01
FOCUS #243: SKU Price ID Dimension

salpaysenturus:
SKU Price ID dimension definition and supporting content

PR #245, opened 2023-11-03
Exception for abbreviation SKU

arrarama:
Added an exception for abbreviation SKU as it is a standard pattern across the industry.

### DIFF
--- a/specification/attributes/column_naming_convention.md
+++ b/specification/attributes/column_naming_convention.md
@@ -30,6 +30,7 @@ Naming convention for columns appearing in billing data.
 ## Exceptions
 
 * Identifiers will use the "Id" abbreviation since this is a standard pattern across the industry.
+* Stock Keep Units will use the "SKU" abbreviation since this is a standard pattern across the industry.
 
 ## Introduced (version)
 


specification/attributes/column_naming_convention.md
```diff
@@ -30,6 +30,7 @@ Naming convention for columns appearing in billing data.
 ## Exceptions
 
 * Identifiers will use the "Id" abbreviation since this is a standard pattern across the industry.
+* Stock Keep Units will use the "SKU" abbreviation since this is a standard pattern across the industry.
```
flanakin, line 33: Have we closed on whether we'll use "SKU" vs. "Sku" in the column ID?  We should be explicit here.

marc-perreaut, line 33: It looks SKU means Stock Keep**ing** Unit.

Reference: https://en.wikipedia.org/wiki/Stock_keeping_unit

salpaysenturus:
For reference, I had issued a change on this doc in #244 with the following verbiage:

> Product offerings that incur charges will use the "SKU" abbreviation because it is a well-understood term both within and outside the industry.

specification/attributes/column_naming_convention.md
salpaysenturus: We have not.  Unfortunately I didn't realize this was a topic of debate -- otherwise I would have made a point of putting it on the agenda of yesterday's VS3 call.

While I had initially assumed "SKU", it seems that "Sku" makes more sense alongside the existing "Id".  There may be more acronyms that come down the pike, but for 1.0 it feels right to me to consistently apply Pascal case to acronyms.

udam-f2:
Closing in favor of the update being made in SkuPriceId PR (#244)

PR #247, opened 2023-11-04
Corrections on some typographic conventions

flowerinthenight:
As titled.


udam-f2 (approved):
Thanks for catching these!

flanakin (approved):
Nice catch! Thanks, @flowerinthenight!

PR #248, opened 2023-11-04
Clean up random uses of "record"

flanakin:
@flowerinthenight noticed we use both "row" and "record" as generic references to a charge. "Row" is used more, so at least normalizing on that. There are a few places where "charge" makes less sense or is needed to differentiate from the monetary value of a "charge".


supporting_content/appendix/quantity_driven_pricing_strategies.md
```diff
@@ -38,49 +38,49 @@ For better comprehension, please refer to the sample price-tiers configuration a
 
 #### UC scenarios for different (sub)types of pricing mechanisms
 
-| Scenario | Description                                                                                                 |
-|:---------|:------------------------------------------------------------------------------------------------------------|
-| S-1      | Tier-based pricing: 12GB of usage falls into two different tiers, resulting in two separate charge records. |
-| S-2      | Volume-based pricing: 12GB of usage falls into the highest tier, resulting in a single charge record.       |
+| Scenario | Description                                                                                              |
+| :------- | :------------------------------------------------------------------------------------------------------- |
+| S-1      | Tier-based pricing: 12GB of usage falls into two different tiers, resulting in two separate charge rows. |
```
flanakin, line 43: This is the only change in this file. The formatting changes were auto-applied on save.

supporting_content/appendix/quantity_driven_pricing_strategies.md
```diff
@@ -38,49 +38,49 @@ For better comprehension, please refer to the sample price-tiers configuration a
...
+| S-2      | Volume-based pricing: 12GB of usage falls into the highest tier, resulting in a single charge row.       |
```
flanakin, line 44: These is the only change in this file. The formatting changes were auto-applied on save.

supporting_content/dimensions/chargefrequency.md
```diff
@@ -57,10 +55,11 @@ Current values observed in billing data for various scenarios:
...
-- An upfront RI is a cost incurred in one billing period, similar to EC2 usage. Although neither will recur into another billing period, the upfront RI will be a 'one-time' charge and EC2 usage will be 'usage-based'. Another separate and distinct upfront RI or record of EC2 usage can be charged in the next period, if they occur. If a upfront RI is bought every month, they will be individual 'one-time' charges and cannot be considered as 'recurring'. 
+- "Usage-Based" value means that charges will be incurred when usage occurs. If there's a charge that happens monthly, even if it's based on usage, it would have a frequency 'recurring' as it's occurrence does not depend on usage consumption but other contractual terms.
+- An upfront RI is a cost incurred in one billing period, similar to EC2 usage. Although neither will recur into another billing period, the upfront RI will be a 'one-time' charge and EC2 usage will be 'usage-based'. Another separate and distinct upfront RI or row of EC2 usage can be charged in the next period, if they occur. If a upfront RI is bought every month, they will be individual 'one-time' charges and cannot be considered as 'recurring'.
```
flanakin, line 59: This is the only change in this file. The formatting changes were auto-applied on save.

udam-f2:
> @flowerinthenight noticed we use both "row" and "record" as generic references to a charge. "Row" is used more, so at least normalizing on that. There are a few places where "charge" makes less sense or is needed to differentiate from the monetary value of a "charge".

Any way to turn off the auto format that seems to be updating all kinds of stuff? I think unless we all agree on a single code style outside of standard markdown lint rules we've enabled, we should avoid auto-formatting. Otherwise different people will auto format differently and every change will come with a bunch of unrelated changes.

udam-f2 (approved):

flanakin:
> Any way to turn off the auto format that seems to be updating all kinds of stuff?
Ideally, we'd have rules defined within our project so it automatically aligns to whatever rules we have. That said, yes, I can turn it off. Do you want me to undo what's already in the PR or leave them? I'm good either way.

salpaysenturus:
My understanding was that "line item" and "charge" are synonyms.  Given that many of these commits change "line item" to "row" instead of "charge", does suggest that "line item" != "charge"?  If so, the glossary needs to reflect that.

flanakin:
> My understanding was that "line item" and "charge" are synonyms. Given that many of these commits change "line item" to "row" instead of "charge", does suggest that "line item" != "charge"? If so, the glossary needs to reflect that.

All the words are relatively synonymous. This was just an attempt to minimize the variations of the less frequently used words. 

It got raised in VS2 because of use of "charge" and "row" in the same context. I looked and found 219 instances of "charge", 10 of "row", 4 of "line item", and 3 of "record" in the core spec. There are some places where we use charge explicitly referencing the monetary value or implicitly the monetary value of a row, so I didn't want to make a huge pass on every use. I just wanted to get rid of the excess words that were used less frequently. We should do a larger pass thru things after 1.0 is out.

PR #251, opened 2023-11-06
Add Support Service Category

yprikhodko:
Use case to allow practitioners easily distinguish support related services and charge items

### DIFF
--- a/specification/dimensions/servicecategory.md
+++ b/specification/dimensions/servicecategory.md
@@ -47,7 +47,7 @@ Allowed values:
 | Security                  | Security monitoring and compliance services.                                                   |
 | Storage                   | Storage services for structured or unstructured data.                                          |
 | Web                       | Services enabling cloud applications to interact via the Internet.                             |
-| Support                   | Platform support related services                                                              |
+| Support                   | Platform services related to customer support.                                                             |
 | Other                     | New or emerging services that do not align with an existing category.                          |
 
 ## Introduced (version)
--- a/specification/dimensions/servicecategory.md
+++ b/specification/dimensions/servicecategory.md
@@ -47,7 +47,7 @@ Allowed values:
 | Security                  | Security monitoring and compliance services.                                                   |
 | Storage                   | Storage services for structured or unstructured data.                                          |
 | Web                       | Services enabling cloud applications to interact via the Internet.                             |
-| Support                   | Platform services related to customer support.                                                             |
+| Support                   | Provider services related to customer support.                                                 |
 | Other                     | New or emerging services that do not align with an existing category.                          |
 
 ## Introduced (version)
--- a/specification/dimensions/servicecategory.md
+++ b/specification/dimensions/servicecategory.md
@@ -47,6 +47,7 @@ Allowed values:
 | Security                  | Security monitoring and compliance services.                                                   |
 | Storage                   | Storage services for structured or unstructured data.                                          |
 | Web                       | Services enabling cloud applications to interact via the Internet.                             |
+| Support                   | Platform support related services                                                              |
 | Other                     | New or emerging services that do not align with an existing category.                          |
 
 ## Introduced (version)


specification/dimensions/servicecategory.md
```diff
@@ -47,6 +47,7 @@ Allowed values:
 | Security                  | Security monitoring and compliance services.                                                   |
 | Storage                   | Storage services for structured or unstructured data.                                          |
 | Web                       | Services enabling cloud applications to interact via the Internet.                             |
+| Support                   | Platform support related services                                                              |
```
salpaysenturus, line 50: We need a full sentence, and it would be best to avoid use of the term "support" when defining this.  Something like this:

```suggestion
| Support                   | Platform assistance related services.                                                             |
```

yprikhodko: We have similar approach across other services where description uses the cost category term, like Storage or Security etc. I think description here is not necessarily definition of the term, but more like expansion of category. Maybe we could phrase it like 'Platform customer support related services'?

salpaysenturus: That would be fine with me.  But maybe to make it a little more English-readable:

```suggestion
| Support                   | Platform services related to customer support.                                                             |
```

salpaysenturus (changes requested):
Line 38, the definition of Multicloud: It would be worth changing the definition there to not start with the word "Support".  (I would have made a more explicit suggestion, but unfortunately Github doesn't allow you to suggest changes outside of the range of changed lines.  [Very annoying](https://github.com/orgs/community/discussions/4452).)

specification/dimensions/servicecategory.md
flanakin, line 50: Looking at the principles for when to add new service categories, 2 apply here:
1. "Each category should be unique and should not be a logical child of another category (e.g., Containers are a subclassification of Compute)."
2. "Each category should have around 3-10 services for the major cloud providers. Avoid categories with only 1 or with over 20 services. No category should have more than 20% of services."

re: #1, isn't support technically a subset of Management and Governance?

re: #2, how many different support services are there? Do we have enough to warrant a separate category? Does consulting fit here?

I don't fundamentally have an issue with this. We can make an exception to the principles if there's explicit value in decoupling this from M&G. (Although that's a slippery slope.) I could see value in that from the perspective of this isn't really a function of the cloud, but of operating in the cloud. But that's also what M&G is about. It makes me wonder if there are other things we'd want to lump together with support. I'm curious what other practitioners think.

salpaysenturus: > isn't support technically a subset of Management and Governance?

Currently, according to [this FOCUS category mapping document](https://docs.google.com/spreadsheets/d/1s8FYVxAkDs2aeHPckEJrjA36Vpn4HKXpkXU0L-O8IAY/edit#gid=2123051243), Support services are mapped to Other for both AWS and Azure.  Do you know who was responsible for authoring that document?

yprikhodko: One of the disadvantages to add support to Management and Governance is support charges will be mixed with usage based services. Also often practitioners need to have special cost allocation rules for support which are different from usage based services hence having it as category would help to quickly focus on it regardless of provider specific service name(s) for support. Otherwise practitioners would need to always know the names of support services for every CSP. AWS has 4 different support service names for Developer, Business,OnRamp and Enterprise support tiers which is more than services in Mobile category for example

salpaysenturus (approved):
👍

cnharris10 (approved):

specification/dimensions/servicecategory.md
flanakin: > Currently, according to this FOCUS category mapping, Support services are mapped to Other for both AWS and Azure. Do you know who was responsible for authoring that document?

We can update that. I'd agree it shouldn't be in Other. Anything in Other feels like either a mapping bug or a gap in FOCUS. In this case, the original intent was M&G. That said, I do like the idea of differentiating from management service costs. 

For context, the only reason I mentioned the principles was to ensure we're either ensure we're changing in a way they support or to amend them to account for a gap in the original logic we had when creating them. I want to protect us from, "Well you added Support, so therefore you should also add Serverless" (or some other random category). Principles can only be upheld if they're upheld but that shouldn't stop us from improving them.

> One of the disadvantages to add support to Management and Governance is support charges will be mixed with usage based services. 

This distinction of different types of services is interesting in and of itself. This feels like something that should be distinguished in some column. It reminds me of people asking for a column to identify IaaS, PaaS, and SaaS separately. I wonder if there's a classification of support, consulting, etc. that would pair well with those in a new column in the future 🤔 Thoughts for another day...

> AWS has 4 different support service names for Developer, Business,OnRamp and Enterprise support tiers

Wouldn't all of those have the same ServiceName but use different SkuPriceId values? From a "service" perspective, they're all the same. The only difference is which support SKU you chose, right?

kk09v (approved):

specification/dimensions/servicecategory.md
udam-f2: I think the main decision that we'll need to agree on is if this is a primary designation OR something that will happen when we introduce a Service Sub Category. I agree with the pain of dealing with support charges from the practitioner perspective. Dealt with it enough times. My main concern is if it fits the definition of a highest level categorization of a service. I admit the asnwer is subjective - as many decisions in this space are. 

I lean towards this is a sub category of the Management and Governance and not a primary category. 

FWIW, other specifications like the common cloud controls (CCC) project in LF are looking to use the FOCUS and are among the people that have called for the sub categorization (e.g. not just database, but relational database, data warehouse etc.)

ljadvey (approved):

rileyjenk (approved):

specification/dimensions/servicecategory.md
yprikhodko: > Wouldn't all of those have the same ServiceName but use different SkuPriceId values? From a "service" perspective, they're all the same. The only difference is which support SKU you chose, right?
They all have different service names and different SKUs

yprikhodko: Management and Governance covers specific services for cloud management observability and control. having support as subcategory will make it different from other categories which include only usage based services spend. 
Also to consolidate only usage based services spend practitioners now would need to filter out Support by sub-category or service name which will require some implementation efforts. In the future there might be sub-categories of Support itself e.g. proactive, reactive etc.

udam-f2 (approved):

specification/dimensions/servicecategory.md
udam-f2: We have ChargeType to differentiate usage based charges vs purchases/tax/adjustments. Service Category isn't to limit to 'usage' based services - instead its "the highest-level classification of a service based on the core function of the service".

Polls can be misleading (not saying this one is) but I lost track of it. I'm fine to not stand in the way of adding this. I suggest we get a thumbs up from @flanakin as well. If needed, conversation can be continued during the preview period.

udam-f2:
Closing due to no activity. Please re-open and bring up for discussion as needed.

PR #252, opened 2023-11-06
FOCUS #252: Identifying custom columns

flanakin:
FOCUS needs a way to differentiate between FOCUS and custom columns that won't conflict as FOCUS evolves and adds more columns over time.

**Proposal**
Prefix all custom columns with an underscore (e.g., `_Foo`) to distinguish them separately from FOCUS columns. FOCUS columns should not be prefixed as they are the official, standardized set of columns that are the primary focal point (see what I did there?) of the cost and usage data.

> [!NOTE]
> _See supporting content for details behind the different considerations that went into this proposal._

**Benefits**
1. FOCUS columns are not treated as "special" or "non-standard" by using a prefix.
2. Custom columns are easily identifiable consistently across providers.
3. Custom columns do not have a long prefix that would be frustrating to type into queries.
4. We don't have to worry about registering or managing vendor prefixes to avoid conflicts (e.g., Microstrategy and Microsoft).
6. Underscores are common and can be used in any client tool or database today.
7. FOCUS and custom columns would be separated when sorted alphabetically, which avoids custom columns being mixed in the middle of FOCUS columns.

**Examples**
- `AvailabilityZone`
- `BilledCost`
- `BillingAccountId`
- `BillingAccountName`
- `BillingPeriodEnd`
- `BillingPeriodStart`
- `ChargeDescription`
- `ChargeFrequency`
- `ChargePeriodEnd`
- `ChargePeriodStart`
- `ChargeCategory`
- `ChargeSubcategory`
- `CommitmentDiscountCategory`
- `CommitmentDiscountId`
- `CommitmentDiscountName`
- `CommitmentDiscountType`
- `EffectiveCost`
- `InvoiceIsuserName`
- `ListCost`
- `ListUnitPrice`
- `PricingMeasurement`
- `PricingModel`
- `PricingQuantity`
- `ProviderName`
- `PublisherName`
- `Region`
- `ResourceId`
- `ResourceName`
- `ResourceType`
- `ServiceCategory`
- `ServiceName`
- `SkuId`
- `SkuPriceId`
- `SubAccountId`
- `SubAccountName`
- `Tags`
- `UsageQuantity`
- `UsageUnit`
- `_AccountName` (existing Microsoft column)
- `_bill_bill_type` (existing AWS column)
- `_cost/myCostOverage` (existing OCI column)
- `_export_time` (existing GCP column)
- `_...`
- `_...`
- `_...`
- `_...`
- `_...`
- `_...`
- `_...`
- `_...`

> [!NOTE]
> _We say that providers SHOULD use the FOCUS naming conventions, but don't dictate it. Hence my examples above using the first custom column I noticed for each dataset._

---

**Examples if we use a vendor prefix**
- `AvailabilityZone`
- `aws_bill_bill_type` (existing AWS column)
- `aws_...`
- `aws_...`
- `aws_...`
- `aws_...`
- `aws_...`
- `aws_...`
- `aws_...`
- `aws_...`
- `aws_...`
- `aws_...`
- `aws_...`
- `aws_...`
- `aws_...`
- `aws_...`
- `aws_...`
- `aws_...`
- `aws_...`
- `aws_...`
- `aws_...`
- `BilledCost`
- `BillingAccountId`
- `BillingAccountName`
- `BillingPeriodEnd`
- `BillingPeriodStart`
- `ChargeDescription`
- `ChargeFrequency`
- `ChargePeriodEnd`
- `ChargePeriodStart`
- `ChargeCategory`
- `ChargeSubcategory`
- `CommitmentDiscountCategory`
- `CommitmentDiscountId`
- `CommitmentDiscountName`
- `CommitmentDiscountType`
- `EffectiveCost`
- `gcp_export_time` (existing GCP column)
- `gcp_...`
- `gcp_...`
- `gcp_...`
- `gcp_...`
- `gcp_...`
- `gcp_...`
- `gcp_...`
- `gcp_...`
- `gcp_...`
- `gcp_...`
- `InvoiceIsuserName`
- `ListCost`
- `ListUnitPrice`
- `ms_AccountName` (existing Microsoft column)
- `ms_...`
- `ms_...`
- `ms_...`
- `ms_...`
- `ms_...`
- `ms_...`
- `ms_...`
- `ms_...`
- `ms_...`
- `ms_...`
- `ms_...`
- `ms_...`
- `ms_...`
- `ms_...`
- `ms_...`
- `ms_...`
- `oci_cost/myCostOverage` (existing OCI column)
- `oci_...`
- `oci_...`
- `oci_...`
- `oci_...`
- `oci_...`
- `oci_...`
- `oci_...`
- `oci_...`
- `oci_...`
- `oci_...`
- `oci_...`
- `PricingMeasurement`
- `PricingModel`
- `PricingQuantity`
- `ProviderName`
- `PublisherName`
- `Region`
- `ResourceId`
- `ResourceName`
- `ResourceType`
- `ServiceCategory`
- `ServiceName`
- `SkuId`
- `SkuPriceId`
- `SubAccountId`
- `SubAccountName`
- `Tags`
- `UsageQuantity`
- `UsageUnit`


mike-finopsorg:
👍 LGTM

@rupagcp / @yprikhodko would like to hear your feedback on this one.

rupagcp:
great idea! should we have the vendor acronym before it, for users to do a comparison query, or easily differentiate fields in a flat list? Example – instead of 'resource.name', 'ResourceName' – it's 'GCP_resource.name' and 'Azure_ResourceName'

mike-finopsorg:
> great idea! should we have the vendor acronym before it, for users to do a comparison query, or easily differentiate fields in a flat list? Example – instead of 'resource.name', 'ResourceName' – it's 'GCP_resource.name' and 'Azure_ResourceName'

https://github.com/FinOps-Open-Cost-and-Usage-Spec/FOCUS_Spec/pull/252/files#diff-9a6b90af36f6588b9e2e91dded61e1c15455ae5a1d55fcb8cd83369a1fd8dd48 goes into the thinking here. the benefit would be if both GCP and Azure chooses to use _ResourceName then there is a positive benefit that its likely this data can be queried across vendors. Best to read the above document and see if you agree or not.

flanakin:
> great idea! should we have the vendor acronym before it, for users to do a comparison query, or easily differentiate fields in a flat list? Example – instead of 'resource.name', 'ResourceName' – it's 'GCP_resource.name' and 'Azure_ResourceName'

I started with vendor prefixes as well. I summarized it in the supporting content, but here's a synopsis:
1. There were concerns about needing to register vendor prefixes to avoid conflicts, which would be a lot of work.
2. Vendor prefixes would get intermixed with the FOCUS columns when sorting alphabetically rather than keeping all FOCUS columns together. (Minor usability issue.)
3. Vendor prefixes would require duplicating columns when multiple providers use the same name, which bloats data size.
4. We could use something like `Provider_`, but that gets long and the casing makes it look like other FOCUS columns, which might be confusing.
5. We could use lowercase to differentiate (e.g., `provider_`) or make it shorter (e.g., `p_`), but those still mix the custom columns in the middle of the provider columns.
6. If we only use a special character, then all the FOCUS and custom columns are 100% separated and it's short enough to not impact usability.

I'm not completely against a vendor prefix. I've always assumed we'd go that route. The main things I like about just underscore is that it keeps it short, consistent, and [from a UX perspective] cleaner and segregated from the FOCUS columns. I've been using a vendor-prefixed version in my own reporting since FOCUS 0.5 and having the prefixed columns in the middle of the FOCUS columns is always annoying when I need to scroll up and down past them. This will get even worse when there are multiple providers in the mix, making it even harder to get to the FOCUS columns. All that said, I personally prefer that over other, especially longer, prefixes.

rupagcp:
Ah i see - thanks for the source link with the details Mike, and the explanation Michael. LGTM !

flanakin:
> Ah i see - thanks for the source link with the details Mike, and the explanation Michael. LGTM !

@rupagcp, can you sign off to ack? Just want to make sure we record it. The only way we can move things forward is with explicit acks 🙂

rupagcp (commented):
Ack / LGTM

arrarama:
Couldn't we do Provider_ with the <Provider> being an acronym that is registered in our exception list? I am worried about 2 providers having the same column name leading to confusion.

lbernalposada:
Agree with @arrarama, how are we making sure we do not have duplicate names?

flanakin:
VS2 discussed this and decided to use the vendor prefix to make it easier to identify the source of the column. I believe that aligns with what y'all were thinking as well. Updated the PR.

supporting_content/attributes/column_naming_convention.md
```diff
@@ -16,3 +16,57 @@ Examples of column names found in provider datasets:
 - Group wanted to get naming convention adopted broadly for provider columns as well, to simplify reporting and analysis by practitioners
 - Different conventions like camel case, snake case, and title case were considered
 - Discussion included different columns naming conventions used by databases
+
+### How to handle custom columns
```
salpaysenturus, line 20: I'm trying to get a sense of these rules.  Would any of the following be allowed: 

- "aws_linelitem/LineItemDescription" (mixed us of "_" and "/")
- "aws_lineitem_LineItemDescription" (not strictly pascal case)
- "aws_LineItemDescription" (does not match precise name in AWS data dictionary)

supporting_content/attributes/column_naming_convention.md
```diff
@@ -16,3 +16,57 @@ Examples of column names found in provider datasets:
...
+  - **Separator character:**
+    - Should be easily identifiable as a separator.
+      - Common: `_`, `-`, `.`, `:`. `@`, `/`, `~`, `|`, `+`, `=`, `;`
+      - Uncommon: `$`, `&`, `#`, `!`, `?`, `*`
+    - Should avoid characters that don't work in most tools or programming languages, since they would need to divert from the spec.
+      - Can be used in variable names: `_`
+      - Can be logically handled via nesting: `.`, `/`, `:`
+      - Limited support: `-`, `$`, `@`, `#`. `~`
+    - Prefer characters that are commonly used in spreadsheet and database column names.
+      - Common: `_`, `.`, `/`
```
salpaysenturus, line 38: Forward slashes and periods are generally frowned up on in database columns, as you have to put brackets around them to make them work in things like SQL queries.... Underscores are best.  Should the spec be more specific and/or stronger in its recommendation to avoid these?

specification/attributes/column_naming_convention.md
```diff
@@ -20,16 +18,20 @@ Naming convention for columns appearing in billing data.
...
-* Column IDs MUST NOT use abbreviations.
-* Column IDs SHOULD NOT use acronyms.
-* Column IDs MUST be alphanumeric with no special characters.
-* Columns that have an ID and a Name MUST have the Id or Name suffix in the Column ID. Display Name for a Column MAY
-  avoid the Name suffix if it is considered superfluous
+- All columns defined by FOCUS MUST follow the following rules:
+  - Column IDs MUST use [Pascal case](https://techterms.com/definition/pascalcase).
+  - Column IDs MUST NOT use abbreviations.
+  - Column IDs SHOULD NOT use acronyms.
+  - Column IDs MUST be alphanumeric with no special characters.
```
salpaysenturus, line 25: What does "special characters" mean?  The composition of special characters changes from one platform to the next, and many such lists include underscores.  The spec should define these or link to an appropriate list.

flanakin: This line is about FOCUS columns and already exists today. We're saying that we cannot use underscores today. The one exception that's part of this PR is a vendor prefix for custom columns. If you think we should change it, can you start a thread in Slack or maybe just a separate PR? Not pushing back. Just don't want to add unrelated changes to this PR to avoid more delays.

supporting_content/attributes/column_naming_convention.md
flanakin: We're not recommending them at all. We explicitly say you can't. This is just supporting content covering the thoughts that led up to the spec. I think we should be good from that perspective.

supporting_content/attributes/column_naming_convention.md
flanakin: All we're saying in this PR is that every vendor MUST prefix their columns with some recognizable lowercase + underscore prefix. We already say they SHOULD use the same rules as we do, but that also means they do any of the 3 options you mentioned. The goal of "SHOULD" was to make provider adoption simpler. In general, we've thought about starting with "SHOULD" and moving some of those to "MUST" in the second release.

specification/attributes/column_naming_convention.md
```diff
@@ -19,16 +18,20 @@ Naming convention for columns appearing in billing data.
...
+  - Columns that have an ID and a Name MUST have the `Id` or `Name` suffix in the Column ID. Display Name for a Column MAY avoid the `Name` suffix if there are no other columns with the same name prefix.
+- Custom columns SHOULD follow the same rules as FOCUS columns listed above.
+- All custom column IDs MUST start with a vendor prefix string and an underscore (e.g., `abc_ColumnName`).
```
marc-perreaut, line 28: My preference would go to slash '/' or colon ': ' rather underscore '_', because underscore would make column names look like a mix of Snake case and Pascale case.
That is my first impression, but maybe I will get used to  '_'.
Curious to know what other think.

specification/attributes/column_naming_convention.md
```diff
@@ -19,16 +18,20 @@ Naming convention for columns appearing in billing data.
...
+  - Vendor prefixes MUST be lowercase, alphanumeric strings and SHOULD NOT be longer than 10 characters in length.
+  - Vendor prefixes MUST be representative of the company or brand that generated the column.
```
marc-perreaut, line 30: There is a discussion about how to avoid prefix conflicts between different vendors and a proposition here:
https://github.com/FinOps-Open-Cost-and-Usage-Spec/FOCUS_Spec/pull/264/files

Another possibility to avoid conflicts: vendors could use URL domain names as vendor prefix, for example: azr.io, aws.io, etc ... Vendors might need to buy domain names to get short prefixes.

What's also interesting with this:
- As a practitioner,  when I discover a prefix, I can go to the website to discover which vendor it is
- A a vendor, I can document my FOCUS column on the website

specification/attributes/column_naming_convention.md
flanakin: The concern there is that most databases don't support them. Look at the supporting content. All the considerations are there. That said, I know slash will likely come in when we do grouping, so that's also something to consider.

specification/attributes/column_naming_convention.md
flanakin: Can you leave this feedback on that PR? I'd like to separate registration from at least having a prefix.

specification/attributes/column_naming_convention.md
```diff
@@ -19,16 +19,18 @@ Naming convention for columns appearing in billing data.
...
 
-* Column IDs MUST use [Pascal case](https://techterms.com/definition/pascalcase).
-* Column IDs MUST NOT use abbreviations.
-* Column IDs SHOULD NOT use acronyms.
-* Column IDs MUST be alphanumeric with no special characters.
-* Columns that have an ID and a Name MUST have the Id or Name suffix in the Column ID. Display Name for a Column MAY
-  avoid the Name suffix if it is considered superfluous
+- All columns defined by FOCUS MUST follow the following rules:
+  - Column IDs MUST use [Pascal case](https://techterms.com/definition/pascalcase).
+  - Column IDs MUST NOT use abbreviations.
```
cnharris10, line 24:
```suggestion
  - Column IDs MUST NOT use abbreviations beyond exceptions listed within the Column Naming Convention guidelines. 
```

cnharris10 (approved):

rileyjenk (approved):

specification/attributes/column_naming_convention.md
flanakin: Does this mean we need to add this caveat for every exception? I was assuming the exception was enough to call out that it's overriding the defined rules 🤔 Ultimately, I'm fine either way.

specification/attributes/column_naming_convention.md
```diff
@@ -3,7 +3,7 @@
 Column IDs provided in cost data following a consistent naming convention reduces friction for FinOps practitioners
 that consume the data for analysis, reporting, and other use cases.
```
flanakin, line 4:
```suggestion
Column IDs provided in cost data following a consistent naming convention reduces friction for FinOps practitioners that consume the data for analysis, reporting, and other use cases.
```

specification/attributes/column_naming_convention.md
```diff
@@ -19,16 +19,18 @@ Naming convention for columns appearing in billing data.
...
-* Columns that have an ID and a Name MUST have the Id or Name suffix in the Column ID. Display Name for a Column MAY
-  avoid the Name suffix if it is considered superfluous
+- All columns defined by FOCUS MUST follow the following rules:
+  - Column IDs MUST use [Pascal case](https://techterms.com/definition/pascalcase).
+  - Column IDs MUST NOT use abbreviations.
+  - Column IDs SHOULD NOT use acronyms.
+  - Column IDs MUST be alphanumeric with no special characters.
+  - Columns that have an ID and a Name MUST have the `Id` or `Name` suffix in the Column ID. Display Name for a Column MAY avoid the Name suffix if there are no other columns with the same name prefix.
+- Custom columns SHOULD follow the same rules as FOCUS columns listed above.
+- All custom columns MUST be prefixed with an underscore (`_`) to distinguish them from FOCUS columns and avoid conflicts in future releases.
```
flanakin, line 29:
```suggestion
* All columns defined by FOCUS MUST follow the following rules:
  * Column IDs MUST use [Pascal case](https://techterms.com/definition/pascalcase).
  * Column IDs MUST NOT use abbreviations.
  * Column IDs SHOULD NOT use acronyms.
  * Column IDs MUST be alphanumeric with no special characters.
  * Columns that have an ID and a Name MUST have the `Id` or `Name` suffix in the Column ID. Display Name for a Column MAY avoid the Name suffix if there are no other columns with the same name prefix.
* Custom columns SHOULD follow the same rules as FOCUS columns listed above.
* All custom columns MUST be prefixed with an underscore (`_`) to distinguish them from FOCUS columns and avoid conflicts in future releases.
* All FOCUS columns MUST be first in provided dataset.
  * Custom columns MUST be listed after all FOCUS columns and MUST NOT be intermixed.
  * FOCUS and custom columns MAY be sorted alphabetically, but MUST remain separate.
```

supporting_content/attributes/column_naming_convention.md
flanakin: UPDATE: Based on feedback in Slack (and since nobody seems to want to sign off on the vendor prefix), I removed the vendor prefix. Now the requirement is to add a single underscore (`_`). Everything else is the same. Resolving this, but feel free to reopen or leave another comment.

specification/attributes/column_naming_convention.md
flanakin: Given this is separate from the prefix, I'm going to "won't fix" this one. Let's chat in Slack. If you feel strongly, I'm happy to push out a separate PR.

specification/attributes/column_naming_convention.md
```diff
@@ -19,16 +18,21 @@ Naming convention for columns appearing in billing data.
...
+  * Column IDs MUST use [Pascal case](https://techterms.com/definition/pascalcase).
+  * Column IDs MUST NOT use abbreviations.
+  * Column IDs SHOULD NOT use acronyms.
+  * Column IDs MUST be alphanumeric with no special characters.
+  * Columns that have an ID and a Name MUST have the `Id` or `Name` suffix in the Column ID. Display Name for a Column MAY avoid the Name suffix if there are no other columns with the same name prefix.
+* Custom columns SHOULD follow the same rules as FOCUS columns listed above.
+* All custom columns MUST be prefixed with an underscore (`_`) to distinguish them from FOCUS columns and avoid conflicts in future releases.
+* All FOCUS columns MUST be first in provided dataset.
+  * Custom columns MUST be listed after all FOCUS columns and MUST NOT be intermixed.
+  * FOCUS and custom columns MAY be sorted alphabetically, but MUST remain separate.
```
flanakin, line 31:
```suggestion
* All FOCUS columns MUST be first in the provided dataset.
  * Custom columns MUST be listed after all FOCUS columns and MUST NOT be intermixed.
  * Columns MAY be sorted alphabetically but custom columns MUST be after all FOCUS columns.
```

specification/attributes/column_naming_convention.md
```diff
@@ -19,16 +18,21 @@ Naming convention for columns appearing in billing data.
...
-* Column IDs MUST be alphanumeric with no special characters.
-* Columns that have an ID and a Name MUST have the Id or Name suffix in the Column ID. Display Name for a Column MAY
-  avoid the Name suffix if it is considered superfluous
+* All columns defined by FOCUS MUST follow the following rules:
+  * Column IDs MUST use [Pascal case](https://techterms.com/definition/pascalcase).
+  * Column IDs MUST NOT use abbreviations.
+  * Column IDs SHOULD NOT use acronyms.
+  * Column IDs MUST be alphanumeric with no special characters.
+  * Columns that have an ID and a Name MUST have the `Id` or `Name` suffix in the Column ID. Display Name for a Column MAY avoid the Name suffix if there are no other columns with the same name prefix.
+* Custom columns SHOULD follow the same rules as FOCUS columns listed above.
```
udam-f2, line 27: The concept of a custom column is not defined anywhere. Would provider-defined column be a little more clearer?
```suggestion
* Provider-defined columns SHOULD follow the same rules as FOCUS columns listed above.
```

specification/attributes/column_naming_convention.md
```diff
@@ -19,16 +18,21 @@ Naming convention for columns appearing in billing data.
...
+* All custom columns MUST be prefixed with an underscore (`_`) to distinguish them from FOCUS columns and avoid conflicts in future releases.
```
udam-f2, line 28:
```suggestion
* All provider-defined columns MUST be prefixed with an underscore (`_`) to distinguish them from FOCUS columns and avoid conflicts in future releases.
```

specification/attributes/column_naming_convention.md
```diff
@@ -19,16 +18,21 @@ Naming convention for columns appearing in billing data.
...
+* All FOCUS columns MUST be first in the provided dataset.
+  * Custom columns MUST be listed after all FOCUS columns and MUST NOT be intermixed.
```
udam-f2, line 30:
```suggestion
  * Provider-defined columns MUST be listed after all FOCUS columns and MUST NOT be intermixed.
```

specification/attributes/column_naming_convention.md
```diff
@@ -19,16 +18,21 @@ Naming convention for columns appearing in billing data.
...
+  * Columns MAY be sorted alphabetically but custom columns MUST be after all FOCUS columns.
```
udam-f2, line 31:
```suggestion
  * Columns MAY be sorted alphabetically but provider-defined columns MUST be after all FOCUS columns.
```

specification/attributes/currency_code_format.md
```diff
@@ -2,7 +2,7 @@
 
 Columns that contain currency information in cost data following a consistent format reduces friction for FinOps practitioners that consume the data for analysis, reporting, and other use cases.
 
-All columns capturing a currency value, defined in the FOCUS specification, MUST follow the requirements listed below. Provider-generated, currency-related columns SHOULD adopt the same format requirements over time.
+All columns capturing a currency value, defined in the FOCUS specification, MUST follow the requirements listed below. Custom currency-related columns SHOULD also follow the same formatting requirements.
```
udam-f2, line 5:
```suggestion
All columns capturing a currency value, defined in the FOCUS specification, MUST follow the requirements listed below. Provider-defined currency-related columns SHOULD follow the same formatting requirements.
```

specification/attributes/datetime_format.md
```diff
@@ -2,7 +2,7 @@
 
 Columns that provide date and time information conforming to specified rules and formatting requirements ensure clarity, accuracy, and ease of interpretation for both humans and systems.
 
-All columns capturing a date/time value, defined in the FOCUS specification, MUST follow the formatting requirements listed below. Provider-generated, date/time-related columns SHOULD adopt the same format requirements over time.
+All columns capturing a date/time value, defined in the FOCUS specification, MUST follow the formatting requirements listed below. Custom date/time-related columns SHOULD also follow the same formatting requirements.
```
udam-f2, line 5:
```suggestion
All columns capturing a date/time value, defined in the FOCUS specification, MUST follow the formatting requirements listed below. Provider-defined date/time-related columns SHOULD follow the same formatting requirements.
```

specification/attributes/null_handling.md
```diff
@@ -3,7 +3,7 @@
 Cost data rows that don't have a value that can be presented for a column must be handled in a consistent way to reduce
 friction for FinOps practitioners that consume the data for analysis, reporting, and other use cases.
 
-All columns defined in the FOCUS specification MUST follow the null handling requirements listed below. Provider-generated columns SHOULD adopt these same null handling requirements over time.
+All columns defined in the FOCUS specification MUST follow the null handling requirements listed below. Custom columns SHOULD also follow the same formatting requirements.
```
udam-f2, line 6:
```suggestion
All columns defined in the FOCUS specification MUST follow the null handling requirements listed below. Provider-defined columns SHOULD follow the same formatting requirements.
```

supporting_content/attributes/column_naming_convention.md
udam-f2, line 20:
```suggestion
### How to handle provider-defined columns
```

supporting_content/attributes/column_naming_convention.md
```diff
@@ -16,3 +16,57 @@ Examples of column names found in provider datasets:
...
+      - Examples: `Custom_`, `External_`, `Provider_`, `Vendor_`
+    - A lowercase constant string would distinguish it from FOCUS column naming conventions, but could still make column names exceedingly long.
+      - Examples: `custom_`, `provider_`, `vendor_`
+    - An abbreviated or very short constant string would keep names short, but goes against the no abbreviations rule.
+      - Examples: `alt_`, `aux`, `col_`, `dim_`, `ext_`, `m_`, `oth_`, `p_`, `src_`
+    - Using a single character alone would be the shortest and would distinguish it from the FOCUS columns that don't have a prefix.
+      - Examples: `_`, `.`, `~`, `$`, `#`, `@`
+      - Using a special character first would also keep all provider columns together outside of the core FOCUS columns when columns are sorted alphabetically.
+      - Some programming languages use an underscore (`_`) as a convention to indicate that a property is private. While these values are not "private", it might be a good way to indicate that these columns are not part of the core FOCUS columns.
+    - BEST: **None** (underscore only).
```
udam-f2, line 64:
```suggestion
* Providers need to add additional details that are not available in FOCUS columns to support additional FinOps scenarios.
* FOCUS needs to allow the addition of these columns without the risk of collision with future FOCUS columns.
* Considerations:
  * **Placement (prefix vs. suffix):**
    * A prefix would be easier to identify as a provider-defined column even with limited space.
    * A suffix would require expanding the column or explicitly looking at the end of the column name, which can be difficult is some client tools where the end of the name gets cut off.
    * BEST: **Prefix**.
  * **Separator character:**
    * Should be easily identifiable as a separator.
      * Common: `_`, `-`, `.`, `:`. `@`, `/`, `~`, `|`, `+`, `=`, `;`
      * Uncommon: `$`, `&`, `#`, `!`, `?`, `*`
    * Should avoid characters that don't work in most tools or programming languages, since they would need to divert from the spec.
      * Can be used in variable names: `_`
      * Can be logically handled via nesting: `.`, `/`, `:`
      * Limited support: `-`, `$`, `@`, `#`. `~`
    * Prefer characters that are commonly used in spreadsheet and database column names.
      * Common: `_`, `.`, `/`
      * Uncommon: `-`, `:`. `~`, `+`, `$`, `#`
    * BEST: **Underscore** (`_`).
  * **Identifying string (FOCUS vs. provider-specific vs. constant vs. none):**
    * A FOCUS prefix allows providers to add FOCUS columns to existing datasets without breaking anyone, which could make adoption simpler
      * Examples: `FOCUS_`, `focus_`, `f_`
      * This would double the size of the dataset as new columns are added.
      * It's unlikely that any provider will add FOCUS columns to an existing datasets without an explicit opt-in, which means we likely wouldn't see the benefit of this approach since practitioners still need to opt-in and change their reporting.
      * This makes the spec feel like it's a supplemental thing, rather than the primary thing that should be used.
      * Practitioners will need to explicitly switch to the new columns anyway, so the value of augmenting existing provider-specific datasets provides little to no value.
      * In general, this isn't approach any other open specification tasks for its primary dataset.
      * Prefixing provider-defined columns establishes FOCUS as _the_ schema and provider-defined columns as the exceptional thing that should be caveated.
    * A provider ID would make the source of the data easily identifiable, but requires every provider to have new columns which would bloat the column count of a consolidated dataset.
      * Examples: `ali_`, `aws_`, `gcp_`, `ibm_`, `ms_`, `oci_`, `tencent_`, `vmw_`
      * If we use a provider prefix, there are concerns about conflicting names that would require a registration system, which is a lot of work.
      * Ultimately, there shouldn't be a need to use provider-defined prefixes since provider-specific columns will never conflict due to data being in separate rows.
    * A constant string would allow providers to reuse column names, but would make column names longer. This might also look like a FOCUS column if capitalized in the same way.
      * Examples: `Custom_`, `External_`, `Provider_`, `Vendor_`
    * A lowercase constant string would distinguish it from FOCUS column naming conventions, but could still make column names exceedingly long.
      * Examples: `custom_`, `provider_`, `vendor_`
    * An abbreviated or very short constant string would keep names short, but goes against the no abbreviations rule.
      * Examples: `alt_`, `aux`, `col_`, `dim_`, `ext_`, `m_`, `oth_`, `p_`, `src_`
    * Using a single character alone would be the shortest and would distinguish it from the FOCUS columns that don't have a prefix.
      * Examples: `_`, `.`, `~`, `$`, `#`, `@`
      * Using a special character first would also keep all provider columns together outside of the core FOCUS columns when columns are sorted alphabetically.
      * Some programming languages use an underscore (`_`) as a convention to indicate that a property is private. While these values are not "private", it might be a good way to indicate that these columns are not part of the core FOCUS columns.
    * BEST: **None** (underscore only).
```

supporting_content/attributes/column_naming_convention.md
flanakin: I changed all the "provider" references to "custom" because there will be other tools and libraries that will generate custom columns that don't come from cloud or service providers.

supporting_content/attributes/column_naming_convention.md
```diff
@@ -5,14 +5,68 @@
 Examples of column names found in provider datasets:
 
 | Provider  | Data set                 | Column                |
-|-----------|--------------------------|-----------------------|
+| --------- | ------------------------ | --------------------- |
 | AWS       | CUR                      | bill_invoicing_entity |
 | GCP       | Big Query Billing Export | resource.global_name  |
 | Microsoft | Cost details             | ResourceId            |
```
arrarama, line 11:
```suggestion
| Microsoft | Cost details             | ResourceId            |
| OCI       | Cost reports             | product/resourceId    | 
```

specification/attributes/column_naming_convention.md
udam-f2: Sorry, trying to understand if this was addressed somehow or if there's a explanation for why we think 'Custom columns' is clear..

specification/attributes/column_naming_convention.md
marc-perreaut: Ok, but some column names of some databases like Oracle are case insensitive by default, while FOCUS uses Pascal case.
Moreover, AWS columns in CUR contain slash.
So I believe data ingestion would do some transformation anyway to cope with database naming constraints.
But Ok I understand the rationale, thanks!

specification/attributes/column_naming_convention.md
yprikhodko: Columns starting with '_' might require special treatment by different DB/query engines hence cause unnecessary issues for practitioners. For example '_' needs to be encapsulated in Amazon Athena and it's one of the commonly used services for querying cost and usage data on AWS (https://docs.aws.amazon.com/athena/latest/ug/tables-databases-columns-names.html#names-that-begin-with-an-underscore). My opinion that using pattern 'provider_customname' would help to avoid this. 
I understand that you won't be able to group all custom columns by sorting, but at the same time all modern IDEs or query editors have search filters which allow quickly search for columns from particular provider by name

specification/attributes/column_naming_convention.md
```diff
@@ -19,16 +18,21 @@ Naming convention for columns appearing in billing data.
...
-  avoid the Name suffix if it is considered superfluous
+* All columns defined by FOCUS MUST follow the following rules:
+  * Column IDs MUST use [Pascal case](https://techterms.com/definition/pascalcase).
+  * Column IDs MUST NOT use abbreviations.
+  * Column IDs SHOULD NOT use acronyms.
+  * Column IDs MUST be alphanumeric with no special characters.
+  * Columns that have an ID and a Name MUST have the `Id` or `Name` suffix in the Column ID. Display Name for a Column MAY avoid the Name suffix if there are no other columns with the same name prefix.
+* Custom columns SHOULD follow the same rules as FOCUS columns listed above.
+* All custom columns MUST be prefixed with an underscore (`_`) to distinguish them from FOCUS columns and avoid conflicts in future releases.
+* All FOCUS columns MUST be first in the provided dataset.
```
sarahcloudpm, line 29: Why are we being particular about the order? This seems odd.

BenShy (approved):

specification/attributes/column_naming_convention.md
flanakin: Usability. This has come up several times (even in 0.5).

specification/attributes/column_naming_convention.md
flanakin: Custom columns will come from many sources, not just "providers".

specification/attributes/column_naming_convention.md
flanakin:
```suggestion
* All FOCUS columns SHOULD be first in the provided dataset.
```

specification/attributes/column_naming_convention.md
flanakin, line 31:
```suggestion
  * Custom columns SHOULD be listed after all FOCUS columns and SHOULD NOT be intermixed.
  * Columns MAY be sorted alphabetically but custom columns SHOULD be after all FOCUS columns.
```

arrarama (approved):

specification/attributes/column_naming_convention.md
flanakin, line 28:
```suggestion
* All custom columns MUST be prefixed with a consistent `x_` prefix to identify them as external, custom columns and distinguish them from FOCUS columns to avoid conflicts in future releases.
```

supporting_content/attributes/column_naming_convention.md
```diff
@@ -5,14 +5,69 @@
...
+      * Examples: `ali_`, `aws_`, `gcp_`, `ibm_`, `ms_`, `oci_`, `tencent_`, `vmw_`
+      * If we use a provider prefix, there are concerns about conflicting names that would require a registration system, which is a lot of work.
+      * Ultimately, there shouldn't be a need to use custom prefixes since provider-specific columns will never conflict due to data being in separate rows.
+    * A constant string would allow providers to reuse column names, but would make column names longer. This might also look like a FOCUS column if capitalized in the same way.
+      * Examples: `Custom_`, `External_`, `Provider_`, `Vendor_`
+    * A lowercase constant string would distinguish it from FOCUS column naming conventions, but could still make column names exceedingly long.
+      * Examples: `custom_`, `provider_`, `vendor_`
+    * An abbreviated or very short constant string would keep names short, but goes against the no abbreviations rule.
+      * Examples: `alt_`, `aux`, `col_`, `dim_`, `ext_`, `m_`, `oth_`, `p_`, `src_`
```
flanakin, line 60:
```suggestion
      * Examples: `alt_`, `aux`, `col_`, `dim_`, `ext_`, `m_`, `oth_`, `p_`, `src_`, `x_`
```

supporting_content/attributes/column_naming_convention.md
```diff
@@ -5,14 +5,69 @@
...
+    * Using a single character alone would be the shortest and would distinguish it from the FOCUS columns that don't have a prefix.
+      * Examples: `_`, `.`, `~`, `$`, `#`, `@`
+      * Using a special character first would also keep all provider columns together outside of the core FOCUS columns when columns are sorted alphabetically.
+      * Some programming languages use an underscore (`_`) as a convention to indicate that a property is private. While these values are not "private", it might be a good way to indicate that these columns are not part of the core FOCUS columns.
+    * BEST: **None** (underscore only).
```
flanakin, line 65:
```suggestion
    * BEST: **x_**.
```

flanakin:
Updated to `x_` (as in "external") due to a few database systems not supporting underscore as a leading character. This convention was suggested similar to what HTTP headers do. Note that if any provider, vendor, or library wants to add more, they can also add their own identifier after this one (e.g., `x_abc_MyAwesomeColumn`). The only requirement is that it must start with `x_`.

udam-f2 (approved):
LGTM

anatri1:
Ack. LGTM.

PR #269, opened 2023-11-12
Region should be nullable

flanakin:


joaopedromatias:
Some CloudFront operations are not region specific billed, like `Invalidation` on cache hits

specification/dimensions/region.md
```diff
@@ -2,7 +2,9 @@
 
 A Region is a provider assigned identifier for an isolated geographic area where a resource is provisioned in and/or a service is provided from. Region is commonly used for scenarios like analyzing cost and prices based on where resources are deployed.
 
-The Region column MUST be present in the billing data. This column MUST be of type String and MUST NOT contain null values. Region values MUST be consistent within the provider and MUST be the same values used to indicate the region when provisioning or purchasing the resource or service.
+Region MUST follow the requirements listed below:
+
+Region MUST be present in the billing data and MUST be of type String. Region MUST NOT be null when a resource or service is operated in or managed from a distinct region. Region values MUST be consistent within the provider and MUST be the same values used to indicate the region when provisioning or purchasing the resource or service.
```
udam-f2, line 7:
```suggestion
Region MUST be present in the billing data and MUST be of type String. Region MUST NOT be null when a resource or service is operated in or managed from a distinct region. Region values MUST be consistent within the provider and MUST be the same values used to indicate the region when provisioning or purchasing the resource or service.
```

udam-f2 (approved):

specification/dimensions/region.md
flanakin: Doh! Sorry. I was debating making this a list instead of a paragraph, but then realized it's not as many rules as I thought. Looks like I forgot to clean up the header. Thanks! Good catch!

PR #273, opened 2023-11-14
Fix build errors

flanakin:


flanakin:
Udam covered this in his PR

PR #274, opened 2023-11-14
Remove pricing model from wd chargetype

aqu-erp:
Also updated the example table to include correct values for chargetype, chargefrequency, pricingcategory

Also fixed build error with seeing the earlier PR


flanakin (approved):

udam-f2 (approved):
You might rebase. The first issue was fixed. Supporting content will be now in a different file but you'll just have to manually merge it in probably.

PR #288, opened 2023-11-17
PricingSubcategory

flanakin:
Completing the breakdown of PricingCategory that we discussed over the past couple weeks.

Related #330


specification/columns/pricingsubcategory.md
```diff
@@ -0,0 +1,70 @@
+# Pricing Subcategory
+
+Pricing Subcategory describes the pricing model used for a charge within a specific [Pricing Category](#pricingcategory). It can be useful for distinguishing between charges that share the same Pricing Category, but have different pricing rules that may result in a lower prices, like flat-rate compared to volume-based discounts.
```
SanjnaSrivatsaVMware, line 3: Not entirely sure if the words "pricing model" are self-explanatory. Is there possibly another way to say that might be less 'left to interpretation' ?

flanakin: Hmm... 🤔 Will put some thought into it. I used the same thing we have for PricingCategory, but we can update both places. Anyone else have thoughts or ideas?

specification/columns/commitmentdiscountcategory.md
```diff
@@ -32,6 +32,7 @@ Allowed values:
 |:--------|:-------------------------------------------------------------------------|
 | Spend   | Commitment-based discounts that require a predetermined amount of spend. |
 | Usage   | Commitment-based discounts that require a predetermined amount of usage. |
+| Other   | Commitment-based discounts that do not fall into another category.       |
```
cnharris10, line 35: Is this just to make sure the column remains applicable for the future?  I don't know of any commitment models that fall outside of "Usage" or "Spend" offhand.

specification/columns/pricingsubcategory.md
```diff
@@ -0,0 +1,70 @@
...
+
+Allowed values when PricingCategory is "On-Demand":
+
+| Value    | Description                                                                                                           |
+| :------- | :-------------------------------------------------------------------------------------------------------------------- |
+| Standard | On-demand charges priced at a flat rate for a SKU that does not support volume-based or tiered pricing.               |
+| Tiered   | On-demand charges priced at a specific volume-based or tiered level.                                                  |
+| Bundled  | On-demand charges priced at a free or reduced rate due to a companion SKU or service that was also used or purchased. |
+| Preview  | On-demand charges priced at a free or reduced rate for a prerelease version of the SKU being used or purchased.       |
+| Other    | On-demand charges priced in a way not covered above.                                                                  |
```
cnharris10, line 51: Private Rates?

cnharris10 (changes requested):

specification/columns/commitmentdiscountcategory.md
flanakin: Correct, this is just about covering a new scenario that doesn't fit into the existing buckets. There are no other known commitment discount types today.

specification/columns/pricingsubcategory.md
flanakin: Is that the same as negotiated discounts? If so, that would still be "standard" pricing, just with reduced rates, right?

udam-f2:
Let's discuss if there's a need for a pricing subcategory in v1.0 and re-open accordingly. This hasn't yet gone through the standard process of git issue, and discussion in TFs. If it was a minor change, that would be fine but for introducing a new column, we need to go through the standard process.

flanakin:
This is part of #330. Let's not close it until we close that conversation please.

specification/columns/pricingsubcategory.md
marc-perreaut: I find that pricing model is a common concept, but to me the issue is PricingCategory and PricingSubcategory do not actually describe the pricing model but refer to categories and sub categories of used pricing models.

So maybe simply:
> Pricing Subcategory describes the sub category in the pricing model used for a charge within a specific Pricing Category

If the term pricing model is too abstract, maybe the term pricing type could be used instead.

specification/columns/pricingsubcategory.md
```diff
@@ -0,0 +1,70 @@
...
+| Allows nulls    | True           |
+| Data type       | String         |
+| Value format    | Allowed values |
+
+Allowed values when PricingCategory is "On-Demand":
+
+| Value    | Description                                                                                                           |
+| :------- | :-------------------------------------------------------------------------------------------------------------------- |
+| Standard | On-demand charges priced at a flat rate for a SKU that does not support volume-based or tiered pricing.               |
+| Tiered   | On-demand charges priced at a specific volume-based or tiered level.                                                  |
```
marc-perreaut, line 48: I see Tiered, Bundled, Preview as pricing dimensions, but not as pricing sub categories: they could combine together and apply to other pricing categories than On-Demand, even though I do not have an example in mind.

flanakin:
Deferring to post-1.0

PR #292, opened 2024-01-22
Fedora support

xkahn:
Add Fedora Linux install instructions. Other Linux distributions should be trivial to add support for. The only difference should be `apt-get` instead of `dnf` in the package command line.

However, I did not test this, so I have left it as a task for someone else.


xkahn:
The CLA is dumb and useless here.

1. This contribution is trival and shouldn't fall under copyright. 
2. I am not a corporate user and thus can't even sign the thing.

In other words, feel free to take the contribution without attribution. But also... fix your CLA-bot. It's silly.

tobrien:
@xkahn - thanks for the contribution.  And apologies for the EasyCLA barrier.   Because this is a standard we have to be very thorough for IP review.   More thorough than most other projects.

But, I acknowledge that it's a frustrating surprise.  Someone's going to take a quick look at this, but wanted to make sure to say thanks (even if you had a bad EasyCLA experience.)

udam-f2:
Closing due to no activity. Please re-open and bring up for discussion as needed.

PR #323, opened 2024-02-05
Numeric columns should not use 0 when no value

flanakin:
Adding additional clarity to ensure numeric columns do not use `0` in place of a null when there is no value.

### DIFF
--- a/specification/attributes/null_handling.md
+++ b/specification/attributes/null_handling.md
@@ -20,7 +20,7 @@ Indicates how to handle columns that don't have a value.
 ## Requirements
 
 * Columns MUST use NULL when there isn't a value that can be specified for a nullable column.
-* String columns MUST NOT use empty strings or placeholder values such as "Not Set" or "Not Applicable" in columns, regardless of whether the column allows nulls or not.
+* Columns MUST NOT use empty strings or placeholder values such as "Not Set" or "Not Applicable" in columns, regardless of whether the column allows nulls or not.
 * Numeric columns MUST use NULL and MUST NOT use 0 when there isn't a value, regardless of whether the column allows nulls or not.
 
 ## Exceptions
--- a/specification/attributes/null_handling.md
+++ b/specification/attributes/null_handling.md
@@ -20,8 +20,7 @@ Indicates how to handle columns that don't have a value.
 ## Requirements
 
 * Columns MUST use NULL when there isn't a value that can be specified for a nullable column.
-* Columns MUST NOT use empty strings or placeholder values such as "Not Set" or "Not Applicable" in columns, regardless of whether the column allows nulls or not.
-* Numeric columns MUST use NULL and MUST NOT use 0 when there isn't a value, regardless of whether the column allows nulls or not.
+* Columns MUST NOT use empty strings or placeholder values such as 0 for numeric columns or "Not Applicable" for string columns, regardless of whether the column allows nulls or not.
 
 ## Exceptions
 
--- a/specification/attributes/null_handling.md
+++ b/specification/attributes/null_handling.md
@@ -20,7 +20,8 @@ Indicates how to handle columns that don't have a value.
 ## Requirements
 
 * Columns MUST use NULL when there isn't a value that can be specified for a nullable column.
-* Columns MUST NOT use empty strings or placeholder values such as "Not Set" or "Not Applicable" in columns, regardless of if the column allows nulls or not.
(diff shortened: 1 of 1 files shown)


specification/attributes/null_handling.md
```diff
@@ -20,7 +20,8 @@ Indicates how to handle columns that don't have a value.
 ## Requirements
 
 * Columns MUST use NULL when there isn't a value that can be specified for a nullable column.
-* Columns MUST NOT use empty strings or placeholder values such as "Not Set" or "Not Applicable" in columns, regardless of if the column allows nulls or not.
+* String columns MUST NOT use empty strings or placeholder values such as "Not Set" or "Not Applicable" in columns, regardless of whether the column allows nulls or not.
```
ijurica, line 23:
```suggestion
* Columns MUST NOT use empty strings or placeholder values such as "Not Set" or "Not Applicable" in columns, regardless of whether the column allows nulls or not.
```
If we omit 'String,' the requirement is applicable to Date/Time and other data types?!

specification/attributes/null_handling.md
```diff
@@ -20,7 +20,8 @@ Indicates how to handle columns that don't have a value.
...
+* Numeric columns MUST use NULL and MUST NOT use 0 when there isn't a value, regardless of whether the column allows nulls or not.
```
udam-f2, line 24: I think the 'regardless of whether the column allows null or not' won't work if we're proposing people use null. If it's not allowing null, a valid value will have to be provided and often, the valid value will be 0 (e.g. cost columns). If the column allows NULL, then we don't want people using 0s when the column is not applicable.

specification/attributes/null_handling.md
rileyjenk: I question this as there are valid entries that may legitimately be "Not set". Can this be simplified saying that the spec calls for the use of Null to specify no data and does not allow for empty string.

specification/attributes/null_handling.md
udam-f2: This also means rows like tax would have a unit price of $0 and quantity of 0. Which seems like a bad idea (and also breaks P*Q.

flanakin: That was there before so I just left it. I believe the point is that, even if the spec "requires" a value, there may be a scenario where a provider can't supply it. And in that case, they MUST NOT use a placeholder. At least that's how I read it. The line I added about numbers was to ensure they don't use `0` in that case, since it would break validation scenarios. If there's no value, it needs to be null to convey that there's no value.

specification/attributes/null_handling.md
```diff
@@ -20,7 +20,8 @@ Indicates how to handle columns that don't have a value.
 ## Requirements
 
 * Columns MUST use NULL when there isn't a value that can be specified for a nullable column.
-* Columns MUST NOT use empty strings or placeholder values such as "Not Set" or "Not Applicable" in columns, regardless of if the column allows nulls or not.
+* Columns MUST NOT use empty strings or placeholder values such as "Not Set" or "Not Applicable" in columns, regardless of whether the column allows nulls or not.
+* Numeric columns MUST use NULL and MUST NOT use 0 when there isn't a value, regardless of whether the column allows nulls or not.
```
flanakin, line 24:
```suggestion
* Columns MUST NOT use empty strings or placeholder values such as 0 for numeric columns or "Not Applicable" for string columns, regardless of whether the column allows nulls or not.
```

specification/attributes/null_handling.md
flanakin: @rileyjenk The first bullet in the requirements conveys that null must be used. Is that good enough for what you're looking for? I think if there's a valid scenario for "Not Set", then that wouldn't be a "placeholder", so that should be fine.

udam-f2 (approved):
This is helpful. I thought it might be better if we explicitly called rules for Nullable vs non-nullable cases but we can revisit that later too.

specification/attributes/null_handling.md
udam-f2: @rileyjenk @ijurica are you satisfied with the above explanation? If not, can you add a suggestion/comment here?

rileyjenk: @udam-f2  @ijurica  @flanakin is this now in conflict with the decisions made today? Namely we let the provider specify if they are going to allow nulls or empty string? 

That being said I like this.

PR #331, opened 2024-02-15
FOCUS feature level introduction - mandatory / conditional / optional

udam-f2:


specification/compliance.md
```diff
@@ -0,0 +1,19 @@
...
+
+Performing these capabilities will ultimately benefit organizations by enabling efficient, data-driven decision-making. The essential requirements needed to perform the important FinOps capabilities defined above are hereby referred to as the 'FOCUS Essential' requirements.
+
+## Meeting FOCUS Requirements
+
+FOCUS [Columns](#columns) designated as FOCUS Essential are required to be present in all compliant datasets and comply with the requirements (e.g. nullability and various other semantics) specified in the normative prose of such columns.
```
marc-perreaut, line 15: Non-essential columns should also comply with the requirements, if they are present, shouldn't they?

specification/columns/availabilityzone.md
```diff
@@ -1,8 +1,8 @@
 # Availability Zone
 
-[Availability Zone](#glossary:availability-zone) is a provider assigned identifier for a physically separated and isolated area within a Region that provides high availability and fault tolerance. Availability Zone is commonly used for scenarios like analyzing cross-zone data transfer usage and the corresponding cost based on where [*resources*](#glossary:resource) are deployed.
+[Availability Zone](#glossary:availability-zone) is a provider-assigned identifier for a physically separated and isolated area within a Region that provides high availability and fault tolerance. Availability Zone is commonly used for scenarios like analyzing cross-zone data transfer usage and the corresponding cost based on where [*resources*](#glossary:resource) are deployed.
 
-The AvailabilityZone column SHOULD be present in the billing data. This column MUST be of type String and MAY contain null values.
+The AvailabilityZone column MUST be present in the billing data when the provider supports deploying resources within an Availability Zone. This column MUST be of type String and MAY contain null values.
```
mbundgus, line 5: Should this clarify that availabilityzone values should be populated / not null, if the indicated service is deployed in an availability zone?

mbundgus: Is this a multi-value field?  e.g. can a service be charged in more than one availability zone?

specification/compliance.md
ahullah: oops i may have made my change as a commit rather than a comment @udam-f2 LMK if you want me to rollback??

specification/compliance.md
ahullah, line 1: oops i may have made my change as a commit rather than a comment @udam-f2  LMK if you want me to rollback??

specification/compliance.md
```diff
@@ -0,0 +1,19 @@
...
+
+FOCUS aims to help FinOps practitioners perform common FinOps capabilities by using a generic set of instructions. FOCUS accomplishes this by defining compliance requirements that account for the provider-level differences mentioned above.
+
+## Enabling FinOps Capabilities For Heterogeneous Cost Datasets
+
+Cloud Service Providers (CSPs) generate detailed cost and usage data in near real-time enabling an extensive set of FinOps capabilities to drive accountability and efficiency for dynamic consumption-based costs.  Non-CSP providers typically provide billing datasets that lack the same level of detail. The purpose of FOCUS Essential's reduced data set is to allow easier access for non-CSP providers to be incorporated in the full FOCUS data cube without breaking any primary FinOps capabilities, performing invoice reconciliation, data ingestion, and data normalization. Simplified versions of additional FinOps capabilities like cost reporting, chargeback, showback, and budgeting/forecasting can also be performed using a less-detailed billing dataset when essential requirements are met.
```
cnharris10, line 9:
```suggestion
Cloud Service Providers (CSPs) and some SaaS providers generate detailed cost and usage data in near real-time enabling an extensive set of FinOps capabilities to drive accountability and efficiency for dynamic consumption-based costs.  Non-CSP providers typically provide billing datasets that lack the same level of detail. The purpose of FOCUS Essential's reduced data set is to allow easier access for non-CSP providers to be incorporated in the full FOCUS data cube without breaking any primary FinOps capabilities, performing invoice reconciliation, data ingestion, and data normalization. Simplified versions of additional FinOps capabilities like cost reporting, chargeback, showback, and budgeting/forecasting can also be performed using a less-detailed billing dataset when essential requirements are met.
```

specification/compliance.md
```diff
@@ -0,0 +1,19 @@
...
+
+Performing these capabilities will ultimately benefit organizations by enabling efficient, data-driven decision-making. The essential requirements needed to perform the important FinOps capabilities defined above are hereby referred to as the 'FOCUS Essential' requirements.
+
+## Meeting FOCUS Requirements
+
+FOCUS [Columns](#columns) designated as FOCUS Essential are required to be present in all compliant datasets and comply with the requirements (e.g. nullability and various other semantics) specified in the normative prose of such columns.
+
+Non-FOCUS Essential columns defined in the specification may be omitted from the billing dataset based on the semantics defined in the normative prose of such columns. For example, columns related to commitment-based discounts can specify that they can be omitted when a provider doesn't support commitment-based discounts. Conversely, the providers that do support this discount construct are required to provide this column and meet all requirements defined in the normative prose of those columns.
```
cnharris10, line 17:
```suggestion
Non-FOCUS Essential columns defined in the specification may be omitted from the billing dataset based on the semantics defined in the normative prose of such columns. For example, columns related to commitment-based discounts can specify that they can be omitted when a provider doesn't support commitment-based discounts. Conversely, the providers that do support this discount construct are required to provide this column and meet all requirements defined in the normative prose of those columns.

However, a provider MUST NOT provide alternate cost and usage dimensions where FOCUS Essential and/or FOCUS Non-Essential columns CAN be applied.
```

specification/columns/availabilityzone.md
udam-f2: That's a great question.. Right now as it stands, it would just be a string value that the provider specifies. So they could find a way to convey the different AZ values. I think your question applies to both AZ and Region. Mind creating a new issue in GH about this clarification for AZ so at least we can discuss and figure out if there's a different way this needs to be handled? I'm guessing what we have now (value chosen by provider and not mandated by FOCUS) would work but I wonder if the language needs any further refinement.

specification/compliance.md
udam-f2: Reworded the whole thing.. captured your point about some SaaS.

specification/compliance.md
udam-f2: @cnharris10 is there a change here?  I'm not seeing a diff. Resolving as I've reworded the whole section. Can you review and add feedback.

specification/compliance.md
udam-f2: @ahullah I ended up rewording the whole section. I'm worried about talking about data cubes. Could you check the new updates and see what needs to be updated further?

rupagcp:
Not sure "Charge Frequency" is essential for Invoice Recon. It's useful but it's not core (especially if you look at all other fields in the list.

specification/columns/listcost.md
```diff
@@ -26,7 +26,7 @@ Cost calculated by multiplying List Unit Price and the corresponding Pricing Qua
 | Constraint      | Value                   |
 |:----------------|:------------------------|
 | Column type     | Metric                  |
-| Column required | True                    |
+| FOCUS Essential | True                    |
```
rileyjenk, line 29: Does this require that the vendor publish a list cost...though that may not actually be published? I see this as a stumbling block for many vendors to adopt even single line item focus data.

rileyjenk: Question answered in discussion today. @ijurica  perhaps we ensure it is clear that in the absence of a list price it matches BilledCost.

rileyjenk (approved):

ijurica:
It seems a bit odd to have information about the PricingUnit and PricingQuantity without the SkuId. If the PricingUnit and PricingQuantity are provided, it implies that the Provider is measuring and charging for something (a SKU). The existence of the item being measured and charged implies that the ID for that item exists and could be provided, right?

I suggest including the SKU ID into the list of Essential FOCUS columns and replacing current references to SKU Price ID with references to SKU ID wherever applicable.
This requires a modification to the SKU ID and SKU Price ID columns, where SKU ID becomes the primary column, acting as the unique identifier for the item being measured and charged, emphasizing the importance of the SKU itself, rather than relying solely on the association with a SKU Price. Subsequently, SKU Price IDs would relate to monetary values associated with a particular SKU.

If we choose to move forward with this adjustment, the SKU and SKU Price specs/definitions will require a bit more work, rephrasing, etc.

jpradocueva:
@ahullah, @marc-perreaut, @mbundgus, @ljadvey this PR has been moved to a `5 days Review & Approval`, meaning we need at least 3 members to review and approve the PR. Could you please review and provide/approve the PR? Thanks

ijurica:
An overview of _problematic_ columns cross-references, derived from the current column specifications and the list of essential columns, is accessible [here](https://docs.google.com/spreadsheets/d/1eWILtXdZo9Y1aNOXqmAV0PTGm5IGF9E9oXvJckP81dw/edit#gid=1316382674) - issues related to UCs when a column that may not be included is referenced in another column, even one that is part of the essential dataset. 

Some of these issues can probably be easily addressed, while others require more attention, such as the redefinition of SkuId and SkuPriceId.

ljadvey (approved):

marc-perreaut:
> It seems a bit odd to have information about the PricingUnit and PricingQuantity without the SkuId. If the PricingUnit and PricingQuantity are provided, it implies that the Provider is measuring and charging for something (a SKU). The existence of the item being measured and charged implies that the ID for that item exists and could be provided, right?
> 
> I suggest including the SKU ID into the list of Essential FOCUS columns and replacing current references to SKU Price ID with references to SKU ID wherever applicable. This requires a modification to the SKU ID and SKU Price ID columns, where SKU ID becomes the primary column, acting as the unique identifier for the item being measured and charged, emphasizing the importance of the SKU itself, rather than relying solely on the association with a SKU Price. Subsequently, SKU Price IDs would relate to monetary values associated with a particular SKU.
> 
> If we choose to move forward with this adjustment, the SKU and SKU Price specs/definitions will require a bit more work, rephrasing, etc.

Indeed, SkuPriceId is needed for pricing reconciliation, but do all providers have it?
However, I believe PricingUnit and PricingQuantity are already good enough to enable FinOps activities like cost optimization.

marc-perreaut:
I would have liked to include the Region in FOCUS essentials, because the region is essential in different use cases where one needs to know "what is running where" (multi-region deployments, data residency, ...), but it might be a barrier for some providers, so it's fine.

marc-perreaut (approved):
Everything is fine and in line with the discussions, great work!

marc-perreaut:
> An overview of _problematic_ columns cross-references, derived from the current column specifications and the list of essential columns, is accessible [here](https://docs.google.com/spreadsheets/d/1eWILtXdZo9Y1aNOXqmAV0PTGm5IGF9E9oXvJckP81dw/edit#gid=1316382674) - issues related to UCs when a column that may not be included is referenced in another column, even one that is part of the essential dataset.
> 
> Some of these issues can probably be easily addressed, while others require more attention, such as the redefinition of SkuId and SkuPriceId.

Let's take the example of the PricingUnit (part of FOCUS essential), which refers to SkuPriceId (not part of FOCUS essential). The specification days:

> The value [of PricingUnit] MUST NOT be null if SkuPriceId is not null and MUST be null if SkuPriceId is null.

To me, this means semantically that PricingUnit and SkuPriceId relate to each other, and this implies they must be consistent together (this might be explicitly specified). But I believe there is no conceptual issue if the column SkuPriceId is absent, because it just means the SkuPriceId information is not available, while the PricingUnit information is available.

ahullah:
added quick fix for failing buildpdf

specification/columns/usageunit.md
```diff
@@ -2,13 +2,13 @@
...
 
-The UsageUnit column MUST be present in the billing data. This column MUST be of type String. The value MUST NOT be null if [SkuPriceId](#skupriceid) is not null and MUST be null if SkuPriceId is null. Units of measure used in UsageUnit SHOULD adhere to the values and format requirements specified in the [UnitFormat](#unitformat) attribute. The UsageUnit column MUST NOT be used to determine values related to any pricing or cost metrics.
+The UsageUnit column MUST be present in the billing data when the provider supports the measurement of usage. This column MUST be of type String. UsageUnit MUST NOT be null if [ChargeCategory](#chargecategory) is "Usage" or "Purchase" and MAY be null for other ChargeCategory values. Units of measure used in UsageUnit SHOULD adhere to the values and format requirements specified in the [UnitFormat](#unitformat) attribute. The UsageUnit column MUST NOT be used to determine values related to any pricing or cost metrics.
```
flanakin, line 5: This feels a bit abstract. Can we bind it to UsageQuantity being present?
```suggestion
The UsageUnit column MUST be present in the billing data when [UsageQuantity](#usagequantity) is present. This column MUST be of type String. UsageUnit MUST NOT be null if [ChargeCategory](#chargecategory) is "Usage" or "Purchase" and MAY be null for other ChargeCategory values. Units of measure used in UsageUnit SHOULD adhere to the values and format requirements specified in the [UnitFormat](#unitformat) attribute. The UsageUnit column MUST NOT be used to determine values related to any pricing or cost metrics.
```

specification/columns/usagequantity.md
```diff
@@ -1,27 +1,27 @@
...
 
-UsageQuantity MUST be present in the billing data. This column MUST be of type Decimal and MUST conform to [Numeric Format](#numericformat). The value MAY be negative in cases where [ChargeSubcategory](#chargesubcategory) is "Refund". This column MUST NOT be null if [SkuPriceId](#skupriceid) is not null and MUST be null if SkuPriceId is null.
+UsageQuantity column MUST be present in the billing data when the provider supports the measurement of usage. This column MUST be of type Decimal and MUST conform to [Numeric Format](#numericformat) requirements. The value MAY be negative in cases where [ChargeSubcategory](#chargesubcategory) is "Refund". This column MUST NOT be null if [ChargeCategory](#chargecategory) is "Usage" or "Purchase" and MAY be null for other ChargeCategory values.
```
flanakin, line 5: This feels a bit abstract. Can we bind it to ChargeCategory = Usage or when there's a difference between pricing and usage?
```suggestion
UsageQuantity column MUST be present in the billing data when the provider supports usage-based charges where [ChargeCategory](#chargecategory) is "Usage" or when [PricingQuantity](#pricingquantity) does is based on [block pricing](#glossary:block-pricing) or measured in a different unit. This column MUST be of type Decimal and MUST conform to [Numeric Format](#numericformat) requirements. The value MAY be negative in cases where [ChargeSubcategory](#chargesubcategory) is "Refund". This column MUST NOT be null if [ChargeCategory](#chargecategory) is "Usage" or "Purchase" and MAY be null for other ChargeCategory values.
```

specification/columns/skuid.md
```diff
@@ -1,27 +1,27 @@
...
 
-The SkuId column MUST be present in the billing data. This column MUST be of type String. The SkuId MUST NOT be null when [SkuPriceId](#skupriceid) is not null. SkuId MUST equal SkuPriceId when a provider does not support an overarching SKU ID construct.
+The SkuId column MUST be present in the billing data when the provider publishes a SKU list. This column MUST be of type String. The SkuId MUST NOT be null when [SkuPriceId](#skupriceid) is not null. SkuId MUST equal SkuPriceId when a provider does not support an overarching SKU ID construct.
```
flanakin, line 5: Shouldn't it be required when a provider offers multiple SKUs? There are many companies who don't "publish" a list of SKUs as an API. But that shouldn't negate the importance of identifying which of their multiple SKUs they're being charged for, right?
```suggestion
The SkuId column MUST be present in the billing data when the provider supports multiple SKUs. This column MUST be of type String. The SkuId MUST NOT be null when [SkuPriceId](#skupriceid) is not null. SkuId MUST equal SkuPriceId when a provider does not support an overarching SKU ID construct.
```

specification/columns/skupriceid.md
```diff
@@ -2,29 +2,29 @@
...
+The SkuPriceId column MUST be present in the billing data when the provider publishes a SKU price list. This column MUST be of type String. SkuPriceId MUST define a single unit price used for calculating the charge. The [ListUnitPrice](#listunitprice) MUST be associated with the SkuPriceId in the provider-published *price list*. The SkuPriceId MUST NOT be null when [ChargeCategory](#chargecategory) is "Purchase" or "Usage". SkuPriceId MUST NOT be null when [ChargeSubcategory](#chargesubcategory) is "Refund" and the refund is related to charges with a specific SkuPriceId. A given value of SkuPriceId MUST be associated with one and only one [SkuId](#skuid), except in cases of commitment discount flexibility.
```
flanakin, line 5: Similar to SkuId, shouldn't this be required whenever a SKU can support multiple prices?
```suggestion
The SkuPriceId column MUST be present in the billing data when the provider has any [SkuId](#skuid) that can have multiple prices for different variations of the SKU. This column MUST be of type String. SkuPriceId MUST define a single unit price used for calculating the charge. The [ListUnitPrice](#listunitprice) MUST be associated with the SkuPriceId in the provider-published *price list*. The SkuPriceId MUST NOT be null when [ChargeCategory](#chargecategory) is "Purchase" or "Usage". SkuPriceId MUST NOT be null when [ChargeSubcategory](#chargesubcategory) is "Refund" and the refund is related to charges with a specific SkuPriceId. A given value of SkuPriceId MUST be associated with one and only one [SkuId](#skuid), except in cases of commitment discount flexibility.
```

specification/columns/region.md
flanakin, line 1: I'm guessing this was a bad merge. You probably want to delete this file since it was renamed to RegionId, right?

specification/columns/resourcetype.md
```diff
@@ -1,8 +1,8 @@
...
+Resource Type describes the kind of [*resource*](#glossary:resource) the charge applies to. A Resource Type is commonly used for scenarios like identifying cost changes in groups of similar *resources* and may include values like Virtual Machine, Data Warehouse, and Load Balancer.
 
-The ResourceType column MUST be present within billing data.  This column MUST be of type String and MUST NOT be null when a corresponding [ResourceId](#resourceid) is not null.  When a corresponding ResourceId value is null, the ResourceType column value MUST also be null.  Providers MUST use a consistent value-format and a set of values for ResourceType values within their billing datasets.
+The ResourceType column MUST be present in the billing data when the provider supports billing based on provisioned resources and supports the categorization of resources. This column MUST be of type String and MUST NOT be null when a corresponding [ResourceId](#resourceid) is not null. When a corresponding ResourceId value is null, the ResourceType column value MUST also be null. Providers MUST use a consistent value format and a set of values for ResourceType values within their billing datasets.
```
flanakin, line 5: nit: "Categorization" sounds like something different than the description of the type of resource.
```suggestion
The ResourceType column MUST be present in the billing data when the provider supports billing based on provisioned resources and supports more than one type of resource. This column MUST be of type String and MUST NOT be null when a corresponding [ResourceId](#resourceid) is not null. When a corresponding ResourceId value is null, the ResourceType column value MUST also be null. Providers MUST use a consistent value format and a set of values for ResourceType values within their billing datasets.
```

specification/columns/usagequantity.md
marc-perreaut: From my practitioner perspective, if a provider omits the UsageQuantity column because it's the same as PricingQuantity, I would not be able to make the distinction with the case where the provider does not support the measurement of usage. So, in the case the UsageQuantity is the same as PricingQuantity, I would prefer UsageQuantity is present and populated with the same as PricingQuantity.

Moreover, the presence of the column should not be related to the value of the ChargeCategory, because this value may differ between records.

specification/columns/skuid.md
marc-perreaut: If a provider supports multiple SKUs, the provider would highly probably publish a SKU list, wouldn't he?
If a provider has a unique SKU, and if the provider publishes it, I find it good to see the provider unique SKU in FOCUS data, because the provider may plan to have more SKUs later, which would then appear in FOCUS data.

jpradocueva:
The Maintainers, indicate that is ready for approval after some minor reviews.

jpradocueva:
This PR will be put forward for approval on Thursday, March 21.

specification/columns/listcost.md
udam-f2: We're not specifying that it needs to be 'public' anywhere. Many providers don't publish rates publicly but they will privately tell you the starting point for price/cost (list price/cost) and then go from there to a negotiated price/cost for you.

udam-f2: "The List Unit Price represents the suggested provider-published unit price for a single Pricing Unit of the
associated SKU, exclusive of any discounts."

specification/columns/availabilityzone.md
```diff
@@ -1,27 +1,27 @@
...
 
-[Availability Zone](#glossary:availability-zone) is a provider assigned identifier for a physically separated and isolated area within a Region that provides high availability and fault tolerance. Availability Zone is commonly used for scenarios like analyzing cross-zone data transfer usage and the corresponding cost based on where [*resources*](#glossary:resource) are deployed.
+An [*availability zone*](#glossary:availability-zone) is a provider-assigned identifier for a physically separated and isolated area within a Region that provides high availability and fault tolerance. Availability Zone is commonly used for scenarios like analyzing cross-zone data transfer usage and the corresponding cost based on where [*resources*](#glossary:resource) are deployed.
 
-The AvailabilityZone column SHOULD be present in the billing data. This column MUST be of type String and MAY contain null values.
+The AvailabilityZone column MUST be present in the billing data when the provider supports deploying resources or services within an *availability zone*. This column MUST be of type String and MAY contain null values when a charge is not specific to an *availability zone*.
```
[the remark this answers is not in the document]
udam-f2: In that case, we have to describe this scenario of when to include and when to exclude in every column which seems unnecessary to me. I don't see an issue if a provider provides a nullable FOCUS column with nulls when it doesn't apply to them. For that reason, I'm ok without but if others feel it's needed, we can go an update all columns to specify the exclusion case. Thoughts?

specification/columns/availabilityzone.md
```diff
@@ -1,27 +1,27 @@
...
-A provider assigned identifier for a physically separated and isolated area within a Region that provides high availability and fault tolerance.
+A provider-assigned identifier for a physically separated and isolated area within a Region that provides high availability and fault tolerance.
 
 ## Content constraints
 
 | Constraint      | Value            |
 |:----------------|:-----------------|
 | Column type     | Dimension        |
-| Column required | False            |
+| FOCUS Essential | False            |
```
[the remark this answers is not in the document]
udam-f2: We introduce 'FOCUS Essential' in section 2 of the spec now (part of the PR). The hope is it becomes a easy marketing term that we can refer to for meeting a basic requirement set for FOCUS (without making a statement about anything beyond the essential columns - e.g. you can meet essential requirements and still fail FOCUS compliance). But yes, it will require some reading/thinking to know these are the must-have columns. If there is a different way to show this, I'm open to it.

For the options presented above: my challenge with anything that has 'required' is that in FOCUS Essential = false columns, the column can still be required IF the provider meets the criteria. E.g. Commitment commitment-based discount columns (e.g. CommitmentDiscountId) are not essential because many providers don't support commitment-based discounts. However, for providers that do support commitment-based discounts, it is required to be provided (see normative prose of the changes in this PR). So I worry about the term 'required' being used. Compliance level Essential can equal Full for providers that don't need to specify any columns outside of the essential. So that has some challenges too.

Once FinOps Foundation training and framework start referencing FOCUS, I think it won't be much of a stretch to think people in FinOps won't understand what 'FOCUS Essential' is without needing to reference section 2 of the spec.

Thoughts?

specification/columns/billingcurrency.md
```diff
@@ -1,27 +1,27 @@
...
 
-## Display name
+## Display Name
 
 Billing Currency
 
 ## Description
 
-Represents the currency that a charge was billed in.
+Represents the currency that a charge was billed.
```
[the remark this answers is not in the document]
udam-f2: Grammarly disagrees. I'm open to both and am not an expert. Who's right? Can others with expertise chime in please?

Human vs AI ;)

specification/columns/billingcurrency.md
```diff
@@ -1,27 +1,27 @@
 # Billing Currency
 
-[Billing Currency](#glossary:billing-currency) is an identifier that represents the currency that a charge for [*resources*](#glossary:resource) or [*services*](#glossary:service) was billed in. Billing Currency is commonly used in scenarios where costs need to be grouped or aggregated.
+[*Billing currency*](#glossary:billing-currency) is an identifier that represents the currency that a charge for [*resources*](#glossary:resource) or [*services*](#glossary:service) was billed. Billing Currency is commonly used in scenarios where costs need to be grouped or aggregated.
```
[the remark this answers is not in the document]
udam-f2: Same reason as above. Who wins? Human or AI/Grammarly :D ?

specification/columns/usagequantity.md
udam-f2: > This feels a bit abstract. Can we bind it to ChargeCategory = Usage or when there's a difference between pricing and usage?

Agree that it needs to not be abstract - and open to make it clearer. 

However, as described, anyone with block pricing regardless of if the provider bills based on 'usage' would have to include this column. The usage* columns were recently broadened to include usage and purchase instead of just usage (which I'm personally not a fan of). In that world, how do we write it in a way that it's not a barrier for entry (and a confusion for customers) when usage* columns are included with values? Might reach out to you both in slack to see if we can drive consensus on this.

specification/columns/skuid.md
udam-f2: I'm in the camp of if a provider has multiple SKUs and SKUPriceIds, AND customers need to understand those details, they will have to present a SKU / SKUPrice list to their customers (regardless of the billing data). If a provider doesn't give those details, there's an assumption that it's because their services don't need that level of detail in billing data (or things like ChargeDescription are sufficient). At that point, requiring this column means a provider now is putting random IDs that can't be referenced anywhere and therefore have no meaning.

Maybe the issue here is the word 'publish' which might hint it's public. Would something like 'makes a SKU list available' or 'presents a SKU list' be more agreeable?

specification/columns/resourcetype.md
udam-f2: Agree on the word categorization can be confused with our 'Category' columns. I struggled with these exclusion clauses. 

My question is how tightly do we tighten the screws here to make it required. We don't have a way to define what the scope of a 'type' is (what if a provider says storage volumes and virtual machines are both of resource type 'cloud resource'?).. do we communicate that it's almost up to the provider to decide unless they explicitly have a typing construct/mechanism in place? Maybe I'm overcomplicating this. Need a bit more convincing.. :)

specification/columns/availabilityzone.md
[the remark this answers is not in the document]
marc-perreaut: I find FOCUS essential is nice to convey the message that this subset enables essential FinOps activities, and that's the minimal subset a provider must support to claim to support FOCUS.

Furthermore, if a provider supports a non-essential column, is there a known or foreseen case where the provider would make the column absent in the FOCUS dataset? And if yes, would it be an issue? If yes, would stating the dataset is not FOCUS compliant help to solve this issue?

specification/columns/availabilityzone.md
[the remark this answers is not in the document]
marc-perreaut: From my practitioner perspective, I would find unclear that a column is present but always populated with null values, because the provider does not support it. Indeed, I could not differentiate the case where a value is null because the provider does not support the column and the case where the provider supports the column but it's the case where the value is null. If a column is present, I would expect the provider supports it and provides non-null values in the relevant cases.

Meanwhile, specifying the exclusion cases for every column might be heavy. Maybe this could be specified in the nullable attribute? If a column is nullable, null values must not be used if the provider does not support the column: the column must be made absent.

specification/columns/skuid.md
marc-perreaut: With the introduction of the concept of FOCUS essentials and the discussion in https://github.com/FinOps-Open-Cost-and-Usage-Spec/FOCUS_Spec/pull/331#discussion_r1531676285 and https://github.com/FinOps-Open-Cost-and-Usage-Spec/FOCUS_Spec/pull/331#discussion_r1531645744, I am starting to think that the sentences "The XXX column MUST be present in the billing data when the provider supports ..." are not really useful:
- FOCUS Essential columns must be present in any FOCUS dataset. We have selected FOCUS essential columns so that none should be a barrier for a provider.
- Any other column may be present or not in a FOCUS dataset. If present, the values must comply with the specification, in particular the cases where the value must not be null. I don't see and I wonder whether there is a case where a provider would support a column but would make it absent.

specification/columns/listcost.md
flanakin: I interpreted "provider-published" to mean that I must publicly "publish" prices. If we remove that, the sentence still conveys the same point. Maybe we just simplify?

> _The List Unit Price represents the suggested unit price for a single Pricing Unit of the
associated SKU, exclusive of any discounts._

specification/columns/availabilityzone.md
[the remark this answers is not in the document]
flanakin: > my challenge with anything that has 'required' is that in FOCUS Essential = false columns, the column can still be required IF the provider meets the criteria.

In this case, "Required: When supported" would be more explicit than "FOCUS Essential: False".

The idea of "Compliance level: Essential" was trying to keep the marketing aspect of "Essential" while accepting that we'll likely need multiple compliance levels. That makes "Essential" a lot more acceptable as added complexity since it's one of the several levels of compliance.

That said, "Essential" is not a compliance level based on this description. A compliance level is something you can support without doing anything more. So regardless of whether I support commitment discounts or not, I could declare Essential compliance without those columns. However, if we say you MUST add those columns, then that means it's never possible to be Essential-compliant without being "fully-compliant" (accepting that full compliance allows for not including unsupported columns).

Unless I'm missing something, this minimizes the value of "Essential" as a marketing term since it's not a statement of compliance. It just conveys to practitioners which columns will be in every FOCUS dataset but it won't be useful when determining what a provider supports. From that perspective, everyone who supports FOCUS would be "Essential".

It might be useful to look at how others use compliance levels. WCAG (accessibility) is my go-to. Sites can either be A, AA, or AAA based on meeting specific guidelines. This is easy for website owners to abide by and absolutely critical for end users who might rely on something that's only supported in AAA. For them, they know that they cannot use a product unless it's AAA. That's immensely powerful. If we follow this model, we might have compliance levels like:
1. `Basic` = "Essential" (or "always" required) columns only, regardless of what's supported. (Alt: Core, Essential)
2. `Standard` = Core + all columns that map to features supported by the provider (e.g., AZ, CBD). (Alt: Feaure)
3. `Extended` = Every column, regardless of what's supported. (Alt: Full, Complete)
4. `Complete` = Every column, regardless of what's supported, with all MUST and SHOULD rules applied. (Alt: Comprehensive)

🤔

specification/columns/usagequantity.md
flanakin: > as described, anyone with block pricing regardless of if the provider bills based on 'usage' would have to include this column

That was my goal 🙃 When there's a difference between usage and pricing, you must have both. That feels like the same as including CommitmentDiscount* columns when commitment discounts are supported.

Fwiw, I agree with Marc that it's best to always have UsageQuantity since that's necessary to do usage-based analysis without block pricing across providers. But I'm happy to defer to practitioners.

specification/columns/skuid.md
flanakin: Microsoft supports availability zones but doesn't have that in the cost data, so we might not include the column.

flanakin: "Publishes" and "makes available" both feel very ambiguous. I would imagine practitioners would want this to identify similar rows, regardless of whether they have the IDs or not. They can be learned based on this data. Having a separate list makes them even more useful, but there's value regardless.

specification/columns/resourcetype.md
flanakin: That's a good point... Someone could put "Resource" for everything 😒 Or someone could put "Windows Server 2024 Virtual Machine with Premium SSD" 😖 Given this is a "Type", I would be fine leaving it up to the provider and have practitioners push them for the right level of granularity, but either way "more than one type of resource" is still applicable. The question is more about whether we should be explicit about how granular "type" must be.

specification/columns/availabilityzone.md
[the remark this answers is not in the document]
udam-f2: Totally aligned on the dataset-level presence of a column vs row-level nullability of the column. 

At the end of the day, once the datasets are combined as it's the promise of FOCUS, you're always looking at the union of columns. So while in a single provider-level dataset you _might be_ able to quickly guess if a provider might be supporting a concept/column or not, you can't easily do that once you're looking at the higher-level integrated dataset. Many providers are also now supporting creating custom projections so I'd argue the presence of a column can't always be counted on because you might only export a subset.

If you all are still aligned, I'm happy to add or you can suggest an update in the nullability attribute OR possibly in section 2 related to compliance.

specification/columns/usagequantity.md
udam-f2: Not sure I understand. Many providers don't support usage-based billing but might have block pricing (and bill up-front for what you bought). They would now have to provide usage* columns which would provide completely inaccurate information. This is the reason we moved this to non-essential based on the discussion in the task forces.

specification/columns/skuid.md
udam-f2: > Microsoft supports availability zones but doesn't have that in the cost data, so we might not include the column.

They could.. that would just not meet FOCUS compatibility.

udam-f2: > * Any other column may be present or not in a FOCUS dataset. If present, the values must comply with the specification, in particular the cases where the value must not be null. I don't see and I wonder whether there is a case where a provider would support a column but would make it absent.

To be clear, this isn't meant to say they have the column but are hiding it. It's completely possible that a provider supports resources but doesn't provide resource-level billing. One major CSP was in this bucket till recently. The sentence is supposed to convey (maybe needs refining?) that when the provider's general capabilities support capability X (which is a concept in FOCUS), those relevant FOCUS columns must be made available.

specification/columns/resourcetype.md
udam-f2: but that means someone with just one category would possibly not provide or provide null? which isn't as helpful when aggregating across datasets. Don't want to hold up based on this. Are we ok with that semantic?

Can we go with something less specific:
".. when the provider supports billing based on provisioned resources and supports typing constructs for resources"
or this:
".. when the provider supports billing based on provisioned resources"

specification/compliance.md
```diff
@@ -0,0 +1,23 @@
...
+
+Cloud Service Providers (CSPs) and some SaaS providers generate detailed cost and usage data in near real-time, which enables a wide range of FinOps capabilities. These capabilities help organizations make data-driven decisions to maximize the value they extract from the providers.
+
+Most other providers typically offer billing datasets that lack the same level of detail. However, those less-detailed datasets meeting an essential set of requirements can enable some important FinOps capabilities like invoice reconciliation, data ingestion, and data normalization. Simplified versions of additional FinOps capabilities like cost reporting, chargeback, showback, and budgeting/forecasting can also be performed using less-detailed billing datasets when these essential requirements are met.
+
+The essential set of data requirements needed to perform the most important FinOps capabilities is referred to as the 'FOCUS Essential' requirements.
+
+## Meeting FOCUS Requirements
+
+*This section is normative.*
```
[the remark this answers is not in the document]
udam-f2: This is a pattern we're taking from many other specs that follow the BCP 14 language - it clearly tells people that when you see MUST/SHOULD etc. that they have a special meaning not just formatting is off.

specification/columns/availabilityzone.md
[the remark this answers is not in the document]
udam-f2: FOCUS Essential is not a compliance level. It's a way to indicate that a minimum set of important columns are available in the dataset. 

Compliance levels should be handled post 1.0. Many good patterns as you've all called out. To me, it's orthogonal to this discussion. Today, you can have essential columns and columns that are required for subsets of providers based on their capabilities.

specification/columns/listcost.md
udam-f2: > The List Unit Price represents the Provider suggested unit price for a single Pricing Unit of the
associated SKU, exclusive of any discounts.

How about this?

specification/columns/skupriceid.md
ijurica: As I recall, SkuId relies on SkuPriceId and if there are no multiple prices SkuId matches the skuPriceId. 
 Both are present or not, depending on the SKU Price list being published.

One more question - ListUnitPrice is not necessarily published (one of the latest conversations), but the SkuPriceId  is still present if SKU Price list is published - we are not changing that, right?

specification/columns/availabilityzone.md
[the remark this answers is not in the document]
flanakin: I had that thought as well... Multi-cloud or multi-dataset is still a smaller subset of overall practitioners who will use FOCUS. I see the overarching value of FOCUS being clearer expectations for practitioners whether they use one or more datasets. With that context, I would lean towards optimizing for the single dataset scenario, noting that it's irrelevant when merged with other datasets that have different levels of support.

But again, I'm looking for others to chime in. This is probably a good, quick survey we can knock out in one of the meetings.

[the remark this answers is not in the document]
marc-perreaut: Indeed, the merge of FOCUS datasets raises questions.

The specification says:

> The value [of PricingUnit] MUST NOT be null if SkuPriceId is not null and MUST be null if SkuPriceId is null.

After the merge, if absent columns are handled with null values, this statement is not true for the charges (rows) of providers that do not support SkuPriceId. Another option for the merge is for a practitioner to put special values for missing columns, for example "<not supported>". Maybe there are other options.

It would be interesting to get others' view in a meeting.

specification/columns/listcost.md
ahullah: Created discussion topic #366 to continue identification of a resolution for this

specification/columns/region.md
ahullah: oops ... sorry folks

specification/columns/availabilityzone.md
flanakin, line 5: I'm not okay with changing AvailabilityZone from SHOULD to MUST. Can we revert it back?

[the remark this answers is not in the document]
udam-f2: The group decided that we make the normative prose about the presence of the data in the charge record. In this case, the group decided that it should remain a SHOULD to prevent providers from going out of compliance for not providing AZ. Column presence is defined in the schema section.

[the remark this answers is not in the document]
ahullah: Follow up discussion topic raised as issue #367

ahullah: follow up discussion raised in issue #367

specification/compliance.md
```diff
@@ -0,0 +1,23 @@
+# FOCUS Compliance
+
+The FOCUS specification is designed to assist FinOps practitioners in performing common FinOps capabilities by utilizing a standardized set of instructions, regardless of the billing data generator. It does so by establishing compliance criteria that account for provider-level variances in the cost and usage datasets providers produce. This section outlines the two primary types of cost datasets that FinOps practitioners currently work with and explains how FOCUS can enhance the interoperability between these different datasets.
```
flanakin, line 3: > This section outlines the two primary types of cost datasets that FinOps practitioners currently work with and explains how FOCUS can enhance the interoperability between these different datasets.

What "two types of cost datasets" is this referring to? This isn't clear.

specification/columns/listcost.md
```diff
@@ -1,10 +1,10 @@
...
+List Cost represents the cost calculated by multiplying the [*list unit price*](#glossary:list-unit-price) and the corresponding [Pricing Quantity](#pricingquantity). List Cost is denominated in the [Billing Currency](#billingcurrency) and is commonly used for calculating savings based on various rate optimization activities, by comparing it with [Billed Cost](#billedcost) and [Effective Cost](#effectivecost).
 
-The ListCost column MUST be present in the billing data and MUST NOT be null. This column MUST be of type Decimal, MUST conform to [Numeric Format](#numericformat), and be denominated in the BillingCurrency. When a ListUnitPrice is not null, multiplying the ListUnitPrice by PricingQuantity MUST produce the ListCost.
+The ListCost column MUST be present in the billing data and MUST NOT be null. This column MUST be of type Decimal, MUST conform to [Numeric Format](#numericformat) requirements, and be denominated in the BillingCurrency. Multiplying the *list unit price* by PricingQuantity MUST produce the ListCost.
```
ijurica, line 5: Do we refer to the glossary instead of directly mentioning the actual columns in the normative paragraph, only in the List Cost and ListUnitPrice specs? Taking into account that it is normative, I would prefer the reference to the actual column.
Consider something like this:

```suggestion
The ListCost column MUST be present in the billing data and MUST NOT be null. This column MUST be of type Decimal, MUST conform to [Numeric Format](#numericformat) requirements, and be denominated in the BillingCurrency. When [ListUnitPrice](#listunitprice) is present and is not null, multiplying the ListUnitPrice by PricingQuantity MUST produce the ListCost.
```

specification/columns/listunitprice.md
```diff
@@ -2,13 +2,13 @@
...
 
-The ListUnitPrice column MUST be present in the billing data. This column MUST be a Decimal within the range of non-negative decimal values, MUST conform to [Numeric Format](#numericformat), and be denominated in the BillingCurrency. ListUnitPrice MUST NOT be null if [SkuPriceId](#skupriceid) is not null and MUST be null if SkuPriceId is null. When ListUnitPrice is not null, multiplying ListUnitPrice by [PricingQuantity](#pricingquantity) MUST equal [ListCost](#listcost).
+The ListUnitPrice column MUST be present in the billing data when the provider publishes unit prices exclusive of discounts. This column MUST be a Decimal within the range of non-negative decimal values, MUST conform to [Numeric Format](#numericformat) requirements, and be denominated in the BillingCurrency. When ListUnitPrice is present, it MUST NOT be null if [ChargeCategory](#chargecategory) is "Usage" or "Purchase" and MAY be null for other ChargeCategory values. When ListUnitPrice is not null, multiplying ListUnitPrice by [PricingQuantity](#pricingquantity) MUST equal [ListCost](#listcost).
```
ijurica, line 5:
```suggestion
The ListUnitPrice column MUST be present in the billing data when the provider publishes unit prices exclusive of discounts. This column MUST be a Decimal within the range of non-negative decimal values, MUST conform to [Numeric Format](#numericformat) requirements, and be denominated in the BillingCurrency. When ListUnitPrice is present, it MUST NOT be null if [ChargeCategory](#chargecategory) is "Usage" or "Purchase" and MAY be null for other ChargeCategory values. When ListUnitPrice is present and not null, multiplying ListUnitPrice by [PricingQuantity](#pricingquantity) MUST equal [ListCost](#listcost).
```

Note: not a showstopper

ijurica (approved):

jpradocueva:
@udam-f2, can you please review the last few comments on this PR?

PR #372 has been merged into this PR, describing columns as "Feature Levels" mandatory, optional, and conditional. We will ask the group for Review & Approval as soon as you resolve the few pending comments. Once merged, I will start preparing PRs to address the conditional requirements.

specification/columns/listcost.md
udam-f2: I was debating this. I'm good with the suggestion. 

(previously, I was debating if we should refer to the glossary entry to remove the complication of list unit price being conditional)

specification/columns/usagequantity.md
udam-f2: If there's an issue to discuss here, lets create a different GH issue and work it. As is, its just replacing the reference to a non-mandatory column. Some of these will get updated with the ChargeClass column nuance anyway.

specification/columns/resourcetype.md
udam-f2: Updated to:
The ResourceType column SHOULD be present in the billing data when the provider supports billing based on provisioned resources and supports assigning a type for resources.

specification/compliance.md
udam-f2: Removed this section

specification/columns/listunitprice.md
udam-f2: Not letting me accept here (says out of date). I've manually made the update.

specification/columns/skupriceid.md
udam-f2: Same reasoning as the other instance. Keeping it at if the provider publishes gives some confidence that if it's relevant, the provider will provide this which then requires it to be available in the billing data - regardless of if there's one or more.

specification/columns/usageunit.md
udam-f2: It's the same clause in UsageQuantity and UsageUnit. Otherwise, we unnecessarily create a link where someone has to go look up when the related column is present (which is the same reason - in this case).

specification/columns/listcost.md
```diff
@@ -1,10 +1,10 @@
...
 
-The ListCost column MUST be present in the billing data and MUST NOT be null. This column MUST be of type Decimal, MUST conform to [Numeric Format](#numericformat), and be denominated in the BillingCurrency. When a ListUnitPrice is not null, multiplying the ListUnitPrice by PricingQuantity MUST produce the ListCost.
+The ListCost column MUST be present in the billing data and MUST NOT be null. This column MUST be of type Decimal, MUST conform to [Numeric Format](#numericformat) requirements, and be denominated in the BillingCurrency. When [ListUnitPrice](#listunitprice) is present and not null, multiplying the ListUnitPrice by PricingQuantity MUST produce the ListCost.
 
-In cases where the ListUnitPrice is null, the following applies:
+In cases where the ListUnitPrice is present and is null, the following applies:
 
 * The ListCost MUST be calculated based on the ListCost of the related charges if the charge is calculated based on other charges (e.g. [ChargeCategory](#chargecategory) is "Tax").
```
ijurica, line 9:
```suggestion
* The ListCost of a charge calculated based on other charges (e.g., when the [ChargeCategory](#chargecategory) is "Tax") MUST be calculated based on the ListCost of those related charges.
```

specification/columns/listcost.md
```diff
@@ -1,10 +1,10 @@
...
 * The ListCost MUST match the [BilledCost](#billedcost) if the charge is unrelated to other charges (e.g. [ChargeSubcategory](#chargesubcategory) is "Credit").
```
ijurica, line 10:
```suggestion
* The ListCost of a charge unrelated to other charges (e.g., when the [ChargeSubcategory](#chargesubcategory) is "Credit") MUST match the [BilledCost](#billedcost).
```

jpradocueva:
Approved by FOCUS Members on April 4th.

mike-finopsorg (approved):

specification/columns/availabilityzone.md
```diff
@@ -1,27 +1,27 @@
...
-A provider assigned identifier for a physically separated and isolated area within a Region that provides high availability and fault tolerance.
+A provider-assigned identifier for a physically separated and isolated area within a Region that provides high availability and fault tolerance.
 
 ## Content constraints
 
 | Constraint      | Value            |
 |:----------------|:-----------------|
 | Column type     | Dimension        |
-| Column required | False            |
+| Feature level   | Conditional      |
```
flanakin, line 24: "Feature level" is not clear. Can we find a better description?

Issue #342, opened 2024-02-28
[Content Suggestion] Tag key appearance


AWS-ZachErdman:
### Description
Should the list of tag keys included in the JSON object be the same list across all line items in a FOCUS export? Even when not all line items have a value for every tag? 
It’s not quite clear, but this particular requirement in "2.37 Tags" 

> "A Tag key without a specified value MUST have its tag value set to null."

seems to indicate that all appearing tag keys should be present in every line item, even when a line item lacks a value for a tag.

### Proposed approach
We should remove this requirement. There isn't a good reason to force a tag key to appear when it is null.

We should also consider adding a requirement

>"A Tag key without a specified value should not appear in the Tag column JSON object for a given line item".

### Github issue or Reference
N/A

### Context
N/A

udam-f2:
That was not the intention. I believe the intention was to cover the scenario where providers only allow a 'label' to be specified but not a key-value pair. 

@cnharris10 could you work with @AWS-ZachErdman to get clarification for this out as a PR, please.

cnharris10:
Yes, the intention here is to effectively capture tags that do not contain a value. In the case where a tag is this single entity, these tags should not be dropped, so guidance is to default their value to null. 

GCP is the prominent example where network tags do not contain a value:
https://cloud.google.com/vpc/docs/add-remove-network-tags

In the case where a tag is a key/value pair, tags keys that are null should be disregarded.

AWS-ZachErdman:
Hi @cnharris10, your comment cleared it up a lot for me, thank you!

I think adding/modifying something like the two lines you said would help a lot.

> In the case where a tag is a single entity and has been assigned to a resource, the tag key should appear in the tags column with the value set to null.
> In the case where a tag is a key/value pair, tag keys for which the value is null SHOULD NOT be included in the tags columns.

The only part I'd want to confirm, is that no cloud provider has key-value tags with "meaningful" null values for a given key where they should be included in the tags column.

cnharris10:
Sounds good, I'll propose an amendment with this and tag you. Thanks for contributing.

flanakin:
Why do we need to declare that tags with a null value should not be in the data? I don't agree with that. I like the first sentence clarifying that label-only tags should have a null value, but the second doesn't seem to be needed since a key-value tag system may also support null values. Microsoft does, for instance. I'm assuming others do as well.

AWS-ZachErdman:
[...]

PR #349, opened 2024-03-01
Related to Issue #342 - Tag key differentation when a supporting value can or cannot exist.

cnharris10:
Related to Issue: https://github.com/FinOps-Open-Cost-and-Usage-Spec/FOCUS_Spec/issues/342

Further clarifying that:
1. Tag keys that cannot support values are given a `true` (boolean) value
--> (ex: [GCP single-entity tags](https://cloud.google.com/vpc/docs/add-remove-network-tags)) `vpc` becomes `"vpc":true` and is included in `Tags`
3. Tag keys that exist but aren't set on a resource aren't included as tags
--> If tag `team` exists but is not set on resource id 'abc', associated `Tags` will _not_ contain `"team":true`

### DIFF
--- a/specification/columns/tags.md
+++ b/specification/columns/tags.md
@@ -26,7 +26,7 @@ The following is an example of one user-defined tag and one provider-defined tag
 ```json
     {
         "foo":"bar",
-        "acme/foo": "bar"
+        "acme/foo":"bar"
     }
 ```
 
--- a/specification/columns/tags.md
+++ b/specification/columns/tags.md
@@ -9,7 +9,7 @@ The Tags column adheres to the following requirements:
 * The Tags column MUST contain user-defined and provider-defined tags.
 * The Tags column MUST only contain finalized tags.
 * The Tags column MUST be in [Key-Value Format](#key-valueformat).
-* A Tag key that supports a corresponding, non-null value tied to a resource SHOULD NOT be included.
+* A Tag key that supports a corresponding, non-null value tied to a resource SHOULD be included.
 * A Tag key that does *not* support a corresponding value, sometimes referred to as a *label*, MUST have a corresponding true (boolean) value set.
 * If Tag finalization is supported, providers MUST publish tag finalization methods and semantics within their respective documentation.
 * Providers MUST NOT alter user-defined Tag keys or values.
--- a/specification/columns/tags.md
+++ b/specification/columns/tags.md
@@ -9,8 +9,8 @@ The Tags column adheres to the following requirements:
 * The Tags column MUST contain user-defined and provider-defined tags.
 * The Tags column MUST only contain finalized tags.
 * The Tags column MUST be in [Key-Value Format](#key-valueformat).
-* A Tag key that supports a corresponding value but that value is null MUST NOT be included.
-* A Tag key that does *not* support a corresponding value, sometimes referred to as a *label*, MUST have a corresponding null value set.
+* A Tag key that supports a corresponding value but that value is null SHOULD NOT be included.
+* A Tag key that does *not* support a corresponding value, sometimes referred to as a *label*, MUST have a corresponding true (boolean) value set.
 * If Tag finalization is supported, providers MUST publish tag finalization methods and semantics within their respective documentation.
 * Providers MUST NOT alter user-defined Tag keys or values.
 
--- a/specification/columns/tags.md
+++ b/specification/columns/tags.md
@@ -9,7 +9,7 @@ The Tags column adheres to the following requirements:
 * The Tags column MUST contain user-defined and provider-defined tags.
 * The Tags column MUST only contain finalized tags.
 * The Tags column MUST be in [Key-Value Format](#key-valueformat).
-* A Tag key that supports a corresponding value but that value is null SHOULD NOT be included.
+* A Tag key that supports a corresponding, non-null value tied to a resource SHOULD NOT be included.
 * A Tag key that does *not* support a corresponding value, sometimes referred to as a *label*, MUST have a corresponding true (boolean) value set.
 * If Tag finalization is supported, providers MUST publish tag finalization methods and semantics within their respective documentation.
 * Providers MUST NOT alter user-defined Tag keys or values.
--- a/specification/columns/tags.md
+++ b/specification/columns/tags.md
@@ -9,7 +9,7 @@ The Tags column adheres to the following requirements:
 * The Tags column MUST contain user-defined and provider-defined tags.
 * The Tags column MUST only contain finalized tags.
 * The Tags column MUST be in [Key-Value Format](#key-valueformat).
-* A Tag key that supports a corresponding value and that value is null MUST NOT be included.
+* A Tag key that supports a corresponding value but that value is null MUST NOT be included.
 * A Tag key that does *not* support a corresponding value, sometimes referred to as a *label*, MUST have a corresponding null value set.
 * If Tag finalization is supported, providers MUST publish tag finalization methods and semantics within their respective documentation.
 * Providers MUST NOT alter user-defined Tag keys or values.
--- a/specification/columns/tags.md
+++ b/specification/columns/tags.md
@@ -9,7 +9,8 @@ The Tags column adheres to the following requirements:
 * The Tags column MUST contain user-defined and provider-defined tags.
 * The Tags column MUST only contain finalized tags.
 * The Tags column MUST be in [Key-Value Format](#key-valueformat).
-* A Tag key without a specified value MUST have its tag value set to null.
+* A Tag key that supports a corresponding value and that value is null MUST NOT be included.
+* A Tag key that does *not* support a corresponding value, sometimes referred to as a *label*, MUST have a corresponding null value set.
 * If Tag finalization is supported, providers MUST publish tag finalization methods and semantics within their respective documentation.
 * Providers MUST NOT alter user-defined Tag keys or values.
 
--- a/specification/columns/tags.md
+++ b/specification/columns/tags.md
@@ -9,7 +9,8 @@ The Tags column adheres to the following requirements:
 * The Tags column MUST contain user-defined and provider-defined tags.
 * The Tags column MUST only contain finalized tags.
 * The Tags column MUST be in [Key-Value Format](#key-valueformat).
-* A Tag key that supports a corresponding, non-null value tied to a resource SHOULD be included.
+* A Tag key with a non-null value for a given resource SHOULD be included in the tags column.
+* Depending on the provider's tag finalization process, a Tag key with a null value for a given resource MAY be included in the tags column.
 * A Tag key that does *not* support a corresponding value, sometimes referred to as a *label*, MUST have a corresponding true (boolean) value set.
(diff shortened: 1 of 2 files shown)


rileyjenk:
Are we saying that {"name":"peter", "version":NULL} and  {"name":"peter" } are equivalent. My concern here is that we are disagreeing with most technologies/languages that say otherwise. I would say we allow either as both are valid and let the provider decide on which they follow.

Single value tags are not key value tags and often function as default true value tags. Can we specify that?
Example: 
['unicorns','dragons','UFOs'] translates to {'unicorns':TRUE, 'dragons':TRUE, 'UFOs':True'}.

cnharris10:
@rileyjenk so your suggestions are:

1. Change value-only tags to have "true" as a default value (instead of null)
2. Do not explicitly restrict providers (maybe say SHOULD NOT) from providing k/v tags that would have a null value (even though they won't anyways)

Is this correct?

rileyjenk:
@cnharris10 yes that is correct.

specification/columns/tags.md
```diff
@@ -9,7 +9,8 @@ The Tags column adheres to the following requirements:
 * The Tags column MUST contain user-defined and provider-defined tags.
 * The Tags column MUST only contain finalized tags.
 * The Tags column MUST be in [Key-Value Format](#key-valueformat).
-* A Tag key without a specified value MUST have its tag value set to null.
+* A Tag key that supports a corresponding, non-null value tied to a resource SHOULD NOT be included.
```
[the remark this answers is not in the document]
cnharris10: Correct. For example, if an account has 100 total tags created across 10 resources, the only tags that should be included with a resource's line item are the N tags associated with that resource - not all 100 tags.

specification/columns/tags.md
```diff
@@ -9,7 +9,8 @@ The Tags column adheres to the following requirements:
 * The Tags column MUST contain user-defined and provider-defined tags.
 * The Tags column MUST only contain finalized tags.
 * The Tags column MUST be in [Key-Value Format](#key-valueformat).
-* A Tag key without a specified value MUST have its tag value set to null.
+* A Tag key that supports a corresponding, non-null value tied to a resource SHOULD be included.
+* A Tag key that does *not* support a corresponding value, sometimes referred to as a *label*, MUST have a corresponding true (boolean) value set.
```
AWS-ZachErdman, line 13: I'm good with line 13.

specification/columns/tags.md
```diff
@@ -9,7 +9,8 @@ The Tags column adheres to the following requirements:
 * The Tags column MUST contain user-defined and provider-defined tags.
 * The Tags column MUST only contain finalized tags.
 * The Tags column MUST be in [Key-Value Format](#key-valueformat).
-* A Tag key without a specified value MUST have its tag value set to null.
+* A Tag key that supports a corresponding, non-null value tied to a resource SHOULD be included.
```
AWS-ZachErdman, line 12: I'm mostly good with line 12. It gives freedom to the cloud provider to decide what tags to include. Here is a suggested rewording to make it more clear the tag should have a non-null value and what should be included:

> A Tag key-value pair where the value is non-null for a given resource SHOULD be included in the tags column.

This also may raise the question to the reader "What about tag keys that have null values?". We could add a line that maintains the freedom for cloud providers:

> A Tag key-value pair where the value is null for a given resource MAY be included in the tags column, depending on the provider's tag finalization process.

specification/columns/tags.md
```diff
@@ -25,7 +26,7 @@ The following is an example of one user-defined tag and one provider-defined tag
 ```json
     {
         "foo":"bar",
-        "acme/foo": "bar"
+        "acme/foo":"bar"
```
AWS-ZachErdman, line 29: Shall we expand the example to include examples of:
1. labels
2. tags with null values being optionally included

specification/columns/tags.md
cnharris10: I like both but just without `-value pair` because it is defined as a KeyValueFormat attribute already.

Does this edits work? @AWS-ZachErdman 

```
A Tag key where its value is non-null for a given resource SHOULD be included in the tags column.
A Tag key where its value is null for a given resource MAY be included in the tags column, depending on the provider's tag finalization process.
```

AWS-ZachErdman: looks good to me!

specification/columns/tags.md
```diff
@@ -20,12 +22,13 @@ Provider-defined Tags additionally adhere to the following requirements:
 
 ## Provider-Defined vs. User-Defined Tags
 
-The following is an example of one user-defined tag and one provider-defined tag, respectively, with tag key, `foo`.  The first tag is user-defined and not prefixed. The second tag is provider-defined and prefixed with `acme/`, which the provider has specified as a reserved tag key prefix.
+The following is an example of one user-defined tag, one provider-defined tag, respectively, with tag key, `foo`, and one tag key without a supported value.  The first tag is user-defined and not prefixed. The second tag is provider-defined and prefixed with `acme/`, which the provider has specified as a reserved tag key prefix. The third tag with tag key, `baz`, does not support a value, so `null` is assigned as its value.
```
AWS-ZachErdman, line 25: > The third tag with tag key, `baz`, does not support a value, so `null` is assigned as its value.

If it does not *support* a value, does that mean it is label and should be "true"?

My semantics may be wrong, but to me, not *supporting* a value means "label", whereas not *having* a value means "null".

AWS-ZachErdman (changes requested):

specification/columns/tags.md
AWS-ZachErdman: I'm happy with everything else, but interested in your thinking here

cnharris10: Sorry, just a mix up from the previous definition.  Fixed now.

AWS-ZachErdman (approved):
Looks great!

jpradocueva:
Approved by the group on Mar 14.

specification/columns/tags.md
```diff
@@ -9,7 +9,9 @@ The Tags column adheres to the following requirements:
 * The Tags column MUST contain user-defined and provider-defined tags.
 * The Tags column MUST only contain finalized tags.
 * The Tags column MUST be in [Key-Value Format](#key-valueformat).
-* A Tag key without a specified value MUST have its tag value set to null.
+* A Tag key with a non-null value for a given resource SHOULD be included in the tags column.
+* Depending on the provider's tag finalization process, a Tag key with a null value for a given resource MAY be included in the tags column.
+* A Tag key that does *not* support a corresponding value, sometimes referred to as a *label*, MUST have a corresponding true (boolean) value set.
```
flanakin, line 14: Can we make sure we have an ack from practitioners who use labels that this is desirable? Just want to confirm a Boolean value is what they want/expect.

flanakin (approved):

specification/columns/tags.md
```diff
@@ -20,12 +22,13 @@ Provider-defined Tags additionally adhere to the following requirements:
 
 ## Provider-Defined vs. User-Defined Tags
 
-The following is an example of one user-defined tag and one provider-defined tag, respectively, with tag key, `foo`.  The first tag is user-defined and not prefixed. The second tag is provider-defined and prefixed with `acme/`, which the provider has specified as a reserved tag key prefix.
+The following is an example of one user-defined tag, one provider-defined tag, respectively, with tag key, `foo`, and one tag key without a supported value.  The first tag is user-defined and not prefixed. The second tag is provider-defined and prefixed with `acme/`, which the provider has specified as a reserved tag key prefix. The third tag with tag key, `baz`, does not support a value, so `true` (boolean) is assigned as its value.
```
udam-f2, line 25:
```suggestion
This example illustrates three different tagging scenarios. The first two are illustrate when the provider supports both keys and values, while the third is for supporting keys only. The first tag is user-defined and doesn't have a provider prefix. The second tag is provider-defined and has a prefix of `acme/`, which is reserved by the provider. The third tag has a tag key of `baz` and its value is assigned the boolean value `true` since the tag doesn't support a value.
```

The first sentence was a bit hard to read for me. Trying a different way of writing this - accept only if it makes a significant improvement.

specification/columns/tags.md
```diff
@@ -9,7 +9,9 @@ The Tags column adheres to the following requirements:
 * The Tags column MUST contain user-defined and provider-defined tags.
 * The Tags column MUST only contain finalized tags.
 * The Tags column MUST be in [Key-Value Format](#key-valueformat).
-* A Tag key without a specified value MUST have its tag value set to null.
+* A Tag key with a non-null value for a given resource SHOULD be included in the tags column.
+* Depending on the provider's tag finalization process, a Tag key with a null value for a given resource MAY be included in the tags column.
```
udam-f2, line 13:
```suggestion
* A Tag key with a null value for a given resource MAY be included in the tags column depending on the provider's tag finalization process.
```

udam-f2 (approved):
Minor editorial suggestions for consideration by the group.

specification/columns/tags.md
rupagcp: +1, would like to see what labels and tags looks like on a single line item as well

jpradocueva:
This PR will be put forward for approval on Thursday, March 21.

specification/columns/tags.md
AWS-ZachErdman: I'm happy with Udam's suggestion.

specification/columns/tags.md
udam-f2: Example for a label now available

specification/columns/tags.md
```diff
@@ -9,7 +9,9 @@ The Tags column adheres to the following requirements:
 * The Tags column MUST contain user-defined and provider-defined tags.
 * The Tags column MUST only contain finalized tags.
 * The Tags column MUST be in [Key-Value Format](#key-valueformat).
-* A Tag key without a specified value MUST have its tag value set to null.
+* A Tag key with a non-null value for a given resource SHOULD be included in the tags column.
+* A Tag key with a null value for a given resource MAY be included in the tags column depending on the provider's tag finalization process.
+* A Tag key that does *not* support a corresponding value, sometimes referred to as a *label*, MUST have a corresponding true (boolean) value set.
```
flanakin, line 14: We need to update the supporting content to account for where we're landing with labels. Currently, we have 2 questions documented that feel outdated:
- What does it look like interpreting labels using Excel formulas only? Would users be required to dl a plugin, use vbscript, etc.
- Create a poll around calling the dimension “Tags” or “Labels”?

specification/columns/tags.md
flanakin: We discussed in the call and had a small ack. If anyone is concerned, please leave a comment or create a new issue.

specification/columns/tags.md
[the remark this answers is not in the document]
udam-f2: I put this up for review at the member meeting today and folks did not vote for changing it. Let's open a separate issue if we want to still make this change - but for now, we're moving forward without to get this long-running PR checked in.

specification/columns/tags.md
udam-f2: Michael said at the maintainer meeting he's good with the logic that this is following. Resolving.

PR #360, opened 2024-03-11
354 General Structure for Metadata and Logical sections for metadata vers…

rileyjenk:
…ion and schema


udam-f2:
@rileyjenk Thanks for getting this started. 

Would like to pressure-test the idea of us getting into the 'how' (in terms of how providers make this data available) as opposed to 'what' they make available. Thus far, we've stuck to the 'what' and not worried about the 'how' to avoid getting into implementation details. As long as the required and preferred data is provided using the terminology and semantics we specify, it could be a file, database, api etc. Do we feel differently here?

rileyjenk:
@udam-f2 Yeah I agree with what you are saying. I like the idea of despite what provider I am working with, the calls to get the metadata are the same (depending on mechanism of choice). We could soften this to be not the **what** and the **how** but provide the what, and convert the how to an example.

cnharris10 (changes requested):
All of this requires more discussion but my main feedback centers around taking a more abstracted view around metadata. I think it's fine for us to define aspects of metadata that are necessary vs. desired. This could be a list of columns, list of files, corresponding account, checksum signature, etc. but given the various methods of delivery, I'd caution us to try to differentiate between DB's vs. Blob stores vs. API's vs. ...

specification/metadata/metadata_schema.md
```diff
@@ -0,0 +1,75 @@
...
+The FOCUS metadata schema provides a list of the columns present in the FOCUS dataset provided.   
+
+
+## Logical Section: Schema
+
+## Available Meta Data
+ * ### schema_id
+   * Use: unique identifier for the schema
+   * Type: STRING
+ * ### index
```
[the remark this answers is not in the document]
rileyjenk: @cnharris10 I like this. I am fine pulling it ensuring that we indicate that the order in the array is the schema order...which is implied.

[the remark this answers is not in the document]
rileyjenk: @cnharris10 and yes much much more discussion is needed. I landed on the initial draft being more prescriptive. I land in the camp that if the meta-data section is not more specific, we end up with metdata that is substantially different provider to provider.

specification/metadata/metadata_schema.md
```diff
@@ -0,0 +1,75 @@
+# FOCUS Metadata Schema
+The FOCUS metadata schema provides a list of the columns present in the FOCUS dataset provided.   
+
+
+## Logical Section: Schema
+
+## Available Meta Data
```
flanakin, line 7:
```suggestion
## Available Metadata
```

specification/metadata/metadata_schema.md
```diff
@@ -0,0 +1,75 @@
...
+ * ### schema_id
```
flanakin, line 8: Shouldn't the metadata schema use the same column attributes as the FOCUS schema? 
```suggestion
 * ### SchemaId
```

specification/metadata/metadata_schema.md
```diff
@@ -0,0 +1,75 @@
...
+   * Use: unique identifier for the schema
```
flanakin, line 9: This isn't clear. What's an example?

specification/metadata/metadata_schema.md
```diff
@@ -0,0 +1,75 @@
...
+   * Type: STRING
+ * ### index
+   * use: provides the index number for the column to specify the order of the columns, 0 based
```
flanakin, line 12: We should probably define these with MUST and SHOULD statements.

specification/metadata/metadata_schema.md
```diff
@@ -0,0 +1,75 @@
...
+   * type: INTEGER
+ * ### column_name
+   * use: provides the date the schema was created
```
flanakin, line 15: Mismatch?

specification/metadata/metadata_schema.md
```diff
@@ -0,0 +1,75 @@
...
+   * type: STRING
+ * ### data_type (provided by API delivery mechanism only)
```
flanakin, line 17: Once I got here, I realized we should probably use an industry standard schema format rather than defining our own. Thoughts? JSON Schema should work, right?

specification/metadata/metadata_schema.md
```diff
@@ -0,0 +1,75 @@
...
+| schema_id | ordinal | column_name                    | data_type |
+|----------------|---------|--------------------------------|-----------|
+| 1234           | 1       | "Name of included FOCUS column" | STRING    |
+| 1234           | 2       | "Name of included FOCUS column" | LONG      |
+| 1234           | 3       | "Name of included FOCUS column" | DATETIME  |
+
+
+### API
+
+#### Endpoint: <api_root>/FOCUS/metadata/schema/{id}
```
flanakin, line 46: Is this an required URL path or just conveying it should be a simple GET request? What if someone doesn't use REST?

specification/metadata/metadata_schema.md
```diff
@@ -0,0 +1,75 @@
...
+   * use: provides the date the schema was created
+   * type: STRING
+ * ### data_type (provided by API delivery mechanism only)
+   * use: provides the endpoint for the schema, including the id
+   * type: STRING
+   * content_restrictions: <format for specifying datatype needed here>
+
+
+
+## Methods of Delivery
```
flanakin, line 24: What about a file as part of the export job?

specification/metadata/metadata_schema.md
rileyjenk: yeah will fix.

specification/metadata/metadata_schema.md
rileyjenk: Will clarify.

marc-perreaut:
Sharing some thoughts:
- Is there is a use case where different versions for FOCUS data and FOCUS meta-data would be useful?
- The meta-data could include the date when data was exported, knowing that cost data evolves over time.
- Meta-data could contain scope information, so that it avoids to parse FOCUS data to the scope information: billing data range, billing account, provider
- The delivery of meta-data might remain specific to the provider, in the same way the delivery of FOCUS data might remain specific to the provider.

specification/metadata/metadata_schema.md
rileyjenk: We likely will narrow this down to examples. Debate is out on that.

rileyjenk: @flanakin these are now all examples rather than specifying "How"

specification/metadata/metadata_schema.md
rileyjenk: As discussed today we are choosing to delay this till a future version. Current content specifies what metadata we want.

specification/metadata/metadata_schema_reference.md
```diff
@@ -0,0 +1,31 @@
+# Reference in FOCUS Data
+The provider must provide a reference to the schema utilized in the FOCUS dataset. Becuase methods of delivery vary, the reference to the schema utilized may be built in, such as a database, where schema is an included component of the tables. Others such as a file, may not be built in and therefore need to provide an ID to the schema used.
```
mike-finopsorg, line 2:
```suggestion
The provider must provide a reference to the schema utilized in the FOCUS dataset. Because methods of delivery vary, the reference to the schema utilized may be built in, such as a database, where schema is an included component of the tables. Others such as a file, may not be built in and therefore need to provide an ID to the schema used.  
```

specification/metadata/metadata_provider.md
```diff
@@ -0,0 +1,38 @@
...
+    * use: Human readable name of the entity that is generating the billing data 
+    * type: STRING
+* ### provider_prefix
+    * use: Shortened name of the provider supplied to be used as the prefix for tags and may be used to modify the name of custom columns.
+    * type: STRING
+
+
+
+
+## Example Provider Metadata
```
mike-finopsorg, line 19: What does everyone feel about these API examples. Part of me wants them moved to supporting content and not be in the first cut of this section for 1.0?

specification/metadata/metadata_structure.md
```diff
@@ -0,0 +1,25 @@
+# Introduction
```
udam-f2, line 1: I would move some of this to the mdpp since otherwise when stitched together, these all get nested (there's a brief description of metadata in that file and also here).

specification/metadata/metadata.mdpp
```diff
@@ -0,0 +1,13 @@
+# Metadata
+
+The FOCUS specification defines a metadata structure that is to be supplied by data providers to facilitate practioners in data ingestion, data locations, schema of the FOCUS dataset.
+
+!INCLUDE "metadata_structure.md",1
+!INCLUDE "metadata_provider.md",1
```
udam-f2, line 6: Nesting with mdpp will be a lot more harder if we want to manage subsections. I'd like to see if we can keep it more flat - it would be consistent with what we did with columns (where we previously had metrics and dimensions).

mike-finopsorg: I am working on a proposal to rework the layout of this content. Hoping to have it available by tomorrow.

specification/metadata/metadata_provider.md
```diff
@@ -0,0 +1,38 @@
+# FOCUS Metadata Provider
```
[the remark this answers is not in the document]
rileyjenk: Yeah I am ok with that. I bounced back and forth because we talked about provider prefix vs data Generator.

specification/metadata/metadata_provider.md
rileyjenk: I am fine moving them as long as we include clear links to the examples.

mike-finopsorg:
I opened #390 as an alternative option to content layout. It is a more simplified structure that is closer aligned to other sections of the specification. Please review and provide feedback as to the preferred direction of this. cc:// @rileyjenk happy to catch up when you are back around do align on these two PRs

specification/metadata/metadata_generator.md
```diff
@@ -0,0 +1,37 @@
...
+
+## Logical Section: Schema Generator
+
+## Schema Definition Metadata
+
+* ### billing_generator
+    * use: Human readable name of the entity that is generating the billing data 
+    * type: STRING
+* ### provider_prefix
+    * use: Shortened name of the provider supplied to be used as the prefix for tags and may be used to modify the name of custom columns.
```
flanakin, line 12: What happens when there are multiple prefixes for different scenarios? I need to look at Tags but I recall internalizing scenarios where we would use different prefixes and not a single value.

specification/metadata/metadata_schema.md
```diff
@@ -0,0 +1,64 @@
...
+
+## Logical Section: Schema
+
+
+## Available Meta Data:
+ * ### schema_id
+   * Use: unique identifier for the schema
+   * Type: STRING
+ * ### FOCUS_version
+   * Use: the declared version of the FOCUS data within the schema
```
flanakin, line 12: Should there be a separate provider version? I will absolutely have multiple versions of 1.0 as the platform evolves.

specification/metadata/metadata_schema.md
```diff
@@ -0,0 +1,64 @@
...
+    endpoint: <api_root>/FOCUS/metadata/schemas 
+    method: GET 
+####
+
+#### Response 
+```
+{
+	[
+		{
+			"schema_id": "1234",
```
flanakin, line 45: nit: If we define this as JSON, I would really like to use camel-casing to align to JSON standards.

specification/metadata/metadata_schema_column_definition.md
```diff
@@ -0,0 +1,87 @@
+# FOCUS Metadata Schema Column Definition
+The FOCUS metadata schema column definition provides a list of the columns present in the FOCUS dataset along with metadata about the columsn. The intent of the schema metadata is to aid pracitioners to understand the structure of the FOCUS data, without the need to parse the included dataset. 
+
+
+## Logical Section: Schema 
+### Subsection: Schema Column Definition
+
+## Schema Definition Metadata
+
+ * ### column_name
```
flanakin, line 10: Should this be an ID to align to the spec?

specification/metadata/metadata.mdpp
```diff
@@ -0,0 +1,13 @@
+# Metadata
+
+The FOCUS specification defines a metadata structure that is to be supplied by data providers to facilitate practioners in data ingestion, data locations, schema of the FOCUS dataset.
```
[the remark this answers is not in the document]
marc-perreaut: Would it be an issue to make it a MUST?

specification/metadata/metadata_generator.md
```diff
@@ -0,0 +1,37 @@
...
+The FOCUS metadata provider logical section provides metadata about the provider of the FOCUS data. 
+
+## Logical Section: Schema Generator
+
+## Schema Definition Metadata
+
+* ### billing_generator
+    * use: Human readable name of the entity that is generating the billing data 
+    * type: STRING
+* ### provider_prefix
```
marc-perreaut, line 11: Conceptually, the provider prefix belongs to the provider, not to the data generator: two data generators for the same provider should have the same prefix.
Maybe we should not go as far as splitting provider metadata and generator metadata, but maybe the specification could say that the provider prefix must be the same for all data generators addressing the same provider.

specification/metadata/metadata_generator.md
marc-perreaut, line 12: I believe the column prefix and the tag prefix might be different, so they might two different meta-data columns.
There might be cases where some providers have different prefixes for tags, so a list type might be needed (JSON array of strings here).

specification/metadata/metadata_schema.md
```diff
@@ -0,0 +1,64 @@
+# FOCUS Metadata Schema 
+Each FOCUS dataset must have a metadata about the schema associated with it. The schema metadata provides information about the structure of the data provided. In the event that the provider modifies the columns provided in the FOCUS dataset, metadata about the schema must be provided, including column definitions.
```
marc-perreaut, line 2: Does the schema belong to the provider or to the data generator?

I would go the simple path where we don't split the metadata between provider and data generator: there would be only data generator metadata, and then it would be the data generator schema.

I would tend to think

specification/metadata/metadata_schema.md
marc-perreaut: Provider version or data generator version or both?

specification/metadata/metadata_schema_column_definition.md
```diff
@@ -0,0 +1,87 @@
+# FOCUS Metadata Schema Column Definition
+The FOCUS metadata schema column definition provides a list of the columns present in the FOCUS dataset along with metadata about the columsn. The intent of the schema metadata is to aid pracitioners to understand the structure of the FOCUS data, without the need to parse the included dataset.
```
marc-perreaut, line 2: typo: columsn

specification/metadata/metadata_schema_column_definition.md
```diff
@@ -0,0 +1,87 @@
...
+
+## Schema Definition Metadata
+
+ * ### column_name
+   * use: provides the date the schema was created
+   * type: STRING
+ * ### data_type
+   * use: data type of the column as provided and must be provided for all columns in the FOCUS dataset
+   * type: STRING
+ * ### max_length
```
marc-perreaut, line 16: Maybe max_string_length / MaxStringLength to be explicit

specification/metadata/metadata_schema_column_definition.md
```diff
@@ -0,0 +1,87 @@
...
+   * use: provides a maximum character length for columns whose data type where character length is relevant. This is typically provided for String type columns
+   * type: Integer
+* ### precision
```
marc-perreaut, line 19: Maybe number_precision/NumberPrecision to be explicit

specification/metadata/metadata_schema_column_definition.md
```diff
@@ -0,0 +1,87 @@
...
+  * use: provide the maximum number of digits for the values in the column. This is typically provided for numeric columns
+  * type: Integer 
+* ### scale
```
marc-perreaut, line 22: Maybe number_scale/NumberScale to be explicit

specification/metadata/metadata_schema_column_definition.md
```diff
@@ -0,0 +1,87 @@
...
+   * use: provide the maximum number of digits after the decimal point. This is typically provided for numeric columns
+   * type: Integer
+* ### encoding
```
marc-perreaut, line 25: Maybe string_encoding / StringEncoding to be clear

ahullah (approved):

specification/metadata/metadata_schema_column_definition.md
rileyjenk: Yeah I am good with that.

specification/metadata/metadata_schema_column_definition.md
rileyjenk: Fixed.

specification/metadata/metadata_schema.md
rileyjenk: Yeah I am going to do a commit to do that all in one pass.

specification/metadata/metadata.mdpp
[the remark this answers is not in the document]
rileyjenk: I would like to make this a MUST.

projectlocale (approved):
My only concern is that on balance the this part of the spec would be MUST as a group of attributes if chosen to use. I don't think that's implicit, but perhaps I am missing something about the spec's design.

Issue #336, opened 2024-02-27
Add the capability to calculate savings resulting from negotiation activities


ijurica:
### Type
> Potentially multiple columns

### Description
> Practitioners require the capability to calculate savings resulting from negotiation activities. To facilitate this, we suggest exploring the following options:
> - NegotiatedUnitPrice and NegotiatedCost columns
> - NegotiatedDiscount column
> - Considering other potential alternatives

### Related Issues
> - #310 
> - #237 (related PR: #238)

### Definition of done

- [ ] Rationalize vendor-neutral, cross-cloud naming
- [ ] Complete spec template and include naming (code name, display name), constraints, guidelines, compatibility with major providers etc.

kk09v:
One thing I've been thinking about is that if we are capturing multiple discount types applying to the same resource/record, do we need to specify the order of operations for which the discounts are applied? A simple example: if I have 2 discounts for 10% each applied to one resource, I get a 19% discount on that resource 1 - ((1-0.1) * (1-0.1));  So which discount gets the 10% (applied first) and which gets the 9% (applied second) or do they both get 9.5% (somehow)?

With contractual pricing, I think the answer is clear that contractual discounts take effect first before other discounts, but even this is messy, as the below scenario tries to illustrate.

An example of this using fake, round numbers for illustration:
- A company negotiates a 20% discount (vs list) on PAYGo usage of a VM
- At list price, a commitment grants you a 30% discount vs PAYGo of that same VM
- The company only receives a 10% discount of of the reservation price of that VM
Math time
- Lp = List Price of PAYGo | Cp = Contractual Price of PAYGo | Lc = List price of a Commitment | Cc = Contractual price of a commitment
- Cc = 0.9 Lc | Lc = 0.7 Lp ∴ Cc = 0.63 Lp Commitments under contractual pricing are a 37% discount from list
- Cp = 0.8 Lp | Cc = 0.63 Lp ∴ Cc = 0.7875 Cp Commitments under contractual pricing are a 21.25% discount from PAYGo contractual pricing, which is what the company will be looking at when evaluating whether to purchase a commitment

[...]

PR #364, opened 2024-03-13
FOCUS #336: Negotiated/Contracted List Unit Price and Cost 

ijurica:
- Negotiated List Unit Price specification and supporting content
- Negotiated List Cost specification and supporting content
- glossary update - _Negotiated List Unit Price included_


ijurica:
## Open issues/Discussion topics:

- **Column naming `Negotiated*` vs `OnDemand*`vs `NegotiatedList*`**
  - See https://github.com/FinOps-Open-Cost-and-Usage-Spec/FOCUS_Spec/issues/336#issuecomment-1975093319, posted by @flanakin 
  - **Suggestion:** publish a poll, provide PROs and CONs

- **UCs and sample data**
  - Compile a Google Spreadsheet containing FOCUS columns relevant for the discussion, list UCs, and include some sample data.

- **Discount handling discussion, addressing both Negotiated and Commitment based discounts**
   - See https://github.com/FinOps-Open-Cost-and-Usage-Spec/FOCUS_Spec/issues/336#issuecomment-1969864067, posted by @kk09v 

- **FOCUS Essential**
  - Currently both columns column MUST be present in the billing data when the provider supports negotiated pricing concept

specification/columns/negotiatedlistcost.md
```diff
@@ -0,0 +1,37 @@
+# Negotiated List Cost
```
marc-perreaut, line 1: Maybe the name Negotiated List Cost is a bit long, and it would imply to rename the column List Cost into Public List Cost, to avoid ambiguities.
Is there any issue with the shorter name Negotiated Cost?

specification/columns/negotiatedlistunitprice.md
```diff
@@ -0,0 +1,32 @@
+# Negotiated List Unit Price
+
+The Negotiated List Unit Price represents the agreed-upon unit price for a single [Pricing Unit](#pricingunit) of the associated SKU, inclusive of negotiated discounts, if present, and exclusive of any other discounts. This price is denominated in the [Billing Currency](#billingcurrency). The Negotiated List Unit Price is commonly used for calculating savings based on negotiation activities. If negotiated discounts are not applicable, the Negotiated List Unit Price defaults to the [List Unit Price](#listunitprice).
```
marc-perreaut, line 3: >  inclusive of negotiated discounts, if present, and exclusive of any other discounts

Maybe there is a discussion here on the phrasing, and actually the scope.
Here is a way to view things:
- Providers have a public offer, which include discounts for some usages, like commitment-based discounts.
- Customers negotiate with providers, leading to negotiated discounts, which may apply on top of the discounts of the public offer. For example, a customer may negotiated an additional discount on commitment-based discounts.
In that perspective, what is "negotiated discounts, if present, and exclusive of any other discounts" might not be clear on whether it includes or not commitment-based discounts.

As a practitioner, it's important to know the negotiated cost without commitment-based discounts, for 2 use cases:
- Calculate the savings made thanks to the commitments. This relates to the Effective Savings Rate: https://www.finops.org/wg/finops-kpis/?prod_kpis%5Bpage%5D=2#modal-kpi-9650
- Calculate the cost of a workload without the commitments, so as to understand whether effective (amortized) cost variations of the workload are linked to a change in the distribution of the commitment discount or a change in usage.
I am used to call "on-demand cost" the cost of a workload without the commitments. Is it the intent of the negotiated list cost of this PR, or does the negotiated cost intends something different?

supporting_content/columns/negotiatedlistunitprice.md
```diff
@@ -0,0 +1,64 @@
...
+### GCP column mappings related issues
+
+* **price.effective_price** is present only in detail export and thus available only for small subset of services/resources. In order to resolve this information we should rely on **BigQuery Pricing Data Export** and/or **Cloud Billing API**, which provide price related structs (out-of-scope of this document).
+* Similar to cost, regardless of the data source we use, the resolved price does not include reductions provided in credits, which aligns with the definition of NegotiatedListUnitPrice.
+
+### Microsoft column mappings related issues
+
+* [Azure Billing Enterprise APIs - PriceSheet | Microsoft Learn](https://learn.microsoft.com/en-us/rest/api/billing/enterprise/ provides information for EA.
+* TODO: Check if there is a Microsoft Custom Prices REST API, which might provide custom/negotiated prices for all account types.
+* **unitPrice** (The price per unit for the charge.) includes commitment based discounts and thus is not applicable for all use cases. Furthermore,
```
marc-perreaut, line 52: To my understanding, for Azure EA, the UnitPrice excludes commitment-based discounts, and so it can be used to calculate the (negotiated) cost without the commitment discounts (Microsoft documentation calls it the "pay-as-you-go cost").

https://learn.microsoft.com/en-us/azure/cost-management-billing/reservations/understand-reserved-instance-usage-ea#determine-reservation-savings
> Multiply the UnitPrice value with Quantity values to get estimated pay-as-you-go costs, if the reservation discount didn't apply to the usage.

https://learn.microsoft.com/en-us/azure/cost-management-billing/savings-plan/utilization-cost-reports#determine-savings-resulting-from-savings-plan
>  Multiply the UnitPrice value with Quantity values to get estimated pay-as-you-go costs if the savings plan discount didn't apply to the usage.

specification/columns/negotiatedlistcost.md
flanakin: +1 
Let's break this down... 

re: "Negotiated":
The vast majority of organizations don't have negotiated discounts. And given this is the basis for everything they will look at, that means all documentation will revolve around this as the way to quantify commitment savings. Given we've already differentiate between committed and non-committed (aka "on-demand") usage in PricingCategory as "On-Demand", we should stick with consistent terminology here and use "OnDemand".

re: "List":
This feels unnecessary. Everything is discounted from list, so if we add it here, wouldn't it make sense in every price/cost column? I would argue that's overkill.

specification/columns/negotiatedlistunitprice.md
flanakin: +1
This also aligns with PricingCategory == "On-Demand" (ad-hoc usage that was not committed).

specification/columns/negotiatedlistcost.md
ijurica: Could you please take a look at the [FOCUS 1.0 columns and negotiation activities - Definition and Naming options GSpreadsheet](https://docs.google.com/spreadsheets/d/1VlDx7wGH1atBTksQV-sR2ubxbDQjOBdt7cJsQH0SapQ/edit#gid=749340528)? It contains some options, pros, and cons. 

Any feedback, suggestions, or additional criteria would be greatly appreciated. Thank you!

marc-perreaut: To help for the naming discussion, do we foresee other types of costs than List, Negotiated, Billed, Effective?

specification/columns/negotiatedlistunitprice.md
ijurica: - Please let me know if this version encapsulates what we are looking for and provides more clarity in terms of scope. 
  Any suggestions are welcome :) 
  > The XXX Unit Price represents the agreed-upon unit price for a single Pricing Unit of the associated SKU, inclusive of negotiated discounts, if present, **while exclusive of any commitment-based discounts or any other discounts**...

- Regarding the column names, let's discuss and vote. Please take a look at the [GSpreadsheet](https://docs.google.com/spreadsheets/d/1VlDx7wGH1atBTksQV-sR2ubxbDQjOBdt7cJsQH0SapQ/edit#gid=749340528). As previously mentioned, any feedback, suggestions, or additional criteria would be greatly appreciated.

ijurica: my bad :(

Here is what I should have posted earlier:

> The XXX Unit Price represents the agreed-upon unit price for a single Pricing Unit of the associated SKU, inclusive of negotiated discounts, if present, **while excluding negotiated commitment-based discounts or any other discounts.**

emphasizing the exclusion of the negotiated commitment based discounts and any other...

flanakin: specification/columns/negotiatedlistunitprice.md

marc-perreaut: Maybe we can also make explicit that the the negotiated discount, which is referred to here, is the on the list price:

> "...inclusive of negotiated discounts on the list unit price based on the provider agreement, while excluding other discounts like bundled or commitment-based discounts."

Maybe the sentence starts to be heavy, but it gives more precise contours.

tobrien (approved):

supporting_content/columns/contractedcost.md
```diff
@@ -0,0 +1,51 @@
...
+
+## Example provider mappings
+
+Current column mappings found in available data sets:
+
+| Provider  | Data set                | Column                                                                                     |
+|:----------|:------------------------|:-------------------------------------------------------------------------------------------|
+| AWS       | CUR;                    | Not available                                                                              |
+| GCP       | BigQuery Billing Export | Not available? or **cost** (the closest match);<br>*see Discussion / Scratch space - GCP column mappings for more details*        |
+| Microsoft | Cost details            | Not available?<br>CostInBillingCurrency is probably the closest match<br>*see Discussion / Scratch space - Microsoft column mappings related issues for more details* |
```
flanakin, line 11:
```suggestion
| Microsoft | Cost details (actual/amortized) | Not available<br>Can calculate using UnitPrice * Quantity |
| Microsoft | Cost details (FOCUS `1.0-preview(v1)`) | x_OnDemandCost |
```

specification/columns/negotiatedlistcost.md
marc-perreaut: Discussion was done about the name and the vote result is Contracted Cost => resolving conversion

supporting_content/columns/contractedunitprice.md
```diff
@@ -0,0 +1,64 @@
...
+
+## Example provider mappings
+
+Current column mappings found in available data sets:
+
+| Provider  | Data set                     | Column                   |
+|:----------|:-----------------------------|:-------------------------|
+| AWS       | CUR;                         | Not available            |
+| GCP       | BigQuery Billing Export;<br>BigQuery Pricing Data Export;<br>Cloud Billing API | Not available in BigQuery Billing Export<br>**price.effective_price** available only in detail export, i.e. available for a subset of resources/services<br>*see Discussion / Scratch space - GCP column mappings related issues for more details* |
+| Microsoft | Cost details                 | Not available?<br>**unitPrice** is probably the closest match but not applicable in all use cases<br>*see Discussion / Scratch space - Microsoft column mappings related issues for more details* |
```
flanakin, line 11:
```suggestion
| Microsoft | Cost details (actual/amortized) | UnitPrice<br>*see Discussion / Scratch space - Microsoft column mappings related issues for more details* |
| Microsoft | Cost details (FOCUS `1.0-preview(v1)`) | x_OnDemandUnitPrice |
```

salpaysenturus:
I'm not sure where to put this comment, so I'll put it here.

I am concerned about the dissonance between these two facts:

- Through a vote, the team decided that "Negotiated" is not the best term to use for these columns.
- "Negotiated discount" is common parlance [throughout the FOCUS spec](https://github.com/search?q=repo%3AFinOps-Open-Cost-and-Usage-Spec%2FFOCUS_Spec%20negotiated%20discounts&type=code) and the [FinOps website](https://www.google.com/search?sca_esv=06eb692d555481e2&sxsrf=ACQVn09WZOkdUe5d_sKnJIgbfdBjCC1kuA:1711054197068&q=site:finops.org+%22negotiated%22+discounts&sa=X&ved=2ahUKEwjX6-3MnYaFAxWrHjQIHSB4CrkQ5t4CegQIHRAB&biw=1537&bih=958&dpr=2).

I would really love to see these terms aligned.  For example, I would love to be able to say one of the following:

- The _Contracted_ Cost column represents the difference between list cost and _contracted_ discounts.  
- The _Negotiated_ Cost column represents the difference between list cost and _negotiated_ discounts.

Let me frame it with a question.  **Can we start using the term "contracted discounts" instead of "negotiated discounts" in FinOps parlance?**  "Contracted" and "negotiated" feel like synonyms that, to my mind, only serve to introduce confusion.

To my mind, it seems that "negotiated discounts" is the accepted FinOps parlance for this concept, and so the names of the columns in this PR should represent that -- but I also want to respect the mandate of the vote we took and avoid dragging us backwards.  

At the very least, I want there to be a paper trail around the decision, however it lands.  If the team feels that the need to align these terms is unnecessary, then great -- at least it's there for reference when someone asks about it.

specification/columns/contractedcost.md
```diff
@@ -0,0 +1,37 @@
+# Contracted Cost
+
+Contracted Cost represents the cost calculated by multiplying [*contracted unit price*](#glossary:contracted-unit-price) and the corresponding [Pricing Quantity](#pricingquantity). Contracted Cost is denominated in the [Billing Currency](#billingcurrency) and is commonly used for calculating savings based on negotiation activities, by comparing it with [List Cost](#listcost). If negotiated discounts are not applicable, the Contracted Cost defaults to the List Cost.
+
+The ContractedCost column MUST be present in the billing data and MUST NOT be null. This column MUST be of type Decimal, MUST conform to [Numeric Format](#numericformat) requirements, and be denominated in the BillingCurrency. When ContractedUnitPrice is present and is not null, multiplying the ContractedUnitPrice by PricingQuantity MUST produce the ContractedCost.
+
+In cases where the ContractedUnitPrice is present and null, the following applies:
```
ahullah, line 7: we might want to pick this up in #366 but i dont think there is any scenario where this value should be NULL

specification/columns/contractedunitprice.md
```diff
@@ -0,0 +1,32 @@
...
+## Content Constraints
+
+| Constraint      | Value                                |
+|:----------------|:-------------------------------------|
+| Column type     | Metric                               |
+| FOCUS Essential | False                                |
+| Allows nulls    | True                                 |
+| Data type       | Decimal                              |
+| Value format    | [Numeric Format](#numericformat)     |
+| Number range    | Any valid non-negative decimal value |
```
ahullah, line 28: would we need to support a negative unit rate in the event of refunds?

ahullah (changes requested):

specification/columns/contractedunitprice.md
ijurica: my thoughts...

**UnitPrices are always non-negative** and in case of refunds:

-  the **Quantity is negative** when ChargeCategory is Usage or Purchase and ChargeClassification is Correction (Correction indicated refund) (the` C = P * Q` math works)
-  the Quantity is null when ChargeCategory is Adjustments (the math doesn't apply)

specification/columns/contractedcost.md
ijurica: For instance, in case of Adjustments (bulk refunds) SkuPriceId is null and thus Unit Prices, Quantities, etc. as well.

specification/columns/contractedcost.md
```diff
@@ -0,0 +1,37 @@
...
+
+ContractedCost
+
+## Display name
+
+Contracted Cost
+
+## Description
+
+Cost calculated by multiplying [*contracted unit price*](#glossary:contracted-unit-price) and the corresponding Pricing Quantity.
```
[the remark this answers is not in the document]
ijurica: my thoughts...

When discussing ListCost (very similar to ContractedCost), we experimented with several versions and ultimately settled on this **simplified approach, focusing on use cases where this Cost makes sense**, i.e. there exists a corresponding UnitPrice). Edge cases such as Tax are addressed in the normative paragraph, and it seamed to me that that's good enough.

**The general approach I tried to follow:** 

- **Description** captures the essence. 
- **Definition** provides more details and a better understanding (When it comes to List or Contracted Cost, the definition also points to the corresponding Unit Price). 
- Edge cases covered in the **Normative paragraph**.

salpaysenturus:
Thanks for the thoughtful reply, @ijurica.  :)  My concern stems from a fact you've mentioned: 

> Both Negotiated Discount and Negotiated UnitPrice/Cost are well-established terms resulting from negotiation activities.

In this case, we are electing to avoid the "well-established term" for these columns and introducing a net-new FinOps term in order to be more inclusive of organizations that perform no negotiation (though I would argue that such orgs will not care about these columns in the first place, and the term Negotiated would help explain that in a self-documenting fashion).  Therefore, the new user/implementer of FOCUS data must then 1) learn the standard itself (e.g. "what is the Contracted Unit Price column?"), and 2) learn new FinOps terms that the specification introduces (e.g. "Wait... what does 'Contracted' mean?  Oh, it's a synonym of Negotiated... kind of.").  To my mind, this raises the bar of difficulty for FOCUS adoption.

Obviously FOCUS introduces net-new terms in other places (I'm looking at you, SKU Price) -- but those terms have been introduced in circumstances when no already widely-accepted parlance existed within the subject matter.

As I've said above, I know I'm in the minority on this topic -- so this will be the last time I grouse about it.  But I believe that this name choice will introduce more confusion than it alleviates.

specification/columns/contractedcost.md
ahullah: Discussed on 28.3.24, there is a conflict between  quantity and unit price requirements when sku price id is not present and the need to not introduce artificial variance when the 4 cost columns are not in sync... ie one column contains a value but others are NULL. we will follow up in this in #366

specification/columns/contractedunitprice.md
ahullah: Discussed on 28.3.24, there is a conflict between  quantity and unit price requirements when sku price id is not present and the need to not introduce artificial variance when the 4 cost columns are not in sync... ie one column contains a value but others are NULL. we will follow up in this in #366

ahullah (approved):

flanakin (approved):

specification/columns/contractedcost.md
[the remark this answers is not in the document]
jpradocueva: @aqu-erp, Andrew, could you please review the comments from Irena and indicate if you are ok with the explanations? Thanks.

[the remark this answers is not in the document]
aqu-erp: LGTM

aqu-erp (approved):

jpradocueva:
Approved by the group on April 4th.

specification/columns/columns.mdpp
```diff
@@ -23,6 +23,8 @@ The FOCUS specification defines a group of columns that provide qualitative valu
 !INCLUDE "invoiceissuer.md",1
 !INCLUDE "listcost.md",1
 !INCLUDE "listunitprice.md",1
+!INCLUDE "contractedcost.md",1
```
udam-f2, line 26: nit: We need these to be in alphabetical order i believe.

udam-f2 (approved):
LGTM

specification/columns/columns.mdpp
ijurica: good catch! resolved.

PR #368, opened 2024-03-22
Clarify what qualifies as a tag and what different types of tags are

AWS-ZachErdman:
We incorrectly suggested that **labels** are usually **key-only tags**. In GCP, labels are key-value pairs https://cloud.google.com/compute/docs/labeling-resources

This PR corrects this by not trying to give a specific name to key-value tags and key-only tags.

This PR also attempts to provide some guidance on what qualifies as a tag (to prevent this from becoming a dumping ground for key-value data) and suggests some other names for tags.

### DIFF
--- a/specification/columns/tags.md
+++ b/specification/columns/tags.md
@@ -1,6 +1,6 @@
 # Tags
 
-The Tags column represents the set of tags assigned to [*tag sources*](#glossary:tag-source) that also account for potential provider-defined or user-defined tag evaluations. Tags are commonly used for scenarios like adding business context to billing data to identify and accurately allocate charges. Tags may also be referred to by providers by other terms, such as labels or networks tags.
+The Tags column represents the set of tags assigned to [*tag sources*](#glossary:tag-source) that also account for potential provider-defined or user-defined tag evaluations. Tags are commonly used for scenarios like adding business context to billing data to identify and accurately allocate charges. Tags may also be referred to by providers using other terms such as labels.
 
 A tag becomes [*finalized*](#glossary:finalized-tag) when a single value is selected from a set of possible tag values assigned to the tag key.  When supported by a provider, this can occur when a tag value is set by provider-defined or user-defined rules.
 
--- a/specification/columns/tags.md
+++ b/specification/columns/tags.md
@@ -1,6 +1,6 @@
 # Tags
 
-The Tags column represents the set of tags assigned to [*tag sources*](#glossary:tag-source) that also account for potential provider-defined or user-defined tag evaluations.  In FOCUS, "Tag" is a generic term to describe metadata that can be attached to a resource, account, etc. by a user or provider to help identify it. Tags may also be referred to by providers by other terms, such as labels or networks tags. Tags are commonly used for scenarios like adding business context to billing data to identify and accurately allocate charges. 
+The Tags column represents the set of tags assigned to [*tag sources*](#glossary:tag-source) that also account for potential provider-defined or user-defined tag evaluations. Tags are commonly used for scenarios like adding business context to billing data to identify and accurately allocate charges. Tags may also be referred to by providers by other terms, such as labels or networks tags.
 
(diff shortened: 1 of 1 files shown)


specification/columns/tags.md
```diff
@@ -1,6 +1,6 @@
 # Tags
 
-The Tags column represents the set of tags assigned to [*tag sources*](#glossary:tag-source) that also account for potential provider-defined or user-defined tag evaluations.  Tags are commonly used for scenarios like adding business context to billing data to identify and accurately allocate charges.
+The Tags column represents the set of tags assigned to [*tag sources*](#glossary:tag-source) that also account for potential provider-defined or user-defined tag evaluations.  In FOCUS, "Tag" is a generic term to describe metadata that can be attached to a resource, account, etc. by a user or provider to help identify it. Tags may also be referred to by providers by other terms, such as labels or networks tags. Tags are commonly used for scenarios like adding business context to billing data to identify and accurately allocate charges.
```
cnharris10, line 3: Couple suggestions:
1. I'd propose moving sentence, "In FOCUS, "Tag" is ..." to adjust the "Tag" term in the [Glossary](https://github.com/FinOps-Open-Cost-and-Usage-Spec/FOCUS_Spec/blob/working_draft/specification/glossary.md?plain=1#L127-L129) and adding a link to this term in the very first sentence of the description.
2.  I like this sentence, "Tags may also be referred to ..." but after "Tags are commonly used ..."

cnharris10 (changes requested):
A couple of suggestions to help tie the description in the Tags definition with the Glossary.  Let me know your thoughts.

specification/columns/tags.md
AWS-ZachErdman: Great suggestions, thanks Chris! Tags was already in the glossary so I just removed it here. Also repositioned the sentence.

cnharris10 (approved):
Nice additions 👏

specification/columns/tags.md
```diff
@@ -1,6 +1,6 @@
 # Tags
 
-The Tags column represents the set of tags assigned to [*tag sources*](#glossary:tag-source) that also account for potential provider-defined or user-defined tag evaluations.  Tags are commonly used for scenarios like adding business context to billing data to identify and accurately allocate charges.
+The Tags column represents the set of tags assigned to [*tag sources*](#glossary:tag-source) that also account for potential provider-defined or user-defined tag evaluations. Tags are commonly used for scenarios like adding business context to billing data to identify and accurately allocate charges. Tags may also be referred to by providers by other terms, such as labels or networks tags.
```
kk09v, line 3:
```suggestion
The Tags column represents the set of tags assigned to [*tag sources*](#glossary:tag-source) that also account for potential provider-defined or user-defined tag evaluations. Tags are commonly used for scenarios like adding business context to billing data to identify and accurately allocate charges. Tags may also be referred to by providers using other terms such as labels.
```

Network tags are networking configurations, not tags as we had discussed. If a provider decides to include them, there's no harm, but as a practitioner, I would not expect that networking or security configurations (e.g. security marks, which are in a similar category) to be included in my consumption data.
https://cloud.google.com/vpc/docs/add-remove-network-tags

There is definitely a possibility of other terms for tags/labels, but I have yet to find one other than "annotations" which has a broader definition and would include the network tags/security marks listed above. GCP does have system_labels, but I still think about those as labels with system specifying their origin (similar to project.labels specifying labels on the project).

AWS-ZachErdman: Thanks Karl - I read through the link you provided and I'm still a little confused by network tags. I can see that they are very explicitly related to networking and would understand why they would have less use in a billing report, but all of the language does seem to refer to them pretty explicitly as "tags".

Nonetheless, I think it's fair to just remove them from this list and let GCP decide whether they want to include them in their exports :)

Makes sense about other names - I'm good with your suggestion and committed it!

udam-f2 (approved):

jpradocueva:
@kk09v, Karl, please check if you are satisfied with how your comments have been addressed. If you agree, please mark the PR as approved for the changes. This PR goes for R&A on Thursday, March 28.

jpradocueva:
Approved by the group on March 28.

flanakin (approved):

jpradocueva:
Approved by the group on April 4th

PR #369, opened 2024-03-23
updateing AZ requirements text

ahullah:
Change to make availability zone a mandatory field


cnharris10 (approved):

udam-f2:
@ahullah any reason why this is a separate PR? Already in the essentials PR?

ahullah:
I thought we agreed in the last meeting to revert the essentials pr to the original text so it removed one of the merge blockers...and to handle this separately as there were a number of us that thought this should be in the spec....???

ahullah:
> @ahullah any reason why this is a separate PR? Already in the essentials PR?

I thought we agreed in the last meeting to revert the essentials pr to the original text so it removed one of the merge blockers...and to handle this separately as there were a number of us that thought this should be in the spec....???

flanakin (changes requested):
I would request that we hold off on this until after GA. The original reason for leaving this as a SHOULD was because it doesn't impact cost and we didn't want it to block adoption. I feel the same holds true, especially in a world where we don't have compliance levels.

ahullah:
We are going to close this PR following discussion on 9/4/24 and revisit this post v1 GA with the expectation that AZ will become a MUST column for all providers that support the concept post v1 GA

ahullah:
Withdrawing PR

Issue #330, opened 2024-02-14
Charge and Pricing category cleanup


flanakin:
📅 _**NOTE:** This issue merges issues #305, #306, and #308 and will be discussed together in Task Force 2 on Wednesdays at 9pm UTC._

### Type

> Dimension: ChargeSubcategory, PricingSubcategory
> Normalized: Yes
>
> Related dimensions: ChargeCategory, PricingCategory

### Description

> This issue is consolidating change requests across 3 areas:
> - ChargeSubcategory values
> - Commitment utilization
> - Pricing categorization

#### ChargeSubcategory

> **Context:** [ChargeSubcategory](https://github.com/FinOps-Open-Cost-and-Usage-Spec/FOCUS_Spec/blob/working_draft/specification/columns/chargesubcategory.md) represents the highest-level breakdown of a specific type of charge.
> 
> **Definition:** _Charge Subcategory indicates what kind of usage or adjustment the row represents. Charge Subcategory is a supplemental detail to the Charge Category). It provides additional context to describe the primary category of charge._
>
> Scenarios:
>
> 1. ☑️ **What is the highest-level breakdown of usage charges?**
>    In 1.0 preview, we decided to indicate whether or not a commitment was applied and used as that highest-level breakdown. But is that right? Should there be a different breakdown of usage that is more about the usage itself than whether a commitment discount was applied or not? This is important for GA because if we agree that there is a better breakdown, we should move "Used Commitment" and "Unused Commitment" to a different column. We do not need to finalize what goes into ChargeSubcategory right now if we do this. But we should make that decision before GA.
>
>    **Definition:** _When Charge Category is "Usage" and the charge is related to a commitment, the Charge Subcategory indicates whether the row represents committed usage or is an amortized charge for the unused portion of the commitment. Charge Subcategory is commonly used for scenarios like calculating commitment utilization when Charge Category is "Usage"._
>
>    Example information hierarchy based on current ChargeSubcategory values:
>    * ChargeCategory = `Usage`
>      * ChargeSubcategory = `On-Demand`
>      * ChargeSubcategory = `Unused Commitment`
>      * ChargeSubcategory = `Used Commitment`
>        * PricingCategory = `Commitment-Based`
>
[...]

PR #382, opened 2024-04-03
FOCUS #330: CommitmentDiscountStatus column

flanakin:
Related to #330


thecloudman:
Commitment Status

specification/columns/commitmentdiscountusage.md
```diff
@@ -0,0 +1,38 @@
+# Commitment Discount Usage
```
cnharris10, line 1: Some others?
- Commitment Discount Application --> Applied, Not Applied
- Commitment Discount Amortized --> Amortized, Not Amortized

AWS-ZachErdman:
Did we consider just making these values in ChargeCategory? Is it helpful to create an entirely new column just for this breakdown?

specification/columns/columns.mdpp
```diff
@@ -19,6 +19,7 @@ The FOCUS specification defines a group of columns that provide qualitative valu
 !INCLUDE "commitmentdiscountid.md",1
 !INCLUDE "commitmentdiscountname.md",1
 !INCLUDE "commitmentdiscounttype.md",1
+!INCLUDE "commitmentdiscountusage.md",1
```
thecloudman, line 22: Suggested Column Name: **Commitment Status.**

Value Examples:

- Commitment Status > Used, Not Used
- Commitment Status > Used, Unused

specification/columns/commitmentdiscountusage.md
flanakin: "Application" might work, but we'd have to keep the values as is. We're not tracking whether a commitment discount was applied. We're tracking whether this portion of the commitment discount was utilized or not since the unutilized portion is effectively waste and lost savings potential.

Same applies for "Amortized" and "Not Amortized". Everything that has a value would be amortized.

flanakin: More options:
- Commitment Discount Status
- Commitment Discount Utilization
- Commitment Discount Consumption

I prefer "Status".

specification/columns/columns.mdpp
flanakin: We said we want to keep "Discount" in the name to make it specific to commitment discounts, so the only tweak I would suggest is "Commitment Discount Status". I do prefer this tho.

specification/columns/commitmentdiscountusage.md
```diff
@@ -0,0 +1,38 @@
...
+## Description
+
+Indicates whether the charge corresponds with the consumption of a *commitment-based discount* or the unused portion of the committed amount.
+
+## Content constraints
+
+| Constraint      | Value          |
+| :-------------- | :------------- |
+| Column type     | Dimension      |
+| Column required | True           |
```
flanakin, line 24:
```suggestion
| Feature level   | Conditional    |
```

specification/columns/commitmentdiscountusage.md
```diff
@@ -0,0 +1,38 @@
+# Commitment Discount Usage
+
+Commitment Discount Usage indicates whether the charge corresponds with the consumption of the [*commitment-based discount*](#glossary:commitment-based-discount) identified in the CommitmentDiscountId column or the unused portion of the committed amount.
+
+The CommitmentDiscountUsage column MUST be present in the billing data. This column MUST be of type String, MUST be null when [CommitmentDiscountId](#commitmentdiscountid) is null, and MUST NOT be null when CommitmentDiscountId is not null. The CommitmentDiscountCategory MUST be one of the allowed values.
```
flanakin, line 5:
```suggestion
The CommitmentDiscountUsage column SHOULD be present in the billing data when the provider supports *commitment-based discounts*. This column MUST be of type String, MUST be null when [CommitmentDiscountId](#commitmentdiscountid) is null, and MUST NOT be null when CommitmentDiscountId is not null. The CommitmentDiscountCategory MUST be one of the allowed values.
```

specification/columns/commitmentdiscountusage.md
ahullah: i will throw my vote behind status

ahullah: sorry 'Commitment Discount Status' to be exact

udam-f2:
4/11: The content here is agreed to by the members. The name of the column is up for discussion still. @flanakin will send out a poll and notify the members to get input on the suggestions that have been provided. Once the naming is agreed to, we are ready for merge.

specification/columns/commitmentdiscountusage.md
projectlocale: I agree with Commitment Discount Status. In particular I would suggest that if in a future change adding additional values becomes an issue of discussion such as the ones Chris suggested, Status will be more future proof (and may very well be needed).

projectlocale (approved):
With 1.0 it's too early to make changes that narrow this to something like Utilization simply because more values may need to be added in the future and we don't know two things:
1) The future values that will be needed. In which case a broader connotation works in our favor.
2) If we'll be under pressure to introduce subtly new concepts without bloating the columns. In which case we'll be grateful for the ontological flexibility to fan out the values more evenly.

Edit: Just clarifying that I'm referring to voting for Commitment Discount Status.

specification/columns/commitmentdiscountusage.md
AWS-ZachErdman: Do we think it's necessary for 'Discount' to be in the name?

I was hoping to be able to use "Used" and "Unused" for On-Demand Capacity Reservation (ODCR) line items ([AWS](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/ec2-capacity-reservations.html), [Azure](https://learn.microsoft.com/en-us/azure/virtual-machines/capacity-reservation-overview)) but ODCRs aren't typically associated with discounts so this column wouldn't make sense for that.

AWS-ZachErdman: Actually that suggestion was probably bad. We may want to be able to see a separate status for both Spend commitments and ODCRs since in AWS you can have an SP apply to an ODCR.

["Note that Savings Plans does not provide a capacity reservation. You can however reserve capacity with On Demand Capacity Reservations and pay lower prices on them with Savings Plans."](https://docs.aws.amazon.com/whitepapers/latest/cost-optimization-reservation-models/savings-plans.html)

So maybe you could have a line where an ODCR was "Unused" but an SP applied and was "Used"

AWS-ZachErdman:
> > Did we consider just making these values in ChargeCategory? Is it helpful to create an entirely new column just for this breakdown?
> 
> @AWS-ZachErdman, yes, we discussed this in the very beginning. Practitioners said they want to see the unused amount included when they filter ChargeCategory == "Usage".

Interesting, so this means to get the actual amount of service usage they would need to write a query like:

SELECT sum(UsageQuantity)
WHERE ChargeCategory = "Usage" AND CommitmentDiscountStatus = "Used Commitment"

specification/columns/commitmentdiscountusage.md
cnharris10: I'm fine with "Commitment Discount Status"

cnharris10 (approved):

flanakin:
> So this means to get the actual amount of service usage they would need to write a query like:
> 
> SELECT sum(UsageQuantity) WHERE ChargeCategory = "Usage" AND CommitmentDiscountStatus = "Used Commitment"

Technically, we don't need to look at ChargeCategory at all.

```sql
SELECT sum(UsageQuantity)
WHERE CommitmentDiscountStatus = "Used"
```

marc-perreaut (approved):
Concerning the remaining conversation about the name, both CommitmentDiscountUsage and CommitmentDiscountStatus are good to me.

Another proposition could be CommitmentDiscountPortion: the CommitmentDiscountPortion is the used one or the unused one.

specification/columns/commitmentdiscountusage.md
ijurica, line 5:
```suggestion
The CommitmentDiscountUsage column MUST be present in the billing data when the provider supports *commitment-based discounts*. This column MUST be of type String, MUST be null when [CommitmentDiscountId](#commitmentdiscountid) is null, and MUST NOT be null when CommitmentDiscountId is not null. The CommitmentDiscountCategory MUST be one of the allowed values.
```
Reminder: 
Noticed that a conversation regarding the Feature level constraint is marked as resolved, but the 'Column required' still present.

specification/attributes/discount_handling.md
```diff
@@ -31,13 +31,12 @@ Indicates how to include and apply discounts to usage charges or rows.
...
   * The EffectiveCost MUST include the portion of the amortized purchase cost that applies to this row.
-  * ChargeSubcategory MUST be "Used Commitment" for rows that received a reduced price from that commitment.
-  * If a commitment is not fully utilized, the provider MUST include a row that represents the unused portion of the commitment for that charge period. ChargeSubcategory MUST be "Unused Commitment".
-  * The sum of the EffectiveCost for all "Used Commitment" and "Unused Commitment" rows for each CommitmentDiscountId over the entire duration of the commitment MUST be the same as the total BilledCost of the commitment-based discount.
+  * CommitmentDiscountUsage MUST be "Used" for rows that received a reduced price from that commitment.
+  * If a commitment is not fully utilized, the provider MUST include a row that represents the unused portion of the commitment for that charge period. CommitmentDiscountUsage MUST be "Unused".
+  * The sum of the EffectiveCost for all rows where CommitmentDiscountUsage is "Used" or "Unused" for each CommitmentDiscountId over the entire duration of the commitment MUST be the same as the total BilledCost of the commitment-based discount.
```
flanakin, line 39:
```suggestion
  * CommitmentDiscountStatus MUST be "Used" for rows that received a reduced price from that commitment.
  * If a commitment is not fully utilized, the provider MUST include a row that represents the unused portion of the commitment for that charge period. CommitmentDiscountStatus MUST be "Unused".
  * The sum of the EffectiveCost for all rows where CommitmentDiscountStatus is "Used" or "Unused" for each CommitmentDiscountId over the entire duration of the commitment MUST be the same as the total BilledCost of the commitment-based discount.
```

specification/columns/columns.mdpp
```diff
@@ -20,6 +20,7 @@ The FOCUS specification defines a group of columns that provide qualitative valu
...
```
flanakin, line 23:
```suggestion
!INCLUDE "commitmentdiscountstatus.md",1
```

jpradocueva:
The group agreed to change the name to `CommitmentDiscountStatus`.
Approved by Group on April 18.

specification/use_case_library.md
```diff
@@ -36,7 +36,7 @@ The following table contains a set of commonly performed FinOps scenarios that w
...
 | FinOps Practitioner | Managing Commitment Based Discounts | As a FinOps Practitioner, I want to track all commitment based discounts purchased for a time period | Provider<br>BillingAccountID<br>CommitmentDiscountId<br>CommitmentDiscountType<br>BilledCost<br>ChargePeriodStart<br>ChargeCategory |
-| FinOps Practitioner | Managing Commitment Based Discounts | As a FinOps Practitioner, I want to track unused commitment charges in any given time period so that I consider them in my future commitment planning or remedy them | ChargeSubcategory (filter)<br>CommitmentDiscountID<br>BilledCost<br>ChargePeriodStart |
+| FinOps Practitioner | Managing Commitment Based Discounts | As a FinOps Practitioner, I want to track unused commitment charges in any given time period so that I consider them in my future commitment planning or remedy them | CommitmentDiscountUsage (filter)<br>CommitmentDiscountID<br>BilledCost<br>ChargePeriodStart |
```
flanakin, line 39:
```suggestion
| FinOps Practitioner | Managing Commitment Based Discounts | As a FinOps Practitioner, I want to track unused commitment charges in any given time period so that I consider them in my future commitment planning or remedy them | CommitmentDiscountStatus (filter)<br>CommitmentDiscountID<br>BilledCost<br>ChargePeriodStart |
```

specification/columns/commitmentdiscountusage.md
flanakin: Changed to CommitmentDiscountStatus

specification/columns/commitmentdiscountstatus.md
```diff
@@ -0,0 +1,38 @@
+# Commitment Discount Status
+
+Commitment Discount Status indicates whether the charge corresponds with the consumption of the [*commitment-based discount*](#glossary:commitment-based-discount) identified in the CommitmentDiscountId column or the unused portion of the committed amount.
+
+The CommitmentDiscountStatus column MUST be present in the billing data. This column MUST be of type String, MUST be null when [CommitmentDiscountId](#commitmentdiscountid) is null, and MUST NOT be null when CommitmentDiscountId is not null. The CommitmentDiscountCategory MUST be one of the allowed values.
```
flanakin, line 5:
```suggestion
The CommitmentDiscountStatus column MUST be present in the billing data when the provider supports *commitment-based discounts*. This column MUST be of type String, MUST be null when [CommitmentDiscountId](#commitmentdiscountid) is null, and MUST NOT be null when CommitmentDiscountId is not null. The CommitmentDiscountCategory MUST be one of the allowed values.
```

specification/columns/commitmentdiscountstatus.md
```diff
@@ -0,0 +1,38 @@
...
+## Description
+
+Indicates whether the charge corresponds with the consumption of a *commitment-based discount* or the unused portion of the committed amount.
+
+## Content constraints
+
+| Constraint      | Value          |
+| :-------------- | :------------- |
+| Column type     | Dimension      |
+| Column required | True           |
```
flanakin, line 24:
```suggestion
| Feature level   | Conditional    |
```

udam-f2 (approved):

PR #384, opened 2024-04-04
Update 1.0 references to 1.0-preview

flanakin:
While we still need to decide whether to reference pre-release versions in the spec, the current spec already does and many of them are wrong. This PR fixes those that are wrong to at least be accurate. I figure we can have a separate conversation about excluding pre-release versions and update everything based on that. This PR only fixes what's currently wrong in the spec today and isn't making a statement about whether or not pre-release versions should be in the spec.


ahullah (approved):

udam-f2:
4/11: Member group reviewed and approved this for check-in. There's concern about referencing unreleased/RC versions in spec documentation. It mentioned that this would be discussed later.

mike-finopsorg (approved):

PR #390, opened 2024-04-09
Metadata structure simple

mike-finopsorg:
Metadata structure simplified for Version 1.0 specification
Moved examples to supporting content for now
Added normative text to schema items
Use structure more similar to columns / attributes


mike-finopsorg:
[spec.pdf](https://github.com/FinOps-Open-Cost-and-Usage-Spec/FOCUS_Spec/files/14911937/spec.pdf)
Added generated spec file for easier reference to how this will look in the document ^^

specification/metadata/providerprefix.md
```diff
@@ -0,0 +1,17 @@
+# Provider Prefix
+
+Used as the prefix for the name of custom columns, useful for the consumer of the data to quickly identify custom column extensions to the FOCUS schema implemented by the provider.
```
ijurica, line 3: REMINDER: 
align with/adjust the following column naming conventions requirement:

> All custom columns MUST be prefixed with a consistent x_ prefix to identify them as external, custom columns and distinguish them from FOCUS columns to avoid conflicts in future releases.

udam-f2: I feel it's worth discussing if x_ is better than $ProviderPrefix$_ (now that we have a way to get that).

specification/metadata/providertagprefix.md
```diff
@@ -0,0 +1,17 @@
+# Provider Tag Prefix
+
+Prefix for Provider-Defined tags in the FOCUS dataset. Useful for the consumer to identify which tags are Provider-Defined vs User-Defined.
```
udam-f2, line 3:
```suggestion
The Provider Tag Prefix defines the prefix used in the tag name of provider-defined [tags](#tags). This metadata is useful for the consumer to identify which tags are provider-defined vs user-defined.
```

specification/metadata/providertagprefix.md
```diff
@@ -0,0 +1,17 @@
...
+
+The ProviderTagPrefix SHOULD be provided when the provider supports tags and provides Provider-Defined tags. The ProviderTagPrefix MUST be of type String. The ProviderTagPrefix SHOULD be easily associated with the provider who generated the FOCUS dataset.
```
udam-f2, line 5:
```suggestion
The ProviderTagPrefix SHOULD be provided when the provider supports provider-defined tags. The ProviderTagPrefix MUST be of type String. The ProviderTagPrefix SHOULD be easily associated with the provider who generated the FOCUS dataset.
```

specification/metadata/providerprefix.md
udam-f2, line 3: Do we need two different things here OR would it be simpler for practitioners if providers just used one value for both? One downside I can think of is if there's more than one value for tag prefixes.

@rileyjenk @mike-finopsorg

specification/metadata/providertagprefix.md
```diff
@@ -0,0 +1,17 @@
+# Provider Tag Prefix
```
udam-f2, line 1: Are there scenarios where this is a set of values that providers may use as prefixes? Supporting an array gives more flexibility. We could say its a comma separated string if > 1 and keep it string.

specification/metadata/focusversion.md
```diff
@@ -0,0 +1,17 @@
+# FOCUS Version
+
+The declared FOCUS version of the FOCUS data within the schema.
```
udam-f2, line 3:
```suggestion
The version of FOCUS utilized for building the dataset.
```

specification/metadata/creationdate.md
```diff
@@ -0,0 +1,17 @@
+# Creation Date
+
+Date the schema was created.
+
+The CreationDate MUST be present in the metadata. This MUST be of type Date/Time and MUST NOT contain null values. CreationDate MUST conform to [Date/Time Format](#date/timeformat).
```
udam-f2, line 5: Leave as a SHOULD? In the case of GCP, they have a column for this in their actual data.

specification/metadata/schema/datatype.md
```diff
@@ -0,0 +1,17 @@
+# Data Type
+
+The data type of the column provided in the FOCUS dataset.
```
udam-f2, line 3: I think we had a discussion that this needs to be the specific datatype used by the provider (not high level data type like what FOCUS defines - e.g. String/Numeric etc.)

specification/metadata/schema/schema.mdpp
```diff
@@ -0,0 +1,15 @@
+
+# Schema Definition
+
+The FOCUS metadata schema definition provides a list of the columns present in the FOCUS dataset along with metadata about the columns. The intent of the schema metadata is to aid pracitioners to understand the structure of the FOCUS data, without the need to parse the included dataset.
+
+# Schema Definition
+
+The FOCUS metadata schema definition provides a list of the columns present in the FOCUS dataset along with metadata about the columns. The intent of the schema metadata is to aid pracitioners to understand the structure of the FOCUS data, without the need to parse the included dataset.
+
```
udam-f2, line 9:
```suggestion
```

specification/metadata/schemaid.md
```diff
@@ -0,0 +1,17 @@
+# Schema ID
+
+Unique identifier for the schema.
```
udam-f2, line 3: Too vague

specification/metadata/focusversion.md
udam-f2: Here I think we need to be mindful that it's not necessarily conveying compliance. I assume many providers will likely not be 100% compliant but will provide data starting with a particular FOCUS version and incrementally add support.

udam-f2 (commented):
In ProviderTagPrefix, I've added a bit more text to make this read more like the rest of the spec. If you feel it's useful, feel free to make the updates to the rest.

specification/metadata/metadata.mdpp
```diff
@@ -0,0 +1,11 @@
+# Metadata
+
+The FOCUS specification defines a metadata structure that is to be supplied by data providers to facilitate practioners in data ingestion, data locations, schema of the FOCUS dataset.
```
marc-perreaut, line 3: > is to be supplied

MAY, SHOULD or MUST?

I would vote for a MUST.

specification/metadata/billinggenerator.md
```diff
@@ -0,0 +1,17 @@
+# Billing Generator
```
marc-perreaut, line 1: Maybe Data Generator, or even simply FOCUS Generator?

FOCUS relates to Cost and Usage (not billing).

specification/metadata/creationdate.md
marc-perreaut: There is value for practitioners to get this information in a unique way for all providers.
Moreover, it's more efficient for this information to be at metadata level rather than at data level (same info for all rows of the generated dataset).
Practitioners may add this info as a column in their datasets at ingestion time.

specification/metadata/focusversion.md
```diff
@@ -0,0 +1,17 @@
...
+
+The FOCUSVersion MUST be provided in the metadata. FOCUSVersion MUST be of type String and MUST NOT contain null values. FOCUSVersion MUST match one of the published versions of the FOCUS specification. FOCUSVersion MUST match the version of the FOCUS specification that the FOCUS dataset conforms to.
+
+## Metadata ID
+
+FOCUSVersion
```
marc-perreaut, line 9: FocusVersion?

Same rationale as for SkuId, SkuPriceId

specification/metadata/providerprefix.md
marc-perreaut: https://github.com/FinOps-Open-Cost-and-Usage-Spec/FOCUS_Spec/pull/360/files#r1560050489

specification/metadata/providertagprefix.md
marc-perreaut: Maybe introduce a list type attribute (JSON array)?, which would be here an array of strings.

specification/metadata/schema/encoding.md
```diff
@@ -0,0 +1,17 @@
+# Encoding
```
marc-perreaut, line 1: Maybe String Encoding, to be clear and explicit?

specification/metadata/schema/precision.md
```diff
@@ -0,0 +1,17 @@
+# Precision
```
marc-perreaut, line 1: Maybe Number Precision, to be clear and explicit?

specification/metadata/schema/scale.md
```diff
@@ -0,0 +1,17 @@
+# Scale
```
marc-perreaut, line 1: Maybe Number Scale, to be clear and explicit?

specification/metadata/providerprefix.md
```diff
@@ -0,0 +1,17 @@
+# Provider Prefix
```
marc-perreaut, line 1: Maybe Provider Column Prefix, to be clear and explicit?

specification/metadata/providerprefix.md
marc-perreaut, line 3: Conceptually, the provider prefix belongs to the provider, not to the data generator: two data generators for the same provider should have the same prefix.
Maybe we should not go as far as splitting provider metadata and generator metadata, but maybe the specification could say that the provider prefix must be the same for all data generators addressing the same provider.

specification/metadata/schemaid.md
marc-perreaut: Does the schema belong to the provider or to the data generator?

I would go the simple path where we don't split the metadata between provider and data generator: there would be only data generator metadata, and then it would be the data generator schema.

specification/metadata/creationdate.md
```diff
@@ -0,0 +1,17 @@
+# Creation Date
+
+Date the schema was created.
```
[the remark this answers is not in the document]
mike-finopsorg:
```suggestion
Date the related FOCUS dataset was created.
```

specification/metadata/billinggenerator.md
flanakin: +1
"Data Generator" will be more forward-looking as we expand FOCUS to other datasets.

specification/metadata/creationdate.md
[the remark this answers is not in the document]
flanakin: If it is, should we use "Date Generated" to align with the "generator" column?

specification/metadata/creationdate.md
flanakin: I would go for a MUST, regardless of whether this is included in the dataset.

specification/metadata/focusversion.md
flanakin: +1

specification/metadata/focusversion.md
```diff
@@ -0,0 +1,17 @@
+# FOCUS Version
```
flanakin, line 1: We also need a DataGeneratorVersion (or something like that) to account for data generators creating multiple versions of their support of a specific FOCUS version (e.g., with new custom columns).

specification/metadata/providertagprefix.md
flanakin: I could definitely see having multiple tag prefixes and would like to see that accounted for.

specification/metadata/schema/datatype.md
flanakin: I would like to see this with a set of allowed values.

specification/metadata/schema/precision.md
flanakin: "Numeric Precision"?

specification/metadata/schema/scale.md
flanakin: I thought "precision" was the number of decimal places. Scale sounds like ones vs. thousands vs millions.

specification/metadata/schemaid.md
flanakin: Is this covered by providing a version of the schema? I was thinking of that in the metadata section.

specification/metadata/focusversion.md
rileyjenk: Yeah that is why I have this as a metadata signifying the version declared in that schema. If this is saying that they are following a specific version in general, the problem is that how does the practitioner know the version for older data. That is why I have it attached to the schema itself.

specification/metadata/creationdate.md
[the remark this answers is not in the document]
mike-finopsorg: Talking to @rileyjenk the intent was to have a date for when the metadata was created. I guess if the metadata is not provided 1:1 with the dataset then this date might not make sense in the format of data generation.

specification/metadata/providerprefix.md
mike-finopsorg: oh I need to re-read that section I had thought we went with a $ProviderPrefix$_. Let's decide if we are going to reopen this conversation or just align this to the current x_ at this weeks maintainers meeting.

mike-finopsorg: I thought about this then I was worried the value that a provider has used in tags (e.g. `AWS:`) isn't 1:1 to a provider's chosen column prefix (e.g. `aws_`). I guess we can discuss this further when we close out on if we are going to stick with x_ or move towards $ProviderPrefix$_. I am guessing this is possibly a significant conversation to reopen at this stage in the game and it might resolve this indirectly.

specification/metadata/providertagprefix.md
mike-finopsorg: Good point, I will make a suggestion for this.

specification/metadata/schema/datatype.md
mike-finopsorg: Open to suggestions here

specification/metadata/metadata.mdpp
mike-finopsorg:
```suggestion
The FOCUS specification defines a metadata structure that MUST be supplied by data providers to facilitate practitioners in data ingestion, data locations, schema of the FOCUS dataset.
```

specification/metadata/providerprefix.md
mike-finopsorg: Updated

specification/metadata/billinggenerator.md
mike-finopsorg: Updated

specification/metadata/focusversion.md
mike-finopsorg: Updated

specification/metadata/focusversion.md
mike-finopsorg: Added DataGeneratorVersion

specification/metadata/schema/scale.md
mike-finopsorg: I had to check it also.

"Precision is the number of digits in a number. Scale is the number of digits to the right of the decimal point in a number" - https://learn.microsoft.com/en-us/sql/t-sql/data-types/precision-scale-and-length-transact-sql?view=sql-server-ver16

specification/metadata/metadata.mdpp
mike-finopsorg: @rileyjenk added some of your entity conversation here. Please feel free to add more suggestions.

projectlocale (approved):
Nothing to add on this one. Any of the outcomes discussed are fine as long as the data type is defined and never null. Expanding the requirement to include allowed values is probably the right move, but I wonder if this is going to be the sort of thing that leads to restricting what an allowed value should be. I would suggest keeping this as a MUST and considering a should as more information becomes available and there is time to tighten up the definition in future revisions. That way adding should guarantees that in future versions the spec doesn't break if the scope of this is re-examined.

specification/metadata/creationdate.md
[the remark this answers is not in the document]
mike-finopsorg: Confirmed this is intended to be a date for the creation of this schema.

specification/metadata/schema/datatype.md
mike-finopsorg: Im struggling to understand how we can have this as specific and an allowable list of values.

specification/metadata/schemaid.md
mike-finopsorg: We need to decide between using SchemaID or DataGeneratorVersion as the identifier. I don't think we should have both and Im leaning towards SchemaID with more of the text pulled in from DataGeneratorVersion.  cc:// @rileyjenk

specification/metadata/schema/datatype.md
marc-perreaut: Maybe the data type reference of the provider (BigQuery, MS SQL, ...) could be specified as a column in metadata: allowed values for column data types would be data types in this reference.

specification/metadata/datageneratorversion.md
```diff
@@ -0,0 +1,17 @@
+# Data Generator Version
+
+A version specific to the data generator for providers which support multiple FOCUS datasets with different provider columns, or for providers who release multiple versions of their generated datasets between FOCUS Versions.
```
marc-perreaut, line 3: I understand that a data generator would have different options to include or not different columns in the dataset, and that the generator version would include (reflect) the usage of these options. Is that the intention?

If yes, wouldn't it be better to have separate columns for the version (released software version) of the data generator and for its usage (options used when calling the generator)?

specification/metadata/metadata.mdpp
```diff
@@ -0,0 +1,6 @@
+# Metadata
+
+The FOCUS specification defines a metadata structure that is to be supplied by data providers to facilitate practioners use of FOCUS data. This meta data includes general information about the data generator and the schema of the FOCUS dataset. FOCUS Metadata SHOULD be provided in a format that is accessible programmatically, such as: a file, website, api, table.
```
marc-perreaut, line 3: typo: practioner

supporting_content/metadata/metadata.md
```diff
@@ -0,0 +1,131 @@
...
+#### Example Request:
+
+    Method: GET 
+    Endpoint : <api_root>/FOCUS/metadata/data_generator
+####
+
+#### Response
+```
+{
+	"DataGenerator": "awesome_corp",
```
marc-perreaut, line 17: Add DataGeneratorVersion?

specification/metadata/datageneratorversion.md
```diff
@@ -0,0 +1,17 @@
...
+
+The DataGeneratorVersion SHOULD be provided in the metadata, when multiple datasets using the same FOCUS version are provided with different provider columns. DataGeneratorVersion MUST be of type String and MUST NOT contain null values.
```
marc-perreaut, line 5: I would vote for a "MUST be provided", because it's needed for practitioners to identify different versions, for example in case of bug fix.

supporting_content/metadata/metadata.md
marc-perreaut: Add also the Schema ID, which is used by this DataGeneratorVersion?

supporting_content/metadata/metadata.md
```diff
@@ -0,0 +1,131 @@
...
+                     NumericPrecision: 20,
+                     NumberScale: 10
+                 },
+         ]
+}`
+```
+
+## Example Schema Reference Metadata
+
+In this example, when the provider returns the FOCUS data they include in the response a reference to the schema utilized in the FOCUS dataset.
```
marc-perreaut, line 116: Could be also useful to include the DataGeneratorVersion used to generate this FOCUS dataset?

specification/metadata/metadata.mdpp
ijurica: Question: How are billing data sets and the corresponding metadata correlated?

specification/metadata/datageneratorversion.md
mike-finopsorg: Im not following your suggestion. Can you please be more concrete with a suggested edit

specification/metadata/metadata.mdpp
mike-finopsorg: We spoke about this in detail on the TF2 call today. At this stage we are not being prescriptive of the correlation approach but we do see this as a future item for discussion.

specification/metadata/datageneratorversion.md
mike-finopsorg:
```suggestion
The DataGeneratorVersion MUST be provided in the metadata, when multiple datasets using the same FOCUS version are provided with different provider columns. DataGeneratorVersion MUST be of type String and MUST NOT contain null values.
```

specification/metadata/metadata.mdpp
mike-finopsorg:
```suggestion
The FOCUS specification defines a metadata structure that is to be supplied by data providers to facilitate practitioners use of FOCUS data. This meta data includes general information about the data generator and the schema of the FOCUS dataset. FOCUS Metadata SHOULD be provided in a format that is accessible programmatically, such as: a file, website, api, table.
```

supporting_content/metadata/metadata.md
```diff
@@ -0,0 +1,127 @@
...
+}`
+```
+
+## Example Schema Metadata
+
+In this example, the billing data includes two different structures of data. An older schema for data previously provided, and a newer schema for more recent data. Both schemas metata is provided. 
+
+### API
+
+#### Endpoint: <api_root>/FOCUS/metadata/schemas
```
marc-perreaut, line 29: As the concept of Schema ID has been removed, in favor of keeping only the Data Generator Version, I wonder whether an API would rather list the Data Generator Versions and the associated schemas, than list the Schemas.

specification/metadata/schemaid.md
marc-perreaut: We might need both Data Generator Versions and Schema ID: there can be different Data Generator Versions for the same Schema ID, for example if there are bug fixes in the Data Generator.

From practitioner perspective:
- The Schema ID is important for data ingestion to manage column changes.
- The Data Generator Version is important to spot potential data issues that can be linked to different Data Generator Versions.

marc-perreaut: If one was to choose between Schema ID and Data Generator Version, I would vote for the Schema ID, as this is directly impacting the data ingestion. The Data Generator Version is more software management and can be managed outside FOCUS metadata.

specification/metadata/datageneratorversion.md
marc-perreaut: Maybe simply:

> A version number that identifies a unique state (or release?) of the Data Generator. Versioning enables practitioners to identify which FOCUS version is supported by which Data Generator Version. Multiple Data Generator Versions may support the same FOCUS version, corresponding to different releases to fix bugs or to introduce additional provider columns.

Maybe also:
> The Data Generator Version SHOULD use semantic versioning. 

NOTE: I would tend to think the Data Generator Version could be removed in favor of the Schema ID: https://github.com/FinOps-Open-Cost-and-Usage-Spec/FOCUS_Spec/pull/390#discussion_r1570240393

specification/metadata/data_generator/datageneratorversion.md
```diff
@@ -0,0 +1,17 @@
+# Data Generator Version
```
rileyjenk, line 1: This construct doesn't work. How do they reference the schema as this is not the schema id. The data generator version is insufficient to reference the schema. This change breaks the schema's metadata usefullness.

rileyjenk (changes requested):

supporting_content/metadata/metadata.md
rileyjenk: The dropping of schemaId breaks the usefullnes of the schema metadata. How is data generator version going to help. It is a version of what? A version of AWS?

jpradocueva:
Approved by Group on April 18.
An Issue will be created to bring back SchameID.

specification/metadata/data_generator/datageneratorversion.md
mike-finopsorg: Agreed to merge this PR and open a related issue for this discussion

supporting_content/metadata/metadata.md
mike-finopsorg: Agreed to merge this PR and open a related issue for this discussion

supporting_content/metadata/metadata.md
mike-finopsorg: Agreed to merge this PR and open a related issue for the version vs schemaId discussion

udam-f2 (approved):

specification/metadata/schemaid.md
mike-finopsorg: Opening related PR to discuss this further

rileyjenk (approved):
I am removing my request for changes as discussed in our last meeting. Further modifications will be handled in a separate PR.

PR #410, opened 2024-04-18
FOCUS #330: Update chargeclass value from "Regular" to "Standard"

thecloudman:
Pricing category had the value "on-demand" changed to "Standard", so recommend keeping consistency in naming and have charge class value as "Standard" instead of "Regular"

### DIFF
--- a/specification/columns/chargeclass.md
+++ b/specification/columns/chargeclass.md
@@ -30,7 +30,7 @@ Allowed values:
 
 | Value      | Description                          |
 | :--------- | :------------------------------------|
-| Regular    | Standard charges for services used or purchased. |
+| Standard    | Standard charges for services used or purchased. |
 | Correction  | Modification to one or more previous charges, like refunds and credit modifications. |
 
 ## Introduced (version)


ahullah:
I don't have an objection here... makes sense to improve consistency

ahullah (approved):

flanakin (approved):
I feel like there's a more intentional word, but this is an improvement to "Regular". That always felt like a placeholder until we found something more appropriate.

mike-finopsorg (approved):

udam-f2:
I believe @marc-perreaut was going to suggest 'Original' for which there was a good amount of support. His logic was it suggests this is the original charge as opposed to the 'correction' that comes after the fact. He mentioned this was already added as a comment in the PR but I'm not sure I see it - adding as a comment to not lose the suggestion. 

FYI @mike-finopsorg @flanakin @ahullah

marc-perreaut:
> This morning in TF1 @marc-perreaut suggested 'Original' for which there was a good amount of support. His logic was it suggests this is the original charge as opposed to the 'correction' that comes after the fact. He mentioned this was already added as a comment in the PR but I'm not sure I see it - adding as a comment to not lose the suggestion.
> 
> FYI @mike-finopsorg @flanakin @ahullah

Sorry, I actually commented in other PR #412
https://github.com/FinOps-Open-Cost-and-Usage-Spec/FOCUS_Spec/pull/412/files#r1576446160

Also maybe a suggestion for a more explicit name for ChargeClass could be ChargeOccurrence.
Indeed, a charge occurrence may be either the original occurrence or a correction occurrence.

ahullah:
Ok I'm just trying to reconcile that with the proposed 'percent' category we were thinking of adding to charge class to accommodate services that are not charged on a unit rate..... Actually thinking about this should original/regular actually be 'unit' as we are normality accounting for services consumed in units....???

ahullah:
Apologies I merged this too soon will submit a separate PR to incorporate @marc-perreaut comment and the TF1 discussion to move this column to Original

ahullah:
follow on update to 'Original' now in PR #422

Issue #366, opened 2024-03-22
SkuId, SkuPriceId, and Pricing-related columns review due to ChargeCategory and ChargeClass cleanup


ahullah:
### Description
We need to clarify what default values are expected in List, Contracted, Billed & Effective costs in the event that a service provider does not have a publicly published price book

### Proposed approach
This proposal will discuss incorporating appropriate default values in these metrics to ensure data is not 'skewed' when aggregating across multiple providers

### Github issue or Reference
picks up from discussion in PR #331

ahullah:
Including Discussion from PR

[specification/columns/listcost.md](https://github.com/FinOps-Open-Cost-and-Usage-Spec/FOCUS_Spec/pull/331/files#diff-b7c0c9eeddf417d9d4a5f582e25af2646b81defd4dc4be6587bcd6f9be76c827)
@@ -26,7 +26,7 @@ Cost calculated by multiplying List Unit Price and the corresponding Pricing Qua
| Constraint      | Value                   |
|:----------------|:------------------------|
| Column type     | Metric                  |
| Column required | True                    |
| FOCUS Essential | True                    |
Member
@[rileyjenk](https://github.com/rileyjenk) rileyjenk [2 weeks ago](https://github.com/FinOps-Open-Cost-and-Usage-Spec/FOCUS_Spec/pull/331#discussion_r1513230022)
Does this require that the vendor publish a list cost...though that may not actually be published? I see this as a stumbling block for many vendors to adopt even single line item focus data.

Member
@[rileyjenk](https://github.com/rileyjenk) rileyjenk [2 weeks ago](https://github.com/FinOps-Open-Cost-and-Usage-Spec/FOCUS_Spec/pull/331#discussion_r1513235714)
Question answered in discussion today. @ijurica perhaps we ensure it is clear that in the absence of a list price it matches BilledCost.

Member
Author
@[udam-f2](https://github.com/udam-f2) udam-f2 [2 weeks ago](https://github.com/FinOps-Open-Cost-and-Usage-Spec/FOCUS_Spec/pull/331#discussion_r1513254449)
@rileyjenk What's the scenario where a provider doesn't have a list cost? Billed cost is based off of ListCost + discounts. Or are you suggesting that providers may not want to publish the List Cost and also make it available in the billing data to keep that private?

Member
[...]

PR #412, opened 2024-04-19
v1.x: Adding % derived charge identification to Charge Class

ahullah:
Adding % derived charge identification to Charge Class.... This was discussed as a helpful addition to PR #366


specification/columns/chargeclass.md
```diff
@@ -1,6 +1,6 @@
 # Charge Class
 
-A Charge Class indicates whether the row represents a regular charge or a correction to one or more previous charges. Charge Class is commonly used to differentiate refunds from regularly incurred charges.
+A Charge Class indicates whether the row represents a regular charge unit based usage charge, a derivative percentage charged based on usage of one or more otehr services or a correction to one or more previous charges. Charge Class is commonly used to differentiate refunds from regularly incurred charges.
```
flanakin, line 3: I'm not quite following. Is this trying to differentiate an independent quantity vs. a percentage-based quantity? If it's percentage-based, would that include taxes? If it's percentage-based, how can we see what the percentage is? I think I like this conceptually but haven't fully grokked it.

FYI - Typo in "otehr services".

specification/columns/chargeclass.md
```diff
@@ -32,6 +32,7 @@ Allowed values:
 | :--------- | :------------------------------------|
 | Regular    | Standard charges for services used or purchased. |
```
flanakin, line 33: We should change "Regular" if we're differentiating it from charges that are based on a percentage of other charges. Would "Direct" make sense?

specification/columns/chargeclass.md
ijurica: I agree that we need to think through this more and provide more precise descriptions for each value and column description.

Taxes would fall under this additional ChargeClass category. Based on the discussions at TF-1/TF-2, in the case of tax-related charges, SkuId, SkuPriceId, PricingQuantity, etc., MUST NOT be provided. The reasoning behind this decision is that specifying what's expected in those columns for tax-related charges would introduce more confusion than value. Additionally, some providers are not providing those charge records at all. 
So, regarding taxes, this would simply indicate that it's not a standalone charge but a percentage of the cost of one or multiple Regular (or Direct) charges, which may not necessarily be of the same SKU.

So, for taxes, this would simply provide the indication that it's not a standalone charge but a percentage of the cost of one ore multiple Regular (or Direct) charges. 

A column providing information about the actual percentage (%) is still on our TODO list, and I assume it's part of the 1.x scope.

As Riley pointed out, there are Percentage charges based on one or multiple charges related to the same SkuId and SkuPriceId. However, I'm not sure what should be provided in PricingQuantity and UsageQuantity. This is another use case that we need to think through and discuss further...

specification/columns/chargeclass.md
ijurica: I like the term Direct

specification/columns/chargeclass.md
```diff
@@ -32,6 +32,7 @@ Allowed values:
 | :--------- | :------------------------------------|
 | Regular    | Standard charges for services used or purchased. |
 | Correction  | Modification to one or more previous charges, like refunds and credit modifications. |
+| Percentage  | Denotes a charge that is billed NOT based on direct unit consumption but is based on a % cost of another service e.g. Support Costs |
```
marc-perreaut, line 35: May a charge be Regular or Correction and also based on a Percentage?
Shouldn't Percentage be another column?

specification/columns/chargeclass.md
marc-perreaut: Another proposition (brainstorming) for Regular: Original.

Indeed, a charge maybe be either the original charge or a correction charge.

marc-perreaut: Thinking further, ChargeOccurence might be more explicit than ChargeClass.

Indeed, a charge occurence may be either the original occurrence or a correction occurence.

jpradocueva:
TF1 agreed to move this topic to v1.x on Apr 23.

udam-f2:
> TF1 agreed to move this topic to v1.x on Apr 23.

There was debate about if this fits at the same level original/standard and correction. Those seem like very high-level categorizations and % based seems more like how your charge was priced (and feels like something not mutually exclusive when compared to the existing two. Feels like this is something that should be discussed as a possible pricing category value?

jpradocueva:
@irena, TF-2 agreed to create a new Issue and close this PR.

Issue #409, opened 2024-04-18
UsageQuantity/Unit should be required based on PricingQuantity differentiation


flanakin:
### Type
> Dimension: UsageQuantity

### Description
> The UsageQuantity column requirements say it is only required if the provider supports the "measurement of usage":
>
> > _UsageQuantity column MUST be present in the billing data when the provider supports the measurement of usage._
>
> This means someone could technically have variable, usage-based charges without actually conveying how much was used. This feels wrong. I believe we should change this to be dependent on differentiation from PricingQuantity and PricingUnit. Maybe something like:
>
> > UsageQuantity column MUST be present in the billing data when the provider **supports charges where UsageQuantity may not match the PricingQuantity**.
>
> UsageUnit has the same condition. That should be updated to be based on UsageQuantity to be more clear:
>
> > UsageUnit column MUST be present in the billing data when **UsageQuantity is present**.

### Definition of done

- [x] Complete spec template and include naming (code name, display name), constraints, guidelines, compatibility with major providers etc.

jpradocueva:
The group suggested adding examples for both options.

jpradocueva:
The group had a lengthy discussion and, at the end, decided:
To create two PRs:
* One which explicitly locks UsageQ to usage lines, and
* The other which just keeps it open to purchase lines but updates some language to avoid “measurement of usage”.

PR #413, opened 2024-04-22
FOCUS #366: SkuId, SkuPriceId, and Pricing-related columns review due to ChargeCategory and ChargeClass cleanup

ijurica:
ChargeCategory and ChargeClass cleanup impacted multiple columns. Based on current ChargeCategory references, the following columns must be revisited:

- SkuId
- SkuPriceId
- ContractedUnitPrice
- ListUnitPrice
- Pricingcategory
- PricingUnit
- PricingQuantity
- UsageQuantity
- UsageUnit
- ListCost
- ContractedCost
- EffectiveCost

PR #418, opened 2024-04-25
#409 Path 1 - Usage Unit/Quantity restricted to usage charges and renamed to Consumed Unit/Quantity

aqu-erp:
PR 1 of 2

Based on the conversation today, it was agreed that (non-pricing) quantities pertaining to purchase or SaaS charges need additional development. 

The changes in this PR to UsageQuantity and UsageUnit definitions intend to:

- Rename the column to Consumed Quantity and ConsumedUnit
- Narrow the application of UsageQuantity and UsageUnit to usage charges
- Retain flexibility to revisit quantities related to purchase charges
- Mitigate unclean or irrelevant values in the interim and allow future additive changes


ijurica:
must be aligned with the PR #413

ijurica (approved):

AWS-ZachErdman:
I prefer this path as I think UsageQuantity is a fundamental FinOps concept and should reflect the amount of a resource consumed

udam-f2:
For the record, I prefer this option (path 1) as well. UsageQuantity describes something very clear - 'usage' and going with other more general names like 'ActualQuantity', 'MeasurementQuantity', or another one of the proposed options - to either provide distinct quantities OR to capture usage OR purchase quantities in a single column is not a net benefit as it confuses what the value means.

If the group feels strongly that the new option is better, I'm ok to go with it.

specification/columns/usagequantity.md
```diff
@@ -1,8 +1,8 @@
 # Usage Quantity
 
-The Usage Quantity represents the volume of a given SKU associated with a [*resource*](#glossary:resource) or [*service*](#glossary:service) used or purchased, based on the [Usage Unit](#usageunit). Usage Quantity is often derived at a finer granularity or over a different time interval when compared to the [Pricing Quantity](#pricingquantity) (complementary to [Pricing Unit](#pricingunit)) and focuses on *resource* and *service* consumption, not pricing and cost.
+The Usage Quantity represents the volume of a given SKU associated with a [*resource*](#glossary:resource) or [*service*](#glossary:service) used, based on the [Usage Unit](#usageunit). Usage Quantity is often derived at a finer granularity or over a different time interval when compared to the [Pricing Quantity](#pricingquantity) (complementary to [Pricing Unit](#pricingunit)) and focuses on *resource* and *service* consumption, not pricing and cost.
```
marc-perreaut, line 3:
```suggestion
The Usage Quantity represents the volume of a given SKU associated with a [*resource*](#glossary:resource) or [*service*](#glossary:service) used, based on the [Usage Unit](#usageunit). Usage Quantity is often derived at a finer granularity or over a different time interval when compared to the [Pricing Quantity](#pricingquantity) (complementary to [Pricing Unit](#pricingunit)) and focuses on *resource* and *service* consumption, not pricing.
```
As a charge is about cost, I would just remove "and cost".
Though, that was part of the original text, so might not be removed.

udam-f2:
> I am not fundamentally against this path, but I find it unfortunate to have these columns dedicated to ChargeCategory=Usage, while it could serve also use cases like Purchase, and so further similar columns might be introduced.
> 
> I agree the names UsageUnit / UsageQuantity can be misleading if the columns apply for other use cases than ChargeCategory=Usage, but I would prefer to go the path of finding a better name for the column.
> 
> I understand the intent of these columns is to get more standardized practitioner-friendly data than the PricingUnit / PricingQuantity, so in the end the columns work in pair (null if and only if the other is null). Maybe to make things simple and clear, the column names could be StandardizedPricingUnit / StandardizedPricingQuantity.

Hey @marc-perreaut, there was a long discussion/debate about this over many meetings. Ultimately, we seemed to have agreed on coming up with a separate column that either represents what you are charged for in distinct units OR a 'factor' column that can help remove the 'block' portion - which would apply to usage/purchase and possibly other ChargeCategory values. This will be discussed and introduced post v1.0. The decision that impacted v1.0 was to limit UsageQuantity to 'Usage' only and rename it to ConsumedQuantity. I've updated the PR to reflect the decision.

The worksheet that helped get agreement on the need to not merge both usage and purchase is here: https://docs.google.com/spreadsheets/d/1zA0brhrEntfWlzt5VNcNLBFnKPEiarajTF84o4ATeEw/edit#gid=1134244055
Rows 30/31/52/54/22 helped guide that decision - given that in the merged scenario, step pricing would not be able to show the distinct quantity you were charged for differently from what you used.

rileyjenk (approved):

thecloudman:
I prefer this option. To me it leaves no doubt what Consumed Quantity/Unit are for and guides well for the usage by practitioners.

jpradocueva:
The group approved this PR on Thursday, May 9, call.

mike-finopsorg (approved):

Issue #414, opened 2024-04-22
Is FOCUS use of "Service" aligned with the FinOps tools and services capability?


flanakin:
### Type
> Dimensions: ServiceName, ServiceCategory

### Description
> FinOps Framework introduced a new "FinOps tools and services" capability which states:
>
> > _"For purposes of describing this capability, a FinOps Tool refers to any of the cloud provider or third party software packages that support any of the FinOps capabilities. Tools range from broad FinOps platforms that support most FinOps Capabilities to targeted specialty tools which deeply support a specific activity or scope of cloud usage. A FinOps Service refers to a third-party service to deliver outcomes related to the FinOps capabilities. These services may include training, consultancy, managed services or even the outsourcing of parts of your FinOps practice altogether."_
>
> This doesn't exactly align to our use of the term "service". Should we make any changes to account for that? Given the "For the purpose of describing this capability" intro, I'm not sure the capability is trying to take an official stance on terms so maybe we're fine. Or maybe we want to call out that our use of "Service" covers both "tools" and "services"? Just wanted to bring this up for discussion.

### Definition of done

- [x] Rationalize vendor-neutral, cross-cloud naming

mike-finopsorg:
Thanks @flanakin, I will take some time today to review this compared to the framework and comment here.

mike-finopsorg:
I read through the spec and if anything I think that the FinOps Capability should be updated slightly to avoid the service (cloud service) vs services (services provider) confusion. The spec correctly uses `services providers` as SaaS/Consulting/Reseller and `service` as an offering from a provider.

jpradocueva:
This topic was moved to v1.1.

shawnalpay:
@mike-finopsorg @flanakin Reading through this, it feels like we can close.  Lmk if you agree/disagree; otherwise, will close soon.

shawnalpay:
Discussed in May 8 Members meeting.  @flanakin agreed that we will not act on this.  Closing.

PR #420, opened 2024-04-25
v1.1: Added clarity on what we mean by service

mike-finopsorg:
Discussed in  #414 we need to clarify what FOCUS means by `service`

### DIFF
--- a/specification/overview.md
+++ b/specification/overview.md
@@ -72,6 +72,7 @@ The following principles were considered while building the specification.
 * Where possible, use consistent names that will naturally create associations between related columns in the specification.
 * Column naming must strictly follow the [column naming conventions](#columnnamingconvention).
 * Use established standards (e.g., ISO8601 for dates, ISO4217 for currency).
+* The use of the term `service` used throughout this specification will be used to reference offerings by cloud service providers/SaaS Vendors/Internal Platforms. [Managed Service Providers (MSPs)](#glossary:managed-service-provider) will be used when refering to MSPs or Resellers.
 
 ## Typographic Conventions
 


ahullah (approved):

specification/overview.md
```diff
@@ -72,6 +72,7 @@ The following principles were considered while building the specification.
 * Where possible, use consistent names that will naturally create associations between related columns in the specification.
 * Column naming must strictly follow the [column naming conventions](#columnnamingconvention).
 * Use established standards (e.g., ISO8601 for dates, ISO4217 for currency).
+* The use of the term `service` used throughout this specification will be used to reference offerings by cloud service providers/SaaS Vendors/Internal Platforms. [Managed Service Providers (MSPs)](#glossary:managed-service-provider) will be used when refering to MSPs or Resellers.
```
flanakin, line 75: Fwiw, my only concern was about `ServiceName` and `ServiceCategory`. Both of those will apply to MSPs as well if they have extra charges, right? I think the only inconsistency with "FinOps tools and services" is that FOCUS uses "service" to represent _both_ "tools" and "services".

jpradocueva:
The group agreed to move it to v1.1

ljadvey (approved):

shawnalpay:
Discussed in Oct 31 Members call.  Will follow up with @mike-finopsorg and @flanakin to determine whether we carry this forward or not.

jpradocueva:
### Comment from Members' call on Oct 31:
**Analysis**: This issue, dating back to April, involves defining "service" for consistency in documentation, but it remains unresolved and somewhat outdated.
**Agreements:** Shawn will follow up with Mike and  Michael to decide if this issue should be updated or closed.

Issue #425, opened 2024-04-30
Consistency Review: ChargeClass definition review


ijurica:
### Description
> We need to revisit the ChargeClass column definition

### Proposed approach
> - Provide a more precise definition for "Correction": currently defined as a modification to one or more previous charges, such as refunds and credit modifications. It should also specify that it's a correction of a charge that was invoiced in a previous or one of the previous billing periods.
> - Review value(s) for non-corrections: see the ongoing conversation in the related PR

### Github issue or Reference
> some concerns raised in PR: https://github.com/FinOps-Open-Cost-and-Usage-Spec/FOCUS_Spec/pull/422

ijurica:
I just noticed that there is an existing issue https://github.com/FinOps-Open-Cost-and-Usage-Spec/FOCUS_Spec/issues/423 under post 1.0 scope, addressing the Correction value specification.

PR #432, opened 2024-05-01
FOCUS #425: ChargeClass (Path 1) - spec updates and associated columns adjustments 

ijurica:
- ChargeClass spec change:
  - Clarification indicating that Corrections apply specifically to charges invoiced in previous billing periods.
  - Allowed values reduced to 'Correction'; otherwise, null is expected.

- Consequential spec changes of the following columns:
  - ContractedUnitPrice
  - ListUnitPrice
  - PricingCategory
  - PricingQuantity
  - PricingUnit
  - SkuId
  - SkuPriceId
  - UsageQuantity
  - UsageUnit

- Changelog adjustments


CHANGELOG.md
```diff
@@ -39,19 +39,19 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
   - Clarified that effective cost does not mix or "blend" costs across multiple charges.
 - `ListUnitPrice` column updates:
   - Column is conditional and only required when the provider publishes a price list that excludes discounts.
-  - Must not be null when `ChargeClass` is "Regular" and `ChargeCategory` is "Usage" or "Purchase".
+  - Must not be null when `ChargeClass` is not "Correction" and `ChargeCategory` is "Usage" or "Purchase".
```
flanakin, line 42: What about in-billing-period corrections? (This question applies to all of these `not "Correction"` changes. I think we want it to stay null if it's a correction within the billing period, right?) 

I feel like this requires us to have a value for in-period corrections so these statements can be correct. This feels like a blocker, right? This is the only reason I'm not signing off as-is.

specification/columns/chargeclass.md
```diff
@@ -1,8 +1,8 @@
 # Charge Class
 
-A Charge Class indicates whether the row represents a regular charge or a correction to one or more previous charges. Charge Class is commonly used to differentiate refunds from regularly incurred charges.
+A Charge Class indicates whether the row represents a correction to one or more [*charges*](#glossary:charge) that have been invoiced in previous billing periods. Charge Class is commonly used to differentiate refunds from regularly incurred charges.
 
-The ChargeClass column MUST be present and MUST NOT be null. This column is of type String and MUST be one of the allowed values.
+The ChargeClass column MUST be present in the billing data. This column MUST be of type String and MUST be set to "Correction" in case of a correction to one or more charges already invoiced in previous billing periods. Otherwise, it MUST be null.
```
flanakin, line 5: Minor phrasing nit and making the last sentence a bit more explicit to hopefully minimize ambiguity.
```suggestion
The ChargeClass column MUST be present in the billing data. This column MUST be of type String and MUST be "Correction" when the row represents a correction to one or more charges invoiced in previous billing periods. ChargeClass MUST be null when it is not a correction or when it is a correction within the current billing period.
```

supporting_content/columns/skupriceid.md
```diff
@@ -40,13 +40,13 @@ The following table serves as the basis for reviewing the SkuPriceId spec, as we
 
 | ChargeCategory | ChargeClass | perSku/bulk                       | SkuId            | SkuPriceId       |
 |----------------|-------------|-----------------------------------|------------------|------------------|
-| Usage          | Regular     | MUST be perSku and perSkuPrice    | MUST not be null | MUST not be null |
+| Usage          | (null)      | MUST be perSku and perSkuPrice    | MUST not be null | MUST not be null |
```
flanakin, line 43: nit: Can we leave this in supporting content to convey the values we've discussed? Alternatively, we could just keep this as a separate note and call out options for future consideration.
```suggestion
| Usage          | Regular/Standard/Original/Direct/(null) | MUST be perSku and perSkuPrice    | MUST not be null | MUST not be null |
```

CHANGELOG.md
ijurica: Good catch! We need to allow nulls for corrections that occur during the same billing period as well.!

It seems like we should provide the ability to identify/reference both scenarios, meaning we should provide two allowed values for corrections after all.

thecloudman: would a correction create a new billing line for the previous charge or would the previous charge line item be updated?

AWS-ZachErdman: Do we think it's necessary to identify corrections that occur within the same billing period?

flanakin: @AWS-ZachErdman, based on our convos, different providers may handle this scenario differently. We don't think FOCUS needs to explicitly state how it should be handled today, but we do need to make sure that FOCUS is capable of representing the data however the provider emits charge rows. I may be missing a scenario, but we talked about:
1. Rewrite the original row (only 1 record post-correction)
2. Ledger-based amendments (old record + negation of the old record + all-inclusive new record)
3. Streaming data (old record + 1 or more delta records)

AWS-ZachErdman: Thanks @flanakin I understand your comment now. And yes, aligned with your correction scenarios.

I realize this is about the definition of ListUnitPrice. Should we just remove here (ListUnitPrice and also ContractUnitPrice definitions) the parts of the requirements that reference ChargeClass? Or is it adding something Im not realizing

udam-f2: My personal view is corrections being separated out within billing periods makes things harder for billing data consumers (when providers restate the billing period). I've never cared to look for what was the first line a provider sent vs the next revision for the same line item - so my answer is based on that. Others might care to find differences - but that also means they're constantly having to aggregate to get the full picture.

For providers that send delta updates, it's unavoidable and therefore a users are trained to look for those deltas throughout the month. They can identify the original or subsequent updates based on when those charges came in.

ijurica: Naming aside, if we revert to  `Allow nulls: False` and the 3 allowed values - `Original`, `Current-cycle Correction`, and `Past-cycle Corrections`:

- When providers restate the billing period, you would only receive `Original` and `Past-cycle Corrections`.
- For providers that already send delta updates, you would receive charges with all 3 values.

If we specify all 3 allowed ChargeClass values, the normative for SkuId, SkuPriceId, PricingUnit, etc., is clear and straightforward. If not, we cannot use ChargeClass to identify when those columns MUST NOT be null.

Example: 
> It MUST NOT be null when ChargeClass is "Current-cycle Correction" or "Past-cycle Corrections," and ChargeCategory is "Usage" or "Purchase.

**Question:** Assuming we'll find suitable terms for all 3 allowed values :), are you strongly opposed to reverting to `Not-nullable` and having `Original` and two types of Corrections as allowed values?

ijurica:
Latest suggestion, based on previous conversations

# Charge Class

Charge Class indicates whether the row represents the original [*charge*](#glossary:charge), a correction to one or more previous *charges* that occurred in the same [*billing period*](#glossary:billing-period), or a correction to one or more *charges* that have been invoiced in a previous *billing period*. Charge Class is commonly used to differentiate refunds from regularly incurred charges.

The ChargeClass column MUST be present in the billing data. This column MUST be of type String, MUST NOT be null, and MUST be one of the allowed values.

## Column ID

ChargeClass

## Display Name

Charge Class

## Description

Indicates whether the row represents the original *charge*, a correction to one or more *charges* that occurred in the same *billing period*, or a correction to one or more *charges* that have been invoiced in a previous *billing period*.

## Content Constraints

| Constraint      | Value          |
|:----------------| :--------------|
| Column type     | Dimension      |
| Feature level   | Mandatory      |
| Allows nulls    | False          |
| Data type       | String         |
| Value format    | Allowed values |

Allowed values:

| Value                    | Description |
|:-------------------------|:------------|
| Original                 | Original charges for services used or purchased.                                        |
| Current-cycle Correction | Correction to one or more previous charges that occurred in the same billing period.    |
| Past-cycle Correction    | Correction to one or more charges that have been invoiced in a previous billing period. |

## Introduced (version)

1.0

ijurica:
Alongside the latest ChargeClass column spec suggestion, here is the normative paragraph suggestion for ListUnitPrice as well.

> The ListUnitPrice column MUST be present in the billing data when the provider publishes unit prices exclusive of discounts. This column MUST be a Decimal within the range of non-negative decimal values, MUST conform to [Numeric Format](#numericformat) requirements, and be denominated in the BillingCurrency. It MUST NOT be null when [ChargeClass](#chargeclass) is "Original" and [ChargeCategory](#chargecategory) is "Usage" or "Purchase", MUST be null when ChargeCategory is "Tax", and MAY be null for all other combinations of ChargeClass and ChargeCategory. When ListUnitPrice is present and is not null, multiplying ListUnitPrice by [PricingQuantity](#pricingquantity) MUST equal [ListCost](#listcost), except in cases of ChargeClass "Correction", which may address PricingQuantity or any cost discrepancies independently.

ijurica:
PR #435 prepared based on following proposals:

- https://github.com/FinOps-Open-Cost-and-Usage-Spec/FOCUS_Spec/pull/432#issuecomment-2091145490
- https://github.com/FinOps-Open-Cost-and-Usage-Spec/FOCUS_Spec/pull/432#issuecomment-2091161283

udam-f2 (approved):

rileyjenk (approved):

rileyjenk:
Thanks for the walk through @ijurica

supporting_content/columns/skupriceid.md
```diff
@@ -38,15 +38,15 @@ Current values observed in billing data for various scenarios:
 
 The following table serves as the basis for reviewing the SkuPriceId spec, as well as price, cost, quantity metrics, etc., impacted by the ChargeCategory and ChargeClass columns cleanup
```
ijurica, line 39: Introducing a minor irrelevant change to check if the issue with checks still remains.

```suggestion
The following table serves as the basis for reviewing the SkuPriceId spec, as well as price, cost, quantity metrics, etc., impacted by the ChargeCategory and ChargeClass columns cleanup.
```

CHANGELOG.md
ijurica: The team decided on a nullable column approach, where Correction is the sole allowed value

specification/columns/chargeclass.md
ijurica: suggested change applied.

thecloudman (changes requested):
I put in a change to the wording here to reflect corrections only. I'm not sure I selected the right processes for this change which may have just changed it without a PR.

jpradocueva:
The group approved this PR on Thursday, May 9, call.

specification/columns/chargeclass.md
```diff
@@ -1,8 +1,8 @@
 # Charge Class
 
-A Charge Class indicates whether the row represents a regular charge or a correction to one or more previous charges. Charge Class is commonly used to differentiate refunds from regularly incurred charges.
+Charge Class indicates whether the row represents a correction to one or more [*charges*](#glossary:charge) invoiced in a previous billing period. Charge Class is commonly used to differentiate corrections from regularly incurred charges.
```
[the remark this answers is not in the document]
jpradocueva: @ijurica, @thecloudman could you please resolve this comment? The group already agreed on the PR. Irena, if you are ok with the changes, we can proceed to merge it.

ijurica:
@thecloudman suggested change already committed. 
could you please review and approve the PR. thanks!

thecloudman (approved):

PR #433, opened 2024-05-01
426-consistency-review-chargecategory-definition-and-description-review

kotzenjh:
Parts of the initial definition are no longer applicable since changes to the allowed values list (credit is no longer part of the adjustment category but a separate category) and the introduction of additional columns. It has beeb raised for example, how this column would be used for separate charges that may require special handling from regular usage.

### DIFF
--- a/specification/columns/chargecategory.md
+++ b/specification/columns/chargecategory.md
@@ -1,6 +1,6 @@
 # Charge Category
 
-A Charge Category is the highest level classification for the type of charge that the billing row represents. The Charge Category is commonly used to identify prepaid purchases separately from usage-based charges or to separate charges that may require special handling from regular usage.
+Charge Category represents the highest-level classification of a charge based on its nature, distinguishing between usage-based charges, purchases (one-time or recurring), after-the-fact adjustments, credits, and taxes.
 
 The ChargeCategory column MUST be present in the billing data and MUST NOT be null. This column is of type String and MUST be one of the allowed values.
 
@@ -14,7 +14,7 @@ Charge Category
 
 ## Description
 
-Indicates whether the row represents an upfront or recurring fee, cost of usage that already occurred, an after-the-fact *adjustment* (e.g., credits), or taxes.
+Represents the highest-level classification of a charge based on its nature, distinguishing between usage-based charges, purchases (one-time or recurring), after-the-fact adjustments, credits, and taxes.
 
 ## Content Constraints
 
--- a/specification/columns/chargecategory.md
+++ b/specification/columns/chargecategory.md
@@ -14,7 +14,7 @@ Charge Category
 
 ## Description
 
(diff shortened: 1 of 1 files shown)


specification/columns/chargecategory.md
```diff
@@ -1,6 +1,6 @@
 # Charge Category
 
-A Charge Category is the highest level classification for the type of charge that the billing row represents. The Charge Category is commonly used to identify prepaid purchases separately from usage-based charges or to separate charges that may require special handling from regular usage.
+Charge Category represents the highest-level classification of a charge based on its nature, distinguishing between usage-based charges, purchases (one-time or recurring), after-the-fact adjustments, credits, and taxes.
```
flanakin, line 3: I like the simplification of the first sentence, but I would keep the second for consistency. We usually try to include a "commonly used" statement in all columns. (Unless we're explicitly wanting to get rid of those of course.) We also generally try to not list all values explicitly in the definition. (Although, there are a few cases of that.)

I'm suggesting a small tweak to the end of the first part to tie it back to billing, which is the point of this column, I believe. I feel like there's a better way to phrase this, but I do like the direction. I'm also bringing back the commonly used sentence, but simplified it a lot since I don't think we need to go into that much detail. Curious what others think.
```suggestion
Charge Category represents the highest-level classification of a charge based on the nature of how it is billed. Charge Category is commonly used to identify and distinguish between types of charges that may require different handling.
```

kotzenjh: Thanks for the guidance and agreed, as the values are already listed there's no reason to list them again. What about adding. On "commonly used" would this work: This high-level classification facilitates accurate financial tracking, compliance with regulatory requirements, and can aid in effective budget management and financial forecasting.

thecloudman: happy with the suggested wording changes, allowed values are still consistent.

jpradocueva:
The group agreed to this Pull Request on May 2nd. 
@flanakin, @thecloudman, @udam-f2, @mike-finopsorg, we need you to approve this PR to be merged.

thecloudman (approved):
approved

jpradocueva:
@udam-f2, @mike-finopsorg can one of you approve this PR? Thanks

mike-finopsorg (approved):

specification/columns/chargecategory.md
```diff
@@ -14,7 +14,7 @@ Charge Category
 
 ## Description
 
-Indicates whether the row represents an upfront or recurring fee, cost of usage that already occurred, an after-the-fact *adjustment* (e.g., credits), or taxes.
+Represents the highest-level classification of a charge based on its nature, distinguishing between usage-based charges, purchases (one-time or recurring), after-the-fact adjustments, credits, and taxes.
```
udam-f2, line 17:
```suggestion
Represents the highest-level classification of a charge based on the nature of how it is billed.
```

ijurica (approved):

PR #324, opened 2024-02-05
FOCUS #294: PricingQuantity, PricingUnit, UsageQuantity and UsageUnit review

ijurica:
Multiple Pricing and Usage related columns review:

- Usage Quantity
  - definition adjustment
  - normative paragraph adjustment
- Usage Unit
  - definition adjustment
  - normative paragraph adjustment
- Pricing Quantity
  - definition adjustment
- Pricing Unit
  - minor spec adjustment
  - a discussion regarding the differentiation between PricingUnit and UsageUnit, block-pricing, etc., to be included in the supporting content

PR #434, opened 2024-05-02
#409 Path 3 - ActualUnit/Quantity for broader applications  

mike-finopsorg:
Changed Usage to Actual and updated wording to better describe the columns apply to a distinct measure of the SKU and are not limited to usage lines. Added more supporting content to better explain the examples discussed during the process of working this column out. - Updated PR captures the discussion from the TF-2 call 1-May-24.


specification/columns/actualquantity.md
```diff
@@ -0,0 +1,38 @@
...
+
+When [ChargeCategory](#chargecategory) is "Purchase":
+    The value represents a measurement that represents the volume of units purchased of a [*resource*](#glossary:resource) or [*service*](#glossary:service) and should be relevant to any associated [ChargeCategory](#chargecategory) "Usage" records.
+
+When [ChargeCategory](#chargecategory) is "Usage":
+    The measurement represents the actual quantity consumed which may be a different value to the quantity that is charged represented in the [PricingQuantity](#pricingquantity) column or measured in a different unit to that of [pricingunit](#pricingunit).
+
+ActualQuantity column MUST be present in the billing data when the provider supports the measurement of usage or purchases. This column MUST be of type Decimal and MUST conform to [Numeric Format](#numericformat) requirements. The value MAY be negative in cases where [ChargeClass](#chargeclass) is "Correction". This column MUST NOT be null when [ChargeClass](#chargeclass) is "Standard" and [ChargeCategory](#chargecategory) is "Usage" or "Purchase", MUST be null when ChargeCategory is "Tax", and MAY be null for all other combinations of ChargeClass and ChargeCategory.
```
mike-finopsorg, line 11: Need to re-word the following: `ActualQuantity column MUST be present in the billing data when the provider supports the measurement of usage or purchases.` 

Current thinking is `when the provider supports charges where ActualQuantity does not match the PricingQuantity` - thoughts?

mike-finopsorg: What about just making these columns a straight MUST?

specification/columns/actualunit.md
```diff
@@ -0,0 +1,31 @@
+# Actual Unit
+
+The Actual Unit represents a provider-specified measurement unit indicating how a provider measures usage or purchase of a given SKU associated with a [*resource*](#glossary:resource) or [*service*](#glossary:service). Actual Unit complements the [Actual Quantity](#actualquantity) metric. It is often listed at a finer granularity or over a different time interval when compared to the [Pricing Unit](#pricingunit) (complementary to [Pricing Quantity](#pricingquantity)), and focuses on *resource* and *service* consumption, not pricing and cost.
```
aqu-erp, line 3:
```suggestion
The Actual Unit represents a provider-specified measurement unit indicating how a provider measures usage or purchase of a given SKU associated with a [*resource*](#glossary:resource) or [*service*](#glossary:service). Actual Unit complements the [Actual Quantity](#actualquantity) metric. It is often listed at a finer granularity or over a different time interval when compared to the [Pricing Unit](#pricingunit) (complementary to [Pricing Quantity](#pricingquantity)), and focuses on *resource* and *service* measurement in a charge, not pricing and cost.
```

specification/columns/actualunit.md
```diff
@@ -0,0 +1,31 @@
...
+
+The ActualUnit column MUST be present in the billing data when the provider supports the measurement of usage or purchases. This column MUST be of type String. It MUST NOT be null when [ChargeClass](#chargeclass) is "Standard" and [ChargeCategory](#chargecategory) is "Usage" or "Purchase", MUST be null when ChargeCategory is "Tax", and MAY be null for all other combinations of ChargeClass and ChargeCategory. Units of measure used in ActualUnit SHOULD adhere to the values and format requirements specified in the [UnitFormat](#unitformat) attribute. For data size unit, count based units, and time based units the ActualUnit MUST represent an individual unit not multiples/blocks of. The ActualUnit column MUST NOT be used to determine values related to any pricing or cost metrics.
```
aqu-erp, line 5:
```suggestion
The ActualUnit column MUST be present in the billing data when the provider supports charges where the ActualUnit does not match the [Pricing Unit](#pricingunit). This column MUST be of type String. It MUST NOT be null when [ChargeClass](#chargeclass) is "Standard" and [ChargeCategory](#chargecategory) is "Usage" or "Purchase", MUST be null when ChargeCategory is "Tax", and MAY be null for all other combinations of ChargeClass and ChargeCategory. Units of measure used in ActualUnit SHOULD adhere to the values and format requirements specified in the [UnitFormat](#unitformat) attribute. For data size unit, count based units, and time based units the ActualUnit MUST represent an individual unit not multiples/blocks of said unit. The ActualUnit column MUST NOT be used to determine values related to any pricing or cost metrics.
```

specification/columns/actualquantity.md
```diff
@@ -0,0 +1,38 @@
+# Actual Quantity
+
+The Actual Quantity represents the volume of a given SKU associated with a [*resource*](#glossary:resource) or [*service*](#glossary:service) used or purchased, based on the [Actual Unit](#actualunit). Actual Quantity is often derived at a finer granularity or over a different time interval when compared to the [Pricing Quantity](#pricingquantity) (complementary to [Pricing Unit](#pricingunit)) and focuses on the actual amount of a *resource* or *service* used or purchased, not pricing and cost.
+
+When [ChargeCategory](#chargecategory) is "Purchase":
+    The value represents a measurement that represents the volume of units purchased of a [*resource*](#glossary:resource) or [*service*](#glossary:service) and should be relevant to any associated [ChargeCategory](#chargecategory) "Usage" records.
+
+When [ChargeCategory](#chargecategory) is "Usage":
+    The measurement represents the actual quantity consumed which may be a different value to the quantity that is charged represented in the [PricingQuantity](#pricingquantity) column or measured in a different unit to that of [pricingunit](#pricingunit).
```
aqu-erp, line 9:
```suggestion
    The value represents the actual quantity consumed which may be a different value to the quantity that is charged represented in the [PricingQuantity](#pricingquantity) column or measured in a different unit to that of [pricingunit](#pricingunit).
```

specification/columns/actualquantity.md
```diff
@@ -0,0 +1,38 @@
+# Actual Quantity
+
+The Actual Quantity represents the volume of a given SKU associated with a [*resource*](#glossary:resource) or [*service*](#glossary:service) used or purchased, based on the [Actual Unit](#actualunit). Actual Quantity is often derived at a finer granularity or over a different time interval when compared to the [Pricing Quantity](#pricingquantity) (complementary to [Pricing Unit](#pricingunit)) and focuses on the actual amount of a *resource* or *service* used or purchased, not pricing and cost.
+
+When [ChargeCategory](#chargecategory) is "Purchase":
+    The value represents a measurement that represents the volume of units purchased of a [*resource*](#glossary:resource) or [*service*](#glossary:service) and should be relevant to any associated [ChargeCategory](#chargecategory) "Usage" records.
```
aqu-erp, line 6:
```suggestion
    The value represents the volume of units purchased of a [*resource*](#glossary:resource) or [*service*](#glossary:service) which may be different from the quantity represented in the [PricingQuantity](#pricingquantity) column or measured in a different unit to that of [PricingUnit](#pricingunit). It should be relevant to any associated  [ChargeCategory](#chargecategory) "Usage" records.
```

specification/columns/actualquantity.md
```diff
@@ -0,0 +1,38 @@
...
+
+When [ChargeCategory](#chargecategory) is "Purchase":
+    The value represents the volume of units purchased of a [*resource*](#glossary:resource) or [*service*](#glossary:service) which may be different from the quantity represented in the [PricingQuantity](#pricingquantity) column or measured in a different unit to that of [PricingUnit](#pricingunit). It should be relevant to any associated  [ChargeCategory](#chargecategory) "Usage" records.
+
+When [ChargeCategory](#chargecategory) is "Usage":
+    The value represents the actual quantity consumed which may be a different value to the quantity that is charged represented in the [PricingQuantity](#pricingquantity) column or measured in a different unit to that of [pricingunit](#pricingunit).
+
```
mike-finopsorg, line 10: TODO: Need to add details on how to handle aggregation across the column

AWS-ZachErdman: I like the descriptions for ChargeCategories

udam-f2:
Hate to do this but I object to this PR. 

There are a few reasons but the most concerning issue I have is using a term like 'Actual'. It means absolutely nothing. Actual as opposed to what? What should a practitioner think of when they're looking at this OR are using a pricing unit? It's not an actual unit? 

Over the past 13 months, one thing has been clear to me. Most of the time when we really struggle with naming, its because we've fundamentally tried to merge too many things. Sometimes its because people are trying to avoid nulls. Other times its because we're describing too many different things that there's no logical name that fits. This unfortunately is another one of those times. By merging usage/purchase and burndown/non-burndown scenarios and each semantic into one column just to avoid **possibly** needing another column - purchase quantity (**which we haven't even broadly validated is necessary**), we're resorting to creating something that now 'actually' has no clear meaning.

"Usage Quantity is often derived at a finer granularity or over a different time interval when compared to the [Pricing Quantity](#pricingquantity) (complementary to [Pricing Unit](#pricingunit)), and focuses on *resource* and *service* _consumption_, not pricing and cost.". 

^^ That's very clear - by name and definition to hopefully anyone who looks at the spec OR the data. This was debated for weeks and settled on with a broad group of members. The only mistake was the words 'usage or purchase' which came from aligning the text for both pricing unit and usage unit to look similar. People like Irena were primary members of the original work (because pricing/usage were done at the same time) - and can attest to the original intention being for quantifying 'usage' (or consumption).

For these reasons, I think merging this PR will be a net negative to the specification not an addition.

rileyjenk:
I also object to this PR. From the beginning, tracking usage has been a first class concept of FOCUS, though through our discussions this has been challenged. The Finops Foundation has clarified that both cost and usage are top tier capabilities. 

Making the question of "how many x did we use" more confusing is a step in the wrong direction. We understand it but that is because of our direct involvement. For the rest of practitioners out there who are converting to use FOCUS 1.0 compared to the CUR/GCP database/Azure Data, we are adding an unnecessary stumbling block.   

The name word Actual is ambiguous and confusing. https://github.com/FinOps-Open-Cost-and-Usage-Spec/FOCUS_Spec/pull/324 was the first step that introduced confusion (though I am sure we would have found things to straighten out in the consistency review). This PR furthers this. What does actual mean, actual of what, does that mean there are non actual units?

It is far more clear to have UsageUnit and UsageQuantity and PurchaseUnit and PurchaseQuantity.  It is far more easy to explain, document, use, and enhance.  By having both columns answering both the "How much did I use?" and the "How much did I purchase?" are more straightforward, easy to understand, and ultimately more valuable to practitioners.

aqu-erp:
I am open to the above alternatives if we can resolve the below:

If UsageUnit and UsageQuantity are for tracking usage, then we must find new homes for the non-pricing quantities and units for charges that do not consume usage. Charges that do not consume usage include but are not limited to: purchases, refunds, credits (money), and unused commitments. PurchaseQuantity and PurchaseUnit are not satisfactory columns for all these values. 

Until it's clear to me how non-pricing quantities and units for all other charges will be handled outside of UsageQuantity and UsageUnit, I maintain that Path 3 is the cleaner option. Practitioners are accustomed to handling quantities in one column differently based on charge context.

In the absence of a complete solution, Path 3 maintains continuity with real world practice while keeping the door open for additional columns to be added in future versions.

specification/columns/actualquantity.md
marc-perreaut: Agree with straight MUST.

marc-perreaut:
While the term "actual" might not be perfect, I prefer the idea of a common column for purchase and usage, and possibly other scenarios.

I understand the intent of these columns is to get more standardized practitioner-friendly data than the PricingUnit / PricingQuantity, so in the end the columns work in pair (null if and only if the other is null).

Maybe to make things simple and clear, the column names could be StandardizedPricingUnit / StandardizedPricingQuantity.
That's not pretty. Here is another try: UsableUnit and UsableQuantity.

I am not fundamentally against dedicated columns for ChargeCategory=Usage: I am just expressing my preference.

Issue #438, opened 2024-05-07
Builds failing with error: externally-managed-environment


mike-finopsorg:
### Description
> Builds have started failing with the following error:

```WARNING: The directory '/github/home/.cache/pip' or its parent directory is not owned or is not writable by the current user. The cache has been disabled. Check the permissions and owner of that directory. If executing pip with sudo, you should use sudo's -H flag.
error: externally-managed-environment

× This environment is externally managed
╰─> To install Python packages system-wide, try apt install
    python3-xyz, where xyz is the package you are trying to
    install.
    
    If you wish to install a non-Debian-packaged Python package,
    create a virtual environment using python3 -m venv path/to/venv.
    Then use path/to/venv/bin/python and path/to/venv/bin/pip. Make
    sure you have python3-full installed.
    
    If you wish to install a non-Debian packaged Python application,
    it may be easiest to use pipx install xyz, which will manage a
    virtual environment for you. Make sure you have pipx installed.
    
    See /usr/share/doc/python3.12/README.venv for more information.

note: If you believe this is a mistake, please contact your Python installation or OS distribution provider. You can override this, at the risk of breaking your Python installation or OS, by passing --break-system-packages.
hint: See PEP 668 for the detailed specification.
Error: Process completed with exit code 1.
```

### Definition of done
> Builds work again

jpradocueva:
The content of this changed was notified to the FOCUS Members on the May 9th call.

PR #439, opened 2024-05-07
Fix broken builds reported in #438

mike-finopsorg:
In order to unbreak the github actions pipeline, adding temporary fix which will allow the pip install to apply to the system python on the container. This is low risk as the python install is only active for the life of the build. If we want to avoid this we need to do further testing on using a venv for the spec creation. But this PR gets the pipeline unblocked for the pending 1.0 release.


rileyjenk (approved):
LGTM

udam-f2 (approved):

Issue #441, opened 2024-05-08
Commitment purchase, Unused commitmenr and Burndown related questions


ijurica:
### Description

Clarification on some Commitment purchase, Unused commitment, and Burndown questions needed

- Commitment Purchase: 
   - Q1: What do we expect in ChargePeriodEnd column?
     - Assumption: In case of a 1 year commitment `ChargePeriodStart: To` and `ChargePeriodEnd: To + 1yr`
     
 - Unused commitment: 
    - Q2: How many and how often should those be provided? 
       - Assumption: Only one at the end of the commitment period
    - Q3: What do we expect in ChargePeriodStart and ChargePeriodEnd columns?
      - Assumption: `ChargePeriodStart: To` and `ChargePeriodEnd: To + 1yr`
    - Q4: In case of Unused commitment record, do we expect `<commitmenId> `in ResourceId column or CommitmentDiscountId column?
       - Assumption:  `ResourceId: (null)` and `CommitmentDiscountId: <commitmenId>`
    - Q5: Is Unused commitment Purchase or Usage?
       - Assumption: `Usage`, whereby the SkuId and SkuPriceId mathch the SkuId and SkuPriceId of the Purchase

PR #444, opened 2024-05-09
#441 Clarifications for various Commitment discount scenarios

udam-f2:
Resolves the concerns in #441 related to representing Commitment-based discount purchases and corresponding usage (for both CommitmentDiscountStatus "Used" and "Unused") scenarios


cnharris10 (approved):

ijurica (approved):

specification/attributes/discount_handling.md
```diff
@@ -34,9 +34,10 @@ Indicates how to include and apply discounts to usage charges or rows.
 * Purchased discounts (e.g., commitment-based discounts) MUST be amortized.
   * The BilledCost MUST be 0 for any row where the commitment covers the entire cost for the charge period.
   * The EffectiveCost MUST include the portion of the amortized purchase cost that applies to this row.
-  * CommitmentDiscountStatus MUST be "Used" for rows that received a reduced price from that commitment.
-  * If a commitment is not fully utilized, the provider MUST include a row that represents the unused portion of the commitment for that charge period. CommitmentDiscountStatus MUST be "Unused".
   * The sum of the EffectiveCost for all rows where CommitmentDiscountStatus is "Used" or "Unused" for each CommitmentDiscountId over the entire duration of the commitment MUST be the same as the total BilledCost of the commitment-based discount.
+  * The CommitmentDiscountId and ResourceId MUST be set to the ID assigned to the commitment-based discount. ChargeCategory MUST be set to "Purchase" on rows that represent a purchase of a commitment-based discount.
```
ijurica, line 38: I'm not entirely sure if I'll manage to describe the use case we discussed yesterday, but here it is: 

What if a provider is unable to provide Purchase records containing IDs of Reserved Instance (RI) instances (assuming those are the commitment IDs we're seeking and would be provided in Usage records with commitment discounts applied), but only a Purchase record with the original Order ID (which allows retrieval of corresponding RI instance IDs through an API)? 
Would it be acceptable to change MUST' to SHOULD, at least for the time being? 
If not, what do we expect in  CommitmentId and ResourceId columns? null? while that OrderId would still be provided in one of the native columns?

udam-f2: I thought @flanakin was ok with the ID assigned to the discount (which might be ok to be order-id for now? but obviously as we all agreed it would be best if it mapped to the actual discount?). I think it would be weird to say you don't have to provide the ID of the thing you bought - because that's what any consumer of the data would want to map/correlate to the usage records (amortization/coverage etc.). Open to ideas but I do think the language can't be so vague that any ID would meet the requirement.

rileyjenk (approved):

aqu-erp (approved):

mike-finopsorg (approved):

jpradocueva:
This PR was agreed by the FOCUS Members on May 9 call.

specification/attributes/discount_handling.md
flanakin: We don't have an instance ID at the time of purchase. We only have the order ID. They are effectively bulk purchases that result in a single charge.

PR #468, opened 2024-05-20
[ADMIN] remove specific dates for 2025 releases

jpradocueva:


jpradocueva:
This PR was approved by the Members' group on May 23 call.

PR #479, opened 2024-06-04
300 proposal billing account term and description confusion

ahullah:
ok this is a bit daft as its not a code change...... do we want to find a way for the 'glossary' type links to work as hyperlinks?


cnharris10 (approved):

flanakin (approved):

jpradocueva:
The group concluded that the description in the glossary is good enough. There is no bug that the links don't work in GitHub. This topic will be reviewed again in the future. This applied to the links in general.

PR #500, opened 2024-07-08
Replace 'SKU' with 'Stock Keeping Unit' for clarity in the glossary

sengaya:


jpradocueva:
@sengaya, thanks for your input.
The group reviewed your suggestion and agreed to close the PR without action.
The group feels that "SKU" is the commonly used term in the industry and does not want to change it to "Stock Keeping Unit." They agreed to mention "Stock Keeping Unit" in the description but not to change the term itself.

In any case, please contact me via focus@finops.org if you would like to participate in the group discussion. I can provide an overview of the requirements for becoming a member.

Issue #497, opened 2024-07-03
SKU Price Details


jpradocueva:
**Lead Maintainer**: Karl
### Documents: 
* [#497](https://github.com/FinOps-Open-Cost-and-Usage-Spec/FOCUS_Spec/issues/497) SKU Price Details (Issue)
* Folder: [#497-srv-SKU-details-SKU-price-details](https://drive.google.com/drive/folders/1kbNNasX0xZBbLrcs1MQK5s5zWyv9o8CD?usp=drive_link)
* PR [#503](https://github.com/FinOps-Open-Cost-and-Usage-Spec/FOCUS_Spec/pull/503) SKU Details + SKU PriceDetails Columns by [karl.kraft@walmart.com](mailto:karl.kraft@walmart.com) 
* PR [#530](https://github.com/FinOps-Open-Cost-and-Usage-Spec/FOCUS_Spec/pull/530) #503a - Update SkuPriceId per SkuPriceDetails by [karl.kraft@walmart.com](mailto:karl.kraft@walmart.com) 
* Spreadsheet:[ 24.07.17-AzureEA-FOCUS-and-PriceList-sample-dataset](https://docs.google.com/spreadsheets/d/1qoy5OjFeZFIK5fuVUKGiI_yCfN5_9tBxVBaRVlBgXtA/edit?gid=240278343#gid=240278343)
* Document from Joseph:[ SKU Details from Providers](https://docs.google.com/document/d/1zPEUZ5L5wf9YAJ63ymW3tyrOFx-krJBB6sf1bHU_oOs/edit#heading=h.p2oplf92wckb)

### Description
This issue was created to host the following action item:
- [ ] TF2-[#497](https://github.com/FinOps-Open-Cost-and-Usage-Spec/FOCUS_Spec/issues/497): Karl, Andrew: Draft SKU price details, ensuring they align with SKU price ID.
Branch: [sku-details_sku-price-details](https://github.com/kk09v/FOCUS_Spec/tree/sku-details_sku-price-details) by Karl

### Proposed approach
> Describe the approach that your group is proposing

### Github issue or Reference
> If the topic is related to a particular work item, reference the Github issue. If its a specification-wide topic, indicate that.

### Context
> Add any context that may help the community think through this and provide useful feedback

jpradocueva:
### New set of action items:
[...]

PR #503, opened 2024-07-09
FOCUS #497: SkuPriceDetails column

kk09v:
Introduction of _KeyValueFormat_ columns _SkuDetails_ and _SkuPriceDetails_  to capture properties of a **SKU ID** and **SKU Price ID**, respectively.

**Lead Maintainer**: Karl
### Documents: 
* [#497](https://github.com/FinOps-Open-Cost-and-Usage-Spec/FOCUS_Spec/issues/497) SKU Price Details (Issue)
* Folder: [#497-srv-SKU-details-SKU-price-details](https://drive.google.com/drive/folders/1kbNNasX0xZBbLrcs1MQK5s5zWyv9o8CD?usp=drive_link)
* PR [#503](https://github.com/FinOps-Open-Cost-and-Usage-Spec/FOCUS_Spec/pull/503) SKU Details + SKU PriceDetails Columns by [karl.kraft@walmart.com](mailto:karl.kraft@walmart.com) 
* PR [#530](https://github.com/FinOps-Open-Cost-and-Usage-Spec/FOCUS_Spec/pull/530) #503a - Update SkuPriceId per SkuPriceDetails by [karl.kraft@walmart.com](mailto:karl.kraft@walmart.com) 
* Spreadsheet:[ 24.07.17-AzureEA-FOCUS-and-PriceList-sample-dataset](https://docs.google.com/spreadsheets/d/1qoy5OjFeZFIK5fuVUKGiI_yCfN5_9tBxVBaRVlBgXtA/edit?gid=240278343#gid=240278343)
* Document from Joseph:[ SKU Details from Providers](https://docs.google.com/document/d/1zPEUZ5L5wf9YAJ63ymW3tyrOFx-krJBB6sf1bHU_oOs/edit#heading=h.p2oplf92wckb)


kk09v:
OPEN ITEMS:
- [ ] Adding references and back links to definitions of related columns, properties, etc.

kk09v:
Additional Discussion points:
- Evaluation of SHOULD vs MUST normative statements
- Clarifications
  - Do we need clarification that the immutability of properties clause means that flags (e.g. "is_current") and other things like validity dates should not be included?
  - Do we need to define what is meant by "property"
  - Should we further clarify that property refers to key and value unless otherwise specified?
  - How do we explain the sorts of things we want included without relying on examples?
- Phrasing: list of words/phrases that might have a better alternative
  - "all resources"
  - "properties"

specification/columns/skudetails.md
```diff
@@ -0,0 +1,49 @@
...
+
+The **SKU Details** column represents a list of relevant properties shared by all resources having the same [**SKU ID**](#skuid) which are not captured in another FOCUS column. These properties enable a practitioner to more easily calculate metrics such as total units of a service when it is not directly billed in those units (e.g. cores) and thus enables FinOps capabilities such as unit economics.
+
+
+The _SkuDetails_ column adheres to the following requirements:
+
+* The properties (both key and value) contained in the _SkuDetails_ column **MUST** be shared across all resources having the same _SkuId_.
+* The _SkuDetails_ column **MUST** be in [_KeyValueFormat_](#key-valueformat).
+* The _SkuDetails_ column **SHOULD NOT** contain properties which are not applicable to the corresponding _SkuId_.
+* Where _SkuDetails_ contains a property with a value which is a numerical, the value provided **MUST** represent the value for a [_ConsumedUnit_](#consumedunit) of 1.
```
cnharris10, line 11: I don't quite understand this statement.  Is this ensuring that a singular unit is represented as the SKU?

kk09v: Think about an offering that has a unit of measure of 100 bars that has a property that tells me how many foos there 
are. In this case, I don't want the foos per bar; I want the foos per 100 bars since 100 bars is the unit of measure. 

If it helps, think of "bars" as dollars (unit of measure is $100 bills) and "foos" as pennies. When we're talking about a $100 bill, I don't want you to tell me that there are 100 pennies in a dollar; I want you to tell me that there are 10,000 pennies in that $100 bill.

ijurica:
### SKU - SKU Price relationship

Is the following statement correct? 

> _... A given value of SkuPriceId MUST be associated with one and only one SkuId, except in cases of commitment discount flexibility. ..._

When we wrote the SkuPriceId specification, it seemed to be the case, but I now have doubts. It would be beneficial to examine sample cost records to verify or contradict this.

### Sample charge records

Karl has already noted the need to include sample charge records (beyond  sample SKU Details and SKU Price Details values). Here are some key use cases I believe we should address:

- flat-rate pricing, 
- tier-based pricing, 
- volume-based pricing, 
- commitment discount flexibility, 
- free tier, 
- etc. 

_Note: It is very likely that the terms used are not the most appropriate (for internal use only). We should probably address those pricing strategies/models/vegetables 🙂 at some point in the glossary or appendix._

specification/columns/skudetails.md
kk09v, line 11: You're right, I had grabbed the wrong column when referencing this. I blame too many open tabs.
```suggestion
* Where _SkuDetails_ contains a property with a value which is a numerical, the value provided **MUST** represent the value for a [_ConsumedQuantity_](#consumedquantity) of 1.
```

specification/columns/skupricedetails.md
```diff
@@ -0,0 +1,50 @@
...
+The **SKU Price Details** column represents a list of relevant properties shared by all resources having the same [**SKU Price ID**](#skupriceid) which are not captured in another FOCUS column. These properties enable a practitioner to more easily understand the differences that make up the multiple **SKU Price IDs** under a given [**SKU ID**](#skuid).
+
+
+The _SkuPriceDetails_ column adheres to the following requirements:
+
+* The properties (both key and value) contained in the _SkuPriceDetails_ column **MUST** be shared across all resources having the same _SkuPriceId_.
+* The _SkuPriceDetails_ column **MUST** be in [_KeyValueFormat_](#key-valueformat).
+* The _SkuPriceDetails_ column **SHOULD NOT** contain properties which are not applicable to the corresponding _SkuPriceId_.
+* The _SkuPriceDetails_ column **SHOULD NOT** contain properties which are already captured in the [_SkuDetails_](#skudetails) column.
+* Where _SkuPriceDetails_ contains a property with a value which is a numerical, the value provided **MUST** represent the value for a [_ConsumedUnit_](#consumedunit) of 1.
```
kk09v, line 12:
```suggestion
* Where _SkuPriceDetails_ contains a property with a value which is a numerical, the value provided **MUST** represent the value for a [_ConsumedQuantity_](#consumedquantity) of 1.
```

specification/columns/skudetails.md
cnharris10: Ok maybe a slight rewording will help make this more evident.  Curious how others feel.

kk09v:
Here are some of the use cases (phrased as questions I'd want to be able to easily answer) I had in my mind when I was creating these dimensions. I do reference specific CSP offerings in many of these because each CSP has a different approach to pricing/billing which may make it easier or harder to answer a question.

How many cores of D-series (Azure) VMs do we have running in each region?
How much of my GCS bill is for storage of data versus access to the data (e.g. operations, network, etc.)?
How much total "disk" storage do we have provisioned?
How much do we spend on GPUs?
What portion of our N-series VM usage is Spot instances?

_Specific to SkuPriceDetails_
Why am I paying more for _this_ resource than _that_ resource (which look the same to me, as a user)?

ijurica:
@kk09v I've noticed that you've started preparing sample dataset.

I feel like we should prepare more sample records before we close this PR. We should probably choose a couple of records from each CSP. As mentioned before, here are some use case scenarios that I believe we should cover.

- flat-rate pricing,
- tier-based pricing,
- volume-based pricing,
- commitment discount flexibility,
- free tier,
- etc.

Not sure if this helps, but here it is - in [this Google spreadsheet](https://docs.google.com/spreadsheets/d/1qoy5OjFeZFIK5fuVUKGiI_yCfN5_9tBxVBaRVlBgXtA/edit?gid=240278343#gid=240278343), I've extracted some Azure-EA sample records (focusing on a subset of FOCUS and Pricing columns). As mentioned before, I've used sample data provided in the azure-finops-toolkit. In one of the sheets, I've also provided a list of columns available in the FOCUS Dataset, EA-Pricing, and Public Azure Pricing.

robmartin33:
“…which are not captured in another FOCUS column.” Implies that “Region” in Irena’s example would not be in SKU Details.

Are we losing some compatibility over time (constructing a query against multiple versions of FOCUS data in a single database would have to look in a column for new data, but parse SKU Details in older data), clarity (we recreate the problem we have today having to know what collection of columns and SKU Details makes up a unique SKU), and flexibility if we take some things out because they are in columns? As opposed to allowing each vendor to specify HOW THEY define their SKU fully (while still filling out the columns correctly also). 

If all of the SKU Details are in for any SKU, including the data that make that SKU unique, we know they are all always all the attributes. 

I see Udam's point about this being an easy button that might slow down column standardization, but it also does simplify reporting for practitioners in certain use cases wouldn't it? 

If I'm getting this all wrong, please let me know. Happy to take a call.

kk09v:
@robmartin33
> “…which are not captured in another FOCUS column.” Implies that “Region” in Irena’s example would not be in SKU Details.
> 
> Are we losing some compatibility over time (constructing a query against multiple versions of FOCUS data in a single database would have to look in a column for new data, but parse SKU Details in older data), clarity (we recreate the problem we have today having to know what collection of columns and SKU Details makes up a unique SKU), and flexibility if we take some things out because they are in columns? As opposed to allowing each vendor to specify HOW THEY define their SKU fully (while still filling out the columns correctly also).
> 
> If all of the SKU Details are in for any SKU, including the data that make that SKU unique, we know they are all always all the attributes.
> 
> I see Udam's point about this being an easy button that might slow down column standardization, but it also does simplify reporting for practitioners in certain use cases wouldn't it?
> 
> If I'm getting this all wrong, please let me know. Happy to take a call.

Region (in particular) is tricky because Region could be a property of a SKU (in the Azure example, the SKU is D4v3/D4sv3 in East US), could be a property of a SKU Price (where the same SKU deployed in different regions has a different price), or could be a property of a resource (a single SKU/SKU Price in GCP applies to "Americas" where the resource could be deployed any of 4 regions). In my example with the Microsoft data, I faked a property called PricingRegion as an example of what could be put into SkuPriceDetails. I used the value "US East" from the x_SkuRegion column which actually differs from the value in the Region column "East US"

This specific case aside, to your question of "What happens if a new dimension is created for something currently in SkuDetails?" The value SHOULD NOT be removed from the SkuDetails because that would violate the immutability clause in the normative section. The clause you're referencing is not in the normative section; it was intended to signal more about what data should be included rather than what should not be included. I'm trying to avoid someone implementing this reading these columns and asking themselves, "Does this _require_ me to repeat PricingCategory, PricingUnit, etc. in SkuPriceDetails?" The answer to that question is no, but if they decide to do so, that's fine because there's nothing indicating the MUST NOT nor SHOULD NOT repeat those as key/value pairs here.

kk09v:
I added the PascalCase clause to both and reformatted the examples as PascalCase, per the discussion in TF2 today.

I also added PascalCase to the glossary, though it may not meet our criteria for being included because it's well defined externally.  The only other place it's referenced in FOCUS is in specification/attributes/column_naming_and_ordering.md where there's a link to an external site. This was added to the glossary for discussion since I could go either way on including it. If we decide to keep it, we'd link PascalCase to the glossary in these column definitions, otherwise we'll remove that addition from the glossary.

specification/columns/skudetails.md
```diff
@@ -0,0 +1,50 @@
+# SKU Details
+
+The **SKU Details** column represents a list of relevant properties shared by all resources having the same [**SKU ID**](#skuid) which are not captured in another FOCUS column. These properties enable a practitioner to more easily calculate metrics such as total units of a service when it is not directly billed in those units (e.g. cores) and thus enables FinOps capabilities such as unit economics.
+
+
+The _SkuDetails_ column adheres to the following requirements:
+
+* The properties (both key and value) contained in the _SkuDetails_ column **MUST** be shared across all resources having the same _SkuId_.
```
jpradocueva, line 8:
```suggestion
* The properties (both key and value) contained in the _SkuDetails_ column MUST be shared across all resources having the same _SkuId_.
```

jpradocueva: @kk09v, although the group has not confirmed the final editorial rules, there was an agreement about don't use bold for the normative keywords.

specification/columns/skudetails.md
```diff
@@ -0,0 +1,50 @@
...
+* The _SkuDetails_ column **MUST** be in [_KeyValueFormat_](#key-valueformat).
```
jpradocueva, line 9:
```suggestion
* The _SkuDetails_ column MUST be in [_KeyValueFormat_](#key-valueformat).
```

specification/columns/skudetails.md
```diff
@@ -0,0 +1,50 @@
...
+* The _SkuDetails_ column **SHOULD NOT** contain properties which are not applicable to the corresponding _SkuId_.
```
jpradocueva, line 10:
```suggestion
* The _SkuDetails_ column SHOULD NOT contain properties which are not applicable to the corresponding _SkuId_.
```

specification/columns/skudetails.md
```diff
@@ -0,0 +1,50 @@
...
+* Where _SkuDetails_ contains a property with a value which is a numerical, the value provided **MUST** represent the value for a [_ConsumedQuantity_](#consumedquantity) of 1.
+* The key for a property **SHOULD** remain consistent across comparable SKUs having that property and the values for this key **SHOULD** remain in a consistent format.
```
jpradocueva, line 12:
```suggestion
* The key for a property SHOULD remain consistent across comparable SKUs having that property and the values for this key SHOULD remain in a consistent format.
```

specification/columns/skudetails.md
```diff
@@ -0,0 +1,50 @@
...
+* The key for a property **SHOULD** be formatted in PascalCase.
```
jpradocueva, line 13:
```suggestion
* The key for a property SHOULD be formatted in PascalCase.
```

specification/columns/skudetails.md
```diff
@@ -0,0 +1,50 @@
...
+* Properties included in _SkuDetails_ **SHOULD** be immutable once included for a _SkuId_.
```
jpradocueva, line 14:
```suggestion
* Properties included in _SkuDetails_ SHOULD be immutable once included for a _SkuId_.
```

specification/columns/skupricedetails.md
```diff
@@ -0,0 +1,51 @@
+# SKU Price Details
+
+The **SKU Price Details** column represents a list of relevant properties shared by all resources having the same [**SKU Price ID**](#skupriceid) which are not captured in another FOCUS column. These properties enable a practitioner to more easily understand the differences that make up the multiple **SKU Price IDs** under a given [**SKU ID**](#skuid).
+
+
+The _SkuPriceDetails_ column adheres to the following requirements:
+
+* The properties (both key and value) contained in the _SkuPriceDetails_ column **MUST** be shared across all resources having the same _SkuPriceId_.
```
jpradocueva, line 8:
```suggestion
* The properties (both key and value) contained in the _SkuPriceDetails_ column MUST be shared across all resources having the same _SkuPriceId_.
```

specification/columns/skupricedetails.md
```diff
@@ -0,0 +1,51 @@
...
+* The _SkuPriceDetails_ column **MUST** be in [_KeyValueFormat_](#key-valueformat).
```
jpradocueva, line 9:
```suggestion
* The _SkuPriceDetails_ column MUST be in [_KeyValueFormat_](#key-valueformat).
```

specification/columns/skupricedetails.md
```diff
@@ -0,0 +1,51 @@
...
+* The _SkuPriceDetails_ column **SHOULD NOT** contain properties which are not applicable to the corresponding _SkuPriceId_.
```
jpradocueva, line 10:
```suggestion
* The _SkuPriceDetails_ column SHOULD NOT contain properties which are not applicable to the corresponding _SkuPriceId_.
```

specification/columns/skupricedetails.md
```diff
@@ -0,0 +1,51 @@
...
+* The _SkuPriceDetails_ column **SHOULD NOT** contain properties which are already captured in the [_SkuDetails_](#skudetails) column.
```
jpradocueva, line 11:
```suggestion
* The _SkuPriceDetails_ column SHOULD NOT contain properties which are already captured in the [_SkuDetails_](#skudetails) column.
```

specification/columns/skupricedetails.md
```diff
@@ -0,0 +1,51 @@
...
+* Where _SkuPriceDetails_ contains a property with a value which is a numerical, the value provided **MUST** represent the value for a [_ConsumedQuantity_](#consumedquantity) of 1.
```
jpradocueva, line 12:
```suggestion
* Where _SkuPriceDetails_ contains a property with a value which is a numerical, the value provided MUST represent the value for a [_ConsumedQuantity_](#consumedquantity) of 1.
```

specification/columns/skupricedetails.md
```diff
@@ -0,0 +1,51 @@
...
+* The key for a property **SHOULD** remain consistent across comparable SKUs having that property and the values for this key **SHOULD** remain in a consistent format.
```
jpradocueva, line 13:
```suggestion
* The key for a property SHOULD remain consistent across comparable SKUs having that property and the values for this key SHOULD remain in a consistent format.
```

specification/columns/skupricedetails.md
```diff
@@ -0,0 +1,51 @@
...
+* The key for a property **SHOULD** be formatted in PascalCase.
```
jpradocueva, line 14:
```suggestion
* The key for a property SHOULD be formatted in PascalCase.
```

specification/columns/skupricedetails.md
```diff
@@ -0,0 +1,51 @@
...
+* Properties included in _SkuPriceDetails_ **SHOULD** be immutable once included for a _SkuPriceId_.
```
jpradocueva, line 15:
```suggestion
* Properties included in _SkuPriceDetails_ SHOULD be immutable once included for a _SkuPriceId_.
```

jpradocueva (commented):
Although the group has not finalized the editorial guidelines, they have agreed not to use bold formatting for normative keywords like "MUST" and "SHOULD."

cnharris10 (approved):

kk09v:
By vote in TF2 today, people were supportive of including Pascal Case in the glossary. 
- Added a link from these columns to the glossary entry
- Replaced the external link in column_naming_and_ordering with a link to the glossary for Pascal Case

Additionally, there was confusion similar to Rob's above https://github.com/FinOps-Open-Cost-and-Usage-Spec/FOCUS_Spec/pull/503#issuecomment-2249433084 on whether the spec was prohibiting the inclusion of data which is present on other dedicated columns, so a MAY clause was added to each definition for clarity.

specification/columns/skudetails.md
```diff
@@ -0,0 +1,51 @@
+# SKU Details
+
+The **SKU Details** column represents a list of relevant properties shared by all resources having the same [**SKU ID**](#skuid) which are not captured in another FOCUS column. These properties enable a practitioner to more easily calculate metrics such as total units of a service when it is not directly billed in those units (e.g. cores) and thus enables FinOps capabilities such as unit economics.
```
AWS-ZachErdman, line 3: "which are not captured in another FOCUS column."

Can we remove this statement given that we don't want the contents of this column to depend on what other columns are present in FOCUS?

specification/columns/skupricedetails.md
```diff
@@ -0,0 +1,52 @@
...
+
+The **SKU Price Details** column represents a list of relevant properties shared by all resources having the same [**SKU Price ID**](#skupriceid) which are not captured in another FOCUS column. These properties enable a practitioner to more easily understand the differences that make up the multiple **SKU Price ID**s under a given [**SKU ID**](#skuid).
+
+
+The _SkuPriceDetails_ column adheres to the following requirements:
+
+* The properties (both key and value) contained in the _SkuPriceDetails_ column MUST be shared across all resources having the same _SkuPriceId_.
+* The _SkuPriceDetails_ column MUST be in [_KeyValueFormat_](#key-valueformat).
+* The _SkuPriceDetails_ column SHOULD NOT contain properties which are not applicable to the corresponding _SkuPriceId_.
+* The _SkuPriceDetails_ column SHOULD NOT contain properties which are already captured in the [_SkuDetails_](#skudetails) column.
```
flanakin, line 11: FYI - I don't have a way to do this at scale. I get it's a SHOULD NOT, which is fine, but I'm calling it out because I don't see how this will ever be scalable without a significant change across hundreds of teams. I'm assuming other providers are the same.

flanakin, line 11: This presents a significant usability problem for practitioners because they need to intimately understand the difference between SkuId and SkuPriceId, which we don't actually define well in the spec. Splitting attributes across two JSON objects will make it harder for people to find the details they care about because in some cases it may be in one and others in the other. From a usability perspective, this sounds like a horrible idea. 

As a product owner, I would never do this without a usability test. I don't think we should be doing it here either. We should start with one and split it out if there's a reason that doesn't work. If we add 2, then we're creating a usability problem upfront that will be incredibly hard to pull back. It's always easier to add something than remove something, so starting with a single SKU details column is just simpler and safer all-around.

flanakin (changes requested):

specification/columns/skudetails.md
kk09v: Given that it's been confusing to multiple people, I agree that this sentence should be rephrased.

specification/columns/skudetails.md
```diff
@@ -0,0 +1,51 @@
...
+
+The _SkuDetails_ column adheres to the following requirements:
+
+* The properties (both key and value) contained in the _SkuDetails_ column MUST be shared across all resources having the same _SkuId_.
+* The _SkuDetails_ column MUST be in [_KeyValueFormat_](#key-valueformat).
+* The _SkuDetails_ column SHOULD NOT contain properties which are not applicable to the corresponding _SkuId_.
+* The _SkuDetails_ column MAY contain properties which are captured in other dedicated columns.
+* Where _SkuDetails_ contains a property with a value which is a numerical, the value provided MUST represent the value for a [_ConsumedQuantity_](#consumedquantity) of 1.
+* The key for a property SHOULD remain consistent across comparable SKUs having that property and the values for this key SHOULD remain in a consistent format.
+* The key for a property SHOULD be formatted in [PascalCase](#glossary:pascalcase).
```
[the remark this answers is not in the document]
kk09v: People voted for PascalCase last week and I _believe_ this is how it's already formatted in Azure's x_SkuDetails.

kk09v:
Per the direction coming out of TF2 today, I've consolidated the two column definitions into column SkuPriceDetails with the intent to evaluate the need and use cases for separating the 2 columns at a later point.

Minor changes:
- Reordered normative clauses to group logical concepts together
- Rephrased the intro paragraph to reduce confusion

specification/columns/skudetails.md
```diff
@@ -0,0 +1,51 @@
...
+* Properties included in _SkuDetails_ SHOULD be immutable once included for a _SkuId_.
```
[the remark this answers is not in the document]
kk09v: There is a nuance that's lost in this rephrase suggestion, which came out of the conversations while you were on vacation: We can add new properties to SkuDetails, but once they're present, we should not remove them. Open to other suggestions if you don't like "immutable"

specification/columns/skupricedetails.md
kk09v: ACK. Leaving this comment  open for when I try to put together supporting content.

specification/columns/skupricedetails.md
```diff
@@ -0,0 +1,52 @@
...
+The _SkuPriceDetails_ column adheres to the following requirements:
+
+* The _SkuPriceDetails_ column MUST be in [_KeyValueFormat_](#key-valueformat).
+* The key for a property SHOULD be formatted in [PascalCase](#glossary:pascalcase).
+* The properties (both key and value) contained in the _SkuPriceDetails_ column MUST be shared across all resources having the same _SkuPriceId_.
+* The key for a property SHOULD remain consistent across comparable SKUs having that property and the values for this key SHOULD remain in a consistent format.
+* The _SkuPriceDetails_ column SHOULD NOT contain properties which are not applicable to the corresponding _SkuPriceId_.
+* The _SkuPriceDetails_ column MAY contain properties which are already captured in other dedicated columns.
+* Where _SkuPriceDetails_ contains a property with a value which is a numerical, the value provided MUST represent the value for a [_ConsumedQuantity_](#consumedquantity) of 1.
+* Properties included in _SkuPriceDetails_ SHOULD be immutable once included for a _SkuPriceId_.
```
AWS-ZachErdman, line 15: Can we make it clear that some properties are allowed to added - but we don't want existing properties to be changed.

specification/columns/skudetails.md
```diff
@@ -0,0 +1,51 @@
+# SKU Details
+
+The **SKU Details** column represents a list of relevant properties shared by all resources having the same [**SKU ID**](#skuid) which are not captured in another FOCUS column. These properties enable a practitioner to more easily calculate metrics such as total units of a service when it is not directly billed in those units (e.g. cores) and thus enables FinOps capabilities such as unit economics.
+
+
+The _SkuDetails_ column adheres to the following requirements:
+
+* The properties (both key and value) contained in the _SkuDetails_ column MUST be shared across all resources having the same _SkuId_.
```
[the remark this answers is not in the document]
kk09v: We discussed this in Maintainers meeting on 2024-08-12 and the general consensus was that "no data is better than bad data" so we're leaving this as a MUST even if that means that CSPs have no content in this on day 1 because they need to curate what goes into it instead of pumping something that they already have (e.g. Additional_Info in Azure) which may not be a great fit.

specification/columns/skudetails.md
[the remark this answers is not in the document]
kk09v: Noted

specification/columns/skupricedetails.md
kk09v: This is redundant as this clause was eliminated when SKU Details was eliminated in favor of only having the SKU Price Details column.

kk09v:
Did the wordsmithing, attempting to align to the suggestions. The key change was decomposing this statement
> * Properties included in _SkuPriceDetails_ SHOULD be immutable once included for a _SkuPriceId_.

into 3 more explicit statements

> * Additional properties (key-value pairs) MAY be added to _SkuPriceDetails_ for a _SkuPriceId_.
>   * Properties SHOULD NOT be removed from _SkuPriceDetails_ for a _SkuPriceId_, once they have been included.
>   * Individual properties (key-value pairs) SHOULD NOT be modified for a _SkuPriceId_.

@flanakin @AWS-ZachErdman let me know what you think about the above rephrase as you both had different feedback on the original clause.

Other lines were also rephrased:
- Replaced "resources" with "charges" based on how Michael had phrased a suggested change because SKU Price ID represents a charge, not a resource (which may have multiple charges for different components, usage, etc.)
- Rephrased the clause about numerical values per Michael's suggestion to get closer to "plain English"

specification/columns/skupricedetails.md
kk09v: Modified in most recent commit

specification/columns/skudetails.md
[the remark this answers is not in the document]
kk09v: Modified in most recent commit.

specification/columns/skudetails.md
```diff
@@ -0,0 +1,51 @@
...
+* The _SkuDetails_ column MUST be in [_KeyValueFormat_](#key-valueformat).
+* The _SkuDetails_ column SHOULD NOT contain properties which are not applicable to the corresponding _SkuId_.
+* The _SkuDetails_ column MAY contain properties which are captured in other dedicated columns.
+* Where _SkuDetails_ contains a property with a value which is a numerical, the value provided MUST represent the value for a [_ConsumedQuantity_](#consumedquantity) of 1.
```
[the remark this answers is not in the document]
kk09v: Attempted to rephrase the clause in SKU Price Details in most recent commit, based on your feedback here

ljadvey (approved):

rileyjenk (approved):

specification/columns/skupricedetails.md
```diff
@@ -0,0 +1,54 @@
...
+## Description
+
+A set of properties of a **SKU Price ID** which are meaningful and common to all instances of that **SKU Price ID**
+
+## Content Constraints
+
+|    Constraint   |      Value       |
+|:----------------|:-----------------|
+| Column type     | Dimension        |
+| Feature level   | Conditional      |
```
flanakin, line 47: This is listed as conditional, but I don't see any condition for when it's required. When is it required?

specification/columns/skupricedetails.md
```diff
@@ -0,0 +1,54 @@
+# SKU Price Details
+
+The **SKU Price Details** column represents a list of relevant properties shared by all resources having the same [**SKU Price ID**](#skupriceid). These properties enable a practitioner to more easily calculate metrics such as total units of a service when it is not directly billed in those units (e.g. cores) and thus enables FinOps capabilities such as unit economics. They also help a practitioner understand the specifics of a **SKU Price ID**.
+
+
+The _SkuPriceDetails_ column adheres to the following requirements:
+
+* The _SkuPriceDetails_ column MUST be in [_KeyValueFormat_](#key-valueformat).
+* The key for a property SHOULD be formatted in [PascalCase](#glossary:pascalcase).
+* The properties (both keys and values) contained in the _SkuPriceDetails_ column MUST be shared across all charges having the same _SkuPriceId_.
```
flanakin, line 10: I feel like this should be a SHOULD. While I don't want to block on this, I'm not sure whether large cloud providers can feasibly do this. This means that, if we want to add a new attribute, then we have to generate a new SkuPriceId. That would impact many teams and how these things are tracked and managed in these large companies. I suspect that this will end up being a "gap" in any provider due to the scale we're working at.

specification/columns/skupricedetails.md
```diff
@@ -0,0 +1,54 @@
...
+* Properties in _SkuPriceDetails_ which have a numerical value MUST represent the value for a the individual [_ConsumedUnit_](#consumedunit) i.e. a [_ConsumedQuantity_](#consumedquantity) of 1.
+* Additional properties (key-value pairs) MAY be added to _SkuPriceDetails_ for a _SkuPriceId_.
+  * Properties SHOULD NOT be removed from _SkuPriceDetails_ for a _SkuPriceId_, once they have been included.
+  * Individual properties (key-value pairs) SHOULD NOT be modified for a _SkuPriceId_.
+
+## Examples
+
+```json
+{
+    "OperationClass": "A",
```
rileyjenk, line 23: We should likely have a set of examples in the appendix for this.

specification/columns/skupricedetails.md
```diff
@@ -0,0 +1,54 @@
+# SKU Price Details
```
flanakin, line 1: Just to document my position here, I would rather call this "SKU Details" to simplify our overall concepts. I understand it's about the lower level SkuPriceId, but that also adds more complexity to the dataset by adding a third level. I would rather see all SKU properties together in a single bucket without introducing a third level (SKU > properties vs. SKU > Price > properties).

specification/columns/skupricedetails.md
```diff
@@ -0,0 +1,54 @@
...
+
+The **SKU Price Details** column represents a list of relevant properties shared by all resources having the same [**SKU Price ID**](#skupriceid). These properties enable a practitioner to more easily calculate metrics such as total units of a service when it is not directly billed in those units (e.g. cores) and thus enables FinOps capabilities such as unit economics. They also help a practitioner understand the specifics of a **SKU Price ID**.
+
+
+The _SkuPriceDetails_ column adheres to the following requirements:
+
+* The _SkuPriceDetails_ column MUST be in [_KeyValueFormat_](#key-valueformat).
+* The key for a property SHOULD be formatted in [PascalCase](#glossary:pascalcase).
+* The properties (both keys and values) contained in the _SkuPriceDetails_ column MUST be shared across all charges having the same _SkuPriceId_.
+* The key for a property SHOULD remain consistent across comparable SKUs having that property and the values for this key SHOULD remain in a consistent format.
```
flanakin, line 11: Do we want to say it MUST be consistent across rows with the same SkuId? We can add that later in a separate PR. Just thinking out loud.

specification/columns/skupricedetails.md
```diff
@@ -0,0 +1,54 @@
...
+* The _SkuPriceDetails_ column SHOULD NOT contain properties which are not applicable to the corresponding _SkuPriceId_.
```
flanakin, line 12: Shouldn't this be a MUST NOT?
```suggestion
* The _SkuPriceDetails_ column MUST NOT contain properties which are not applicable to the corresponding _SkuPriceId_.
```

specification/columns/skudetails.md
```diff
@@ -0,0 +1,51 @@
+# SKU Details
+
+The **SKU Details** column represents a list of relevant properties shared by all resources having the same [**SKU ID**](#skuid) which are not captured in another FOCUS column. These properties enable a practitioner to more easily calculate metrics such as total units of a service when it is not directly billed in those units (e.g. cores) and thus enables FinOps capabilities such as unit economics.
+
+
+The _SkuDetails_ column adheres to the following requirements:
```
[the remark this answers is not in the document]
flanakin: Resolving this since I added one in SkuPriceDetails

specification/columns/skupricedetails.md
rileyjenk: Feature level is where we specify if it is required, correct?

specification/columns/skupricedetails.md
```diff
@@ -0,0 +1,54 @@
...
+* The _SkuPriceDetails_ column MAY contain properties which are already captured in other dedicated columns.
+* Properties in _SkuPriceDetails_ which have a numerical value MUST represent the value for a the individual [_ConsumedUnit_](#consumedunit) i.e. a [_ConsumedQuantity_](#consumedquantity) of 1.
```
flanakin, line 14: There's a typo with "a the" in here. I still feel like it's a bit confusing so not sure how to change that. I feel like this is more straightforward. I don't have a blocking issue with this text. I just feel it could be a little clearer. Feel free to resolve if it's just me.
```suggestion
* Numeric properties in _SkuDetails_ MUST represent the value for a single [_ConsumedUnit_](#consumedunit).
```

specification/columns/skupricedetails.md
thecloudman: I support this. In the current exports using FOCUS, SkuId and SkuPriceId from Azure are the same. For Google, SkuPriceId is concatenated with additional data to make a new Id that is not the same as SkuId and is blank where SkuId relates to Tax and adjustments or corrections. So if Providers dont have a different Id for each column then SkuId column should be used and SkuDetails would be better to align to SkuId.

specification/columns/skupricedetails.md
```diff
@@ -0,0 +1,54 @@
...
+* Additional properties (key-value pairs) MAY be added to _SkuPriceDetails_ for a _SkuPriceId_.
```
flanakin, line 15: What does "additional" reference? Is this saying we can add to SkuDetails over time, just not remove anything from it? If so, that addresses my primary concern with the MUST statement above. Should we move this up to be a sub-bullet so they're linked more explicitly?

specification/columns/skupricedetails.md
```diff
@@ -0,0 +1,55 @@
+# SKU Price Details
+
+The **SKU Price Details** column represents a list of relevant properties shared by all resources having the same [**SKU Price ID**](#skupriceid). These properties enable a practitioner to more easily calculate metrics such as total units of a service when it is not directly billed in those units (e.g. cores) and thus enables FinOps capabilities such as unit economics. They also help a practitioner understand the specifics of a **SKU Price ID**.
```
flanakin, line 3: I believe this is about "charges" and not resources, given purchases will also have this and resources can have many SkuPriceId rows that may not be directly related to the resource, like bandwidth.
```suggestion
The **SKU Price Details** column represents a list of relevant properties shared by all charges with the same [**SKU Price ID**](#skupriceid). These properties enable practitioners to calculate metrics such as total units of a service when it is not directly billed in those units (e.g. cores) and thus enables FinOps capabilities such as unit economics. They also help a practitioner understand the specifics of a **SKU Price ID**.
```

specification/columns/skupricedetails.md
flanakin: Side note: I believe the SkuId == SkuPriceId in Microsoft data is a bug. I need to investigate. Let's not make any decisions based on that. They should be different.

thecloudman: Ok will wait on your feedback

specification/columns/skupricedetails.md
kk09v: Thank you. I had reflected your terminology feedback in the normative section, but didn't in the introduction. I was looking for a better term for this [since the beginning](https://github.com/FinOps-Open-Cost-and-Usage-Spec/FOCUS_Spec/pull/503#issuecomment-2218426679), but you nailed it.

specification/columns/skupricedetails.md
kk09v: This line is communicating that you shouldn't include a property of {"Cores" : 12} in one SkuId/SkuPriceId and a property of {"vCPUs" : "8.0"} in another related one. I feel that going more strongly on that statement would almost necessitate us to define specific properties we desire to be used.

Reading your comment another way, it's already explicit in the definition of SkuPriceId that it's a child of SkuId: "A given value of SkuPriceId MUST be associated with one and only one SkuId, except in cases of commitment discount flexibility."

specification/columns/skupricedetails.md
kk09v: We previously had 2 columns, SkuDetails and SkuPriceDetails, but after a suggestion and subsequent conversation in TF2, we consolidated them to 1 column. The thought on it being SkuPriceDetails is that if we only have one field, this is more granular and it could hold properties of both SkuId as well as SkuPriceId. SkuDetails wouldn't support any properties of SkuPriceId (e.g. details about tiering start and end values) because that property wouldn't be the same across all SkuPriceIds within that SkuId.

I'm intending a follow-up change to SkuPriceId to indicate that because a number of fields are properties of SkuPriceId (including this and list price, for example), the value for SkuId can be repeated for SkuPriceId, assuming that it doesn't violate any of the corresponding column requirements (one price, one set of properties, etc.)

specification/columns/skupricedetails.md
kk09v: Yes and the condition for it to be present is now included in the normative section.

jpradocueva:
### Action Items, TF-2 call on Aug 21st:
- [ ] [TF1-[#503](https://github.com/FinOps-Open-Cost-and-Usage-Spec/FOCUS_Spec/pull/503)] All Members: To review PR [#503](https://github.com/FinOps-Open-Cost-and-Usage-Spec/FOCUS_Spec/pull/503) and resolve any outstanding comments before the next approval meeting.

jpradocueva:
### Action Items, Members call on Aug 22:
- [ ] [Members-[#503](https://github.com/FinOps-Open-Cost-and-Usage-Spec/FOCUS_Spec/pull/503) ] Karl @kk09v : To address all comments and prepare the PR for review next week.

specification/columns/skupricedetails.md
kk09v: I believe my thinking was that "properties which are not applicable" is loose enough of a statement that it would be hard to say (enforce) a MUST. If your VMs have a property about ephemeral disk, would that property be applicable to D8sv5 which does not have an ephemeral disk? I think that if the D8dsv5 had the property, I'd expect it to be on D8sv5 as well, even though the fact that it doesn't have one could be interpreted to mean that it's not applicable.

However, I sense you're reading this as something egregious like including a property communicating the number of cores on every resource (network, disks, PaaS, etc.). Planning to incorporate this change, but leaving the thread open for a little longer in case more people want to weigh in.

specification/columns/skupricedetails.md
kk09v: That's correct. I'll move it up a few lines and resolve these comments.

specification/columns/skupricedetails.md
kk09v: This is saying that the set of properties (both keys and values) must apply to all occurrences of a SkuPriceId. Resolved by moving other clauses around to provide clarity.

specification/columns/skupricedetails.md
kk09v: A property is the key and value, which is why I was specifying the property's value as numeric. I like the way you've phrased it, so here's my attempt.
"If a property has a numeric value, it must represent the value for a single [_ConsumedUnit_](#consumedunit)."

specification/columns/skupricedetails.md
kk09v: Follow up PR #530 submitted, per above comment

ijurica (approved):

thecloudman (approved):

specification/columns/skupricedetails.md
```diff
@@ -0,0 +1,58 @@
...
+
+
+The _SkuPriceDetails_ column adheres to the following requirements:
+
+* The _SkuPriceDetails_ column MUST be in [_KeyValueFormat_](#key-valueformat).
+* The key for a property SHOULD be formatted in [PascalCase](#glossary:pascalcase).
+* The properties (both keys and values) contained in the _SkuPriceDetails_ column MUST be shared across all charges having the same _SkuPriceId_.
+  * Additional properties (key-value pairs) MAY be added to _SkuPriceDetails_ for a _SkuPriceId_.
+  * Properties SHOULD NOT be removed from _SkuPriceDetails_ for a _SkuPriceId_, once they have been included.
+  * Individual properties (key-value pairs) SHOULD NOT be modified for a _SkuPriceId_.
```
AWS-ZachErdman, line 13: Individual properties (key-value pairs) SHOULD NOT be modified for a given SkuPriceId. They should be the same every time they appear for that SkuPriceId.

specification/columns/skupricedetails.md
```diff
@@ -0,0 +1,58 @@
+# SKU Price Details
+
+The **SKU Price Details** column represents a list of relevant properties shared by all charges with the same [**SKU Price ID**](#skupriceid). These properties enable practitioners to calculate metrics such as total units of a service when it is not directly billed in those units (e.g. cores) and thus enables FinOps capabilities such as unit economics. They also help a practitioner understand the specifics of a **SKU Price ID**.
```
AWS-ZachErdman, line 3: These properties can provide qualitative and quantitative details about the product represented by the SKU. They may enable practitioners to calculate...

jpradocueva:
### Action Items from TF-2 call on Aug 28:
- [ ] [TF2-[#503](https://github.com/FinOps-Open-Cost-and-Usage-Spec/FOCUS_Spec/pull/503)] Karl - @kk09v : To incorporate suggestions from the team and finalize the wording in PR #503.
-[ ] [TF2-[#503](https://github.com/FinOps-Open-Cost-and-Usage-Spec/FOCUS_Spec/pull/503)] Karl - @kk09v : To ensure the final document is consistent, including the use of asterisks for bullet points.
-[ ] [TF2-[#503](https://github.com/FinOps-Open-Cost-and-Usage-Spec/FOCUS_Spec/pull/503)] Karl - @kk09v : To add relevant examples and supporting content in a separate follow-up PR.

specification/columns/skupricedetails.md
```diff
@@ -0,0 +1,58 @@
...
+* The _SkuPriceDetails_ column MUST be in [_KeyValueFormat_](#key-valueformat).
+* The key for a property SHOULD be formatted in [PascalCase](#glossary:pascalcase).
+* The properties (both keys and values) contained in the _SkuPriceDetails_ column MUST be shared across all charges having the same _SkuPriceId_.
+  * Additional properties (key-value pairs) MAY be added to _SkuPriceDetails_ for a _SkuPriceId_.
+  * Properties SHOULD NOT be removed from _SkuPriceDetails_ for a _SkuPriceId_, once they have been included.
+  * Individual properties (key-value pairs) SHOULD NOT be modified for a _SkuPriceId_.
+* The key for a property SHOULD remain consistent across comparable SKUs having that property and the values for this key SHOULD remain in a consistent format.
+* The _SkuPriceDetails_ column MUST NOT contain properties which are not applicable to the corresponding _SkuPriceId_.
+* The _SkuPriceDetails_ column MAY contain properties which are already captured in other dedicated columns.
+* If a property has a numeric value, it must represent the value for a single [_ConsumedUnit_](#consumedunit).
```
udam-f2, line 17: This needs to be pricing unit NOT consumed unit. Pricing unit is what the list unit price correlates to and therefore is the unit of measure that the charge uses.
```suggestion
* If a property has a numeric value, it MUST represent the value for a single [_PricingUnit_](#pricingunit).
```

specification/columns/skupricedetails.md
```diff
@@ -0,0 +1,58 @@
...
+* Additional properties (key-value pairs) MAY be added to _SkuPriceDetails_ for a _SkuPriceId_.
+  * Properties SHOULD NOT be removed from _SkuPriceDetails_ for a _SkuPriceId_, once they have been included.
```
udam-f2, line 19: this is repeated in line 12

specification/columns/skupricedetails.md
```diff
@@ -0,0 +1,58 @@
...
+  * Individual properties (key-value pairs) SHOULD NOT be modified for a _SkuPriceId_.
```
udam-f2, line 20: repeated in line 13

udam-f2 (commented):
nit: I would reorder these so the more definitive statements (MUST) aren't spread out throughout in between other less definitive requirements

specification/columns/skupricedetails.md
kk09v: Yep, this was the conflict I mentioned. All 3 lines got duplicated when merging a suggestion.

specification/columns/skupricedetails.md
kk09v: Took another go at this to incorporate your feedback without breaking into 2 sentences.
"Individual properties (key-value pairs) SHOULD NOT be modified for a _SkuPriceId_ and SHOULD remain consistent over time."

The other aspect of your statement is covered in row 10, which this is a sub-bullet under.

kk09v: Also added "given" to each of these 3 sub-bullets.

kk09v:
> nit: I would reorder these so the more definitive statements (MUST) aren't spread out throughout in between other less definitive requirements

The ordering has gotten flipped around a few times. The general feedback was that grouping clauses together by topic was easier to read and understand than grouping them by level of normativity. I'm not aware of where we take a stand on this in the design guidelines or similar, but happy to conform to guidelines as needed.

jpradocueva:
### Action Items from Members' call on Aug 29:
- [ ] [Members-[#503](https://github.com/FinOps-Open-Cost-and-Usage-Spec/FOCUS_Spec/pull/503)] Karl - @kk09v : Finalize editorial changes and ensure consistent language throughout the document.

kk09v:
Changed verbiage on lines 10 and 11 to reconcile a conflict between the bullet and sub-bullets.

jpradocueva:
### Action Item from Maintainers' call on Sep 2nd:
- [ ] TF1-[#503](https://github.com/FinOps-Open-Cost-and-Usage-Spec/FOCUS_Spec/pull/503) ] Karl, @kk09v : To refine the wording in PR #503 to reflect the agreed-upon approach to consistency in SKU PriceDetails.

specification/columns/skupricedetails.md
```diff
@@ -0,0 +1,55 @@
...
+- The properties (both keys and values) contained in the *SkuPriceDetails* column MUST be shared across all charges having the same *SkuPriceId*, subject to the below provisions.
+  - Additional properties (key-value pairs) MAY be added to *SkuPriceDetails* going forward for a given *SkuPriceId*.
+  - Properties SHOULD NOT be removed from *SkuPriceDetails* for a given *SkuPriceId*, once they have been included.
+  - Individual properties (key-value pairs) SHOULD NOT be modified for a given *SkuPriceId* and SHOULD remain consistent over time.
+- The key for a property SHOULD remain consistent across comparable SKUs having that property and the values for this key SHOULD remain in a consistent format.
+- The *SkuPriceDetails* column MUST NOT contain properties which are not applicable to the corresponding *SkuPriceId*.
+- The *SkuPriceDetails* column MAY contain properties which are already captured in other dedicated columns.
+- If a property has a numeric value, it MUST represent the value for a single [*PricingUnit*](#pricingunit).
+- The *SkuPriceDetils* column MUST be present in the billing data when the provider includes a *SkuPriceID*.
```
ijurica, line 18:
```suggestion
- The *SkuPriceDetails* column MUST be present in the billing data when the provider includes a *SkuPriceId*.
```

Question: 
Should we also specify when it MUST and MUST NOT be null? For instance, _SkuPriceDetails MUST be null when SkuPriceId is null, and MUST NOT be null when SkuPriceId is not null._

kk09v:
Updates since last week for 20240905 meeting:
- Added nullability clauses to make it explicit.
- Made changes to clarify how two clauses relate to one another:
  - The properties (both keys and values) contained in the *SkuPriceDetails* column MUST be shared across all charges having the same *SkuPriceId*, **subject to the below provisions**.
  - Additional properties (key-value pairs) MAY be added to *SkuPriceDetails* **going forward** for a given *SkuPriceId*

jpradocueva:
### Action Items from TF-2 call on Sep 4:
- [ ] [TF2-[#503](https://github.com/FinOps-Open-Cost-and-Usage-Spec/FOCUS_Spec/pull/503)] Karl, @kk09v : Make final edits on capitalization and null conditions before the next review.

udam-f2 (approved):

jpradocueva:
Approved by the Members group on Sep 5.

PR #504, opened 2024-07-11
Proposal for FOCUS v1.1 Scope

jpradocueva:
**Irena** and **Udam** have been working offline to formalize the scope for FOCUS v1.1. This proposal is based on the current discussions and Issues discussed in the three Task Forces (TF-1, TF-2, and TF-3).

In a nutshell, here is what Udam and I are proposing:

> _**SKU and SKU Price Details**_
> _Provide deeper visibility into SKU and SKU Price details (in addition to the current SKU and SKU Price IDs that are available in FOCUS) which enables many standard FinOps use cases from cost reporting to commitment optimization._
>
>_**Deeper service classification**_
>_Move beyond the highest-level service categorization to a sub-categorization for services while adding flexibility for providers to share their native categorizations within FOCUS._
>
>_**Capacity Reservation**_
>_Ability to represent on-demand capacity reservations within FOCUS datasets_
>
>_**Improved metadata support**_
>_Strengthen the correlation between FOCUS-compatible datasets and the corresponding metadata to better support data analysis and ETL use cases._

If everyone agrees, we could add that to the Release Planning as a high-level scope for FOCUS 1.1.

### DIFF
--- a/RELEASE-PLANNING.md
+++ b/RELEASE-PLANNING.md
@@ -12,7 +12,7 @@ This section outlines the planned release schedule and key milestones for the FO
         <td>v1.1</td>
         <td>Nov 11, 2024</td>
         <td>
-            <strong>SKU and SKU Price Details<strong>
+            <strong>SKU and SKU Price Details</strong>
             <ul>
                 <li>Provide deeper visibility into SKU and SKU Price details (in addition to the current SKU and SKU Price IDs that are available in FOCUS) which enables many standard FinOps use cases from cost reporting to commitment optimization.
                 </li>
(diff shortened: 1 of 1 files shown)


cnharris10:
I want to push back on excluding the commitment utilization topic within the v1.1 release.

From today's member call, it was explained that there's ambiguity on how to model this feature around SaaS, so due to this, the topic will continue to be a priority, but not an official commitment.

Given that we've gained clarity from the cake vs. ingredients conversations to unblock moving forward on this topic, I believe there's enough of a headstart to include this topic as part of the 1.1 release that primarily focuses on CSP's without inhibiting a SaaS focus in a future iteration.

jpradocueva:
The group agreed this PR during the Members' call on July 18.

jpradocueva:
@udam-f2, could you please approve this PR? It was agreed by the Members on the July 18 call. Thanks

udam-f2 (approved):

PR #506, opened 2024-07-11
[EDITORIAL] Reclassify ConsumedUnit as Dimension instead of Metric

kk09v:
Correct ConsumedUnit column type to Dimension from Metric (similar to PricingUnit) because it is an attribute of how the consumption is measured and is not able to have mathematical operations applied to it.


cnharris10 (approved):

rileyjenk (approved):

udam-f2 (approved):

jpradocueva:
Approved by the Members' group on July 18.

Issue #493, opened 2024-07-01
[ADMIN] FOCUS Roles & Tasks


jpradocueva:
### Description
The FOCUS project is structured around a framework of roles critical to its operational excellence and strategic direction, including Working Group Members Chair, Task Force Conveners, Maintainers, Editors, Members, and the Program Manager. Each role contributes uniquely, ranging from overseeing daily operations and maintaining documentation standards to leading strategic initiatives and ensuring alignment with project goals. This delineation of responsibilities is crucial for maintaining the project’s integrity, ensuring quality standards, and fostering a collaborative and flexible environment that responds to the evolving demands of the FinOps community.

### Proposed approach
This document, [FOCUS Roles & Tasks](https://docs.google.com/document/d/1bb7GstMwP2Rn4eA7OX0pb8f6d7R4iwKHZyZCc4gTt2g/edit) contains a first draft of FOCUS Roles, Tasks and Responsibilities that need to be analyzed by the members.

### Github issue or Reference
The document containing the description can be found on FOCUS Google Workspace: FOCUS Groups / Discussion-Docs/ [FOCUS Roles & Tasks](https://docs.google.com/document/d/1bb7GstMwP2Rn4eA7OX0pb8f6d7R4iwKHZyZCc4gTt2g/edit)

### Context
It is important to have a clear definition of the tasks and responsibilities within the group.

jpradocueva:
PR [FOUNDATION#10](https://github.com/FinOps-Open-Cost-and-Usage-Spec/foundation/pull/10)

jpradocueva:
Latest list of Roles vs Tasks: [Roles & Tasks](https://docs.google.com/spreadsheets/d/1FOqX3w9l3IjBtUDRJMLv3AVWk4fNuirReYlN5LZ5rls/edit?gid=0#gid=0)

shawnalpay:
@jpradocueva Now that https://github.com/FinOps-Open-Cost-and-Usage-Spec/foundation/pull/10 was merged, can we close this issue?

jpradocueva:
@shawnalpay, I see. the issue didn't close automatically as the PR was in a different repo.

PR #507, opened 2024-07-12
FOCUS #493: ConsumedQuantity and ConsumedUnit update - addressing Unused usage

ijurica:
Currently in normative paragraphs for both ConsumedQuantity and ConsumedUnit:

> _MUST NOT be null if ChargeCategory is "Usage" and ChargeClass not "Correction". This column MUST be null for other ChargeCategory values._

We need to adjust this considering both ConsumedQuantity and ConsumedUnit MUST be null in case of ChargeCategory 'Usage' and CommitmentStatus 'Unused'.

### DIFF
--- a/specification/columns/consumedquantity.md
+++ b/specification/columns/consumedquantity.md
@@ -2,7 +2,16 @@
 
 The Consumed Quantity represents the volume of a given SKU associated with a [*resource*](#glossary:resource) or [*service*](#glossary:service) used, based on the [Consumed Unit](#consumedunit). Consumed Quantity is often derived at a finer granularity or over a different time interval when compared to the [Pricing Quantity](#pricingquantity) (complementary to [Pricing Unit](#pricingunit)) and focuses on *resource* and *service* consumption, not pricing and cost.
 
-ConsumedQuantity column MUST be present in the billing data when the provider supports the measurement of usage. This column MUST NOT be null if [ChargeCategory](#chargecategory) is "Usage" and  [ChargeClass](#chargeclass) is not "Correction". This column MUST be null for other ChargeCategory values. This column MUST be of type Decimal and MUST conform to [Numeric Format](#numericformat) requirements. The value MAY be negative in cases where [ChargeClass](#chargeclass) is "Correction".
+The ConsumedQuantity column adheres to the following requirements:
+
+* ConsumedQuantity MUST be present in the billing data when the provider supports the measurement of usage.
+* ConsumedQuantity MUST be of type Decimal and MUST conform to [Numeric Format](#numericformat) requirements.
+* ConsumedQuantity MUST NOT be null if [ChargeCategory](#chargecategory) is "Usage", unless [ChargeClass](#chargeclass) is "Correction" or [CommitmentStatus](#ommitmentstatus) is "Unused".
+* ConsumedQuantity MAY be null if ChargeCategory is "Usage" and ChargeClass is "Correction"
+* ConsumedQuantity MUST be null if CommitmentStatus is "Unused" or for other ChargeCategory values.
+* ConsumedQuantity value MAY be negative in cases where ChargeClass is "Correction".
+
+The ConsumedQuantity column MUST NOT be used to determine values related to any pricing or cost metrics.
 
 ## Column ID
 
--- a/specification/columns/consumedunit.md
+++ b/specification/columns/consumedunit.md
@@ -2,7 +2,15 @@
 
 The Consumed Unit represents a provider-specified measurement unit indicating how a provider measures usage of a given SKU associated with a [*resource*](#glossary:resource) or [*service*](#glossary:service). Consumed Unit complements the [Consumed Quantity](#consumedquantity) metric. It is often listed at a finer granularity or over a different time interval when compared to [Pricing Unit](#pricingunit) (complementary to [Pricing Quantity](#pricingquantity)), and focuses on *resource* and *service* consumption, not pricing and cost.
 
-The ConsumedUnit column MUST be present in the billing data when the provider supports the measurement of usage. This column MUST be of type String. ConsumedUnit MUST NOT be null if [ChargeCategory](#chargecategory) is "Usage" and [ChargeClass](#chargeclass) is not "Correction". This column MUST be null for other ChargeCategory values. Units of measure used in ConsumedUnit SHOULD adhere to the values and format requirements specified in the [UnitFormat](#unitformat) attribute. The ConsumedUnit column MUST NOT be used to determine values related to any pricing or cost metrics.
+The ConsumedQuantity column adheres to the following requirements:
+
+* ConsumedUnit MUST be present in the billing data when the provider supports the measurement of usage.
+* ConsumedUnit MUST be of type String, and the units of measure used in ConsumedUnit SHOULD adhere to the values and format requirements specified in the [UnitFormat](#unitformat) attribute.
+* ConsumedUnit MUST NOT be null if [ChargeCategory](#chargecategory) is "Usage", unless [ChargeClass](#chargeclass) is "Correction" or [CommitmentStatus](#ommitmentstatus) is "Unused".
+* ConsumedUnit MAY be null if ChargeCategory is "Usage" and ChargeClass is "Correction"
+* ConsumedUnit MUST be null if CommitmentStatus is "Unused" or for other ChargeCategory values.
+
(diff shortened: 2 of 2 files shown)


ahullah:
Hey @ijurica this makes sense to me, it keeps ConsumedQuantity and ConsumedUnit clean for easy comparison with our regular qty fields
> Overview:
> 
> ChargeCategory	ChargeClass	CommitmentStatus	ConsumedQuantity	ConsumedUnit
> Usage	(null)	Used	MUST not be null	MUST not be null
> Usage	Correction	Used	MAY be null	MAY be null
> Usage	(null)	Unused	MUST be null	MUST be null
> Usage	Correction	Unused	MUST be null	MUST be null
> Purchase			MUST be null	MUST be null
> Credit			MUST be null	MUST be null
> Adjustment			MUST be null	MUST be null
> Tax			MUST be null	MUST be null

ijurica:
Although he is on vacation, @rileyjenk  Jenkins provided detailed feedback on the topic. Unfortunately, I did not register this in time 😞 . Original comment added to the Issue and I am adding a copy so that we can continue the conversation here on the PR. 

> In the case of unused resource commitments such as RIs, ConsumedUnit and ConsumedQuantity should not be null, as the intent was for the unit to be representative of the commitment amount that when unused. For example if I purchase an m6g.xlarge RI and don't use it, then theses columns would represent the number of compute hours that went unused.
> 
> Here is a link to the raw example data: https://docs.google.com/spreadsheets/d/13FdQoNtsJbqawxL0hPiTg7ZdNnHUs9goJj_tatsBtME/edit?gid=0#gid=0
> 
> For spend commitments that are time isolated, like savings plans, we need to consider how those are represented. The issues comes into view when say for that same instance it is now covered by a savings plan. if ConsumedUnit is now dollars (which it should not be), then the practitioner is now unable to sum the total number of compute hours for something like ec2 because the units are not the same thing. It would be compute hours.
> 
> It is for this reason that we likely need to consider dedicated columns for commitment quantity column/s for providing this information.
> 
> And the there is SAAS:
> 
> One of the principle areas that we do not support at multiple levels is where a company has a non currency unit that is purchased upfront and then "consumed" like a currency by various usage. The naming conflict is unfortunate but when a "credit", "point", "token", or "" is consumed rather than an actual currency the consumed unit wasn't really intended to be that unit. That column is intended to represent the unit that was used.
> 
> Take for example that you do not have direct to dollars and you have credits. Say that those credits are used up by different features at different rates (e.g. https://docs.snowflake.com/en/user-guide/cost-understanding-compute). If the provider puts the consumed Unit as credits it does not reflect the usage, only the credits. If I wanted to know how many compute hours I used, I would not be able to.
> 
> It is for this reason that in the example of a made up currency unit like credits, tokes, etc... we need specific support for this scenario. We need support for expressing the purchase of those units, and their burn down.

cnharris10:
> Although he is on vacation, @rileyjenk Jenkins provided detailed feedback on the topic. Unfortunately, I did not register this in time 😞 . Original comment added to the Issue and I am adding a copy so that we can continue the conversation here on the PR.
> 
> > In the case of unused resource commitments such as RIs, ConsumedUnit and ConsumedQuantity should not be null, as the intent was for the unit to be representative of the commitment amount that when unused. For example if I purchase an m6g.xlarge RI and don't use it, then theses columns would represent the number of compute hours that went unused.
> > Here is a link to the raw example data: https://docs.google.com/spreadsheets/d/13FdQoNtsJbqawxL0hPiTg7ZdNnHUs9goJj_tatsBtME/edit?gid=0#gid=0
> > For spend commitments that are time isolated, like savings plans, we need to consider how those are represented. The issues comes into view when say for that same instance it is now covered by a savings plan. if ConsumedUnit is now dollars (which it should not be), then the practitioner is now unable to sum the total number of compute hours for something like ec2 because the units are not the same thing. It would be compute hours.
> > It is for this reason that we likely need to consider dedicated columns for commitment quantity column/s for providing this information.
> > And the there is SAAS:
> > One of the principle areas that we do not support at multiple levels is where a company has a non currency unit that is purchased upfront and then "consumed" like a currency by various usage. The naming conflict is unfortunate but when a "credit", "point", "token", or "" is consumed rather than an actual currency the consumed unit wasn't really intended to be that unit. That column is intended to represent the unit that was used.
> > Take for example that you do not have direct to dollars and you have credits. Say that those credits are used up by different features at different rates (e.g. https://docs.snowflake.com/en/user-guide/cost-understanding-compute). If the provider puts the consumed Unit as credits it does not reflect the usage, only the credits. If I wanted to know how many compute hours I used, I would not be able to.
> > It is for this reason that in the example of a made up currency unit like credits, tokes, etc... we need specific support for this scenario. We need support for expressing the purchase of those units, and their burn down.

All of these points are being discussed/considered in TF1 around commitment utilization for CSP's and SaaS.  I think we should combine efforts in one of the TF's to ensure that alignment occurs.  As Irena and I have at least start incorporating OCI and SaaS models, the accounting of these providers makes tracking unused amounts of variable timeframes more complicated.

I'm not entirely sure that an "Unused" row is the holistic answer but instead of other routes like a new ChargeCategory value and/or additional columns.  We should dicuss.

jpradocueva:
### Action items, TF-3, July 12:
- [ ] TF3-[#507](https://github.com/FinOps-Open-Cost-and-Usage-Spec/FOCUS_Spec/pull/507) Irena: to check and correct any inconsistencies in the supporting content of the PR.
- [ ] TF3-[#507](https://github.com/FinOps-Open-Cost-and-Usage-Spec/FOCUS_Spec/pull/507) Team members to provide feedback on the PR, preferably directly on the PR itself.

ijurica:
A fictional FOCUS dataset that we could use as a basis for further Consumed Quantity/Unit discussion is available [here](https://docs.google.com/spreadsheets/d/1C9fSLKDz7kY1ln9Hlu08xd6acHiSmpbLLuQco2Davng/edit?gid=0#gid=0).

jpradocueva:
### Action Item from TF-3, July 19 call:
Action Items:
- [ ] TF3-[#507](https://github.com/FinOps-Open-Cost-and-Usage-Spec/FOCUS_Spec/pull/507) PR [#507](https://github.com/FinOps-Open-Cost-and-Usage-Spec/FOCUS_Spec/pull/507) to be put on hold.
- [ ] TF3-[#507](https://github.com/FinOps-Open-Cost-and-Usage-Spec/FOCUS_Spec/pull/507) Chris and Irena: To lead the commitment utilization discussions and update the PR accordingly.

flanakin:
> Single purpose: When we were deciding whether to use Consumed Quantity and Unit for purchase-related records, one of the main criteria was ... Strive to build columns that serve a single purpose.... Unused usage-related records do not report on consumed but unused usage, similar to how commitment purchase-related records could report on consumable usage.

I don't quite understand why "unused" usage should not be tracked consistently. This makes it more difficult to quantify overall usage. If someone doesn't want unused, then they already have a means to filter it out. This feels like we're making things more complex by adding extra rules to support a simple summation that can never happen "simply" because of the need to group by units and some other column that isn't even in FOCUS yet to make sense of what we're grouping since not all units are created equal.

Do we have a clear understanding of the scenarios where someone would want unused included in the quantity and when they wouldn't?

supporting_content/columns/consumedquantity.md
```diff
@@ -16,3 +16,16 @@ Current column mappings found in available data sets:
...
 * The second column needed for the staircase pricing scenario above, a potential 'DistinctPricingQuantity' column, could be introduced in the future to identify the 60 seconds that you were 'charged' for in singular units. Today, you may have to use pricing quantity - which may have block pricing - therefore you may may need to do math to get 'distinct' units). Alternatively, we could introduce a 'factor' to remove the block portion of the pricing unit. This need was discussed and determined to be something that needs to be thought through and evaluated in detail post v1.0.
+
+## ConsumedQuantity and ConsumedUnit normatives overview
+
+| ChargeCategory | ChargeClass | CommitmentStatus | ConsumedQuantity | ConsumedUnit     |
```
salpaysenturus, line 22:
```suggestion
| ChargeCategory | ChargeClass | CommitmentDiscountStatus | ConsumedQuantity | ConsumedUnit     |
```

ijurica:
This issue is addressed in a separate PR and this PR can be closed.

shawnalpay:
Closing per @ijurica 's comment above that this was addressed in a separate PR.

PR #509, opened 2024-07-12
#397 first drafts

jpradocueva:
First capacity reservation column drafts

Issue #397, opened 2024-04-13
How to represent ODCRs in FOCUS


AWS-ZachErdman:
* **Maintainer Leader**: Alex
### Documents:
* [#397 ](https://github.com/FinOps-Open-Cost-and-Usage-Spec/FOCUS_Spec/issues/397)How to represent ODCRs in FOCUS by [zserdman@amazon.com](mailto:zserdman@amazon.com)(Capacity Reservation)
* Folder: [#397-ODCR](https://drive.google.com/drive/folders/1UzTl4LBBd47g_bAY-sYqGEf6MiZuPhHe?usp=sharing)
[FOCUS OCDRs](https://docs.google.com/spreadsheets/d/1LdgZCBuV6TDehgykQ4m6AluOLE_TjX9kFxKXymXn8I4/edit?gid=1124060342#gid=1124060342)  (this document was not created inside of FOCUS workspace)
* Option 1: ~PR [#509](https://github.com/FinOps-Open-Cost-and-Usage-Spec/FOCUS_Spec/pull/509) How to represent ODCRs in FOCUS by [zserdman@amazon.com](mailto:zserdman@amazon.com)~ DEPRACATED.
* Option 2: ~PR [#515](https://github.com/FinOps-Open-Cost-and-Usage-Spec/FOCUS_Spec/pull/515) How to represent ODCR in FOCUS (Alternative) by [alex.hullah@gs.com](mailto:alex.hullah@gs.com)~ DEPRACATED.
* Option 3: PR [#523](https://github.com/FinOps-Open-Cost-and-Usage-Spec/FOCUS_Spec/pull/523) Capacity Reservation Proposal for ID & Status 
* AWS: [On-Demand Capacity Reservations](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/ec2-capacity-reservations.html)
* Azure: [On-Demand Capacity Reservation](https://learn.microsoft.com/en-us/azure/virtual-machines/capacity-reservation-overview)
* Document: [How to represent ODCRs in FOCUS](https://docs.google.com/document/d/1wNKxqfqltfEFN1tdnswsNAqQUbSyXN28UAPUVbxZDpA/edit#heading=h.el7tqolosins)
* Slack: “[tf-odcr](https://f2-focus.slack.com/archives/C07GAV6UA3C)”
* Approval Deadline: ~Aug 29~, Sep 5

### Description
[...]

PR #515, opened 2024-08-02
How to represent ODCRs in FOCUS (Alternate)

ahullah:
Alternative recommendation from Chris for a more flexible/generic way to incorporate ODCRs without requiring single use columns

### DIFF
--- a/specification/columns/relatedresourceid.md
+++ b/specification/columns/relatedresourceid.md
@@ -1,0 +1,31 @@
+# Related Resource ID
+
+Related Resource ID is populated when a usage resource line item is impacted by another line item that modifies its billing behaviour, e.g. when the charge has consumed a [*capacity reservation*](#glossary:capacity-reservation).
+
+The RelatedResourceID column MUST be present in the billing data when the provider supports *capacity-reservations*. This column MUST be of type String, MUST be null when the charge is not related to another line item that impacts its billing behaviour, and MUST NOT be null when the charge is related to another line item that impacts its billing behaviour e.g a capacity reservation. The RelatedResourceID MUST match the ResourceID of thhe line item that has impacted the billing behaviour.
+
+## Column ID
+
+RelatedResourceID
+
+## Display Name
+
+Related Resource ID
+
+## Description
+
+Indicates whether the charge represents either the consumption of a *capacity reservation* or when a *capacity reservation* is unused.
+
+## Content constraints
+
+| Constraint      | Value            |
+| :-------------- | :--------------- |
+| Column type     | Dimension        |
+| Feature level   | Conditional      |
+| Allows nulls    | True             |
+| Data type       | String           |
+| Value format    | \<not specified> |
+
+## Introduced (version)
+
+1.1
--- a/specification/columns/relatedresourceid.md
+++ b/specification/columns/relatedresourceid.md
@@ -2,7 +2,7 @@
 
 Related Resource ID is populated when a usage resource line item is impacted by another line item that modifies its billing behaviour, e.g. when the charge has consumed a [*capacity reservation*](#glossary:capacity-reservation).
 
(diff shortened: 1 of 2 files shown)


jpradocueva:
### Summary:
The group discussed two alternative approaches for representing On-Demand Capacity Reservations (ODCR) in FOCUS: creating new columns for capacity reservations or using a related resource ID.
### Discussion:
* Alex presented two approaches:
   * PR [#509](https://github.com/FinOps-Open-Cost-and-Usage-Spec/FOCUS_Spec/pull/509): Creating new columns specific to capacity reservations
   * PR [#515](https://github.com/FinOps-Open-Cost-and-Usage-Spec/FOCUS_Spec/pull/515): Using a generic related resource ID
* The merits of each approach were debated, considering future-proofing and flexibility.
* The related resource ID approach was seen as more flexible but potentially more complex.
### Questions & Answers:
* Q: Is the related resource ID meant to cover various types of relationships?
* A: Yes, but it needs to be specific to avoid ambiguity.
* Q: Should we stick to single-use columns for clarity?
* A: This approach might be more straightforward for initial implementation.
### Agreements Reached:
* Further discussion is needed to decide the best approach.
* The importance of clearly defining the relationships and their allowable values was acknowledged.
### Action Items, TF-3, Aug 2nd call:
- [ ]  All Members: Review and provide feedback on both approaches to determine the best path forward.

AWS-ZachErdman:
this makes sense as an alternative for adding an ID column, but this PR is missing the use cases of:
1/ used vs unused status
2/ capacity total

specification/columns/relatedresourceid.md
```diff
@@ -0,0 +1,31 @@
+# Related Resource ID
+
+Related Resource ID is populated when a usage resource line item is impacted by another line item that modifies its billing behaviour, e.g. when the charge has consumed a [*capacity reservation*](#glossary:capacity-reservation).
+
+The RelatedResourceID column MUST be present in the FOCUS data when the provider supports *capacity-reservations*. This column MUST be of type String, MUST be null when the charge is not related to another line item that impacts its billing behaviour, and MUST NOT be null when the charge is related to another line item that impacts its billing behaviour e.g a capacity reservation. The RelatedResourceID MUST match the ResourceID of thhe line item that has impacted the billing behaviour.
```
AWS-ZachErdman, line 5: We need to allow that some providers will support Capacity reservations, but will not support labeling of the usage line items with the ARN of a capacity reservation (which is all providers currently).

It's for this reason that I dont think this column is strictly necessary yet.

specification/columns/relatedresourceid.md
```diff
@@ -0,0 +1,31 @@
+# Related Resource ID
```
flanakin, line 1: This assumes there can only ever be a single relationship. This feels like it won't scale. We could add it as a JSON object, but it seems like we'd need to put more thought into it before going that route.

jpradocueva:
### Agreements Reached during TF-3, Aug 16:
* PR [#515](https://github.com/FinOps-Open-Cost-and-Usage-Spec/FOCUS_Spec/pull/515) will not be pursued further.
* FOCUS will be on developing a column for ODCR status, with the ID column introduced as a "SHOULD" rather than a "MUST."
* JSON was ruled out as an option for storing ODCR data due to the complexity and lack of standardization.
### Action Items:
- [ ] [TF3-[#397](https://github.com/FinOps-Open-Cost-and-Usage-Spec/FOCUS_Spec/issues/397)] George to draft a PR for the new “ID” and “Status” columns related to ODCRs.
* [ ] [TF3-[#397](https://github.com/FinOps-Open-Cost-and-Usage-Spec/FOCUS_Spec/issues/397)] to consult AWS for further guidance on the feasibility of providing a CR ID in the billing data.

jpradocueva:
Topic superseded by https://github.com/FinOps-Open-Cost-and-Usage-Spec/FOCUS_Spec/pull/523

PR #543, opened 2024-09-19
Cleanup Metadata PR to fix content bugs/errors, broken links, reduce nesting, formatting

udam-f2:
Fixes include fixing missing content linking, broken links, formatting, and some content bugs.


jpradocueva:
### Action Items from TF-1 call on Sep 24:
- [ ] [TF1-[#543](https://github.com/FinOps-Open-Cost-and-Usage-Spec/FOCUS_Spec/pull/543) ] Chris, @cnharris10 : Finalize the metadata cleanup PR and ensure all issues are addressed before the next meeting.

jpradocueva:
### Action items from Members' call on Sep 26:
- [ ] [TF1-[ #543](https://github.com/FinOps-Open-Cost-and-Usage-Spec/FOCUS_Spec/pull/543)] Members to review the changes offline and approve or provide feedback.

shawnalpay:
@rileyjenk @udam-f2 Are `provider` and `generator` synonyms?  I ask because I'm starting at the top of the metadata section of the PDF (direct link to latest for convenience [here](https://github.com/FinOps-Open-Cost-and-Usage-Spec/FOCUS_Spec/actions/runs/11113240942/artifacts/1997455085)), and it starts with (emphasis mine):

> The FOCUS specification defines a metadata structure to be supplied by data **providers** to facilitate practitioners' use of FOCUS data. This metadata includes general information about the data **generator** and the schema of the FOCUS dataset.

Does `generator` perhaps deserve a glossary entry in the same way that `provider` does?  I recognize that we did not introduce this term with this section, but I did get tripped up by this upon review, and I wonder if other readers might experience the same.

jpradocueva:
### Action Items from TF-1 call on October 1st:
- [ ] [TF1-[#543](https://github.com/FinOps-Open-Cost-and-Usage-Spec/FOCUS_Spec/pull/543)] Team: Review the updated PR before the next meeting.

shawnalpay:
Is there any way to avoid page breaks like this in the PDF generation, or is that just the way it has to be?

Screenshot 2024-10-02 at 9 20 02 AM

shawnalpay:
Another odd page break.  Anything to be done about it, or no?

Screenshot 2024-10-02 at 9 25 51 AM

udam-f2:
> Is there any way to avoid page breaks like this in the PDF generation, or is that just the way it has to be?
> 

Good catch, but unfortunately this seems to be an artifact of building on Ubuntu (which also uses the latest pandoc). My local build doesn't have this issue. I tried moving the github pandoc to several prior version including the version i have locally and cannot get the build to produce things the same way my mac (pandoc 3.1.12.1) does it. Locally, I don't see the second problem and the first one is very slightly visible.

I think we'll have to live with this for now unless Mike or someone with pandoc expertise can find a solution. We can force page breaks but that can be risky as other content above pushes or pulls content up/down - the page break may just add unnecessary whitespace.

shawnalpay:
> My local build doesn't have this issue.

Is it "allowed" for the release candidate (and final spec) to be locally generated, or does it have to run through the Github pandoc?

ijurica (approved):

ljadvey (approved):

jpradocueva:
image

jpradocueva:
### Action Items from the Member's call on October 3rd:
- [ ] [Maintainers-[#543](https://github.com/FinOps-Open-Cost-and-Usage-Spec/FOCUS_Spec/pull/543) ] Udam, @udam-f2  to validate the PDF generation process for future updates.

PR #552, opened 2024-09-25
Service Subcategory language inconsistency fixes

udam-f2:
Fixes based on discussions at the consistency review meeting on 9/25.


specification/columns/servicesubcategory.md
```diff
@@ -2,10 +2,12 @@
...
 
-* The ServiceSubcategory column MUST be present in a FOCUS dataset and MUST NOT be null.
-* This column is of type String and MUST be one of the allowed values.
-* Each Service Subcategory value MUST have one and only one Service Category parent.
-* Though a given *service* can have multiple purposes, each *service* SHOULD have one and only one Subcategory that best aligns with its primary purpose.
+The ServiceSubcategory column adheres to the following requirements:
+
+* ServiceSubcategory is RECOMMENDED to be present in a FOCUS dataset and MUST NOT be null.
+* ServiceSubcategory is of type String and MUST be one of the allowed values.
+* Each ServiceSubcategory value MUST have one and only one ServiceCategory parent.
```
ijurica, line 9: just a suggestion

```suggestion
* Each ServiceSubcategory value MUST have one and only one parent ServiceCategory.
```

ijurica (approved):

specification/columns/servicesubcategory.md
ijurica: Question: 
Could/should we specify that the **parent is defined by FOCUS**? Meaning that providers could conform to this requirement while continuously providing a ServiceSubcategory: "Search" with a ServiceCategory: "Web"  ;)

salpaysenturus: I think that's a good point, Irena!

specification/columns/servicesubcategory.md
```diff
@@ -2,10 +2,12 @@
...
+* Though a given *service* can have multiple purposes, each *service* SHOULD have one and only one ServiceSubcategory that best aligns with its primary purpose.
```
ijurica, line 10: **One more question:** 
What do you think about **referencing the `ServiceName` column as well, rather than just the term in the glossary**? 
Note: For `ServiceCategory`, we also referenced only the term in the glossary and not the `ServiceName` column

I’m sure there’s a better way to do it if we decide to... 

```suggestion
* Though a given *service* can have multiple purposes, each *service* ([ServiceName](#servicename) SHOULD have one and only one ServiceSubcategory that best aligns with its primary purpose.
```

arrarama: How would this work with services that depend on other services? Example Data Science and Kubernetes may depend on compute?

jpradocueva:
### Action Items from Members' call on Sep 26:
- [ ] [TF1- [#552](https://github.com/FinOps-Open-Cost-and-Usage-Spec/FOCUS_Spec/pull/552)] Team: Update terminology and include glossary links where necessary.

specification/columns/servicesubcategory.md
udam-f2: I'm in favor of keeping this as is. I see the point but this is consistently the way we've done this for Charge/Billing Periods, SKUs etc. If we do want to move forward with this, I might suggest the following:

(Same change would need to be made for ServiceCategory if we're going forward with this) 

```suggestion
* Though a given *service* can have multiple purposes, each *service* (identified by columns like [ServiceName](#servicename) SHOULD have one and only one ServiceSubcategory that best aligns with its primary purpose.
```

udam-f2: > How would this work with services that depend on other services? Example Data Science and Kubernetes may depend on compute?

Hey @arrarama .. the goal of ServiceCategory and Service Subcategory is to categorize based on the primary purpose of that service (regardless of the underlying services used). E.g. a relational database service is built based on compute, storage, and networking foundational services. However, its primary purpose is to serve database needs - thus it should use the database service category.

There's a separate discussion about being charged for different SKUs within the service - like compute or storage related SKUs that have pricing attached. That would eventually be captured at the SKU level (SkuCategory or something along those lines) whereas this category is for the service-level.

specification/columns/servicesubcategory.md
udam-f2:
```suggestion
* Each ServiceSubcategory value MUST have one and only one parent ServiceCategory as specified in the allowed values below.
```

jpradocueva:
### Action Items from TF2 call on October 1st:
- [ ] [TF1-[#552](https://github.com/FinOps-Open-Cost-and-Usage-Spec/FOCUS_Spec/pull/552)] Udam, @udam-f2 : Implement the agreed language changes.

jpradocueva:
Approve on Members' call on October 3

PR #558, opened 2024-09-26
Repaired the table format in supporting content for SubaccountID

shawnalpay:
Someone in the FOCUS User Group was asking about Resource Group, and I pointed them to the supporting content for SubaccountID.  There, I noticed that the table format was broken.  I just repaired it.


jpradocueva:
Approved by the group on Sep 26.

Issue #606, opened 2024-10-15
[DISCUSSION]: 24.09.27 - Columns – Editorial PRs 


jpradocueva:
### Document:
* Spreadsheet: [24.09.27 - Columns – Editorial PRS](https://docs.google.com/spreadsheets/d/1w797BvZFsswVPjD99gON1w5YXebgO9Jk9ZNOPmD67cY/edit?usp=drive_link) created by Joaquin.

### Proposed Approach

The group agreed to start applying the Editorial Guidelines after finalizing the development of Release v1.1.

jpradocueva:
### Action Items assigned during TF-2 call on Oct 15:
- [ ] [TF-1-[#606](https://github.com/FinOps-Open-Cost-and-Usage-Spec/FOCUS_Spec/issues/606)] Joaquin, @jpradocueva: Continue creating PRs (as per spreadsheet) changing verbose normative statements to bullet-point sentences.

jpradocueva:
### Summary Members' Meeting Oct 17:
Primary Issue: Control of editorial changes and column updates across multiple pull requests. Spreadsheet: [24.09.27 - Columns – Editorial PRS](https://docs.google.com/spreadsheets/d/1w797BvZFsswVPjD99gON1w5YXebgO9Jk9ZNOPmD67cY/edit?usp=drive_link) 
Core Problem: Managing changes across numerous columns and creating a baseline for version 1.2.
Divergent Views: No significant opposition, but a thorough review is required to ensure consistency.
Final Agreement: Editorial pull requests will continue to be reviewed by the group, with a target for completion within two weeks.
Action Items:
- [ ] [Members-[#606](https://github.com/FinOps-Open-Cost-and-Usage-Spec/FOCUS_Spec/issues/606)] Irena @ijurica , Joaquin @jpradocueva : Prepare and finalize editorial changes over the next two weeks.

jpradocueva:
### Action Items from Members' call on Oct 31:
- [ ] [[#606](https://github.com/FinOps-Open-Cost-and-Usage-Spec/FOCUS_Spec/issues/606) ] All members: Review the listed PRs in the spreadsheet and provide approval or feedback.
[...]

Issue #557, opened 2024-09-26
[Work_Item] Refine normative requirements for all existing spec content


ijurica:
## 1. **Problem Statement** *
> Describe the problem, issue, use case, or opportunity that this work item addresses.
> Include practitioner quotes illustrating real examples a) of questions being asked by practitioners and b) value unlocked by answering these questions, if available.
> - What is the problem?: Explain the context and why it needs resolution.
> - Impact: Describe how the problem affects users, systems, or the project.

The normative requirements for two columns in FOCUS 1.0 were interpreted differently by two dataset producers. While these critical issues were addressed in the FOCUS 1.1 release, further analysis indicates additional room for improvement. A comprehensive review of all columns and attributes is needed to ensure that the normative requirements are clear, consistent, unambiguous, and complete, preventing future misinterpretation by dataset producers or consumers.

Impact: Inconsistent interpretation of normative requirements can lead to production of non-conforming datasets, errors in data validation, and reduced data quality. By refining these requirements, we can ensure uniformity, reduce ambiguity, and improve dataset quality across providers and consumers.

## **Use Case**
NA

## 2. **Objective** *
> State the objective of this work item. What outcome is expected?
> * Success Criteria: Define how success will be measured (e.g. metrics and KPIs).

The goal of this work item is to review and refine the normative requirements for all columns and attributes in the FOCUS dataset, ensuring they are unambiguous, consistent, and aligned with editorial guidelines. 

A new guideline (template) will also be proposed for formatting requirements, including suggestions for grouping technical and business logic where applicable. This will enhance readability, reduce ambiguity, and clarify which requirements can be automated versus those requiring manual validation.

Success Criteria:

- Review and Update: All normative requirements are reviewed and updated as needed.
- Resolve Inconsistencies: Any inconsistencies regarding constraints, value ranges, nullability, etc., are resolved.
- Minimize Ambiguity: Normative statements are clear, unambiguous, and compliant with editorial guidelines, with any remaining ambiguity reduced to a minimum.
- Clarity: Ensure that the normative statements are comprehensible and can be effectively interpreted by both dataset producers and consumers.

## 3. **Supporting Documentation** *
> Include links to supporting documents such as:
> - Data Examples: [Link to data or relevant files; DO NOT share proprietary information]
> - Related Use Cases or Discussion Documents: [Link to discussion]
> - PRs or Other References: [Link to relevant references]

Prerequisites:

- Issue #606
[...]

PR #654, opened 2024-11-21
[EDITORIAL] ListCost: convert normative statement into bullet points

jpradocueva:
This PR is related to issue https://github.com/FinOps-Open-Cost-and-Usage-Spec/FOCUS_Spec/issues/606 and converts normative statements into bullet points.

### DIFF
--- a/specification/columns/listcost.md
+++ b/specification/columns/listcost.md
@@ -2,7 +2,11 @@
 
 List Cost represents the cost calculated by multiplying the [*list unit price*](#glossary:list-unit-price) and the corresponding [Pricing Quantity](#pricingquantity). List Cost is denominated in the [Billing Currency](#billingcurrency) and is commonly used for calculating savings based on various rate optimization activities, by comparing it with [Contracted Cost](#contractedcost), [Billed Cost](#billedcost) and [Effective Cost](#effectivecost).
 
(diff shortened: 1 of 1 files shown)


ijurica (approved):

kk09v:
This change looks good, but I just wanted to note that this definition could use some review:

- ListUnitPrice is conditional on a price list being published, but ListCost is mandatory and not nullable. 
  - There is no guidance for how ListCost should be calculated when ListUnitPrice is not present.
- The first reference to List Unit Price links to the glossary rather than the column.

ijurica:
> This change looks good, but I just wanted to note that this definition could use some review:
> 
> * ListUnitPrice is conditional on a price list being published, but ListCost is mandatory and not nullable.
>   
>   * There is no guidance for how ListCost should be calculated when ListUnitPrice is not present.
> * The first reference to List Unit Price links to the glossary rather than the column.

**Adding a couple of notes:** 

- Guidance for how ListCost should be calculated**
  - ListUnitPrice MUST be null when ChargeCategory is "Tax" and I'm not sure if, in case of tax, we can even describe how this should or could be calculated. (there is so much I don’t know about taxes 😃) The question is: should we rephrase this or attempt to describe it, if possible?
  - By the way, we used a similar phrase in Contracted Cost.

- Link to glossary term
  - The link to the glossary was also used in Contracted Cost. I'm not saying that it was the right decision, but it was intentional 😄. The decision was made to link the glossary term because those unit prices might not always be present.

Overall, **I suggest we discuss and address this as part of #557** (copying this comment there).

cnharris10 (approved):

jpradocueva:
The group agreed to the work in stages. First is to convert the normative statements into bullet points and then enhance the bullet points according to #557. The PR is approved.

Issue #668, opened 2024-12-18
[MAINTENANCE]: Build jobs have started failing


mike-finopsorg:
### Description

Build job has started to fail with random errors about unable to access temp file.

### Definition of Done

Builds working again

mike-finopsorg:
```
Error: Failed loading page file:///__w/FOCUS_Spec/FOCUS_Spec/specification/toPdfViaTempFile386-0.source (sometimes it will work just to ignore this error with --load-error-handling ignore)
Exit with code 1, due to unknown error.
Error producing PDF.
```

mike-finopsorg:
Modifications made, builds are working again

PR #669, opened 2024-12-18
Build job failing #668

mike-finopsorg:
Found issues with wkhtmltopdf working inside pandoc Likely an update to pandoc is not compatible with wkhtmltopdf As wkhtmltopdf is not actively maintained I separated the build into two stages, make HTML then convert to PDF which appears to work ok.


jpradocueva (approved):
Ok to merge.

shawnalpay (approved):

PR #671, opened 2024-12-18
670 actions deprecation

mike-finopsorg:
Bump version of github actions used in Actions pipelines. Both actions/checkout@v3 and actions/upload-artifact@v3.1.2 need to be updated to v4. Testing on forked repo it appears this is safe for us to perform.

### DIFF
--- a/.github/workflows/candidate_release.yml
+++ b/.github/workflows/candidate_release.yml
@@ -12,7 +12,7 @@ jobs:
       image: pandoc/extra:latest-ubuntu
     steps:
       - name: Check out repository code
-        uses: actions/checkout@v4
+        uses: actions/checkout@v4.2.2
       - name: Install prequirements
         shell: bash
         run: |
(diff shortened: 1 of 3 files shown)


shawnalpay (approved):
@mike-finopsorg Does the minor version matter?  You've updated to version `4`, but [the most recent version of actions/checkout](https://github.com/actions/checkout/releases) is `4.2.2` -- so do we perhaps want to upgrade all the way and use `actions/checkout@v4.2.2`?

I'll leave that to you to decide, as it works fine as-is.  Approving.

gparker-at-sf (approved):

ljadvey (approved):

mike-finopsorg:
> @mike-finopsorg Does the minor version matter? You've updated to version `4`, but [the most recent version of actions/checkout](https://github.com/actions/checkout/releases) is `4.2.2` -- so do we perhaps want to upgrade all the way and use `actions/checkout@v4.2.2`?
> 
> I'll leave that to you to decide, as it works fine as-is. Approving.

Yeah good point. I went ahead and tested the latest point release of both actions and updated the PR.

Issue #619, opened 2024-10-28
[Work_Item] Add SKU and SKU Price properties (e.g. term, tier, SKU region)


ijurica:
## 1. **Problem Statement** *
> Describe the problem, issue, use case, or opportunity that this work item addresses.
> Include practitioner quotes illustrating real examples a) of questions being asked by practitioners and b) value unlocked by answering these questions, if available.
> - What is the problem?: Explain the context and why it needs resolution.
> - Impact: Describe how the problem affects users, systems, or the project.

Although the SKU Price Details column provides relevant SKU and SKU Price properties, and the SKU Meter helps practitioners understand the functionalities being metered by different SKUs, **the existing FOCUS specification still lacks a user-friendly approach to presenting other prominent SKU properties** (such as SKU terms, tiered pricing details, SKU regions/locations, size, etc.), which are essential for understanding and comparing SKUs both within and across providers. 

While SKU Price Details offers these properties in a key-value format, the keys are provider-specific and vary between providers, requiring extensive parsing and mapping, which is often challenging and reduces usability.

Over the past months, both focus members and user group members have expressed interest in various SKU and SKU Price properties. Some of these have been added to the backlog as individual issues and compiled in the *Closed [SKU, SKU Price] Properties-Related Issues Consolidation List*. The issues in this list are now closed and superseded by this work item, which seeks to address SKU details in a comprehensive manner, identifying the most prominent properties while incorporating requirements from that consolidated list.

## Use Case(s)
❗  Follow instructions on [how to craft a use case](https://docs.google.com/document/d/1oMV78K7cGmQ5cjcAv_aZBe-sQzV9zatvJVTEsE7aL7M/edit?tab=t.0) -- need both names and descriptions

## 2. **Objective** *
> State the objective of this work item. What outcome is expected?
> * Success Criteria: Define how success will be measured (e.g. metrics and KPIs).

The objective is to provide a standardized and user-friendly way to access the most prominent SKU and SKU Price properties, enabling practitioners to understand, distinguish, and compare SKUs and SKU Prices more effectively, both within and across providers. This work item aims to simplify the presentation of SKU details so that essential properties are readily accessible, without requiring extensive parsing or the need to map provider-specific keys.

Success Criteria:

- Prominent SKU and SKU Price properties are presented in a standardized manner (e.g., FOCUS columns), minimizing the need to parse the SKU Price Details column and map provider-specific keys.
[...]

PR #683, opened 2025-01-22
WI #619: `SKU Price Details` preferred keys

kk09v:


kk09v:
NOTE: this is a working draft, converted to a PR for easier discussion.

supporting_content/columns/skupricedetails.md
```diff
@@ -0,0 +1,40 @@
...
+| Category | ProposedKey | Explanation | Concerns |
+| ---- | ---- | ---- | ---- |
+| Size & Shape | SizeVcpu | This is to capture the size in terms of cores / vCPUs per the Pricing Unit of the SKU | Time Normalization |
+| Size & Shape | SizeGib | This is to capture the size in terms of GiB per Pricing Unit; this was intentionally kept generic to allow it to be used for multiple different types of SKUs | Time Normalization, Unit Size, a SKU may have 2 different things that could be measured in GiB (e.g. memory + ephemeral disk) |
+| Size & Shape | SizeName | This is to capture the name of the size/shape so that it can be used for aggregation independent of other factors that make up the SKU (e.g. region) | 
+| Size & Shape | SkuVersionName | This is to capture the name of the version or family to aggregate like SKUs | |
+| Size & Shape | SkuXUnits | This is to capture units specific to a PaaS offering which often conatain a measure of processing power (RUs, PTUs, DBUs, etc.) | X is a placeholder for another word |
+
+
+| ID | Type | CSP | Description | SizeVcpu | SizeGib | SizeName | SkuVersionName | SkuXUnits |
```
flanakin, line 23: nit: Description == SkuMeter?
```suggestion
| ID | Type | ProviderName | SkuMeter | SizeVcpu | SizeGib | SizeName | SkuVersionName | SkuXUnits |
```

supporting_content/columns/skupricedetails.md
```diff
@@ -0,0 +1,40 @@
...
+- For some SKUs, the Size & Shape property _feels_ duplicative with the unit of measure (pricing unit, consumed unit). 
+  - It is often the case that the unit of measure will be time based, whereas these Size & Shape properties represent what you have for that period of time. E.g. 1 Hour (unit of measure) * 16 cores (SizeVcpu) = 16 core-hours
+ 
+| Category | ProposedKey | Explanation | Concerns |
+| ---- | ---- | ---- | ---- |
+| Size & Shape | SizeVcpu | This is to capture the size in terms of cores / vCPUs per the Pricing Unit of the SKU | Time Normalization |
+| Size & Shape | SizeGib | This is to capture the size in terms of GiB per Pricing Unit; this was intentionally kept generic to allow it to be used for multiple different types of SKUs | Time Normalization, Unit Size, a SKU may have 2 different things that could be measured in GiB (e.g. memory + ephemeral disk) |
+| Size & Shape | SizeName | This is to capture the name of the size/shape so that it can be used for aggregation independent of other factors that make up the SKU (e.g. region) |
```
flanakin, line 18: I have a bunch of reports that have a column for size. To support building those reports in Power BI, I will need to extract this into a dedicated column, which means I will start referencing `x_SkuSize` as a custom column that is extracted from `SkuPriceDetails.Size`. The same would happen across APIs and databases that practitioners and vendors build when they do care about size. While this is fine, it makes it feel like FOCUS doesn't support my needs because I had to reference a custom column. You could make this argument about anything in here, but this one feels super common. Thoughts?

supporting_content/columns/skupricedetails.md
```diff
@@ -0,0 +1,40 @@
...
+  - Where the region chosen impacts pricing, it would constitute a separate SKU
+  - E.g. GCP sku 3E94-1AB5-33F5 "N2 Custom Instance Core running in Americas" the SKU region (proposed key PricingRegion) would be "Americas" whereas the RegionId column would contain the region that the virtual machine is deployed in (e.g. us-east1)
+- Time Normalization: The pricing unit for 1 SKU may be per month and another sku may be per hour, which would make it more difficult to use this field to calculated
+- Unit Size: GB was used to mean either GB and GiB in differnt contexts, prior to ISO 80000-3; there may be some services truly measured in GB (or other Base10 units) which would be difficult to state in GiB (or other Base2 units)
+  - Also one services make sense measured at different scales (e.g. L1 cache is measured in KiB versus disks that are measured in GiB or TiB).
+- With regards to SKU sizing / naming, different CSPs think about the problem in different ways. E.g. in Azure, SizeName would be the property of the sku, but in GCP SizeName would be the property of the resource.
```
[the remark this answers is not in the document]
kk09v: The way I have it structured in my example table below, GCP would not include the SizeName as a property of that given SKU, because it isn't a property of the SKU. The way they communicate this today is that it's in the system_labels (somewhat equivalent to AdditionalInfo).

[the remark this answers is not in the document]
flanakin: Good point. While we may add `SizeName`, they wouldn't specify it because it doesn't apply to them. And with that, they would be conformant because it wouldn't be mandatory. Makes sense.

supporting_content/columns/skupricedetails.md
```diff
@@ -0,0 +1,40 @@
...
+- For some SKUs, the Size & Shape property _feels_ duplicative with the unit of measure (pricing unit, consumed unit). 
+  - It is often the case that the unit of measure will be time based, whereas these Size & Shape properties represent what you have for that period of time. E.g. 1 Hour (unit of measure) * 16 cores (SizeVcpu) = 16 core-hours
+ 
+| Category | ProposedKey | Explanation | Concerns |
+| ---- | ---- | ---- | ---- |
+| Size & Shape | SizeVcpu | This is to capture the size in terms of cores / vCPUs per the Pricing Unit of the SKU | Time Normalization |
+| Size & Shape | SizeGib | This is to capture the size in terms of GiB per Pricing Unit; this was intentionally kept generic to allow it to be used for multiple different types of SKUs | Time Normalization, Unit Size, a SKU may have 2 different things that could be measured in GiB (e.g. memory + ephemeral disk) |
```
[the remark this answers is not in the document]
ijurica: We will certainly discuss the format and names further, but it does seem practical to have generic placeholders for Size Details (Name, Component/Type, Unit, Quantity).

Of course, not everyone may be able to interpret this, but it seems to me that this column is intended for those who should be able to make sense of it. :)

Also, if we're going in this direction, let's not forget to mention that:
- Unit SHOULD conform to the [UnitFormat](https://github.com/FinOps-Open-Cost-and-Usage-Spec/FOCUS_Spec/blob/working_draft/specification/attributes/unit_format.md#data-size-unit-names) requirements
- Quantity MUST conform to [Numeric Format](https://github.com/FinOps-Open-Cost-and-Usage-Spec/FOCUS_Spec/blob/working_draft/specification/attributes/numeric_format.md) requirements
- Etc.

On that note, we should probably add a **reminder** that these 'predefined' keys are quite similar to OPTIONAL FOCUS columns and should therefore be treated in the same way (apart from the description/introductory section, we should also include normative requirements, etc., as we move forward).

supporting_content/columns/skupricedetails.md
kk09v: I can totally understand your use case, but for me, the "SKU Family" is reported on far more often than the Size. The idea with having all of the data in this semi-structured column is that each consumer can extract and structure the data that's relevant/necessary to them without having a hundred of columns of data that could be sparse, redundant, and/or unneeded.

My proposal is that for the 1.2 release, we should focus on enumerating common properties in the SkuPriceDetails column. Then, we can review the adoption, usage, etc. to evaluate promoting some to dedicated column.s

supporting_content/columns/skupricedetails.md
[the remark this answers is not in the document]
kk09v: @flanakin your suggestion (both options) would require a change to _either_ the value format of [SkuPriceId](https://github.com/FinOps-Open-Cost-and-Usage-Spec/FOCUS_Spec/blob/working_draft/specification/columns/skupriceid.md) or the definition of [Key-Value Format](https://github.com/FinOps-Open-Cost-and-Usage-Spec/FOCUS_Spec/blob/working_draft/specification/attributes/key_value_format.md).

"Values in a key-value pair MUST be one of the following types: number, string, true, false, or null."

Ref

supporting_content/columns/skupricedetails.md
udam-f2: Not involved in these conversations so feel free to ignore.

My suggestion is that if we don't have a clear signal (sounds like survey was more about what not how), it's safer to start with a query-friendly format (json column) keep everything in a details column, and see if there's enough demand to move it to separate columns. On the surface, these seem like ones that make sense to move to separate columns, but my sense is that we won't have clear boundaries for what/where to draw the line and will end up adding more and more columns and eventually be in the same boat AWS CUR 1.0 ended up with (bunch of sparse columns). For that reason, I personally would be fine if all SKU details were in a 'properties' complex column, and all resource details were in their own properties column, and similarly for commitment and discounts, etc. Feels like a much more scalable approach where we keep the core columns of FOCUS simple, and the minority of folks interested in those details have to care about the contents of the other columns - if/when necessary.

kk09v:
Just did some cleanup on the list of suggested keys

This is a MSFT oriented list of values. Should we look for commonalities, expand the list to include additional vendor concepts, or mix the list and say that they're examples?
- Commentary GCP access tiers are standard, nearline, coldline, and archive
- GCP has 3 different disk offerings backed by SSDs (balanced, performance, extreme), should we look at the underlying technology or the offering?
- GCP redundancy would include "dual-region"

Still need to define:
Autoscaling
Compliance Certifications
Encryption
Performance Tier
Pricing Tier

supporting_content/columns/skupricedetails.md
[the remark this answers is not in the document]
mike-finopsorg: > SkuPriceDetails = {
    Size: "Medium",
    Processor: { Quantity: 8, Units: "Cores" },
    Memory: { Quantity: 32, Units: "GB" },
    DiskSize: { Quantity: 128, Units: "GB" },
}

This is way easier to use within systems to query, and given a few examples of accessing the data even less technical users should be ok to use.

>  SkuPriceDetails = {
    Size: "Medium",
    Specs: [
        { Component: "Processor", Quantity: 8, Units: "Cores" },
        { Component: "Memory", Quantity: 32, Units: "GB" },
        { Component: "Disk Size", Quantity: 128, Units: "GB" },
    ],
}

This format is more tricky to use in query languages as you have no easy way to "select" the spec element you want. And will require the same JSON element selection as the previous example so no technical benefit. 

> "Values in a key-value pair MUST be one of the following types: number, string, true, false, or null."

Yes it looks like this should also support a nested object, I think this was avoided being added earlier to limit how much we nest objects, but that becomes very restrictive in cases like this.

shawnalpay:
@kk09v to update this PR to reflect that this is only going to be for `SkuPriceDetails`.

specification/columns/skupricedetails.md
```diff
@@ -1,30 +1,32 @@
...
-  * Individual properties (key-value pairs) SHOULD NOT be modified for a given SkuPriceId and SHOULD remain consistent over time.
-* The key for a property SHOULD remain consistent across comparable SKUs having that property and the values for this key SHOULD remain in a consistent format.
-* The SkuPriceDetails column MUST NOT contain properties which are not applicable to the corresponding SkuPriceId.
-* The SkuPriceDetails column MAY contain properties which are already captured in other dedicated columns.
-* If a property has a numeric value, it MUST represent the value for a single [PricingUnit](#pricingunit).
-* The SkuPriceDetails column MUST be present in a [*FOCUS dataset*](#glossary:FOCUS-dataset) when the provider includes a SkuPriceId.
-  * The SkuPriceDetails column MAY be null when SkuPriceId is not null.
-  * The SkuPriceDetails column MUST be null when SkuPriceId is null.
+  * Individual properties SHOULD NOT be modified for a given SkuPriceId and SHOULD remain consistent over time.
+* SkuPriceDetails keys and their values SHOULD remain consistent across comparable SKUs having that property.
```
ijurica, line 13: Phrasing aside, I prefer the original version of this normative requirement because I have some concerns regarding the material change introduced by this revision. While the new version is suitable for SKU-related properties, it is not suitable for Pricing properties.

Maybe we should split this in two requirements - one for SKU-related properties and the other one for Pricing?


```suggestion
* The key for a property SHOULD remain consistent across comparable SKUs having that property and the values for this key SHOULD remain in a consistent format.
```

supporting_content/columns/skupricedetails.md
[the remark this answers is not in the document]
kk09v: Hey Mike, the general consensus from discussion/voting in the Working Group channel is that we should not introduce a new object format (which allows for nested values) in 1.2 and stick with key-value format, so that was removed from the PR last week.

supporting_content/columns/skupricedetails.md
kk09v: Thanks for weighing in Udam. The voting leaned more toward keeping them in the SkuPriceDetails column than having separate columns, but it wasn't conclusively one direction or the other. During the TF2 call yesterday, we agreed to keep them in for 1.2 and rely on practitioner feedback to make determinations on promoting these to dedicated columns in future releases.

There was a wish expressed that they remain in SkuPriceDetails even if they get a dedicated column. This would be allowable and desired per the clauses: 
- Properties SHOULD NOT be removed from SkuPriceDetails for a given SkuPriceId, once they have been included.
- SkuPriceDetails MAY contain properties which are already captured in other dedicated columns.

specification/columns/skupricedetails.md
kk09v: Because the diff got hosed during multiple rewrites. I want to clarify this this suggestion is a revert to the prior phrasing from 1.1  

I agree that this attempt at clearer phrasing substantially changed the meaning of the clause. Thank you for catching this, I had missed it when I read through the proposed changes the first time.

The intention of this clause was that if a provider offers two "comparable" SKUs (e.g. two different virtual machines), we would not want one virtual machine to have a property of {"CoreCount":8} and the other to have a property of {"vCPUs":"eight"}. Ignoring everything else that's wrong with the second example, the keys should be the same and the values should have the same format.

kk09v: For now, I'll commit this suggestion so that the meaning is not changed, but if we can have another go at rephrasing this for clarity because the change in meaning due to rephase is a solid indication that the original phrasing is not clear.

specification/columns/skupricedetails.md
ijurica, line 1: Since we currently have 3 PRs addressing the SkuPriceDetails Normative Requirements, I’m posting a preview so you can get an idea of how it might look if all three are approved.

You will notice that I’ve included a few suggestions on how to modify the 3 normative requirements we just added in this PR.

Additionally, I suggest renaming **Predefined keys** to **FOCUS-defined SKU Price properties** (or something similar), as in the table, we have not only defined the key for a property but also outlined additional requirements.

Note: This is just a suggestion, and I’d really appreciate any feedback, @kk09v , especially from you! 😄 

---

The SkuPriceDetails column adheres to the following requirements:

* SkuPriceDetails MUST be present in a [*FOCUS dataset*](#glossary:FOCUS-dataset) when the provider supports unit pricing concepts and publishes [*price lists*](#glossary:price-list), publicly or as part of contracting.
* SkuPriceDetails MUST be of type String.
* SkuPriceDetails MUST conform to [StringHandling](#stringhandling) requirements.
* SkuPriceDetails MUST conform to [KeyValueFormat](#key-valueformat) requirements.
* Property key SHOULD conform to [PascalCase](#glossary:pascalcase) format.
* SkuPriceDetails nullability is defined as follows:
  * SkuPriceDetails MUST be null when SkuPriceId is null.
  * SkuPriceDetails MAY be null when SkuPriceId is not null.
  * SkuPriceDetails MUST be associated with a given [ResourceId](#resourceid) or [ServiceName](#servicename) used or purchased.
  * SkuPriceDetails MAY include properties that are already captured in other dedicated columns.
  * SkuPriceDetails SHOULD remain consistent over time for a given SkuPriceId.
  * SkuPriceDetails properties for a given SkuPriceId adhere to the following additional requirements:
    * Existing SkuPriceDetails properties SHOULD remain consistent over time.
    * Existing SkuPriceDetails properties SHOULD NOT be removed.
    * Additional SkuPriceDetails properties MAY be added over time.
  * Property key SHOULD remain consistent across comparable SKUs having that property, and the values for this key SHOULD remain in a consistent format.
  * Property value MUST represent the value for a single [PricingUnit](#pricingunit) when the property holds a numeric value.
* FOCUS-defined SKU Price properties adhere to the following additional requirements:
  * Property key MUST equal the predefined key for that property.
  * Property value MUST be of the type specified for that property.
  * Property value MUST represent the value for a single unit of measure specified for that property when the property holds a numeric value.

kk09v:
Per the the TF2 meeting last week and the follow-up conversation on Friday, I made the following changes to the PR:

- Renamed "Predefined Keys" section to "FOCUS-Defined Properties" and changed verbiage in normative section to reflect this
- Added F_ prefix to keys of FOCUS-defined properties
  - Added comment to FOCUS-Defined Properties section specifying that the keys will all be prefixed with F_ t
  - Added normative clause preventing providers from using F_ for keys that are not FOCUS-defined
- Strengthened normative statements for FOCUS-defined properties from SHOULD to MUST
- Modified predefined properties section
  - Renamed StorageTier property to StorageClass
  - Added Series
  - Added InstanceType
  - Added SizeClass

specification/columns/skupricedetails.md
kk09v: To make this easier to compare, I took the existing statements and put them in the same order/indentation to compare the differences and evaluate whether this changed the meaning. Any lines below in ALL CAPS are my summary of a newly added line.

Notes (beginning with line number of your suggested rephrasing, including the opening line and following blank line)
*  7 I'd change this to keys instead of key
* 12 What is intended here? SkuPriceDetails should not be associated with a resource ID. Why would we specify an association to Service Name instead of to SkuId / SkuPriceId? Sorry, I'm just not following the reasoning on this in my first read.
* 14 this is a change in meaning from what was written (addition for 1.2), but I think I'm okay with it as it mostly acts to strengthen the language from "if you include it, you should..." to "you should include it"
`SkuPriceDetails properties that represent common constructs listed in the FOCUS-Defined Properties section below SHOULD use the defined keys.`
`SkuPriceDetails SHOULD include all FOCUS-defined SKU Price properties listed below that are applicable to the corresponding SkuPriceId.`
* 16 seems redundant with lines 17-20
* 26 "the value for a single unit of measure" didn't quite make sense to me. I'd suggest the below. The for a single PricingUnit clause above (22) already handles an aspect of this, though I've included it again since it's a SHOULD on 22.
`Property value MUST represent the value for a single PricingUnit, denominated in the unit of measure specified for that property when the property holds a numeric value.`
* + I added a new clause related to the FOCUS-defined property namespacing between your lines 22 and 23
* - I noticed that a clause was dropped in your suggestion. This is somewhat implied, but I think it's important because it makes it clear that you shouldn't include for example a resource property in here (which would differ across different records for the same SkuPriceId).
`SkuPriceDetails properties (both keys and values) MUST be shared across all charges having the same SkuPriceId, subject to the below provisions.`

---
* SkuPriceDetails MUST be present in a FOCUS dataset when the provider includes a SkuPriceId.
* STRING TYPE
* STRING Handling
* SkuPriceDetails MUST be in KeyValueFormat.
* SkuPriceDetails keys SHOULD be formatted in PascalCase.
* NULL grouping
  * SkuPriceDetails MAY be null when SkuPriceId is not null.
  * SkuPriceDetails MUST be null when SkuPriceId is null.
* PROPERTY GROUPING
  * ASSOCIATED…
  * SkuPriceDetails MUST NOT contain properties which are not applicable to the corresponding SkuPriceId.
  * SkuPriceDetails properties that represent common constructs listed in the FOCUS-Defined Properties section below SHOULD use the defined keys.
  * SkuPriceDetails MAY contain properties which are already captured in other dedicated columns.
  * SKU CONSISTENCY
    * Individual properties SHOULD NOT be modified for a given SkuPriceId and SHOULD remain consistent over time.
    * Properties SHOULD NOT be removed from SkuPriceDetails for a given SkuPriceId, once they have been included.
    * Additional properties MAY be added to SkuPriceDetails going forward for a given SkuPriceId.
  * The key for a property SHOULD remain consistent across comparable SKUs having that property and the values for this key SHOULD remain in a consistent format.
  * SkuPriceDetails properties with a numeric value MUST represent the value for a single PricingUnit.
  * SkuPriceDetails keys MUST NOT begin with the string "F_" unless it is a FOCUS-defined property
* FOCUS PROPERTIES
  * PROPERTY KEYS MUST MATCH
  * Values for FOCUS-defined properties MUST be the specified data type for that property in the section below.
  * Numeric values for FOCUS-defined properties MUST be the specified unit of measure for that property in the section below.

specification/columns/skupricedetails.md
```diff
@@ -43,13 +46,37 @@ A set of properties of a SKU Price ID which are meaningful and common to all ins
...
+| :------------ | :--------------------------------- |
+| Column type   | Dimension                          |
+| Feature level | Conditional                        |
+| Allows nulls  | True                               |
+| Data type     | String                             |
+| Value format  | [KeyValueFormat](#key-valueformat) |
+
+### FOCUS-Defined Properties
+
+The following keys should be used when applicable to facilitate cross-service and cross-provider queries for the same conceptual property. FOCUS-defined keys will be prefixed with "F_" to make them easy to identify and prevent collisions with other properties conveyed by a provider.
```
ijurica, line 59: something to consider

```suggestion
The following keys should be used when applicable to facilitate cross-SKU and cross-provider queries for the same conceptual property. FOCUS-defined keys will be prefixed with "F_" to make them easy to identify and prevent collisions with other properties conveyed by a provider.
```

specification/columns/skupricedetails.md
```diff
@@ -43,13 +46,37 @@ A set of properties of a SKU Price ID which are meaningful and common to all ins
...
+| F_MemorySize               | RAM allocated for processing                                    | Numeric          | Measure: Gibibytes (GiB<sup>1</sup>)                  |
+| F_OperatingSystem          | Operating system family<sup>2</sup>                             | String           | Examples: "Linux", "MacOS", "Windows"                 |
+| F_Redundancy               | Level of redundancy offered by the SKU                          | String           | Examples: "Local", "Zonal", "Global"                  |
+| F_Series                   | Common name for the series and/or generation of the SKU         | String           | Examples: "M5", "Dadv5", "N2D"                        |
+| F_SizeClass                | Logical grouping of size, which may be used for size fexibility | String           | Examples: "Large", "2xlarge", "Metal"                 |
+| F_StorageClass             | Tier or class of storage provided                               | String           | Examples: "Hot", "Archive", "Nearline"                |
```
ijurica, line 75:
```suggestion
| F_StorageClass             | Class or tier of storage provided                               | String           | Examples: "Hot", "Archive", "Nearline"                |
```

specification/columns/skupricedetails.md
ijurica: Cleaned up version:

---

The SkuPriceDetails column adheres to the following requirements:

* SkuPriceDetails MUST be present in a [*FOCUS dataset*](#glossary:FOCUS-dataset) when the provider supports unit pricing concepts and publishes [*price lists*](#glossary:price-list), publicly or as part of contracting.
* SkuPriceDetails MUST be of type String.
* SkuPriceDetails MUST conform to [StringHandling](#stringhandling) requirements.
* SkuPriceDetails MUST conform to [KeyValueFormat](#key-valueformat) requirements.
* Property keys SHOULD conform to [PascalCase](#glossary:pascalcase) format.
* SkuPriceDetails nullability is defined as follows:
  * SkuPriceDetails MUST be null when SkuPriceId is null.
  * SkuPriceDetails MAY be null when SkuPriceId is not null.
  * SkuPriceDetails MAY include properties that are already captured in other dedicated columns.
  * SkuPriceDetails properties for a given SkuPriceId adhere to the following additional requirements:
    * Existing SkuPriceDetails properties SHOULD remain consistent over time.
    * Existing SkuPriceDetails properties SHOULD NOT be removed.
    * Additional SkuPriceDetails properties MAY be added over time.
  * Property key SHOULD remain consistent across comparable SKUs having that property, and the values for this key SHOULD remain in a consistent format.
  * Property value MUST represent the value for a single [PricingUnit](#pricingunit) when the property holds a numeric value.
* FOCUS-defined SKU Price properties adhere to the following additional requirements:
  * Property key MUST equal the predefined key for that property.
  * Property value MUST be of the type specified for that property.
  * Property value MUST represent the value for a single PricingUnit, denominated in the unit of measure specified for that property when the property holds a numeric value.

specification/columns/skupricedetails.md
```diff
@@ -1,30 +1,33 @@
...
-* The SkuPriceDetails column MUST NOT contain properties which are not applicable to the corresponding SkuPriceId.
-* The SkuPriceDetails column MAY contain properties which are already captured in other dedicated columns.
-* If a property has a numeric value, it MUST represent the value for a single [PricingUnit](#pricingunit).
-* The SkuPriceDetails column MUST be present in a [*FOCUS dataset*](#glossary:FOCUS-dataset) when the provider includes a SkuPriceId.
-  * The SkuPriceDetails column MAY be null when SkuPriceId is not null.
-  * The SkuPriceDetails column MUST be null when SkuPriceId is null.
+* SkuPriceDetails MUST NOT contain properties which are not applicable to the corresponding SkuPriceId.
+* SkuPriceDetails MAY contain properties which are already captured in other dedicated columns.
+* SkuPriceDetails properties with a [numeric](#numeric-format) value MUST represent the value for a single [PricingUnit](#pricingunit).
+* SkuPriceDetails keys MUST NOT begin with the string "F_" unless it is a FOCUS-defined property.
```
flanakin, line 17: I remember when this came up on Friday and was holding my comments until we had a chance to discuss, but we ran out of time. I really, really, really don't like this. While I understand the concept, it's the opposite of what we do for custom columns. I would highly prefer something that aligns with the current pattern for custom attributes, if necessary at all.

specification/columns/skupricedetails.md
```diff
@@ -1,30 +1,33 @@
...
+* SkuPriceDetails properties that represent common constructs listed in the FOCUS-Defined Properties section below SHOULD use the defined keys.
+  * Values for FOCUS-defined properties MUST be the specified data type for that property in the section below.
+  * Numeric values for FOCUS-defined properties MUST be the specified unit of measure for that property in the section below.
+* SkuPriceDetails MUST be present in a [*FOCUS dataset*](#glossary:FOCUS-dataset) when the provider includes a SkuPriceId.
+  * SkuPriceDetails MAY be null when SkuPriceId is not null.
```
flanakin, line 22: I've heard some people say that they want an empty property bag instead of null. Thoughts? @shawnalpay may have an opinion 😉

specification/columns/skupricedetails.md
```diff
@@ -1,30 +1,33 @@
...
 ## Examples
 
 ```json
 {
-    "OperationClass": "A",
-    "PricingTier": 2,
-    "CoreHours": 4,
+    "AccessTier": "Archive",
+    "CoreCount": 4,
     "PreimumProcessing": true,
```
flanakin, line 31:
```suggestion
    "PremiumProcessing": true,
```

specification/columns/skupricedetails.md
```diff
@@ -43,13 +46,37 @@ A set of properties of a SKU Price ID which are meaningful and common to all ins
...
+| Data type     | String                             |
+| Value format  | [KeyValueFormat](#key-valueformat) |
+
+### FOCUS-Defined Properties
+
+The following keys should be used when applicable to facilitate cross-service and cross-provider queries for the same conceptual property. FOCUS-defined keys will be prefixed with "F_" to make them easy to identify and prevent collisions with other properties conveyed by a provider.
+
+| Key                        | Description                                                     | Data Type        | Unit of Measure (numeric) or example values (string)  |
+| :------------------------- | :-------------------------------------------------------------- | :--------------- | :---------------------------------------------------- |
+| F_Bandwidth                | Network capacity for data transfer                              | Numeric          | Measure: Megabits per second (Mbps)                   |
```
flanakin, line 63: This "F_" is F_ing killing me 🤢😒 This makes FOCUS feel like a second-class citizen in its own spec. In a world where FOCUS properties are the majority, would we seriously have this across the board? That would be horrible. We should be prioritizing our space and requiring custom prefixes for custom properties. Same as with columns.

specification/columns/skupricedetails.md
flanakin: Should we have preferred values for things like this to ensure consistency when an applicable value exists?

flanakin (commented):
I'm not signing off solely because of the "F_" prefix. If we get rid of that, you've got my vote.

specification/columns/skupricedetails.md
kk09v: I had looked at that, and at first blush there are a lot of similarities.
Azure has Hot, Cool, Cold, Archive
GCP has Standard, Nearline, Coldline, Archive
AWS has Standard, Infrequent Access, and Archive

These sound similar, but the specific capabilities differ and we'd need to specify what constitutes each Class we establish (e.g. should Infrequent Access be grouped with Cool or Cold storage?).

Then there are the special things each offers: Azure Premium, AWS Express One Zone, and auto tiering options like S3 Intelligent Tiering and GCP Autoclass. GCP Autoclass still falls under the existing categories, but I can't speak to how AWS does it.

Finally, we don't have a set of preferred values for any preferred key yet, so we'd need to build a bit more around that. TBH, I'm not actually opposed to this idea, but I don't think we have the time to get to a consensus on that by the deadline given how I've understood the work required.

specification/columns/skupricedetails.md
ijurica: As I mentioned in the meeting, I understand the reasoning behind how it's currently specified, but I’m more inclined to apply the same principle as for column naming.

specification/columns/skupricedetails.md
ijurica: For what it's worth, I really, really prefer null over empty strings and bags 😄 

And here is one of the String Handling requirements:
> * Empty strings and strings consisting solely of spaces SHOULD NOT be used in not-nullable string columns.

specification/columns/skupricedetails.md
shawnalpay: I agree _strongly_ with @flanakin on this one: this is the opposite of what we do for custom columns.  We should have no prefix for FOCUS-defined properties, similar to how we handle FOCUS-defined columns.

specification/columns/skupricedetails.md
shawnalpay: Oh geez.  I agree with @ijurica on this one.  As a practitioner, I might interpret an empty property bag as a data error, and I'd rather get null.  Also aligns with the principle/requirement Irena quoted.

specification/columns/skupricedetails.md
shawnalpay: Strong agree.  An `F_` prefix feels wrong.  Must we have any prefix convention here at all?  Is it perhaps sufficient to just have a list of reserved FOCUS keys that thou shalt not use?

kk09v: Here's my understanding of Riley's thinking when he proposed we prefix the FOCUS columns:
- Providers will ideally be introducing properties that are relevant to their SKUs, independent of the FOCUS-defined properties.
- If a provider adds a property {"XylophoneUnits" : 12} and following year xylophone computing becomes the next big thing to the point that we want to add that as a FOCUS-defined property, then we've created a collision which is problematic because we've both said that for FOCUS-defined property keys they MUST give values in the format we've specified but also that they SHOULD NOT remove/change a property once it's been introduced.
  - By reserving a prefix, we never need to worry about these collisions.

The alternative would be to force a prefix onto any property keys that aren't FOCUS defined. 


I can go either way on which gets prefixed. However, what makes me nervous about this conversation is Michael's statement "In a world where FOCUS properties are the majority" implying that he is not intending to include many properties that aren't FOCUS-defined. I haven't done the math, but I suspect we have a long tail in terms of SKUs with 20% making up 80% of the spend. We've been focused on the 20% of the SKUs, particularly the ones that have commonalities across multiple CSPs for defining these properties. However, all of the PaaS offerings which are unique to each CSP should have their own properties, which would never become a FOCUS-defined property.

aws-letianfeng (changes requested):
1. Does F_InputOutput indicate I/O for network or storage?
2. AWS offers many SKUs that have burstable I/O, and the specification is "up to X". For example, the network performance is up to 25Gbps. How does the numerical fields handle "up to"?

specification/columns/skupricedetails.md
kk09v: I had intentionally left this in there as an example of a theoretical provider-defined property with a Boolean value.

specification/columns/skupricedetails.md
kk09v: Thanks again @ijurica , I just reviewed this before merging and it looks like the F_ prefix clause got missed. I know you were going to look for where to fit that bullet in, but I'll put it next to the other "FOCUS-defined SKU Price Properties clause" for now.

ijurica: @kk09v I wonder if we need two separate requirements - one specifying when the F_ (or x_ 😄 ) prefix MUST be included and another specifying when it MUST NOT be included.

specification/columns/skupricedetails.md
ijurica: Even though it's old and dusty, I'm pulling my Java programmer hat out of the drawer ... and I still prefer `null`. 😄

specification/columns/skupricedetails.md
shawnalpay: @kk09v Just to be clear: you intentionally left a typo...?  `Preimum` instead of `Premium`.

specification/columns/skupricedetails.md
shawnalpay: > The alternative would be to force a prefix onto any property keys that aren't FOCUS defined.

I feel like you did not address the question in my comment.  Can you tell me a little more about why we must force a prefix on _any_ property keys?  

> Must we have any prefix convention here at all? Is it perhaps sufficient to just have a list of reserved FOCUS keys that thou shalt not use?

kk09v:
> 1. Does F_InputOutput indicate I/O for network or storage?
> 2. AWS offers many SKUs that have burstable I/O, and the specification is "up to X". For example, the network performance is up to 25Gbps. How does the numerical fields handle "up to"?

@flanakin This was a property you had added and the explanation seemed reasonable to me, though it's a property I wouldn't specifically use. Do you have an answer to @aws-letianfeng's questions above? If so, I'll add that clarity into the supporting content. If not, I'm happy to drop this property for the current release as it can always be refined and reintroduced later.

My interpretation, without explicitly knowing Michael's specific intent, is:
- The SKU being billed.
  - On a disk, that's the I/O of the disk. 
  - On a NIC, that's the network I/O.
  - On a VM, this could get tricky if they have properties for both maximum disk I/O and network I/O
    - I don't have experience with AWS, so I got way too deep into the describe-instance-types documentation and could not find a disk bandwidth or throughput property of a virtual machine, only of disks themselves. That being said, Azure does have "Premium SSDs" which can only be attached to certain instance types, but even there, I don't see disk I/O listed as a property of the instance.
- Great callout on the burstable SKUs. 
  - In the case of burstable SKU, I'd interpret the value to to be the baseline, which I'll add to the Supporting Content as it would also apply to other properties such as CoreCount for VMs with burstable cores. 
  - I did some quick research into the documentation on AWS [burstable bandwidth](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/ec2-instance-network-bandwidth.html) and it seems to align to this thinking as does the documentation for Azure [burstable B-Series VMs](https://learn.microsoft.com/en-us/azure/virtual-machines/sizes/general-purpose/b-family)

specification/columns/skupricedetails.md
kk09v: Sorry, I find examples and illustrations to be helpful for me to understand things, so I often explain that way. Let me try this again.

- Until a property key is part of the list of FOCUS-Defined properties, it could be adopted by a provider as the key for a provider-defined property.
  - To illustrate this, I started out with placeholder property "XyzProperty.", before making it more concrete by morphing it into "XylophoneUnits"
- If in a later release, FOCUS defines a property with a key that is already in use by a provider, this becomes problematic based on a few clauses:
  - Summarizing [3 clauses](https://github.com/FinOps-Open-Cost-and-Usage-Spec/FOCUS_Spec/blob/000b45b78f5e2643a11cd82ec1be61de32107095/specification/columns/skupricedetails.md?plain=1#L27) Property keys, value formats, and units of measure for FOCUS-defined properties MUST follow what is laid out for that FOCUS-defined property.
  - Now what happens if the FOCUS-defined property differs in description or unit of measure from what was contained in the provider-defined column?
    - E.g. FOCUS wants XylophoneUnits to mean the number of bars on the Xylophone, but the provider had defined it as the number of notes per octave of the scale that the Xylophone is tuned to (BTW, I'm **really** stretching for these examples since xylophone was just my replacement for xyz originally).
    - Properties SHOULD remain consistent over time and SHOULD NOT be removed [link](https://github.com/FinOps-Open-Cost-and-Usage-Spec/FOCUS_Spec/blob/000b45b78f5e2643a11cd82ec1be61de32107095/specification/columns/skupricedetails.md?plain=1#L21)
      - This is only a SHOULD, but depending on your interpretation, the provider would either be removing their property if the value specified in FOCUS is not something they have available or changing the value of the property to match what is defined by FOCUS.
    - The provider also now needs to define a new key for their existing property since they can no longer use the key reserved for FOCUS.
    - Also consider the timing of implementation. Providers are not necessarily adopting the newest version of FOCUS the moment it's released, so unless a practitioner is digging into the metadata to find which version of FOCUS their feed is compatible with and reviewing the changelog/prior versions of FOCUS to determine when the FOCUS-defined property was introduced, there's a strong chance of them misinterpreting the meaning of this property to be aligned to the FOCUS definition for that key.

shawnalpay: I grant you that there may be a collision.  However, it presumes that the working group would not conduct a survey of keys in use by FOCUS providers, and I should hope that we would execute on a good-faith effort to avoid any existing key names.  If we do decide to enact [eminant domain](https://en.wikipedia.org/wiki/Eminent_domain), then I would hope that we do it for a very good reason.

I would very much like to avoid requiring practitioners to have to chop off a prefix during their FOCUS transformations process before getting the data into a data warehouse or semantic model for BI purposes.

I'm willing to stand down on this point if it's a non-starter with the group -- we'll see what the Slack poll generates -- but if so, then my fallback preference would be, as stated previously, to prefix the provider keys, as that has follows our precedent of requiring providers to prefix their non-FOCUS columns with `x_`.

specification/columns/skupricedetails.md
kk09v: Because we're prefixing FOCUS properties and we've said that they must not prefix provider properties with F_, we have one side covered (must not)
By stating
`Property key MUST equal the predefined key for that property.`
and defining the key for the properties to have a F_ prefix, we have the the other side covered (must)

specification/columns/skupricedetails.md
kk09v: This has been taken to a vote in #working-group per the TF2 meeting today, so I'll close this conversation.

specification/columns/skupricedetails.md
ijurica: Although I generally prefer reserved keywords, Karl pointed out my main concern:

> * _The provider also now needs to define a new key for their existing property since they can no longer use the key reserved for FOCUS.._

That's the main reason I opted for `x_`

specification/columns/skupricedetails.md
```diff
@@ -4,31 +4,38 @@ The SKU Price Details column represents a list of relevant properties shared by
...
-* SkuPriceDetails properties that represent common constructs listed in the FOCUS-Defined Properties section below SHOULD use the defined keys.
-  * Values for FOCUS-defined properties MUST be the specified data type for that property in the section below.
-  * Numeric values for FOCUS-defined properties MUST be the specified unit of measure for that property in the section below.
-* SkuPriceDetails MUST be present in a [*FOCUS dataset*](#glossary:FOCUS-dataset) when the provider includes a SkuPriceId.
-  * SkuPriceDetails MAY be null when SkuPriceId is not null.
+* SkuPriceDetails MUST be present in a [*FOCUS dataset*](#glossary:FOCUS-dataset) when the provider supports unit pricing concepts and publishes [*price lists*](#glossary:price-list), publicly or as part of contracting.
+* SkuPriceDetails MUST be of type String.
+* SkuPriceDetails MUST conform to [StringHandling](#stringhandling) requirements.
+* SkuPriceDetails MUST conform to [KeyValueFormat](#key-valueformat) requirements.
+* Property keys SHOULD conform to [PascalCase](#glossary:pascalcase) format.
```
flanakin, line 11: nit: I would like each requirement to be able to stand on its own, when possible. Adding the column name to be explicit.
```suggestion
* SkuPriceDetails property keys SHOULD conform to [PascalCase](#glossary:pascalcase) format.
```

aws-letianfeng:
> > 1. Does F_InputOutput indicate I/O for network or storage?
> > 2. AWS offers many SKUs that have burstable I/O, and the specification is "up to X". For example, the network performance is up to 25Gbps. How does the numerical fields handle "up to"?
> 
> @flanakin This was a property you had added and the explanation seemed reasonable to me, though it's a property I wouldn't specifically use. Do you have an answer to @aws-letianfeng's questions above? If so, I'll add that clarity into the supporting content. If not, I'm happy to drop this property for the current release as it can always be refined and reintroduced later.
> 
> My interpretation, without explicitly knowing Michael's specific intent, is:
> 
> * The SKU being billed.
>   
>   * On a disk, that's the I/O of the disk.
>   * On a NIC, that's the network I/O.
>   * On a VM, this could get tricky if they have properties for both maximum disk I/O and network I/O
>     
>     * I don't have experience with AWS, so I got way too deep into the describe-instance-types documentation and could not find a disk bandwidth or throughput property of a virtual machine, only of disks themselves. That being said, Azure does have "Premium SSDs" which can only be attached to certain instance types, but even there, I don't see disk I/O listed as a property of the instance.

I think we should split F_InputOutput into F_NetworkInputOutput, F_LocalStorageInputOutput (for storage physically attached to the instance), and F_StorageInputOutput (the last one is specific to remote attached storage like Amazon EBS volumes)

> * Great callout on the burstable SKUs.
>   
>   * In the case of burstable SKU, I'd interpret the value to to be the baseline, which I'll add to the Supporting Content as it would also apply to other properties such as CoreCount for VMs with burstable cores.
>   * I did some quick research into the documentation on AWS [burstable bandwidth](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/ec2-instance-network-bandwidth.html) and it seems to align to this thinking as does the documentation for Azure [burstable B-Series VMs](https://learn.microsoft.com/en-us/azure/virtual-machines/sizes/general-purpose/b-family)

Let's be specific in the field names that these are either for baseline or burst. Note that AWS usually publishes network performance spec based on burst limit for EC2 instances. For CPU, AWS publishes both baseline and burst.

specification/columns/skupricedetails.md
ijurica: I've already applied that suggestion in the #798 😄 

@kk09v I’ve also slightly reformulated the Guidelines. (I’m trying to find what works best for multiple columns, and yes - there are exceptions 😄).

JSON columns are specific in that we specify requirements for:
- the column itself (Tags, SkuPriceDetails)
- the elements it is made up of (properties and tags)
- property keys/values and tag keys/values

We should probably specify that each requirement starts with the name of the subject (most often a column name) ?!

kk09v:
- Made x_ prefix for provider-defined keys and updated verbiage
  - Per the vote in WorkingGroup channel
  - Resolves this thread: https://github.com/FinOps-Open-Cost-and-Usage-Spec/FOCUS_Spec/pull/683#discussion_r1999891513
- Broke InputOutput into NetworkInputOutput and DiskInputOutput
  - Per @aws-letianfeng's suggestion
- Renamed Bandwidth to NetworkBandwidth to align to NetworkInputOutput prefix convention established with NetworkInputOutput
- Added footnote specifying baseline or guaranteed values should be used for burstable or otherwise size flexible SKUs
  - Maximum values can be defined in a provider-specified column

specification/columns/skupricedetails.md
```diff
@@ -43,13 +57,39 @@ A set of properties of a SKU Price ID which are meaningful and common to all ins
...
-| Feature level   | Conditional      |
-| Allows nulls    | True             |
-| Data type       | JSON             |
-| Value format    | [Key-Value Format](#key-valueformat) |
+| Constraint    | Value                              |
+| :------------ | :--------------------------------- |
+| Column type   | Dimension                          |
+| Feature level | Conditional                        |
+| Allows nulls  | True                               |
+| Data type     | String                             |
```
ijurica, line 65: Per @UdamDewaraja suggestion and TF-1 discussion reverting back to JSON (as specified in FOCUS 1.1).

_**Note:** This will be revisited in future releases (AI #855)._ 

```suggestion
| Data type     | JSON                               |
```

specification/columns/skupricedetails.md
```diff
@@ -1,31 +1,45 @@
...
-* The key for a property SHOULD remain consistent across comparable SKUs having that property and the values for this key SHOULD remain in a consistent format.
-* The SkuPriceDetails column MUST NOT contain properties which are not applicable to the corresponding SkuPriceId.
-* The SkuPriceDetails column MAY contain properties which are already captured in other dedicated columns.
-* If a property has a numeric value, it MUST represent the value for a single [PricingUnit](#pricingunit).
-* The SkuPriceDetails column MUST be present in a [*FOCUS dataset*](#glossary:FOCUS-dataset) when the provider includes a SkuPriceId.
-  * The SkuPriceDetails column MAY be null when SkuPriceId is not null.
-  * The SkuPriceDetails column MUST be null when SkuPriceId is null.
+* SkuPriceDetails MUST be present in a [*FOCUS dataset*](#glossary:FOCUS-dataset) when the provider supports unit pricing concepts and publishes [*price lists*](#glossary:price-list), publicly or as part of contracting.
+* SkuPriceDetails MUST be of type String.
+* SkuPriceDetails MUST conform to [StringHandling](#stringhandling) requirements.
```
ijurica, line 13: Per @UdamDewaraja suggestion and TF-1 discussion reverting back to JSON (as specified in FOCUS 1.1).

_**Note:** This will be revisited in future releases (AI #855)._ 

```suggestion
```

shawnalpay:
@aws-letianfeng Please re-review per @kk09v's request yesterda.  If you do not provide an explicit approval or change request, your stale review will be dismissed in the next 48 hours.

ijurica (approved):

aws-letianfeng (approved):

specification/columns/skupricedetails.md
```diff
@@ -43,13 +57,39 @@ A set of properties of a SKU Price ID which are meaningful and common to all ins
...
+| Value format  | [KeyValueFormat](#key-valueformat) |
+
+### FOCUS-Defined Properties
+
+The following keys should be used when applicable to facilitate cross-SKU and cross-provider queries for the same conceptual property. Focus-defined keys will appear in the list below and Provider-defined keys will be prefixed with "x_" to make them easy to identify as well as prevent collisions.
+
+| Key                      | Description                                                     | Data Type        | Unit of Measure (numeric) or example values (string)  |
+| :----------------------- | :-------------------------------------------------------------- | :--------------- | :---------------------------------------------------- |
+| CoreCount                | Number of physical or virtual CPUs available<br><sup>1</sup>    | Numeric          | Measure: Quantity of Cores                            |
+| DiskInputOutput          | Storage input/output operations per second<br><sup>1</sup>      | Numeric          | Measure: Input/Output Operations per Second (IOPS)    |
```
UdamDewaraja, line 75:
```suggestion
| DiskIOPS          | Storage input/output operations per second<br><sup>1</sup>      | Numeric          | Measure: Input/Output Operations per Second (IOPS)    |
```

specification/columns/skupricedetails.md
```diff
@@ -43,13 +57,39 @@ A set of properties of a SKU Price ID which are meaningful and common to all ins
...
+| DiskSpace                | Storage capacity available                                      | Numeric          | Measure: Gibibytes (GiB)                              |
+| DiskType                 | Kind of disk used                                               | String           | Examples: "SSD", "HDD", "NVMe"                        |
+| GpuCount                 | Number of GPUs available                                        | Numeric          | Measure: Quantity of GPUs                             |
+| InstanceType             | Common name of the instance including size, shape, series, etc. | String           | Examples: "m5d.2xlarge", "NC24rs_v3", "P50"           |
+| MemorySize               | RAM allocated for processing                                    | Numeric          | Measure: Gibibytes (GiB<sup>2</sup>)                  |
+| NetworkBandwidth         | Network capacity for data transfer<br><sup>1</sup>              | Numeric          | Measure: Megabits per second (Mbps)                   |
+| NetworkInputOutput       | Network input/output operations per second<br><sup>1</sup>      | Numeric          | Measure: Input/Output Operations per Second (IOPS)    |
```
UdamDewaraja, line 82:
```suggestion
| NetworkIOPS       | Network input/output operations per second<br><sup>1</sup>      | Numeric          | Measure: Input/Output Operations per Second (IOPS)    |
```

specification/columns/skupricedetails.md
```diff
@@ -43,13 +57,39 @@ A set of properties of a SKU Price ID which are meaningful and common to all ins
...
+| OperatingSystem          | Operating system family<sup>3</sup>                             | String           | Examples: "Linux", "MacOS", "Windows"                 |
+| Redundancy               | Level of redundancy offered by the SKU                          | String           | Examples: "Local", "Zonal", "Global"                  |
+| Series                   | Common name for the series and/or generation of the SKU         | String           | Examples: "M5", "Dadv5", "N2D"                        |
+| SizeClass                | Logical grouping of size, which may be used for size fexibility | String           | Examples: "Large", "2xlarge", "Metal"                 |
```
UdamDewaraja, line 86: IMO, I think these two are very confusing and are unlike the other properties in this list. These seem like generic 'we can't really describe what all would go in this but here's a key for you to provide whatever you see fit' type of property where as the others seem like they're referring to a specific attribute that isn't so ambiguous. My suggestion is to remove these OR make them very specific like the other things above. E.g., Series could be InstanceFamily which is consistent with the other property above called InstanceType. 

If, instead, this is not an Instance (VM) property (which contradicts the other properties), it feels like we would need a better way to convey what this represents, or providers will put whatever they feel fits here and then this property won't deliver any of the intended value we're hoping for it to deliver.

specification/columns/skupricedetails.md
```diff
@@ -1,31 +1,45 @@
...
+  * SkuPriceDetails MAY include properties that are already captured in other dedicated columns.
+  * SkuPriceDetails properties for a given SkuPriceId adhere to the following additional requirements:
+    * Existing SkuPriceDetails properties SHOULD remain consistent over time.
+    * Existing SkuPriceDetails properties SHOULD NOT be removed.
+    * Additional SkuPriceDetails properties MAY be added over time.
+  * Property key SHOULD remain consistent across comparable SKUs having that property, and the values for this key SHOULD remain in a consistent format.
+  * Property key MUST NOT begin with the string "F_" unless it is a FOCUS-defined property.
+  * Property value MUST represent the value for a single [PricingUnit](#pricingunit) when the property holds a numeric value.
+* FOCUS-defined SKU Price properties adhere to the following additional requirements:
+  * Property key MUST equal the predefined key for that property.
```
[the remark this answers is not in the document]
flanakin: I think this is saying that, if you define a core count, you MUST use "CoreCount" and not "NumberOfCores" or "CpuCount" or something else.

specification/columns/skupricedetails.md
flanakin: We have a rule to not use acronyms. If we did, it would be `DiskIops` based on our current naming standard.

flanakin (approved):

specification/columns/skupricedetails.md
flanakin: We discussed this at length with @rileyjenk. But if we can't close it, I would vote we remove it and discuss for 1.3.
```suggestion
```

specification/columns/skupricedetails.md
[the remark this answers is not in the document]
UdamDewaraja: Given that this is normative, i suggest we have clear guidance here. If it seems unnecessary, maybe we remove it for now? Open to a different way of saying it if the goal is to emphasize what Michael said. 

```suggestion
```

specification/columns/skupricedetails.md
UdamDewaraja: And we could go update this to be one of the supported things? or maybe we don't need to yet because these aren't columns (yet?)

To me, the issue is the term InputOutput as a quantifiable measure means very little ("Hey Azure, what's your DiskInputOutput?")  and the kicker is the description we're clearly describing IOPS. Thanks, agree it needs to be DiskIops according to the way we handle other acronyms. Updated.

specification/columns/skupricedetails.md
shawnalpay: +1 on this change.

specification/columns/skupricedetails.md
shawnalpay: I would be disappointed to see `Series` and `SizeClass` get cut.  `InstanceType` alone tells the basic story, but then the practitioner still has to perform provider-specific transformations to get to these smaller chunks of the story.   As a vendor, we found that allowing our customers to build a report that slices cost by `medium`, `large`, `2xlarge`, and so on to be really important, especially if the data can be enriched by observability (e.g., CPU %, MB %, IOPS) to then start unlocking optimizations (e.g., this cost is all associated with `large` instances that should be upsized).  I suspect @kk09v is nodding violently in agreement as he reads this, but I'll let him chime in.

That said, I won't die on this hill if we can't get consensus on this one.

CHANGELOG.md
```diff
@@ -21,6 +21,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 
 - `CommitmentDiscountType` column updates:
   - Must be a consistent, readable display value.
+- `SkuPriceDetails` column updates:
+  - Created FOCUS-Defined properties.
```
flanakin, line 25:
```suggestion
  - Created FOCUS-defined properties.
```

specification/columns/skupricedetails.md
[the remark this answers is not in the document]
flanakin: Doesn't removing this nullify the entire body of work because it then won't require data generators use the defined keys? I think we need to fix it if there's confusion. Does this help?
```suggestion
  * Property key MUST be one of the predefined key names when there is a logical equivalent for the information provided.
```

specification/columns/skupricedetails.md
flanakin: I get your point. I'm indifferent. InputOutput feels cleaner, but I also get that Iops is more precise. I'm happy to go with whatever the group wants. If we can't close it, then I would just remove it for 1.2 and discuss for 1.3. We'll continue to evolve these keys over time.

UdamDewaraja: Feel free to resolve this based on consensus. I saw a couple of people agreeing to Iops below. I wouldn't let the 'no acronym' rule hold back clarity. We've already made exceptions for similar cases

shawnalpay: Agree with @UdamDewaraja's suggestion that this should be `DiskIops`.  I believe @rileyjenk agrees, based on his comment on [this thread](https://github.com/FinOps-Open-Cost-and-Usage-Spec/FOCUS_Spec/pull/683#discussion_r2017345297).

shawnalpay: We agreed in Mar 31 Maintainers to sustain this change.

specification/columns/skupricedetails.md
shawnalpay: We agreed in Mar 31 Maintainers to sustain this change.

specification/columns/skupricedetails.md
shawnalpay: We agreed in Mar 31 Maintainers to cut the `Series` and `SizeClass` keys from this PR and consider adding them in a future release, potentially along with more precise guidance on what belongs in such keys.

specification/columns/skupricedetails.md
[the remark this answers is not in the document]
shawnalpay: We agreed in Mar 31 Maintainers that @kk09v is now armed with the crux of this suggestion (i.e. "use a FOCUS property key if it's applicable"), and he will wordsmith a more accurate wording.

specification/columns/skupricedetails.md
kk09v: @shawnalpay I disagree with your summary.
Series is being renamed to InstanceSeries (to align with InstanceType prefix) and clarification of the provider mappings has been added to the supporting content.
[AWS](https://github.com/FinOps-Open-Cost-and-Usage-Spec/FOCUS_Spec/blob/26837645fd5e7163cb482de3fdcdb309a4f2d38f/supporting_content/columns/skupricedetails.md?plain=1#L15)
[GCP](https://github.com/FinOps-Open-Cost-and-Usage-Spec/FOCUS_Spec/blob/26837645fd5e7163cb482de3fdcdb309a4f2d38f/supporting_content/columns/skupricedetails.md?plain=1#L23)
[Azure](https://github.com/FinOps-Open-Cost-and-Usage-Spec/FOCUS_Spec/blob/26837645fd5e7163cb482de3fdcdb309a4f2d38f/supporting_content/columns/skupricedetails.md?plain=1#L38)

I'm working to get alignment with the people who provided feedback above (yourself included) to see if we can remove objections before TF2 on Wednesday.

kk09v: SizeClass I don't immediately have a path forward for. I haven't found a specific use outside of AWS, so I'm checking with Riley.

specification/columns/skupricedetails.md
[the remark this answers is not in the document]
kk09v: I clarified this bullet as `Property key MUST match the spelling and casing specified for the FOCUS-defined property.`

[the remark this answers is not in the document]
kk09v: There was an additional concern raised today around forcing inclusion of the FOCUS-defined property. We already had the provision indicating that all the FOCUS-defined properties SHOULD be included when applicable.
`SkuPriceDetails SHOULD include all FOCUS-defined SKU Price properties listed below that are applicable to the corresponding SkuPriceId.`
to which I added a subsequent bullet:
`SkuPriceDetails MUST include the FOCUS-defined SKU Price property when an equivalent property is included as a Provider-defined property.`

So, if a provider is already sending a property `x_vcpus` which represents the number of vCPUs on the SKU as a Provider-defined property, this would require them to also provide `CoreCount` once it's introduced as a FOCUS-defined property.

Technically, the subsequent bullet may be out of order per the normative requirements, but I think that it makes the most sense if this immediately follows the other one.

specification/columns/skupricedetails.md
shawnalpay: (Oh... sorry Karl, I must have misremembered the outcome of the conversation.  I'm still nursing a head cold.)

As we agreed in today's call (at least, I'm pretty sure), `Series` means multiple things.  I quote from your supporting content:

> To remove ambiguity, the InstanceSeries Focus-defined property value will refer to the subset of [Azure Virtual Machine naming convention](https://learn.microsoft.com/en-us/azure/virtual-machines/vm-naming-conventions) excluding the "# of vCPUs" and "*Constrained vCPUs" components, as shown in the [Virtual Machine name breakdowns structure](https://learn.microsoft.com/en-us/azure/virtual-machines/sizes/overview#name-structure-breakdown)

This is a stricture on the meaning of the spec that we have placed outside the spec, and no provider is required to adhere to it.  Thus, the practitioner's understanding of this column may be entirely divorced from what a provider puts there (in this case, Azure).  **I think that these keys need more clarity _within the spec_ as to what values the practitioner can expect to find.**  While I am absolutely willing to accept waiting until the next release to install that clarity, I deeply worry about the practitioners' -- and providers' -- understanding of these two keys in the interim.

UdamDewaraja (approved):

specification/columns/skupricedetails.md
shawnalpay: @kk09v I am unresolving this comment ("sustain" means "accept"!).  Please either apply the suggested change or provide context for why you do not support it.

specification/columns/skupricedetails.md
shawnalpay: @kk09v I am unresolving this comment ("sustain" means "accept"!).  Please either apply the suggested change or provide context for why you do not support it.

specification/columns/skupricedetails.md
UdamDewaraja: "sustain" always confuses me :) 😅

specification/columns/skupricedetails.md
shawnalpay: Upon further review -- looks like @kk09v applied this via a separate commit.  Current version uses `Iops`.  Re-resolving.

specification/columns/skupricedetails.md
```diff
@@ -43,13 +56,38 @@ A set of properties of a SKU Price ID which are meaningful and common to all ins
...
+| Value format  | [KeyValueFormat](#key-valueformat) |
+
+### FOCUS-Defined Properties
+
+The following keys should be used when applicable to facilitate cross-SKU and cross-provider queries for the same conceptual property. Focus-defined keys will appear in the list below and Provider-defined keys will be prefixed with "x_" to make them easy to identify as well as prevent collisions.
+
+| Key                      | Description                                                     | Data Type        | Unit of Measure (numeric) or example values (string)  |
+| :----------------------- | :-------------------------------------------------------------- | :--------------- | :---------------------------------------------------- |
+| CoreCount                | Number of physical or virtual CPUs available<sup>1</sup>        | Numeric          | Measure: Quantity of Cores                            |
+| DiskIops                 | Storage input/output operations per second<sup>1</sup>          | Numeric          | Measure: Input/Output Operations per Second (IOPS)    |
```
shawnalpay, line 74: Did the TF discuss use of the word `Max` in this key name?  Typically, IOPS capabilities are expressed as Max IOPS to delineate from any observability data, where IOPS is a metric and not an attribute.  Examples: [Azure](https://learn.microsoft.com/en-us/azure/virtual-machines/disks-types), [AWS](https://docs.aws.amazon.com/ebs/latest/userguide/ebs-volume-types.html), and [GCP](https://cloud.google.com/compute/docs/disks/performance).
```suggestion
| DiskMaxIops                 | Storage maximum input/output operations per second<sup>1</sup>          | Numeric          | Measure: Input/Output Operations per Second (IOPS)    |
```

specification/columns/skupricedetails.md
```diff
@@ -43,13 +56,38 @@ A set of properties of a SKU Price ID which are meaningful and common to all ins
...
+| DiskSpace                | Storage capacity available                                      | Numeric          | Measure: Gibibytes (GiB)                              |
+| DiskType                 | Kind of disk used                                               | String           | Examples: "SSD", "HDD", "NVMe"                        |
+| GpuCount                 | Number of GPUs available                                        | Numeric          | Measure: Quantity of GPUs                             |
+| InstanceType             | Common name of the instance including size, shape, series, etc. | String           | Examples: "m5d.2xlarge", "NC24rs_v3", "P50"           |
+| InstanceSeries           | Common name for the series and/or generation of the instance    | String           | Examples: "M5", "Dadv5", "N2D"                        |
+| MemorySize               | RAM allocated for processing                                    | Numeric          | Measure: Gibibytes (GiB<sup>2</sup>)                  |
+| NetworkBandwidth         | Network capacity for data transfer<br><sup>1</sup>              | Numeric          | Measure: Megabits per second (Mbps)                   |
```
shawnalpay, line 81: Was there TF discussion of the use of `Bandwidth` vs `Throughput`?  The latter seems like the much more common publicly-facing description / pricing of network capabilities, and as a vendor, that was the term we used.

Examples:
https://learn.microsoft.com/en-us/azure/vpn-gateway/about-gateway-skus
https://docs.aws.amazon.com/storagegateway/latest/vgw/PerfGatewayAWS-common.html
https://cloud.google.com/compute/all-pricing?hl=en

kk09v: It was not discussed as nobody had taken issue with what was proposed. Michael had originally made the suggestion as simply "Bandwidth" which was later renamed NetworkBandwidth when another property was prefixed with Network.

thecloudman (approved):

specification/columns/skupricedetails.md
kk09v: I can go either way on this, depending on what we're looking at, but I understand the argument for Throughput and am leaning more and more that direction.

To get everyone on the same starting page, Bandwidth measures capacity and Throughput measures usage. Here's an example explanation https://www.geeksforgeeks.org/difference-between-bandwidth-and-throughput/

Your [AWS example](https://docs.aws.amazon.com/storagegateway/latest/vgw/PerfGatewayAWS-common.html) is specifically talking about the measurement of actual bytes, not the capacity.
In one of your other [AWS examples](https://docs.aws.amazon.com/ebs/latest/userguide/ebs-volume-types.html), it talks about Max Throughput, so it's a choice between NetworkBandwidth and NetworkMaxThroughput. Those sound like the same thing, but I understand there's a subtle difference.
IEEE uses Bandwidth when talking about the ports and [cables](https://ieeexplore.ieee.org/document/STDGTSU20531) that are carrying the data. However, I understand that unless I'm buying a bare metal instance, I'm not getting a 10/100/1000 port, I'm getting an allocation of network capacity which is dependent on the provider's network stack and thus should be measured by observation of actual performance (aka NetworkMaxThroughput).

specification/columns/skupricedetails.md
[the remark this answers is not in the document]
kk09v: Udam reviewed and approved this PR which I'll interpret to reflect he's happy with the changes. I'm going to mark this conversation resolved.

specification/columns/skupricedetails.md
```diff
@@ -1,31 +1,44 @@
...
+
+The composition of properties associated with a specific *SKU Price* may differ across providers and across *SKUs* within the same provider. However, the exclusion of dynamic or negotiable pricing properties should ensure that all charges with the same SKU Price ID share the same SKU Price Details, i.e., that SKU Price Details remains consistent across different billing periods and billing accounts within a provider.
+
+SKU Price Details helps practitioners understand and distinguish *SKU Prices*, each identified by a SKU Price ID and associated with a used or purchased [*resource*](#glossary:resource) or [*service*](#glossary:service). It can also enable practitioners to calculate metrics such as total units of a SKU when it is not directly priced in those units (e.g. cores) and thus enables FinOps capabilities such as unit economics. Additionally, the SKU Price Details may be used to analyze costs based on pricing properties such as terms and tiers.
```
ijurica, line 7: just a suggestion 😄 

```suggestion
SKU Price Details helps practitioners understand and distinguish *SKU Prices*, each identified by a SKU Price ID and associated with a used or purchased [*resource*](#glossary:resource) or [*service*](#glossary:service). It can also help determine the quantity of units for a property when it holds a numeric value (e.g., CoreCount), even when its unit differs from the one in which the SKU is priced and charged, thus supporting FinOps capabilities like unit economics. Additionally, the SKU Price Details may be used to analyze costs based on pricing properties such as terms and tiers.
```

specification/columns/skupricedetails.md
kk09v: After further pondering, I updated to NetworkMaxThroughput to bring it in line with DiskMaxIops 
@shawnalpay I'd like to make the same change NetworkIops -> NetworkMaxIops, but I wanted to gut check that one with your first.

specification/columns/skupricedetails.md
```diff
@@ -43,13 +56,38 @@ A set of properties of a SKU Price ID which are meaningful and common to all ins
...
+| GpuCount                 | Number of GPUs available                                                 | Numeric          | Measure: Quantity of GPUs                             |
+| InstanceType             | Common name of the instance including size, shape, series, etc.          | String           | Examples: "m5d.2xlarge", "NC24rs_v3", "P50"           |
+| InstanceSeries           | Common name for the series and/or generation of the instance             | String           | Examples: "M5", "Dadv5", "N2D"                        |
+| MemorySize               | RAM allocated for processing                                             | Numeric          | Measure: Gibibytes (GiB<sup>2</sup>)                  |
+| NetworkMaxThroughput     | Network maximum sustained throughput for data transfer<sup>1</sup>       | Numeric          | Measure: Megabits per second (Mbps)                   |
+| NetworkIops              | Network input/output operations per second<sup>1</sup>                   | Numeric          | Measure: Input/Output Operations per Second (IOPS)    |
```
kk09v, line 82:
```suggestion
| NetworkMaxIops           | Network maximum sustained input/output operations per second<sup>1</sup> | Numeric          | Measure: Input/Output Operations per Second (IOPS)    |
```

specification/columns/skupricedetails.md
kk09v: Resolving this conversation and moving that prior comment into it's own suggestion

specification/columns/skupricedetails.md
kk09v: Suggestion to bring NetworkIops in line with DiskMaxIops and NetworkMaxThroughput.

shawnalpay: Definitely.  +1.

jpradocueva:
This PR was approved during the Members’ meeting on April 3rd.

PR #685, opened 2025-01-22
643 - add use case template

beau-nelford-anglepoint:


rileyjenk (changes requested):
@beau-nelford-anglepoint lets add a top level file under use_cases that explains what they are and the informatoin provided.

rileyjenk:
For an example of this check out the appendix example section.

supporting_content/use_cases/compare_billed_cost_per_subaccount_to_budget.md
```diff
@@ -0,0 +1,54 @@
...
+SELECT
+  ProviderName,
+  SubAccountId,
+  SubAccountName,
+  SUM(BilledCost) AS TotalBilledCost
+FROM focus_data_table
+WHERE ChargeCategory = 'Usage'
+  AND ChargePeriodStart >= ? and ChargePeriodEnd <= ?
+  AND ProviderName = ?  
+  AND SubAccountId = ?
```
flanakin, line 38: nit: I'm assuming this is the exact query from the UC library, but shouldn't we leave extraneous filter conditions out? I would love to simplify this as much as possible. People should be able to figure out filtering on their own.
```suggestion
```

supporting_content/use_cases/compare_billed_cost_per_subaccount_to_budget.md
```diff
@@ -0,0 +1,54 @@
...
+## Example SQL Query
+```
+SELECT
+  ProviderName,
+  SubAccountId,
+  SubAccountName,
+  SUM(BilledCost) AS TotalBilledCost
+FROM focus_data_table
+WHERE ChargeCategory = 'Usage'
+  AND ChargePeriodStart >= ? and ChargePeriodEnd <= ?
```
flanakin, line 36: Why are we only showing usage? That doesn't seem like it's something that can be compared to a budget since purchases, adjustments, credits, and possibly taxes would impact budgets.
```suggestion
WHERE ChargePeriodStart >= ? AND ChargePeriodEnd <= ?
```

supporting_content/use_cases/compare_billed_cost_per_subaccount_to_budget.md
```diff
@@ -0,0 +1,54 @@
...
+
+## Description
+Product wants to confirm charges are in line with expectations (budgets) for a business which has all its cost within a single subaccount. The ChargePeriod is used for filtering to capture items that apply to the period only (excludes adjustments from previous periods that were applied in the current BillingPeriod).
+
+## 1st Degree Columns
+* Provider
+* Sub Account ID
+* Sub Account Name
+* Billed Cost
+* Charge Category
```
flanakin, line 19: Not actually needed for this use case
```suggestion
```

supporting_content/use_cases/compare_billed_cost_per_subaccount_to_budget.md
```diff
@@ -0,0 +1,54 @@
...
+Compare billed cost per subaccount to budget
+
+## Use Case ID
+14383
+
+## Supported Specification Version
+1.0
+
+## Description
+Product wants to confirm charges are in line with expectations (budgets) for a business which has all its cost within a single subaccount. The ChargePeriod is used for filtering to capture items that apply to the period only (excludes adjustments from previous periods that were applied in the current BillingPeriod).
```
flanakin, line 12: nit: "Product" as a pronoun is somewhat industry-specific. I would generalize to "Leadership". Also, ChargePeriod and BillingPeriod should be regular words in this context.
```suggestion
Leadership wants to confirm charges are in line with expectations (budgets) for a business which has all its cost within a single subaccount. The charge period is used for filtering to capture items that apply to the period only (excludes adjustments from previous periods that were applied in the current billing period).
```

supporting_content/use_cases/compare_billed_cost_per_subaccount_to_budget.md
```diff
@@ -0,0 +1,54 @@
+# Use Case
+
+Compare billed cost per subaccount to budget
+
+## Use Case ID
+14383
+
+## Supported Specification Version
```
flanakin, line 8: Can we align this to the columns and move it to the end of the doc?
```suggestion
## Introduced (version)
```

supporting_content/use_cases/compare_billed_cost_per_subaccount_to_budget.md
```diff
@@ -0,0 +1,54 @@
+# Use Case
+
+Compare billed cost per subaccount to budget
+
+## Use Case ID
+14383
```
flanakin, line 6: Is this our number or the UC library's number? Should we reset numbers for the spec? I'm not sure how we jumped up to 14.3k use cases already 😜

supporting_content/use_cases/compare_billed_cost_per_subaccount_to_budget.md
```diff
@@ -0,0 +1,54 @@
...
+## Parameters
+* ChargePeriodStart
+* ChargePeriodEnd
+* ProviderName
+* SubAccountId
+
+## Use Case Change Log
+| Version | Change | Description |
+| :-------| :------| :-----------|
+|
```
flanakin, line 54: I would avoid adding a changelog to this doc. Let's keep updates centralized in the main changelog.
```suggestion
```

supporting_content/use_cases/compare_billed_cost_per_subaccount_to_budget.md
```diff
@@ -0,0 +1,54 @@
...
+## Use Case ID
+14383
+
+## Supported Specification Version
+1.0
+
+## Description
+Product wants to confirm charges are in line with expectations (budgets) for a business which has all its cost within a single subaccount. The ChargePeriod is used for filtering to capture items that apply to the period only (excludes adjustments from previous periods that were applied in the current BillingPeriod).
+
+## 1st Degree Columns
```
flanakin, line 14: nit: Should these be column names or IDs?

supporting_content/use_cases/compare_billed_cost_per_subaccount_to_budget.md
```diff
@@ -0,0 +1,54 @@
+# Use Case
+
+Compare billed cost per subaccount to budget
```
flanakin, line 3: @beau-nelford-anglepoint What changed in 1.1 that impacted this use case? None of the columns here are flagged as changed in the 1.1 changelog.

flanakin (approved):

specification/use_cases/use_cases.md
flanakin, line 1: nit: Should this be in the specification folder? Maybe templates should be in a root `templates` folder. And shouldn't the template be a full example with all sections built out?

supporting_content/use_cases/compare_billed_cost_per_subaccount_to_budget.md
rileyjenk: Yeah having it be a introduced version makes more sense.

supporting_content/use_cases/compare_billed_cost_per_subaccount_to_budget.md
rileyjenk: Budgets are not an inherent part of the FOCUS spec. Refinement needed on what the actual supported feature is. For now fine, but we need to distill these down to what we actually support.

rileyjenk (approved):

specification/use_cases/use_cases.md
rileyjenk: I believe it should be. We are going to address location and such in the other PR. This PR merges the two into one.

PR #699, opened 2025-01-30
WI #534: Add Commitment Discount Unit Price and Cost columns

### LINKED ISSUE #534: Expose a CommitmentDiscountUnitPrice and a CommitmentDiscountCost column to expose its unit price and cost to compare against List and Contracted unit rates/costs.
Exposing a commitment discount unit price and cost allows customers to determine whether the effective cost was solely determined by the committed rate or combined with possibly additional discounts. For customers with high commitment discount coverage, this commitment discount unit price and cost information is more useful for forecasting future costs than list or contracted cost/unit prices since these services majority of eligible services are being charged at these committed rates. 

Currently:

1. When a commitment discount covers a resource, its unit price is not exposed by default.

2. A contracted discount can also apply to a commitment discount rate, so showing neither makes deriving the original commitment discount rate, and its additional savings, impossible.

cnharris10:
Authored by @qquigley and @cnharris10.

This pull request proposes adding columns, `CommitmentDiscountUnitPrice` and `CommitmentDiscountCost`, to help practitioners better understand the breakdown of pre/post-negotiated, discounted rates/costs and to use this information to better forecast future cost and usage trends.

Example `CommitmentDiscountUnitPrice` and `CommitmentDiscountCost` data can be found [here](https://docs.google.com/spreadsheets/d/1AYDPZ4rl90PPEbyQJUaGCUSwZ4s_sfy-8sO2KyNq0aM/edit?pli=1&gid=1712247049#gid=1712247049).

image

### DIFF
--- a/specification/columns/commitmentdiscountcost.md
+++ b/specification/columns/commitmentdiscountcost.md
@@ -1,6 +1,6 @@
 # Commitment Discount Cost
 
-Commitment Discount Cost represents the cost calculated by multiplying the [Commitment Discount Unit Price](#glossary:commitmentdiscountunitprice) and the corresponding [Pricing Quantity](#pricingquantity). Commitment Discount Cost is denominated in the [Billing Currency](#billingcurrency) and is commonly used to calculate savings incurred by [*commitment discounts*](#commitment-discount).
+Commitment Discount Cost represents the cost calculated by multiplying the [Commitment Discount Unit Price](#glossary:commitmentdiscountunitprice) and the corresponding [Pricing Quantity](#pricingquantity) inclusive of any [*negotiated discounts*](#glossary:negotated-discount) that may apply. Commitment Discount Cost is denominated in the [Billing Currency](#billingcurrency) and is commonly used to calculate savings incurred by [*commitment discounts*](#commitment-discount).
 
 The CommitmentDiscountCost column adheres to the following requirements:
 
@@ -10,20 +10,16 @@ The CommitmentDiscountCost column adheres to the following requirements:
 * CommitmentDiscountCost MUST be denominated in the *BillingCurrency*.
 * CommitmentDiscountCost MUST be present when [*CommitmentDiscountId*](#glossary:commitmentdiscountid) is present.
 * CommitmentDiscountCost MUST equal [*PricingQuantity*](#pricingquantity) * ([*CommitmentDiscountUnitPrice*](#commitmentdiscountunitprice) - ([*ListUnitPrice*](#listunitprice) - [*ContractedUnitPrice*](#contractedunitprice))).
+* CommitmentDiscountCost MUST equal the [*covering cost*](#glossary:coveringcost) for all [*rows*](#glossary:row) of the same [*charge period*](#glossary:chargeperiod) and [*CommitmentDiscountID*](#commitmentdiscountid) before any applicable [*negotiated discounts*](#glossary:negotiateddiscount) when [*ChargeCategory*](#chargecategory) is "Usage", [*PricingCategory*](#pricingcategory) is "Committed", and [*ChargeClass*](#chargeclass) is not "Correction".
+* CommitmentDiscountCost MAY equal the *covering cost* before any applicable *negotiated discounts* when [*ChargeCategory*](#chargecategory) is "Purchase", depending on the payment schedule of the *commitment discount*.
 * [*BilledCost*](#billedcost) MUST equal CommitmentDiscountCost when [*ChargeCategory*](#chargecategory) is "Purchase" and *ChargeClass* is not "Correction".
 * [*EffectiveCost*](#billedcost) MUST equal CommitmentDiscountCost when *ChargeCategory* is "Usage" and *ChargeClass* is not "Correction".
 * CommitmentDiscountCost MAY be null or any valid decimal value if *ChargeClass* is "Correction".
 
-When [*CommitmentDiscountCategory*](#commitmentdiscountcategory) is "Spend" and *ChargeClass* is not "Correction", the following applies:
+When a provider supports [*commitment discount flexibility*](glossary:commitment-discount-flexibility) for a [*CommitmentDiscountType*](#commitmentdiscounttype) and *ChargeClass* is not "Correction", the following applies:
 
-* When [*ChargeCategory*](#chargecategory)] is "Purchase", CommitmentDiscountCost MUST be the predefined amount of spend committed for each [*charge period*](glossary:chargeperiod) of the *commitment discount's* [*term*].
-* When *ChargeCategory* is "Usage", CommitmentDiscountUnitPrice MUST be the [*effective unit price*](glossary:effective-unit-price) of the covering resource's SKU.
-
-When *CommitmentDiscountCategory* is "Usage" and *ChargeClass* is not "Correction", the following applies:
-
-* CommitmentDiscountUnitPrice MUST be the *effective unit price* of the preselected SKU.
-* When [*commitment discount flexibility*](glossary:commitment-discount-flexibility) applies, the CommitmentDiscountCost of the covering resource's SKU MAY be calculated from a different *CommitmentDiscountUnitPrice* than the purchasing SKU.
-* When *commitment discount flexibility* does not apply, the CommitmentDiscountCost of the covering resource's SKU MUST be calculated from the same *CommitmentDiscountUnitPrice* as the purchasing SKU.
+* When *commitment discount flexibility* applies, the CommitmentDiscountCost of the covering resource's [*SkuId*](#skuid) MAY be calculated from a different *CommitmentDiscountUnitPrice* than the purchasing *SkuId*.
+* When *commitment discount flexibility* does not apply, the CommitmentDiscountCost of the covering resource's *SkuId* MUST be calculated from the same *CommitmentDiscountUnitPrice* as the purchasing *SkuId*.
 
 ## Column ID
 
@@ -35,7 +31,7 @@ Commitment Discount Cost
 
 ## Description
 
-Cost calculated by multiplying Commitment Discount Unit Price and the corresponding Pricing Quantity.
+Cost calculated by multiplying Commitment Discount Unit Price and the corresponding Pricing Quantity inclusive of any *negotiated discounts* that may apply.
 
 ## Usability Constraints
 
--- a/specification/columns/commitmentdiscountcost.md
+++ b/specification/columns/commitmentdiscountcost.md
@@ -1,0 +1,57 @@
+# Commitment Discount Cost
+
+Commitment Discount Cost represents the cost calculated by multiplying the [Commitment Discount Unit Price](#glossary:commitmentdiscountunitprice) and the corresponding [Pricing Quantity](#pricingquantity). Commitment Discount Cost is denominated in the [Billing Currency](#billingcurrency) and is commonly used to calculate savings incurred by [*commitment discounts*](#commitment-discount).
+
+The CommitmentDiscountCost column adheres to the following requirements:
+
+* CommitmentDiscountCost MUST be present in a [*FOCUS dataset*](#glossary:FOCUS-dataset) when the provider supports *commitment discounts*.
+* CommitmentDiscountCost MUST be of type Decimal.
+* CommitmentDiscountCost MUST conform to [Numeric Format](#numericformat) requirements.
+* CommitmentDiscountCost MUST be denominated in the *BillingCurrency*.
+* CommitmentDiscountCost MUST be present when [*CommitmentDiscountId*](#glossary:commitmentdiscountid) is present.
+* CommitmentDiscountCost MUST equal [*PricingQuantity*](#pricingquantity) * ([*CommitmentDiscountUnitPrice*](#commitmentdiscountunitprice) - ([*ListUnitPrice*](#listunitprice) - [*ContractedUnitPrice*](#contractedunitprice))).
+* [*BilledCost*](#billedcost) MUST equal CommitmentDiscountCost when [*ChargeCategory*](#chargecategory) is "Purchase" and *ChargeClass* is not "Correction".
+* [*EffectiveCost*](#billedcost) MUST equal CommitmentDiscountCost when *ChargeCategory* is "Usage" and *ChargeClass* is not "Correction".
+* CommitmentDiscountCost MAY be null or any valid decimal value if *ChargeClass* is "Correction".
+
+When [*CommitmentDiscountCategory*](#commitmentdiscountcategory) is "Spend" and *ChargeClass* is not "Correction", the following applies:
+
+* When [*ChargeCategory*](#chargecategory)] is "Purchase", CommitmentDiscountCost MUST be the predefined amount of spend committed for each [*charge period*](glossary:chargeperiod) of the *commitment discount's* [*term*].
+* When *ChargeCategory* is "Usage", CommitmentDiscountUnitPrice MUST be the [*effective unit price*](glossary:effective-unit-price) of the covering resource's SKU.
+
+When *CommitmentDiscountCategory* is "Usage" and *ChargeClass* is not "Correction", the following applies:
+
+* CommitmentDiscountUnitPrice MUST be the *effective unit price* of the preselected SKU.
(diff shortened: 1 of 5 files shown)


specification/columns/commitmentdiscountunitprice.md
```diff
@@ -0,0 +1,55 @@
...
+* CommitmentDiscountUnitPrice MUST be a Decimal within the range of non-negative decimal values.
+* CommitmentDiscountUnitPrice MUST conform to [Numeric Format](#numericformat) requirements.
+* CommitmentDiscountUnitPrice MUST be denominated in the [*BillingCurrency*](#billingcurrency).
+* CommitmentDiscountUnitPrice MUST be present when [*CommitmentDiscountId*](#commitmentdiscountid) is present.
+* CommitmentDiscountUnitPrice MUST NOT include [*negotiated discounts*](#glossary:negotiated-discount)
+* CommitmentDiscountUnitPrice MAY be null or any valid decimal value if [*ChargeClass*](#chargeclass) is "Correction".
+
+When [*CommitmentDiscountCategory*](#commitmentdiscountcategory) is "Spend" and *ChargeClass* is not "Correction", the following applies:
+
+* When [*ChargeCategory*](#chargecategory) is "Purchase", CommitmentDiscountUnitPrice MUST be the predefined amount of spend committed for each [*charge period*](glossary:chargeperiod) of the *commitment discount's* [*term*](glossary:term).
```
ijurica, line 17: This requirement implies that Quantity = 1, while CommitmentDiscountUnitPrice represents the committed spend per charge period. However, I believe CommitmentDiscountUnitPrice should remain consistent, and PricingQuantity should reflect the actual committed spend, varying depending on the commitment (i.e., P x Q = C).

```suggestion
* When [*ChargeCategory*](#chargecategory) is "Purchase", CommitmentDiscountUnitPrice MUST be the predefined amount of spend committed for each [*charge period*](glossary:chargeperiod) of the *commitment discount's* [*term*](glossary:term).
```

specification/columns/commitmentdiscountunitprice.md
```diff
@@ -0,0 +1,55 @@
...
+* When *ChargeCategory* is "Usage", CommitmentDiscountUnitPrice MUST be the *effective unit price* of the covering resource's SKU.
+
+When *CommitmentDiscountCategory* is "Usage" and *ChargeClass* is not "Correction", the following applies:
+
+* CommitmentDiscountUnitPrice MUST be the *effective unit price* for the preselected SKU.
```
ijurica, line 22: CommitmentDiscountUnitPrice and the effective unit price match only if there are no negotiated discounts, right?

specification/columns/commitmentdiscountunitprice.md
```diff
@@ -0,0 +1,55 @@
...
+* When *commitment discount flexibility* applies, the CommitmentDiscountUnitPrice of the covering resource's SKU MAY be different than the purchasing SKU.
```
ijurica, line 23: I understand the idea and agree with it, but I would suggest linking those unit prices to SKU Price ID (we can rephrase this once we clarify and address SkuId and SkuPriceId). 

It seems to me that we could introduce something similar for the other unit prices as well.

```suggestion
* When *commitment discount flexibility* applies, the CommitmentDiscountUnitPrice of the covering resource's SKU MAY be different than the purchasing SKU.
```

specification/columns/commitmentdiscountunitprice.md
cnharris10: 👍 Will soften the language a bit to include this point

specification/columns/commitmentdiscountunitprice.md
cnharris10: I disagree with this point.  For a usage-based commitment discount, the unit price is the SKU's discounted rate _selected_ by the user for the term.  For a spend-based commitment discount, the unit price is the charge period's committed spend _selected_ by the user for the term.  So, this concludes as:

```
<Number of charge period time blocks> * 
{
    <User-selected SKU with fixed rate>,   # Usage-based Commitment Discount
    <User-selected commitment rate>       # Spend-based Commitment Discount
} = 
Commitment Discount Purchase Total (to amortize)
```

From an AWS perspective, there is no rate associated with a spend-based commitment discount purchase.

For example, for a Compute Savings Plan, the SKU is `2MHX3QZPD4ABTS29` and only has term metadata:

image

and the only unit rates reflected are the covering resource(s):

image

specification/columns/commitmentdiscountunitprice.md
udam-f2: Also only in the case of no-upfront? if you partially/fully paid, the unit price for the hourly usage is not going to equal what is referred to as an 'effective unit price'?

cnharris10: The unit price should always be stable regardless of payment option. The pricing quantity though will be different across payment options.

For example, with a savings plan committing to $1, the varying purchase records' pricing_quantity values per charge period would be:

No upfront: 1.0
Partial Upfront: 0.5
All Upfront: 0 (not included)

against a stable rate:

Effective unit rate: $1

specification/columns/commitmentdiscountunitprice.md
cnharris10: I've incorporated *SkuId* into both columns and we can clarify if needed later on.

specification/columns/commitmentdiscountunitprice.md
cnharris10: I've changed the glossary term: *effective unit price* to *covering cost* to indicate the amount of spend that a purchase (only if no upfront) and usage rows must sum up towards per charge period.  This *covering cost* is derived as:

```
Usage-based: Cost of resource over term / number of hours in a term
Spend-based: User-defined committed rate
```

specification/glossary.md
```diff
@@ -64,6 +64,10 @@ A feature of [*commitment discounts*](#glossary:commitment-discount) that may fu
 
 The agreed-upon unit price for a single [Pricing Unit](#pricingunit) of the associated SKU, inclusive of negotiated discounts, if present, and exclusive of any other discounts. This price is denominated in the [Billing Currency](#glossary:billingcurrency).
 
+<a name="glossary:covering-cost">Covering Cost</a>
+
+The amount of spend applied to each [*charge period*] derived as the total committed purchase amount divided by the number of [*charge periods*](glossary:chargeperiod) within a [*commitment discount's*](#glossary:commitment-discount) [*term*](#glossary:term).
```
ijurica, line 69: I understand that describing this concept is challenging, and I don’t have a specific suggestion at this time ☹, but I believe we should consider rephrasing the Covering Cost definition (or whatever term we decide to use).

We agreed that the sample dataset is okay, but it seems to me that we need to rephrase the Covering Cost definition to ensure the result aligns with the data in the sample dataset for One-Time Spend Commitment Discount Purchases.

Here is my interpretation of Covering Cost (and I may have misunderstood it):

For the All-Upfront and Partial Upfront payment options, for One-Time purchase records, the charge period should match the eligibility duration, i.e., it refers to the entire commitment term, correct? As a result, the Covering Cost for an All-Upfront purchase equals the total committed purchase amount, which equals the CommitmentDiscountUnitPrice. However, this does not seem to align with the sample data.

The entire dataset is available at this [link](https://docs.google.com/spreadsheets/d/1AYDPZ4rl90PPEbyQJUaGCUSwZ4s_sfy-8sO2KyNq0aM/edit?pli=1&gid=1712247049#gid=1712247049), and here is a subset of the relevant columns and rows from the dataset:


| Charge Description                                              | Charge Category | Commitment Discount Category | Pricing Category | Pricing Quantity | Commitment Discount UnitPrice | Commitment Discount Cost | List Cost | List Unit Price | Contracted Cost | Contracted Unit Price | Billed Cost | Effective Cost |
|------------------------------------------------------------------|-----------------|------------------------------|------------------|------------------|-------------------------------|---------------------------|-----------|------------------|-----------------|------------------------|-------------|----------------|
| Purchase (3 years, $1 hourly commitment) - Recurring             | Purchase        | Spend                        | Standard         | 1.00             | 1                             | 1                         | 1.00      | 1.0              | 1.00            | 1.0                    | 0.00        | 0.0            |
| Purchase (3 years, $1 hourly commitment) - Partial Upfront - One-Time | Purchase        | Spend                        | Standard         | 39,220.00        | 1                             | 39,220.00                 | 39,220.00 | 1.0              | 39,220.00       | 1.0                    | 39,220.00   | 0.00           |
| Purchase (3 years, $1 hourly commitment) - Partial Upfront - Recurring | Purchase        | Spend                        | Standard         | 0.50             | 1                             | 0.50                      | 0.50      | 1.00             | 0.50            | 1.00                   | 0.50        | 0.00           |
| Purchase (3 years, $1 hourly commitment) - All Upfront           | Purchase        | Spend                        | Standard         | 78,440.00        | 1                             | 78,440.00                 | 78,440.00 | 1.0              | 78,440.00       | 1.0                    | 78,440.00   | 0.00           |

ijurica: **Additional Question:**

If there is a negotiated discount in place, and it applies to the Spend Commitment Discount Purchase, how would you interpret the term total committed purchase amount - is it inclusive of negotiated discounts or not? Assuming that Commitment Discount Unit Price and Cost are exclusive of any negotiated discounts, we may want to clarify this explicitly in the Covering Cost definition.

specification/columns/commitmentdiscountcost.md
```diff
@@ -0,0 +1,53 @@
+# Commitment Discount Cost
+
+Commitment Discount Cost represents the cost calculated by multiplying the [Commitment Discount Unit Price](#glossary:commitmentdiscountunitprice) and the corresponding [Pricing Quantity](#pricingquantity) inclusive of any [*negotiated discounts*](#glossary:negotated-discount) that may apply. Commitment Discount Cost is denominated in the [Billing Currency](#billingcurrency) and is commonly used to calculate savings incurred by [*commitment discounts*](#commitment-discount).
```
ijurica, line 3: inclusive or exclusive?

specification/glossary.md
cnharris10: It aligns perfectly with the sample data. 😄 

The `CommitmentDiscountUnitPrice` for an All Upfront Spend-based Commitment Discount is: `$1.00`, and the covering cost, using the definition below, is calculated as ...

```
The amount of spend applied to each [*charge period*] derived as the total committed purchase amount divided by the number of charge periods within a commitment discount's term.
```

"the total committed purchase amount" --> `26,280 hours / year * 3 years * $1 committed rate`
"divided by" --> `/`
"the number of charge periods within a commitment discount's term" --> `26,280 hours / year * 3 years`
= `26,280 hours / year * 3 years * $1 committed rate / 26,280 hours / year * 3 years = $1.00`

ijurica: The charge period definition specifies that ...**The charge period for a non-usage charge with time boundaries should match the duration of eligibility**....

Based on this, the charge period for this purchase record matches the duration of eligibility, which results in `NumOfChargePeriods = 1.`

I understand that we expected the number of charge periods to be 26,280 * 3, but if you interpret the charge period definition literally, it leads to 1.

cnharris10: As our data sheet as shown since 1.1 and commitment discount appendix, the `PricingQuantity` reflects the number of hours over the term against the $1.00 `CommitmentDiscountUnitCost`.  

image

ijurica: I agree with the sample data.

What I'm saying is that when I follow the normative requirements for Spend Commitment Discount Purchase and apply the definition of Covering Cost, I don't get the expected result - the one reflected in the sample data.

ijurica: **Question:** 

Is the CommitmentDiscountUnitPrice for Commitment Discount Purchase records always identical to the ListUnitPrice?

It's exclusive of any negotiated discounts, correct? But should we exclude all discounts from this unit price in the case of Commitment Discount Purchases, making the CommitmentDiscountUnitPrice identical to the ListUnitPrice?

cnharris10: > I agree with the sample data.

> What I'm saying is that when I follow the normative requirements for Spend Commitment Discount Purchase and > apply the definition of Covering Cost, I don't get the expected result - the one reflected in the sample data.

Please either suggest _extremely_ specific area of the wording that you do not understand or suggest a revision with additional clarification.  Past that, there's nothing I can do about this comment.

> Question:

> Is the CommitmentDiscountUnitPrice for Commitment Discount Purchase records always identical to the ListUnitPrice?

> It's exclusive of any negotiated discounts, correct? But should we exclude all discounts from this unit price in the case of Commitment Discount Purchases, making the CommitmentDiscountUnitPrice identical to the ListUnitPrice?

ListUnitPrice should absolutely not be identical to CommitmentDiscountUnitPrice give the former is a public, non-discounted rate and separate public, discounted rate.  Making these identical goes against the goals of the Work Item description.

aws-letianfeng:
Shall we define a column like "Gross Effective Cost" to solve this use case? Effective Cost is defined as cost inclusive of the impacts of all reduced rates and discounts, augmented with the amortization of relevant purchases (one-time or recurring) paid to cover future eligible charges. What we are trying to achieve here is to figure out if there's no discount, what would a customer have paid  if no discount applies. I would try to avoid columns that have the definition of "if this, then the value should be populated".

cnharris10:
> Shall we define a column like "Gross Effective Cost" to solve this use case? Effective Cost is defined as cost inclusive of the impacts of all reduced rates and discounts, augmented with the amortization of relevant purchases (one-time or recurring) paid to cover future eligible charges. What we are trying to achieve here is to figure out if there's no discount, what would a customer have paid  if no discount applies. I would try to avoid columns that have the definition of "if this, then the value should be populated".

This is certainly possible to reframe as a PreEffectiveUnitPrice and PreEffectiveCost or similar naming convention. Let's discuss this further.

udam-f2:
Possibly an unpopular opinion: 
This seems, in my personal opinion, like a niche use case scenario to justify introducing not one but two new columns to the spec. Just remember a cost column has to be specified for all rows for it to be useful in comparison use cases. This means for the cost column to be useful, all providers that don't support commitments will have to specify it - which to me seems like a step too far. 

The unit price _could_ provide some value (allows the cost to be calculated for the limited use cases). This category of data IMO would fit perfectly in a Commitment properties column (assuming we can get the terminology issues worked out) along with many other important commitment details (we've talked about term, payment format). An alternative lower-impact way to accomplish this would be as a part of SKUPriceDetails, where when covered by a commitment, those rows can have a key that has the public commitment unit price (use a normalized key) - which seems like it would serve the same use case.

cnharris10:
> Possibly an unpopular opinion: 
> 
> This seems, in my personal opinion, like a niche use case scenario to justify introducing not one but two new columns to the spec. Just remember a cost column has to be specified for all rows for it to be useful in comparison use cases. This means for the cost column to be useful, all providers that don't support commitments will have to specify it - which to me seems like a step too far. Would love to learn more and be better informed.
> 
> 
> 
> The unit price _could_ provide some value (allows the cost to be calculated for the limited use cases). This category of data IMO would fit perfectly in a Commitment properties column (assuming we can get the terminology issues worked out) along with many other important commitment details (we've talked about term, payment format). An alternative lower-impact way to accomplish this would be as a part of SKUPriceDetails, where when covered by a commitment, those rows can have a key that has the public commitment unit price (use a normalized key) - which seems like it would serve the same use case.
> 
> 
> 
> Again, personal opinion, in my role as a practitioner, the folks I was pushing the FinOps data to nor I never worried about prices that didn't include negotiated discounts (even List wasn't that useful other than for lining up with pricing sheets). Cost without negotiated discounts is a hypothetical price that an organization never has to deal with/worry about. If they don't have a discount, then EffectiveCost already captures what's intended here. If they do have a discount, then contracted cost becomes the theoretical highest price anyone can pay (and effective captures with commitment), and the diff tells the rate optimization benefit. Again, would love to learn more and understand why this is worthy of an expensive cost column and another unit price.

The point about not having a unit price / cost value for non-discounted rows is very valid. This definitely needs to be addressed before moving forward. I like @aws-letianfeng idea about generalizing this work item into a pre-discounted effective rate and cost column which would solve the issue you pointed out. Let's discuss further in TF-1 next week.

shawnalpay:
@aws-letianfeng Are you able to attend TF1 tomorrow?  If not, please coordinate a time with @cnharris10 (and potentially @udam-f2) to align on a path forward.

shawnalpay:
@cnharris10 Can we close this PR?  If not, which of its contents be applicable to #961?  We need to either clean up this PR to be mergeable (there are conflicts to resolve), or we need to close and repurpose.

Setting back to draft.

Issue #406, opened 2024-04-18
[WI] Add support for coverage eligibility


cnharris10:
## 1. **Problem Statement** *

> - **What is the problem?**: Explain the context and why it needs resolution.
> - **Impact**: Describe how the problem affects users, systems, or the project.

**Problem:**  It is a non-trivial task to determine whether an on-demand usage-based charge is eligible for a commitment (and which commitment) or not. 

**Impact:** Commitment discount efficiency and rate optimization analysis is inaccessible or unachievable for practitioners with a provider-native dataset, without joining additional data or context. This adds undue overhead to computation of commitment coverage, and undermines decision-making capability, particularly around purchases and renewals. 

Use-case: [Analyze resource costs by SKU](https://focus.finops.org/use-cases/?prod_use_cases%5BrefinementList%5D%5Brelated_personas.post_title%5D%5B0%5D=FinOps%20Practitioner&prod_use_cases%5BrefinementList%5D%5Brelated_personas.post_title%5D%5B1%5D=Finance#modal-column-14595)

For this use-case, additional information around coverage eligibility allows practitioners to determine whether a resource's cost can be reduced or not.

---

## 2. **Objective** *

> - **Success Criteria**: Define how success will be measured (e.g. metrics and KPIs).

**Success Criteria:** 
- [ ] Practitioners are able to discern commitment eligibility of a charge
- [ ] Practitioners are able to identify type of commitment eligibility
- [ ] Practitioners are able to systematically compute commitment coverage and usage

---

## 3. **Supporting Documentation** *
<details>

<summary>
Example restrictions across AWS, Azure, GCP
</summary>

1. AWS
    a. Spot instances
    b. [P5 instance family](https://aws.amazon.com/ec2/pricing/reserved-instances/pricing/)
    c. [Mac instance families](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/ec2-mac-instances.html#:~:text=Mac%20instances%20are%20available%20only%20as%20On%2DDemand%20Instances.%20They%20are%20not%20available%20as%20Spot%20Instances%20or%20Reserved%20Instances.%20You%20can%20save%20money%20on%20Mac%20instances%20by%20purchasing%20a%20Savings%20Plan.)
[...]

PR #747, opened 2025-02-09
WI #647: Adds `Amortization Class` column

### LINKED ISSUE #647: [WI] Prevent double-counting list/contracted cost for commitment discounts
### Problem Statement

When calculating savings, you need to compare the sum of `ListCost` or `ContractedCost` with the sum of `EffectiveCost`. When doing that, commitment discount purchase costs are double counted because both the purchase and usage records have a `ListCost` and `ContractedCost`.

**Use Case**
NA

### Objective

Practitioners have a simple method to calculate savings that doesn't accidentally double-count commitment discount purchase costs.

### Supporting Documentation

This example demonstrates a commitment discount that gets 50% off list price when the organization has an 20% discount on all usage.
ChargeDescription | ChargePeriodStart | ListCost | ContractedCost | BilledCost | EffectiveCost
---|---|---:|---:|---:|---:
Reservation | Jan 1 | $365.00 | $365.00 | $365.00 | $0.00
VM usage | Jan 1 | $2.00 | $1.80 | $0.00 | $1.00
VM usage | Jan 2 | $2.00 | $1.80 | $0.00 | $1.00
VM usage | ... | ... | ... | ... | ...
VM usage | Dec 31 | $2.00 | $1.80 | $0.00 | $1.00
Sum of 365 days | Jan 1-Dec 31 | $1,095.00 | $1,022.00 | $365 | $365

Some options:
1. Cut: ~Split ListCost and ContractedCost into separate columns for actual and amortized cost.~
2. Add another column that identifies amortizable costs to make them easier to filter out.

One proposal for option 2:

`AmortizationCategory` column with values of:
- **Principal** – Original charge that gets amortized over time (e.g., commitment discount purchases).
- **Amortized Charge** – Rows that are a breakdown of the principal payment (e.g., commitment discount usage).

I avoided the words "purchase" and "usage" here since we could have purchases amortized down to recurring purchases in the future.

### Epic or Theme Association

Savings calculation

### Stakeholders

_No response_

### LINKED ISSUE #740: [AI] Complete PR draft for WI #647
### Detailed Description

_No response_

### LINKED ISSUE #817: [AI] Facilitate decision on naming via spreadsheet and/or vote
### Detailed Description

_No response_

JGrubb:
refs #647


specification/columns/amortizationcategory.md
```diff
@@ -0,0 +1,41 @@
...
+| Allows nulls    | False          |
+| Data type       | String         |
+| Value format    | Allowed values |
+
+Allowed values:
+
+| Value            | Description                          |
+| :--------------- | :------------------------------------|
+| Principal        | Original charge that is amortized over time into usage by other line items. |
+| Amortized Charge | Rows that receive the benefit of the original principal payment |
```
shawnalpay, line 37: Does this need to `Amortized Charge`?  Could it just be `Amortized`?

specification/columns/amortizationcategory.md
```diff
@@ -0,0 +1,41 @@
...
+
```
shawnalpay, line 38: Do we need more categories? 
- `N/A` or similar?
- `Coverable` or similar?

shawnalpay:
Potential values, if we expand this Work Item to handle also for #406:

(One level hierarchy)
- Commitment purchase
- Eligible - Covered
- Eligible - Not Covered
- Not Coverable

(Two level hierarchy)
- Commitment purchase
- Coverage Eligible
  - Covered
  - Not Covered
- Not Coverable

(Purchase) Principal: purchases of commitment discounts
(Usage) Amortized: application of commitment discounts to resources
(Usage) Is Not Amortized, but could be covered: resources that could have been covered but weren't
(Usage) N/A: resources that can't be covered

robmartin33:
In terms of identifying eligible for discount items, it seems to me that CSPs all (I think) have guidelines on how they apply coverage in terms of priority of application of types of discounts (least flexible commitments first, RIs before SPs, EC2 SPs before Compute SPs, etc.) and I believe all promise to apply the discounts to things that get the “greatest” discount percentage, and then to small instances over big when ISF applies… AND THEN randomly within whatever is left somehow. It is also likely that there are more complex application rules to account for “strange” conditions of private agreements, but cannot confirm or discuss. 

So if they are doing this work in order to figure out where to apply discounts, it is likely that they are doing some sort of flagging of resources in order to work that all out, and therefore may be able to identify things that were eligible to be covered by at least one thing (or n things), and could identify them...

udam-f2:
Full transparency - I've not been involved in this work and don't have all the context and don't have a good naming suggestion.

Something to think about:
Isn't there a need to handle the double counting scenario for 'negotiated terms' as well? I don't think 'Amortization' is a general term that would work in that case.

For example, I sign with Acme Co. and commit to using $1000$ per month. I pay at the end of the month (including the unused portion). They provide a 'fee' record for $1000$ for that month (Billed Cost). Then, throughout the month, there are usage rows that burn down that negotiated term (Effective Cost). Both rows have list/contracted cost. Does the amortization term make sense for this column when trying to avoid this double-counting scenario?

JGrubb:
Definitely, thanks Udam.  I think this is the gist of Irena's point in today's call and given that the other usecase for this column is determining a line item's eligibility for commitment coverage, I don't think this column name will stick as is.  Not sure yet what to put forth though.

shawnalpay:
Some alternative name options that come to mind:

Coverage Category
Coverage Eligibility
Cost Category
Discount Category
Payment Category

specification/columns/amortizationcategory.md
```diff
@@ -0,0 +1,51 @@
+# Amortization Category
+
+Amortization Category represents the various classifications of a charge
+based on it's eligibility for coverage by a commitment discount, and
```
narlaadbe, line 4: Minor
```suggestion
based on its eligibility for coverage by a commitment discount, and
```

specification/columns/amortizationcategory.md
```diff
@@ -0,0 +1,51 @@
...
+AmortizationCategory
+
+## Display Name
+
+Amortization Category
+
+## Description
+
+Represents the various classifications of a charge
+based on it's eligibility for coverage by a commitment discount, and
```
narlaadbe, line 27:
```suggestion
based on its eligibility for coverage by a commitment discount, and
```

GrinsteadJ:
As we expand support to SaaS providers, we likely also need to clearly indicate whether a line item has been covered by pre-payment or will incur a charge during the invoice period. 

This seems straightforward, but the following example highlights the need.
- A customer is consuming prepaid credits
- Mid-month, they run out of prepaid credits and start accumulating charges on the same deal.
- The customers will need to know which line items were covered by prepaid credits, and which are creating new charges.

If the customer is paying the SaaS provider directly, this isn't that big of a deal. Billed Cost would change from 0 to >0. However, for customers paying via marketplace, Billed Cost will be 0 whether using prepaid credits or not.

So we therefore need to indicate whether a row is prepaid or not. Commitment based discounts are not a consideration here, since there's not necessarily a discount associated with the prepayment. Customers just simply need to know if usage has been paid for already or not.

With this in mind, I think a name like "Payment Category" might be more inclusive to the use cases that can be addressed.

cc @udam-f2

ijurica:
Personally, I'm inclined to reuse the existing columns whenever possible, especially since negotiated and non-negotiated drawdown commitment discounts are conceptually so similar.

@GrinsteadJ , take a look at the [PricingCategory ](https://github.com/FinOps-Open-Cost-and-Usage-Spec/FOCUS_Spec/blob/working_draft/specification/columns/pricingcategory.md) column. For non-negotiated drawdown commitment discounts, such as Reserved Instances and Savings Plans, the Committed category indicates that the charge is covered, and the [CommitmentDiscountId ](https://github.com/FinOps-Open-Cost-and-Usage-Spec/FOCUS_Spec/blob/working_draft/specification/columns/commitmentdiscountid.md) allows you to identify the exact non-negotiated discount applied.

Until version 1.1, [discount handling attribute](https://github.com/FinOps-Open-Cost-and-Usage-Spec/FOCUS_Spec/blob/working_draft/specification/attributes/discount_handling.md) and commitment discount-related columns were not limited to non-negotiated drawdown commitments, and I assumed we planned to use them for negotiated drawdown commitments as well.

In one of our meetings, we mentioned the possibility of introducing a ContractId column (and stopping there - Contracts are out of scope for 1.2 😃). If we decide to proceed with ContractId (as a String or JSON if needed), would the PricingCategory value Committed (or an additional value, if this one is limited to non-negotiated drawdown commitments) provide all the necessary answers? Would you agree? 

Of course, we would also need to enrich the discount handling attribute to account for this behavior.

UdamDewaraja:
It would be good to have an initial discussion to see if this should be mixed up with the pricing category. One challenge i see is commitments and negotiated discounts (with a high-level commitment to a provider) can happen at the same time. Using 'Committed' for this would confusing because even things like 'Standard' would end up being 'Committed' for people with negotiated discounts based on commitments (which seems wrong).

specification/columns/amortizationcategory.md
```diff
@@ -0,0 +1,51 @@
...
+| Data type       | String         |
+| Value format    | Allowed values |
+
+Allowed values:
+
+| Value            | Description                          |
+| :--------------- | :------------------------------------|
+| Commitment Purchase | Original charge that is amortized over time into usage by other line items. |
+| Eligible - Covered | Usage that is discounted or covered by a commitment discount |
+| Eligible - Not Covered | Usage that could be covered by some form of commitment discount, but is not |
```
flanakin, line 46: Removing these since amortization isn't exclusive to commitment discounts. I'm not sure eligibility applies here. I believe it's either amortized or not.
```suggestion
| Amortized Charge | Charge that has an Effective Cost from an amortized Principal row and no Billed Cost. |
```

specification/columns/amortizationcategory.md
```diff
@@ -0,0 +1,51 @@
...
+| Not Applicable | Usage that is ineligible for commitment discounts of any kind. |
```
flanakin, line 47: Do we prefer "Not Applicable" or just plain null? Again, it's not about commitment discounts, but we could say amortization is not applicable. That feels like a placeholder value, which we say we don't support, tho 🤔
```suggestion
```

supporting_content/columns/amortizationcategory.md
```diff
@@ -0,0 +1,52 @@
+# Column: Amortization Category
+
+Objective: we need a simple method to calculate savings that doesn't accidentally double-count commitment discount purchase costs.
```
flanakin, line 3:
```suggestion
Objective: we need a simple method to calculate savings that doesn't accidentally double-count amortized purchase costs.
```

supporting_content/columns/amortizationcategory.md
```diff
@@ -0,0 +1,52 @@
...
+Objective: we need a simple method to identify which line items are and are not eligible for
+commitment discounts, and which actually have commitment discounts applied to them.
```
flanakin, line 5: This feels separate from amortization. I do think this is something we need, but it needs to account for usage and spend commitment discounts. I think that needs to be a separate column (or 2).

supporting_content/columns/amortizationcategory.md
```diff
@@ -0,0 +1,52 @@
...
+
+## Example provider mappings
+
+Current column mappings found in available data sets:
+
+| Provider | Data set                 | Column                    |
+|----------|--------------------------|---------------------------|
+| AWS      | CUR                      | lineItem/lineItemType     |
+| Google Cloud | BigQuery Billing Export | ??N/A because no upfront payments??  |
+| Microsoft | Cost details             | ?? Specific SKUs that designate upfront payments |
```
flanakin, line 15:
```suggestion
| Microsoft    | Cost details             | ChargeType indicates purchases and PricingModel indicates commitment discounts. There is no column that identifies when and how amortization applies. |
```

flanakin (changes requested):

specification/columns/amortizationcategory.md
```diff
@@ -0,0 +1,51 @@
...
+based on it's eligibility for coverage by a commitment discount, and
+whether or not a commitment discount was actually applied.
+
+Amortization Category is used to identify and distinguish between types of charges
+that may require different handling, for example an upfront payment on
+a Savings Plan that is amortized over time into usage by other line items.
+
+The AmortizationCategory column adheres to the following requirements:
+
+* The AmortizationCategory column MUST be present in a [*FOCUS dataset*](#glossary:FOCUS-dataset) and MUST NOT be null.
```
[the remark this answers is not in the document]
JGrubb: I would say "when a provider supports public commitment discounts" something to that effect.  GCP doesn't technically amortize their CUDs into the usage items because they do not do upfront payments of any kind.

specification/columns/amortizationcategory.md
JGrubb: To me it's not a placeholder, it's clearly delineating which line items are not able to be covered by a commitment discount (which I believe is where the eligibility conversation started).  I think `NULL` would do the job if we absolutely don't want to get into the eligibility topic.

specification/columns/amortizationcategory.md
flanakin: But we can't do that since amortization can apply to non-commitment charges.

specification/columns/amortizationcategory.md
flanakin: If you say "not applicable to be amortized", then I don't disagree. Whether we have this or not just depends on whether we want to prirotize being overly explicit or reduce data size when the value doesn't provide much value. I see both sides. I'm only barely on one side of that fence because I hear a lot of complaints about data size. I know customers of other cloud providers aren't as sensitive to that as my customers are, tho.

supporting_content/columns/amortizationcategory.md
flanakin: Monthly fees can still be amortized. Amortization [as I understand it] is the process of taking any charge and splitting it out by the time granularity of the dataset. So if I amortize a charge that has a 7-day applicability period, then I would see 7 amortized rows in a daily dataset and 168 amortized rows in an hourly dataset. Conversely, if I have a monthly dataset, then there'd only be 1 amortized row. All of these are on top of the original/principal/source charge.

specification/columns/amortizationcategory.md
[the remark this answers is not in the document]
JGrubb: Very fair, and it occurred to me on the call just now that they _are_ amortized, you just have to look at an hourly granularity, since the GCP CUD is an hourly-level commitment that benefits resources that are running during that hourly.

It's the same as a no upfront savings plans or reservation at the other two.

specification/columns/amortizationcategory.md
JGrubb: I mean, I would say that the Chris/Datadog scenario where you're paying upfront for the year is still a commitment, so we need a word that covers public discount instruments as well as contracted upfront payments.

supporting_content/columns/amortizationcategory.md
JGrubb: Agree, per above.

specification/columns/amortizationcategory.md
```diff
@@ -0,0 +1,41 @@
...
+| Feature level   | Mandatory      |
+| Allows nulls    | False          |
+| Data type       | String         |
+| Value format    | Allowed values |
+
+Allowed values:
+
+| Value            | Description                          |
+| :--------------- | :------------------------------------|
+| Principal        | Original charge that is amortized over time into usage by other line items. |
```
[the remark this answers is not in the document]
JGrubb: I like "prepayment" personally.  It's the most general value that can apply to the several scenarios discussed so far.

shawnalpay:
A scratchpad for what we discussed in Mar 4 TF1:

Some use cases we might want to handle for:
source charge that will get amortized (i.e. Commitment Discount Purchase)
commitment to use an amount
purchased from cloud provider but usage from SaaS provider
charges that are the after-effect of amortization

Agreed to keep the first three condensed and just keep it a flag, with Not Applicable.

Some possible names:

Prepurchase Source
Prepurchase Target
Not Applicable

Committed Source
Committed Target
Not Applicable

Prepurchase
Prepurchase Usage
Not Applicable

Commitment Purchase
Committed Usage
Not Applicable

(group felt we can't use the word Commitment/Committed in case a provider amortizes every purchase)

Commitment
Commitment Usage
Not Applicable

Prepayment
Prepaid
Not Applicable

shawnalpay:
@JGrubb @flanakin Per our discussion in TF1 today, I am going to put this PR back into Draft status until the following are complete:

- You both are aligned on column name and column values, which theoretically will be found through AI #817.
- The PR is revised to reflect that alignment.

supporting_content/columns/amortizationcategory.md
```diff
@@ -0,0 +1,52 @@
...
+    SUM(ListCost) as TotalCost,
+    AmortizationCategory
+...
+GROUP BY AmortizationCategory
+```
+
+
+
+## Discussion / Scratch space:
+
```
flanakin, line 52:
```suggestion

There are 4 known scenarios for double-counting:
1. Amortization
2. Burndown of a previous purchase with no charge for underutilized purchases
3. Burndown of a promise where no money is charged with a potential "unused" charge at the end of the period
4. Marketplace rows in a cloud provider's dataset and separately in a SaaS provider's dataset being combined by the practitioner

Amortization requirements:
* Must be able to identify (and filter out) the original/source charge.
  * This charge has a BilledCost of some positive value and an EffectiveCost of 0.
  * This charge is the one that will get filtered out when calculating savings.
  * This is typically a purchase but may not always be (e.g., monthly aggregated usage row could be broken down into daily/hourly usage rows).
  * In finance, the original charge is generally referred to as the "principal".
* Must be able to identify (and filter down to) the target or "effective" charges where the principal was amortized out over time.
* All amortized charges should include rows for both the source and target charges to ensure the filtering works as intended.
```

specification/columns/amortizationcategory.md
flanakin: Resolving this since we voted, but we can absolutely bring this back.

flanakin (approved):

flanakin (approved):

shawnalpay:
@JGrubb 
Given the changes @flanakin has performed here today, please review and confirm that this PR is now in a state that works for you!

CHANGELOG.md
```diff
@@ -263,61 +264,65 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
...
-| SkuId                        | 1.0-preview             |                   |
-| SkuMeter                     | 1.1                     |                   |
-| SkuPriceDetails              | 1.1                     |                   |
-| SkuPriceId                   | 1.0-preview             |                   |
-| SubAccountId                 | 0.5                     |                   |
-| SubAccountName               | 0.5                     |                   |
-| Tags                         | 1.0-preview             |                   |
-| UsageQuantity                | 1.0-preview             | Renamed to ConsumedQuantity in v1.0 |
-| UsageUnit                    | 1.0-preview             | Renamed to ConsumedUnit in v1.0 |
+| **Column ID**              | **Revision Introduced** | **Status**                                 |
```
UdamDewaraja, line 267: nit: Please preserve git blame. This whitespace only change has no impact on the content/rendering.

specification/columns/amortizationclass.md
```diff
@@ -0,0 +1,44 @@
+# Amortization Class
```
UdamDewaraja, line 1: There are a few things about this that seem problematic to me
- We're really talking about pre-payment here and the process of burning down pre-payments.
- When you pre-pay a contract (be it SaaS or Cloud or whatever), and burn that down (possibly very unevenly) over a term, are those charges considered 'Amortized Charges'? To me, it doesn't seem that way.
- This says allows nulls False but the normative statements says this should be null when its not Principal or the Amortized Charge.

The original fix may need to happen in EffectiveCost to avoid talking about amortized cost. 

Would something like ChargeCoverageCategory with values like 
- Pre-payment Charge: Billed > 0, effective = 0
- Burndown Charge: Billed=0, effective>0
- Standard Charge: Billed = effective work? 

If too late to take these suggestions, respond here as I can't attend the calls.

specification/columns/amortizationclass.md
```diff
@@ -0,0 +1,44 @@
...
+
+Amortization Class indicates how a row in a [*FOCUS dataset*](#glossary:FOCUS-dataset) relates to [*amortization*](#glossary:amortization). Amortization Class is used to identify and distinguish between the charges before and after *amortization* is applied. Amortization Class is commonly used to quantify cost savings accurately by excluding charges that can be double-counted when aggregated List Cost or Contracted Cost.
+
+The AmortizationClass column adheres to the following requirements:
+
+* AmortizationClass column MUST be present in a [*FOCUS dataset*](#glossary:FOCUS-dataset) when the provider supports *amortization*.
+* AmortizationClass MUST be of type String.
+* AmortizationClass MUST be "Principal" when the row represents an original charge that is amortized across multiple other rows.
+* AmortizationClass MUST be "Amortized Charge" when the row represents an result of *amortization* from a previous charge.
```
gparker-at-sf, line 10: nit - 'an result' should be 'a result'

specification/columns/amortizationclass.md
```diff
@@ -0,0 +1,44 @@
+# Amortization Class
+
+Amortization Class indicates how a row in a [*FOCUS dataset*](#glossary:FOCUS-dataset) relates to [*amortization*](#glossary:amortization). Amortization Class is used to identify and distinguish between the charges before and after *amortization* is applied. Amortization Class is commonly used to quantify cost savings accurately by excluding charges that can be double-counted when aggregated List Cost or Contracted Cost.
```
gparker-at-sf, line 3: is there verbiage missing here? this last sentence does not read fluidly to me, as if it it is missing some important punctuation or another phrase. 

"Amortization Class is commonly used to quantify cost savings accurately by excluding charges that can be double-counted when aggregated List Cost or Contracted Cost."

Specifically this clause, 'excluding charges that can be double-counted when aggregated List Cost or Contracted Cost'

JGrubb: There are a few of these proofreading issues I will take care of today.

supporting_content/columns/amortizationclass.md
```diff
@@ -0,0 +1,65 @@
...
+| ------------ | ----------------------- | ----------------------------------------------------------------------------------------------------------------------------------------------------- |
+| AWS          | CUR                     | lineItem/lineItemType                                                                                                                                 |
+| Google Cloud | BigQuery Billing Export | ??N/A because no upfront payments??                                                                                                                   |
+| Microsoft    | Cost details            | ChargeType indicates purchases and PricingModel indicates commitment discounts. There is no column that identifies when and how amortization applies. |
+| OCI          | Cost reports            | ???                                                                                                                                                   |
+
+## Example usage scenarios
+
+This example demonstrates a commitment discount that gets 50% off list price when the organization has an 20% discount on all usage.
```
gparker-at-sf, line 18: I would expect this example to include the new column being proposed.

gparker-at-sf (changes requested):

specification/columns/amortizationclass.md
flanakin: We discussed this in TF1 last week. We determined that expanding to other scenarios doesn't make sense because those scenarios require different handling. That would be overloading a single column with multiple purposes and indicating which applies would be difficult. We decided to keep the initial scope of only handling amortization.

specification/columns/amortizationclass.md
flanakin: This seems simple...
```suggestion
Amortization Class indicates how a row in a [*FOCUS dataset*](#glossary:FOCUS-dataset) relates to [*amortization*](#glossary:amortization). Amortization Class is used to identify and distinguish between the charges before and after *amortization* is applied. Amortization Class is commonly used to quantify cost savings accurately by excluding charges that can be double-counted when aggregating List Cost or Contracted Cost.
```

specification/columns/amortizationclass.md
flanakin:
```suggestion
* AmortizationClass MUST be "Amortized Charge" when the row represents a result of *amortization* from a previous charge.
```

supporting_content/columns/amortizationclass.md
```diff
@@ -0,0 +1,65 @@
+# Column: Amortization Class
+
+Objective: We need a simple method to calculate savings that doesn't accidentally double-count commitment discount purchase costs that are also represented in usage rows.
```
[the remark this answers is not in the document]
flanakin: Do you have a proposal? Note this is only in supporting content and will not be part of the spec.

supporting_content/columns/amortizationclass.md
```diff
@@ -0,0 +1,65 @@
...
+This example demonstrates a commitment discount that gets 50% off list price when the organization has an 20% discount on all usage.
+
+| ChargeDescription | ChargePeriodStart |  ListCost | ContractedCost | BilledCost | EffectiveCost |
+| ----------------- | ----------------- | --------: | -------------: | ---------: | ------------: |
+| Reservation       | Jan 1             |   $365.00 |        $365.00 |    $365.00 |         $0.00 |
+| VM usage          | Jan 1             |     $2.00 |          $1.80 |      $0.00 |         $1.00 |
+| VM usage          | Jan 2             |     $2.00 |          $1.80 |      $0.00 |         $1.00 |
+| VM usage          | ...               |       ... |            ... |        ... |           ... |
+| VM usage          | Dec 31            |     $2.00 |          $1.80 |      $0.00 |         $1.00 |
+| Sum of 365 days   | Jan 1-Dec 31      | $1,095.00 |      $1,022.00 |       $365 |          $365 |
```
flanakin, line 27:
```suggestion
| ChargeDescription | AmortizationClass | ChargePeriodStart |  ListCost | ContractedCost | BilledCost | EffectiveCost |
| ----------------- | ----------------- | ----------------- | --------: | -------------: | ---------: | ------------: |
| Reservation       | Principal        | Jan 1             |   $365.00 |        $365.00 |    $365.00 |         $0.00 |
| VM usage          | Amortized Charge | Jan 1             |     $2.00 |          $1.80 |      $0.00 |         $1.00 |
| VM usage          | Amortized Charge | Jan 2             |     $2.00 |          $1.80 |      $0.00 |         $1.00 |
| VM usage          | Amortized Charge | ...               |       ... |            ... |        ... |           ... |
| VM usage          | Amortized Charge | Dec 31            |     $2.00 |          $1.80 |      $0.00 |         $1.00 |
| Sum of 365 days   | (sum)            | Jan 1-Dec 31      | $1,095.00 |      $1,022.00 |       $365 |          $365 |
```

supporting_content/columns/amortizationclass.md
flanakin: Good idea!

gparker-at-sf (approved):

supporting_content/columns/amortizationclass.md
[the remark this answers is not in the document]
gparker-at-sf: If not part of the spec, this doesn't need to be blocking -- approved now.

specification/columns/amortizationclass.md
```diff
@@ -0,0 +1,44 @@
...
+
+Indicates how a row relates to amortization to identify and distinguish between the charges before and after amortization is applied.
+
+## Content Constraints
+
+| Constraint    | Value          |
+| :------------ | :------------- |
+| Column type   | Dimension      |
+| Feature level | Conditional    |
+| Allows nulls  | False          |
```
flanakin, line 31:
```suggestion
| Allows nulls  | True           |
```

CHANGELOG.md
```diff
@@ -265,6 +266,7 @@ This table maps the evolution of the specification, showcasing column introducti
 
 | **Column ID**                | **Revision Introduced** | **Status**        |
 |------------------------------|-------------------------|-------------------|
+| AmortizationClass          | 1.2                     |                                            |
```
flanakin, line 269:
```suggestion
| AmortizationClass            | 1.2                     |                   |
```

flanakin: Fix spacing

specification/columns/amortizationclass.md
flanakin: FYI - Discussed again in TF1 on Mar 18

specification/columns/amortizationclass.md
```diff
@@ -0,0 +1,44 @@
...
+| Allows nulls  | True           |
+| Data type     | String         |
+| Value format  | Allowed values |
+
+Allowed values:
+
+| Value            | Description                                                                                                                       |
+| :--------------- | :-------------------------------------------------------------------------------------------------------------------------------- |
+| Principal        | Original charge that is amortized over time across multiple rows.                                                                 |
+| Amortized Charge | Effective charge after the principal amount is amortized over time based on the associated ChargePeriodStart and ChargePeriodEnd. |
```
cnharris10, line 40: I think it makes sense to approach this in iteration, but I do think we should better specify "Amortized Charge" so it doesn't roadblock us later on.  AFAIK, a post-contract commitment discount is an amortized charge is an example of [straight-line amortization](https://en.wikipedia.org/wiki/Amortization_schedule), or equal payments over the term.

In the case of a pre-paid contract (i.e. Datadog) of funds over a term, I think it's valid to argue that the amortization schedule is reliant on the user's usage activity, so this is another form of amortization.

As we start to mix SaaS datasets, I think we should differentiate between different principals and amortization types with more specific values like (these are bad names):
- Principal --> Post-Contract Principal
- Amortized Charge --> Post-Contract Amortized Charge

so we can then include and differentiate values like (these are bad names):
- Contract Principal (i.e. SaaS Upfront Payment)
- Contract Amortized Charge (i.e. SaaS usage)

aws-letianfeng (changes requested):
Need more clarification about 

> * AmortizationClass MUST be "Amortized Charge" when the row represents a result of *amortization* from a previous charge.

What does "represents a result of" mean here? For example, for EC2 Capacity Block for ML, the charge is paid upfront (assume this is Principal). But for each EC2 instance covered by the Capacity Block for ML, there's an hourly on-demand charge for OS licenses. So the effective cost of the EC2 instances covered by Capacity Block is the sum of amortized cost and the on-demand OS license cost. If we intend to consider this as "amortized charge", the definition needs to be reworded as something like below to clarify the meaning "represents a result of"

> * AmortizationClass MUST be "Amortized Charge" when part or all of the AmortizedCost or EffectiveCost values of the row include *amortization* from a previous charge.

specification/columns/amortizationclass.md
flanakin: I don't quite understand the scenario. We might need to talk thru it. But regardless, if it's about amortization, then there's the original charge with BilledCost that gets amortized and then there are the "target" charges that get the amortized amounts with EffectiveCost. For the purpose of identifying how the rows apply amortization, do we care whether it's pre or post? 🤔 I can't say definitively since I don't completely understand the scenario, but if we're creating this to prevent double counting the original and target charges, then we should be able to use the same values for both.

supporting_content/columns/amortizationclass.md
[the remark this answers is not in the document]
flanakin:
```suggestion
Objective: We need a method to calculate savings that accurately accounts for commitment discount purchase costs without creating duplication in the usage rows.
```

shawnalpay:
@UdamDewaraja and @rileyjenk, please quantify your concerns and/or suggestions on this PR.

rileyjenk (changes requested):
The name of this column is confusing as the name indicates that it classifying amortization and not a helper column to avoid double counting. There types/classes of amortization in the finance/accounting world. 

Also we indicate on the ListCost and Contracted Cost columns that they cannot be aggregated in sum without considering purchase vs usage. I don't challenge the need for the column to aid in prevention of double counting. I would prefer we indicate this columns use to prevent double counting.

flanakin:
> What does "represents a result of" mean here? For example, for EC2 Capacity Block for ML, the charge is paid upfront (assume this is Principal). But for each EC2 instance covered by the Capacity Block for ML, there's an hourly on-demand charge for OS licenses. So the effective cost of the EC2 instances covered by Capacity Block is the sum of amortized cost and the on-demand OS license cost. If we intend to consider this as "amortized charge", the definition needs to be reworded as something like below to clarify the meaning "represents a result of"

@aws-letianfeng Does this upfront charge have BilledCost > 0 and EffectiveCost == 0? Is the cost of this broken down hourly for the lifetime of the commitment with those hourly charges having an EffectiveCost > 0 and BilledCost == 0? If all those are true, then it applies. If not, then it doesn't.

flanakin:
> The name of this column is confusing as the name indicates that it classifying amortization and not a helper column to avoid double counting. There types/classes of amortization in the finance/accounting world.

I added the poll results to the supporting content. See https://docs.google.com/spreadsheets/d/1iRHiLCMBH9RPWU4s1GGFr0TiEYUP4SCpkxiHsO4kWHw/edit?gid=1313687443#gid=1313687443

@rileyjenk I'm not opposed to another name. Runners up were `AmortizationCategory` and `AmortizationEffect` but there were more that didn't get any votes.

> Also we indicate on the ListCost and Contracted Cost columns that they cannot be aggregated in sum without considering purchase vs usage. I don't challenge the need for the column to aid in prevention of double counting. I would prefer we indicate this columns use to prevent double counting.

That might be problematic as there are different, overlapping scenarios where double counting can happen. For instance, credit burndown charges can also be commitment discount purchases, so there would be 2 types of double counting scenarios where it might be included or excluded depending on the scenario. Based on this, we need each distinct double counting scenario to be handled in separate, single-purpose columns. I added details to supporting content.

We could do something like `AmortizationDoubleCounting`. I'm not sure what the values would be. We still need to differentiate the original and effective charges.

Maybe something more explicit like `AmortizationRow` with `Billed Charge` and `Effective Charge`?

specification/columns/amortizationclass.md
flanakin: FYI - Added more context to the supporting content to explain why we can't cover all double counting scenarios in one column. Specifically, commitment discount purchases can be both credit burndown targets and original amortization charges.

aws-letianfeng:
> > What does "represents a result of" mean here? For example, for EC2 Capacity Block for ML, the charge is paid upfront (assume this is Principal). But for each EC2 instance covered by the Capacity Block for ML, there's an hourly on-demand charge for OS licenses. So the effective cost of the EC2 instances covered by Capacity Block is the sum of amortized cost and the on-demand OS license cost. If we intend to consider this as "amortized charge", the definition needs to be reworded as something like below to clarify the meaning "represents a result of"
> 
> @aws-letianfeng Does this upfront charge have BilledCost > 0 and EffectiveCost == 0? Is the cost of this broken down hourly for the lifetime of the commitment with those hourly charges having an EffectiveCost > 0 and BilledCost == 0? If all those are true, then it applies. If not, then it doesn't.


For the upfront charge, billed cost >0, effective cost =0
For the hourly charge, billed cost >0 (if the instance is using a paid OS, the billed cost represents the OS licensing fee), effective cost >0 (OS licensing fee + amortization of upfront Capacity Block for ML payment)

The following wording needs to incorporate the second case.

> * AmortizationClass MUST be "Amortized Charge" when the row represents a result of *amortization* from a previous charge.

Suggested edits

> * AmortizationClass MUST be "Amortized Charge" when the row includes *amortization* from a previous charge.

shawnalpay (changes requested):
This is a placeholder change request on behalf of @aws-letianfeng as described [here](https://github.com/FinOps-Open-Cost-and-Usage-Spec/FOCUS_Spec/pull/747#issuecomment-2749466055).  I will dismiss this request once he requests changes himself.

specification/columns/amortizationclass.md
jpickardgreen: I think it might be good to show more examples if possible so I understand what Chris is saying from the SaaS side and agree with the iteration but how would it work from the public cloud side as an example?

shawnalpay:
@ijurica will add changes to usability constraints to relevant columns like `List Cost` to leverage this column.

We will then put this PR forward for Member review despite not full TF consensus.

shawnalpay (changes requested):
@ijurica will add changes to usability constraints to relevant columns like `List Cost` to leverage this column.

We will then put this PR forward for Member review despite not full TF consensus.

marc-perreaut (approved):

shawnalpay:
@flanakin 
As a reminder:

 * This PR has changes requested per @aws-letianfeng [here](https://github.com/FinOps-Open-Cost-and-Usage-Spec/FOCUS_Spec/pull/747#issuecomment-2749466055).
  * @ijurica has changes to make per [this comment](https://github.com/FinOps-Open-Cost-and-Usage-Spec/FOCUS_Spec/pull/747#pullrequestreview-2714409562).

This PR is scheduled for approval tomorrow, but we cannot proceed with approval with the PR in this state.

shawnalpay:
@rileyjenk Please re-review per @flanakin's request from five days ago.  If you do not provide an explicit approval or change request, your stale review will be dismissed before the Members meeting tomorrow.

specification/columns/amortizationclass.md
```diff
@@ -0,0 +1,44 @@
+# Amortization Class
+
+Amortization Class indicates how a row in a [*FOCUS dataset*](#glossary:FOCUS-dataset) relates to [*amortization*](#glossary:amortization). Amortization Class is used to identify and distinguish between the charges before and after *amortization* is applied. Amortization Class is commonly used to quantify cost savings accurately by excluding charges that can be double-counted when aggregating List Cost or Contracted Cost.
+
+The AmortizationClass column adheres to the following requirements:
+
+* AmortizationClass column MUST be present in a [*FOCUS dataset*](#glossary:FOCUS-dataset) when the provider supports *amortization*.
+* AmortizationClass MUST be of type String.
+* AmortizationClass MUST be "Principal" when the row represents an original charge that is amortized across multiple other rows.
+* AmortizationClass MUST be "Amortized Charge" when the row represents a result of *amortization* from a previous charge.
```
aws-letianfeng, line 10: For AWS, EC2 Capacity Block for ML has the following behavior --

For the upfront charge, billed cost >0, effective cost =0
For the hourly charge, billed cost >0 (if the instance is using a paid OS, the billed cost represents the OS licensing fee), effective cost >0 (OS licensing fee + amortization of upfront Capacity Block for ML payment)

Need to clarify what does "represents a result of" in the proposal.

`* AmortizationClass MUST be "Amortized Charge" when the row represents a result of *amortization* from a previous charge.
`
Update to
`* AmortizationClass MUST be "Amortized Charge" when the row includes *amortization* from a previous charge.
`

aws-letianfeng (changes requested):
For AWS, EC2 Capacity Block for ML has the following behavior --

For the upfront charge, billed cost >0, effective cost =0
For the hourly charge, billed cost >0 (if the instance is using a paid OS, the billed cost represents the OS licensing fee), effective cost >0 (OS licensing fee + amortization of upfront Capacity Block for ML payment)

Need to clarify what does "represents a result of" in the proposal.

* AmortizationClass MUST be "Amortized Charge" when the row represents a result of *amortization* from a previous charge. 
Update to
* AmortizationClass MUST be "Amortized Charge" when the row includes *amortization* from a previous charge. 


Flagging the specific line for changes.

UdamDewaraja (changes requested):
Agree with @rileyjenk @GrinsteadJ and a few others that have brought up concerns. This naming is problematic for SaaS. We can't say this works for Cloud and not SaaS. FOCUS is agnostic to providers. 

I support solving the issue of double counting. I don't support this naming or supported values. 

Just the statement "Amortization Class is commonly used to quantify cost savings accurately by excluding charges that can be double-counted when aggregating List Cost or Contracted Cost." seems way off to me. Something named Amortization Class isn't going to be what people think of when they think about avoiding double counting.

specification/columns/amortizationclass.md
flanakin: @letian-wang-92 FYI - you can make a suggestion using the "+/-" file icon in the toolbar when leaving a comment inline.
```suggestion
* AmortizationClass MUST be "Amortized Charge" when the row includes *amortization* from a previous charge.
```

specification/columns/amortizationclass.md
flanakin: As discussed in TF1, this suggestion wouldn't work because it's merging two concepts into one column. One is the type of purchase and the other is how amortization is applied. This would need to be split out into separate columns:

| Scenario | PurchaseType | AmortizationClass |
|-----------|-----------|-----------|
| Commitment discount purchase | Commitment Discount | Principal |
| Commitment discount usage | Commitment Discount | Amortized Charge |
| Amortized Marketplace purchase | Marketplace | Principal |
| Amortized Marketplace usage | Marketplace | Amortized Charge |
| Amortized credit purchase | Credit | Principal |
| Amortized credit usage | Credit | Amortized Charge |

The only reason we need additional values is if there's another scenario where a row is impacted by amortization in a different way than the original charge or the allocated breakdown.

flanakin: More importantly for this comment, I researched amortization and this credit burndown scenario doesn't seem to apply. From what I'm seeing on accounting websites, amortization always has a structured, consistent payment schedule called an "amortization schedule". This is usually monthly, daily, or hourly, like for commitment discounts. While we apply amortization differently than a loan, the amortization schedule is effectively the same where the original purchase ("principal" in accounting terms) is broken down over a structured, consistent usage schedule. This isn't applicable in the credit burndown scenario. While you could use BilledCost for the credit payment and EffectiveCost for the credit usage, it is not amortization and therefore would not have an AmortizationClass value.

I think we need to provide more context on amortization and the use of BilledCost and EffectiveCost in non-amortized scenarios like the credit burndown scenario. We should look at that for 1.3.

specification/columns/amortizationclass.md
```diff
@@ -0,0 +1,44 @@
+# Amortization Class
+
+Amortization Class indicates how a row in a [*FOCUS dataset*](#glossary:FOCUS-dataset) relates to [*amortization*](#glossary:amortization). Amortization Class is used to identify and distinguish between the charges before and after *amortization* is applied. Amortization Class is commonly used to quantify cost savings accurately by excluding charges that can be double-counted when aggregating List Cost or Contracted Cost.
```
flanakin, line 3:
```suggestion
Amortization Class indicates how a row in a [*FOCUS dataset*](#glossary:FOCUS-dataset) relates to [*amortization*](#glossary:amortization). Amortization Class is used to identify and distinguish between the charges before and after *amortization* is applied. Amortization Class is commonly used to quantify cost savings accurately by excluding the original principal charges that are amortized over time, which could otherwise be double-counted when aggregating List Cost or Contracted Cost.
```

flanakin:
> This naming is problematic for SaaS. We can't say this works for Cloud and not SaaS. FOCUS is agnostic to providers.

@UdamDewaraja This column only applied to amortization and works in all amortization scenarios, whether they're cloud or SaaS or other. It doesn't apply to non-amortization scenarios like credit burndown which don't have an amortization schedule. 

We voted on the name. Results are in the supporting content. If anyone has a better name, please suggest one. Blocking without a suggestion doesn't help us close things.

> Just the statement "Amortization Class is commonly used to quantify cost savings accurately by excluding charges that can be double-counted when aggregating List Cost or Contracted Cost." seems way off to me. Something named Amortization Class isn't going to be what people think of when they think about avoiding double counting.

This is a helpful callout. I updated it to be more explicit about the original principal charges that are amortized over time. 

Again, this is not credit burndown or the other scenarios that may have double-counting. The column is not for generic double-counting. It's for amortization-related double-counting only. Those other scenarios are overlapping and cannot be handled in the same column, as we've discussed a few times.

UdamDewaraja:
> Blocking without a suggestion doesn't help us close things.

Apologize, I do need to provide a suggestion for what would work. I think the concerns have been brought up by 3-4 people but we need to spend some time working on this together.

> @UdamDewaraja / @rileyjenk Please take a fresh look at this without factoring in credit burndown, promise burndown, and duplicate marketplace/SaaS datasets that don't apply here. This column is not attempting to solve those overlapping scenarios.

Sorry, but I don't see how only handling some double counting scenarios and not others for List/Contracted is a robust solution especially when:
- the solution put forth is not going to be able to be adopted for the other half
- Is not a problem that HAVE TO solve in 1.2 (we have guidance on excluding purchases or the covered charges in those columns)
- I don't know if the term Amortization even applies in the case of the current cloud - it seems only a subset of even clouds support up-front payments. Others are doing periodic charges and burning it down - is that even classified as amortization? Is a spend commitment (e.g., savings plan) a 'tangible asset that's been purchased'?

rileyjenk (changes requested):
I don't think we are ready to put this in the spec. My concerns are as follows:

- The column name is misleading. 
- There are more scenarios that have amortization impact than just commitment based dicsount purchases, especially in the SaaS world. 

The name of this column is particularly problematic. There are types/methods/approaches to amortization that are a part of the finops world and very much a thing in finance.  The column name at face value seems that it would be indicating an amortization approach and not  what this is. 

Amortization approaches: https://tax.thomsonreuters.com/blog/amortization-in-accounting-101/

Ultimately I don't think we have this solid enough to put in the spec.

shawnalpay:
We brought this PR forward for review and approval in the Mar 27 Members meeting today.  Multiple concerns were raised, summarized in @UdamDewaraja's comments [here](https://github.com/FinOps-Open-Cost-and-Usage-Spec/FOCUS_Spec/pull/747#issuecomment-2758894652) and [here](https://github.com/FinOps-Open-Cost-and-Usage-Spec/FOCUS_Spec/pull/747#issuecomment-2758910464), and @rileyjenk's comment [here](https://github.com/FinOps-Open-Cost-and-Usage-Spec/FOCUS_Spec/pull/747#pullrequestreview-2723120864).

@flanakin, please respond to their concerns here asynchronously, and in the Mar 31 Maintainers meeting, we will discuss whether this PR has a path forward in 1.2.

rileyjenk:
Let me add further clarification here. This columns attempt is to fix a double count scenario for ListCost and ContractedCost as identified in WI [647](https://github.com/FinOps-Open-Cost-and-Usage-Spec/FOCUS_Spec/issues/647). Later WI https://github.com/FinOps-Open-Cost-and-Usage-Spec/FOCUS_Spec/issues/406 was added.

In the case of a purchase that is to be amortized in the traditional sense there are classifications/methods for amortization that are used in the finance world.  The method of amortization that is closest to what we are talking about is called "Units of Production." In this case the value of an asset is depreciated over time by measure the usage of a thing. https://www.wallstreetprep.com/knowledge/units-of-production-method/

Here is the rub though....we aren't just talking about the amortization of purchases of fixed assets. Double counting happens in more scenarios. 

For example double counting happens with token/credit/point purchase, usage, and overage handing. It happens on initial purchases just like RIs/Savings plans.  Also because overage can be handled with the purchase of more token/credit/points legitimate ListCost and ContractedCost that occur for purchase due to exceeding a committed purchase balance. 

Very simple example: (https://docs.google.com/spreadsheets/d/1kQTDK3Sk9BnNcn6Ovyaa37T1aMaXfHaDahsuk1Notn4/edit?usp=sharing , see tab ExampleOfPostOveragePurchases)

|                                                                                            |             |                  |                    |             |                                  |                 |                 |                   |                |
|--------------------------------------------------------------------------------------------|-------------|------------------|--------------------|-------------|----------------------------------|-----------------|-----------------|-------------------|----------------|
|                                                                                            |             |                  |                    |             |                                  |                 |                 |                   |                |
|                                                                                            |             |                  |                    |             |                                  |                 |                 |                   |                |
| Scenario                                                                                   | BilledCost  | BillingPeriodEnd | BillingPeriodStart | ChargeClass | ChargeDescription                | ChargeFrequency | ChargePeriodEnd | ChargePeriodStart | ContractedCost |
| Initial Purchase Token Allocation 200 Credits at $1, list price is $2                      | $200,000.00 | 2025-04-30       | 2025-04-01         | Purchase    | Token Purchase For Period 1      | One-Time        | 2025-04-30      | 2025-04-01        | $200.00        |
|                                                                                            |             |                  |                    |             |                                  |                 |                 |                   |                |
| Day 1 Token Usage, 100 total                                                               | $0.00       | 2025-04-30       | 2025-04-01         | Usage       | Token Consumption - Q Widget SKU | Usage-Based     | 2025-04-02      | 2025-04-01        | $100.00        |
| Day 1 Token Usage, 100 total                                                               | $0.00       | 2025-04-30       | 2025-04-01         | Usage       | Token Consumption - Q Widget SKU | Usage-Based     | 2025-04-03      | 2025-04-02        | $100.00        |
| Day 1 Token Usage, 100 total, allocations depleted, per unit cost falls back to list price | $0.00       | 2025-04-30       | 2025-04-01         | Usage       | Token Consumption - Q Widget SKU | Usage-Based     | 2025-04-04      | 2025-04-03        | $200.00        |
|                                                                                            |             |                  |                    |             |                                  |                 |                 |                   |                |

specification/columns/amortizationclass.md
UdamDewaraja: The last two places I worked at both had accounting people who were debating if things like cloud costs, including reservations, could be capitalized. I see the same arguments online. I don't think we're going to get an easy/clear answer on if we should consider this amortization.

For commitments that are charged hourly or daily even though they are for a year or longer (which I believe is the case for all 3 big CSPs), how does this model the charges? Do you have any sample data for the common commitment scenarios? Does amortization apply to those no up-front commitments?

flanakin:
We discussed this in the maintainer call and decided to push this to 1.3

shawnalpay:
Discussed in Jul 28 Maintainers.  Agreed to close and revisit if/when we pick this work up again in a future release.

PR #798, opened 2025-02-25
WI #557: Normative Requirements Refinement - Part 2

ijurica:
Updating normative requirements for a subset of FOCUS columns in line with the [FOCUS Columns Normative Guidelines](https://github.com/FinOps-Open-Cost-and-Usage-Spec/FOCUS_Spec/blob/557-normative-reqs-refinement-part-1/guidelines/normative_requirements_guidelines.md). 

_**Note**: The remaining 33 columns are addressed in a separate PR (#787)_

## FOCUS 1.2 Release - Column Normative Requirements Guidelines and Refinement

As part of the FOCUS 1.2 release (Work Item #557), our goal was to establish the **Column Normative Requirements Guidelines** and update existing **Column Normative Requirements** to align with these Guidelines.

### Review Request

For those reviewing the **Normative Requirements PRs (i.e., (PR #787 and PR #798))**, please:

* **Review the Guidelines** – Suggest improvements, flag inconsistencies, and clarify any ambiguities.
* **Review the updated Column Normative Requirements** – Ensure they align with the Guidelines and suggest any necessary refinements.
* **Identify areas for additional guidelines** – If, during the column review, you identify aspects that should be explicitly highlighted as guidelines, please propose them for inclusion.
* **Address or flag bugs, oversights, or ambiguities** – As part of the Column Normative Requirements refinement process, we aim to identify and resolve any bugs, oversights, or ambiguities. While some have already been addressed or flagged in these two PRs, this review aims to capture any remaining issues. These should be either:  
  * **Addressed** in the new version of the requirements for the respective columns (by adding GitHub suggestions for updates).  
  * **Flagged** and **Added** to the **Discussion Topics** list for further review (by leaving GitHub comments).  
  * *Note: We acknowledge that some issues may still be missed in this iteration, and we plan to continue this review process in future releases.*  
* **Flag questionable changes** – Please note that during the refinement process, some of the original normative requirements may have been unintentionally altered. If you spot any such discrepancies or questionable changes, flag them (by adding a comment in GitHub) so we can address them—either by correcting any inadvertent errors or confirming that the updated version is preferred.

### Discussion Topics

We will prioritize and assess the complexity of each [Discussion Topic](https://docs.google.com/spreadsheets/d/1rD9wKQ8A18GdklXYSXue8tm7tu23eAAJ-HaAbCWJ_9w/edit?gid=1947065303#gid=1947065303):

* Some will be addressed within **FOCUS 1.2**.
[...]

PR #787, opened 2025-02-25
WI #557: Normative Requirements Refinement - Part 1

ijurica:
Updating normative requirements for a subset of FOCUS columns in line with the attached [FOCUS Columns Normative Guidelines](https://github.com/FinOps-Open-Cost-and-Usage-Spec/FOCUS_Spec/blob/557-normative-reqs-refinement-part-1/guidelines/normative_requirements_guidelines.md). The remaining columns are addressed in a separate PR (#798)

## FOCUS 1.2 Release - Column Normative Requirements Guidelines and Refinement

As part of the FOCUS 1.2 release (Work Item #557), our goal was to establish the **Column Normative Requirements Guidelines** and update existing **Column Normative Requirements** to align with these Guidelines.

### Review Request

For those reviewing the **Normative Requirements PRs (i.e., (PR #787 and PR #798))**, please:

* **Review the Guidelines** – Suggest improvements, flag inconsistencies, and clarify any ambiguities.
* **Review the updated Column Normative Requirements** – Ensure they align with the Guidelines and suggest any necessary refinements.
* **Identify areas for additional guidelines** – If, during the column review, you identify aspects that should be explicitly highlighted as guidelines, please propose them for inclusion.
* **Address or flag bugs, oversights, or ambiguities** – As part of the Column Normative Requirements refinement process, we aim to identify and resolve any bugs, oversights, or ambiguities. While some have already been addressed or flagged in these two PRs, this review aims to capture any remaining issues. These should be either:  
  * **Addressed** in the new version of the requirements for the respective columns (by adding GitHub suggestions for updates).  
  * **Flagged** and **Added** to the **Discussion Topics** list for further review (by leaving GitHub comments).  
  * *Note: We acknowledge that some issues may still be missed in this iteration, and we plan to continue this review process in future releases.*  
* **Flag questionable changes** – Please note that during the refinement process, some of the original normative requirements may have been unintentionally altered. If you spot any such discrepancies or questionable changes, flag them (by adding a comment in GitHub) so we can address them—either by correcting any inadvertent errors or confirming that the updated version is preferred.

### Discussion Topics

We will prioritize and assess the complexity of each [Discussion Topic](https://docs.google.com/spreadsheets/d/1rD9wKQ8A18GdklXYSXue8tm7tu23eAAJ-HaAbCWJ_9w/edit?gid=1947065303#gid=1947065303):

* Some will be addressed within **FOCUS 1.2**.
[...]

Issue #616, opened 2024-10-22
[Work_Item] Document and add initial support for SaaS-centric concepts


shawnalpay:
## 1. **Problem Statement** *
> Describe the problem, issue, use case, or opportunity that this work item addresses.
> Include practitioner quotes illustrating real examples a) of questions being asked by practitioners and b) value unlocked by answering these questions, if available.
> - What is the problem?: Explain the context and why it needs resolution.
> - Impact: Describe how the problem affects users, systems, or the project.

The FOCUS specification, so far, has chiefly been engineered to support cost and usage data as it is generated by cloud service providers (CSP) such as AWS, Azure, GCP, and OCI.  There is a rich landscape of software-as-a-service (SaaS) providers in the FinOps landscape whose billing models differ in various ways from CSPs -- and as currently constructed, FOCUS does not yet provide an onramp for those SaaS providers (unless they presumably create a host of custom columns to support their concepts).

## **Use Case(s)**
❗  Follow instructions on [how to craft a use case](https://docs.google.com/document/d/1oMV78K7cGmQ5cjcAv_aZBe-sQzV9zatvJVTEsE7aL7M/edit?tab=t.0) -- need both names and descriptions

## 2. **Objective** *
> State the objective of this work item. What outcome is expected?
> * Success Criteria: Define how success will be measured (e.g. metrics and KPIs).

Analyze a set of major SaaS providers and determine the common SaaS-centric concepts that could be implemented in a future version of the spec.  Time permitting, we would then attempt to augment the specification to establish an MVP of those concepts within that release, with further support added in future releases.

## 3. **Supporting Documentation** *
> Include links to supporting documents such as:
> - Data Examples: [Link to data or relevant files; DO NOT share proprietary information]
> - Related Use Cases or Discussion Documents: [Link to discussion]
> - PRs or Other References: [Link to relevant references]

Related issue: #337 

Documentation TBD

## 4. **Proposed Solution / Approach**
> Outline any proposed solutions, approaches, or potential paths forward. Do not submit detailed solutions; please keep suggestions high-level.
> 
> - Initial Ideas: Describe potential solution paths, tools, or technologies.
> - Considerations: Include any constraints, dependencies, or risks.
> - Feasibility: Include any information that helps quantify feasibility, such as perceived level of effort to augment the spec, or existing fields in current data generator exports.
[...]

PR #799, opened 2025-02-26
WI #616: Addition of `Pricing Currency` and `Pricing Currency to Billing Currency` columns

shawnalpay:
This pull request creates two new columns for addition in the FOCUS specification:

 * `PricingCurrency`: virtual currency (e.g. tokens, credits) for providers with virtual currency usage charges, and fiat currency (e.g. USD, EUR) otherwise
 * `PricingCurrencyToBillingCurrency`: the conversion factor from Pricing Currency to Billing Currency

The normative requirements presented here follow the format as proposed in WI #557 and PRs #787 and #798.

This PR also revises an existing attribute:

* `CurrencyFormat`: rename from `CurrencyCodeFormat` and expanding definition to include both fiat currencies and virtual currencies

Revised the following columns:

- `BillingCurrency`: point to `CurrencyFormat` instead of `CurrencyCodeFormat`
- `List Cost`: add references to `PricingCurrencyToBillingCurrency`
- `List Unit Price`: add references to `PricingCurrencyToBillingCurrency`; change denomination to `PricingCurrency`
- `Contracted Cost`: add references to `PricingCurrencyToBillingCurrency`
- `Contracted Unit Price`: add references to `PricingCurrencyToBillingCurrency`; change denomination to `PricingCurrency`
- `Pricing Quantity`: add references to `PricingCurrencyToBillingCurrency`

Also added an entry for `Pricing Currency` in the glossary.


supporting_content/columns/pricingdenomination.md
```diff
@@ -0,0 +1,21 @@
...
+
+## Example provider mappings
+
+Current column mappings found in available data sets:
+
+| Provider  | Data set                     | Column                   |
+|:----------|:-----------------------------|:-------------------------|
+| AWS       | Cost and Usage Report        | n/a |
+| GCP       | BigQuery Billing Export      | n/a |
+| Microsoft | Cost Details                 | n/a |
```
ijurica, line 11: **Reminder:** 

* In Microsoft MCA Cost details there are columns that can be directly mapped to PricingDenomination and PricingDenominationToBillingCurrency.
* A few MCA Cost Details columns and sample values relevant to this topic:
  * billingCurrency (EUR)
  * pricingCurrency (USD)
  * exchangeRatePricingToBilling
  * PayGPrice (denominated in pricingCurrency, i.e., USD)

* Corresponding Microsoft FOCUS dataset and values:
  * BillingCurrency (EUR)
  * x_PricingCurrency (USD)
  * x_BillingExchangeRate
  * x_BillingExchangeRateDate

specification/columns/pricingdenomination.md
```diff
@@ -0,0 +1,47 @@
...
+The unit of measure may be one of the following types: 
+
+ * Currency (e.g. USD, EUR).
+ * Non-currency consumable unit (e.g. tokens, credits).
+
+The PricingDenomination column adheres to the following requirements:
+
+* PricingDenomination presence in a [*FOCUS dataset*](#glossary:FOCUS-dataset) is defined as follows:
+  * PricingDenomination MUST be present in a FOCUS dataset when the provider presents prices in a non-currency consumable unit of measure (e.g. credits, tokens).
+  * PricingDenomination MUST be present in a FOCUS dataset when the provider presents prices and bills in different currencies (e.g. prices are presented in USD and billed in EUR).
```
ijurica, line 14: Compute vs. Storage Tradeoff?

### Option A (described in this PR)

* PricingDenomination and PricingDenominationToBillingCurrency  as described in this PR.
* In this case, ListUnitPrice and ContractedUnitPrice are expressed in PricingDenomination when it's present and not null, and in BillingCurrency when PricingDenomination is either not present or is null. Is this correct?

### Option B

* ListUnitPrice and ContractedUnitPrice expressed in PricingDenomination.
* PricingDenomination present as long as the provider supports the unit pricing concept, and equal to BillingCurrency if the provider doesn’t support either of these two concepts.
* PricingDenominationToBillingCurrency present as long as the provider supports the unit pricing concept, and set to 1 if the provider doesn’t support either of these two concepts.

I prefer Option B.

specification/columns/pricingdenomination.md
```diff
@@ -0,0 +1,47 @@
...
+  * PricingDenomination MAY be present in a FOCUS dataset in all other cases.
+* PricingDenomination MUST be of type String.
+* PricingDenomination MUST conform to [Pricing Denomination Format](#pricingdenominationformat) requirements.
+* PricingDenomination nullability is defined as follows:
+  * PricingDenomination MUST NOT be null when the provider presents prices in a non-currency consumable unit of measure (e.g. credits, tokens).
+  * PricingDenomination MUST NOT be null when the provider presents prices and bills in different currencies (e.g. prices are presented in USD and billed in EUR).
+  * PricingDenomination MAY be null in all other cases.
```
shawnalpay, line 21:
```suggestion
* PricingDenomination MUST NOT be null.
```

specification/attributes/pricing_denomination_format.md
```diff
@@ -0,0 +1,30 @@
...
+
+## Description
+
+Formatting for a combination of currency and non-currency columns appearing in a [*FOCUS dataset*](#glossary:FOCUS-dataset).
+
+## Requirements
+
+* Pricing Denomination columns MUST conform to [String Handling](#stringhandling) requirements when [ChargeCategory] is not "Purchase" and the provider presents prices in a non-currency consumable unit of measure (e.g. credits, tokens).
+* Pricing Denomination columns MUST conform to [Currency Code Format](#currencycodeformat) requirements in all other cases.
+
```
shawnalpay, line 23: Per @rileyjenk: be specific about the full writeout of the non-currency consumable unit (e.g. use Bitcoin, don't use BTC).

specification/attributes/pricing_denomination_format.md
```diff
@@ -0,0 +1,30 @@
+# Pricing Denomination Format
```
shawnalpay, line 1: Per @udam-f2: need a name that is less specific to the single column.

specification/attributes/pricing_denomination_format.md
```diff
@@ -0,0 +1,30 @@
...
+
+Columns that contain some combination of currencies (e.g. USD, EUR) and non-currency consumable units of measure (e.g. credits, tokens) in cost data, following a consistent format reduce friction for FinOps practitioners who consume the data for analysis, reporting, and other use cases.
```
shawnalpay, line 3: Per @gparker-at-sf: takes issue with the term `non-currency` to frame credits, tokens, and Riot Points.  
Per @flanakin: perhaps rope all of this into an augmentation of the Currency attribute.

shawnalpay:
I have now applied the following changes, per discussion in today's TF3:

- Use the terms `fiat currency` and `virtual currency` to represent USD/EUR vs tokens/credits.
- Rename `Currency Code Format` to `Currency Format` and expand its definition to allow for virtual currencies.
- Revise `Billing Currency` to leverage that rename.
- Eliminate `Pricing Denomination Format` (no longer necessary, now that we have revised `Currency Format`.
- Rename `Pricing Denomination to Pricing Currency`.
- Rename `Pricing Denomination to Billing Currency` to `Pricing Currency to Billing Currency`.

specification/attributes/pricing_denomination_format.md
shawnalpay: Eliminated this attribute in favor of expanding definition of `Currency Code Format`.  Resolving.

specification/attributes/pricing_denomination_format.md
shawnalpay: Applied both points in most recent version.  Closing.

specification/columns/pricingdenomination.md
shawnalpay: I renamed this column and made substantial changes per our discussion today in TF3.  @ijurica if this is still applicable, could you apply a new comment?  Resolving.

amitkinha:
LGTM

specification/columns/pricingcurrencytobillingcurrency.md
```diff
@@ -0,0 +1,50 @@
+# Pricing Currency to Billing Currency
```
flanakin, line 1: Can we simplify this to `BillingExchangeRate` or `BillingConversionRate`? The directionality should be built into the description/requirements and shouldn't need to be explicitly called out in the name. It makes the column name too long.

specification/columns/pricingcurrencytobillingcurrency.md
flanakin, line 1: Don't we need the date of the exchange/conversion rate?

supporting_content/columns/pricingcurrency.md
```diff
@@ -0,0 +1,21 @@
...
+
+## Example provider mappings
+
+Current column mappings found in available data sets:
+
+| Provider  | Data set                     | Column                   |
+|:----------|:-----------------------------|:-------------------------|
+| AWS       | Cost and Usage Report        | n/a |
+| GCP       | BigQuery Billing Export      | n/a |
+| Microsoft | Cost Details                 | n/a |
```
flanakin, line 11:
```suggestion
| Microsoft | Cost Details                 | PricingCurrency |
```

specification/columns/contractedcost.md
```diff
@@ -1,12 +1,12 @@
 # Contracted Cost
 
-Contracted Cost represents the cost calculated by multiplying [*contracted unit price*](#glossary:contracted-unit-price) and the corresponding [Pricing Quantity](#pricingquantity). Contracted Cost is denominated in the [Billing Currency](#billingcurrency) and is commonly used for calculating savings based on negotiation activities, by comparing it with [List Cost](#listcost). If [*negotiated discounts*](#glossary:negotiated-discount) are not applicable, the Contracted Cost defaults to the List Cost.
+Contracted Cost represents the cost calculated by multiplying [*contracted unit price*](#glossary:contracted-unit-price) with the corresponding [Pricing Quantity](#pricingquantity) and the conversion factor of [Pricing Currency to Billing Currency](#pricingcurrencytobillingcurrency). Contracted Cost is denominated in the [Billing Currency](#billingcurrency) and is commonly used for calculating savings based on negotiation activities, by comparing it with [List Cost](#listcost). If [*negotiated discounts*](#glossary:negotiated-discount) are not applicable, the Contracted Cost defaults to the List Cost.
```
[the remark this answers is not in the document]
shawnalpay: The cost columns shall definitely remain in Billing Currency; for example, Contracted Cost still says:

> This column MUST be of type Decimal, MUST conform to Numeric Format requirements, and be denominated in the BillingCurrency.

For providers that don't have a different Pricing Currency, this conversion factor is 1, so the definition doesn't change.  Does that assuage your concern?  We can discuss further in TF3 on Fri.

specification/columns/pricingcurrency.md
```diff
@@ -0,0 +1,42 @@
+# Pricing Currency
+
+[*Pricing Currency*](#glossary:pricing-currency) is a value that represents the unit of measure in which a charge for [*resources*](#glossary:resource) or [*services*](#glossary:service) was priced. Pricing Currency is commonly used in scenarios where costs need to be grouped or aggregated by some combination of the following currency types:
```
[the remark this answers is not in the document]
shawnalpay: Okay... how about "denomination"?  This is the term we use elsewhere, such as (emphasis mine):

> This column MUST be of type Decimal, MUST conform to [Numeric Format](https://github.com/FinOps-Open-Cost-and-Usage-Spec/FOCUS_Spec/blob/69258a1b6c7a1c58db247572ecc9cba6e555ab10/specification/columns/contractedcost.md#numericformat) requirements, and be **denominated** in the BillingCurrency.

```suggestion
[*Pricing Currency*](#glossary:pricing-currency) is a value that represents the denomination in which a charge for [*resources*](#glossary:resource) or [*services*](#glossary:service) was priced. Pricing Currency is commonly used in scenarios where costs need to be grouped or aggregated by some combination of the following currency types:
```

specification/columns/pricingcurrency.md
```diff
@@ -0,0 +1,42 @@
...
+
+* Fiat currency (e.g. USD, EUR).
+* Virtual currency (e.g. tokens, credits).
```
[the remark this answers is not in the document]
shawnalpay: Strictly speaking, you're right that describing it in the attribute is sufficient.  The idea was to put it here to reduce confusion.  I could sustain striking it if you think the duplication hurts more than helps.

specification/columns/pricingcurrencytobillingcurrency.md
shawnalpay: A few points:

- Definitely don't disagree that it's a long name.  But if we don't have both `Pricing` and `Billing` in there, could it lead to confusion...?  I like names to self-document.
- Given that: I also recognize that "X to Y" doesn't necessarily self-document enough.  The most precise name would be `PricingCurrencyToBillingCurrencyExchangeRate`.  Really rolls off the tongue.
- I'm open to `ExchangeRate` (after all, it's a conversion from one currency to another), but if we use the term Conversion, I'd want `ConversionFactor`.

Given all that, I'm open to `BillingExchangeRate`.  Let's discuss in TF3 and see what the group thinks.

specification/columns/pricingcurrencytobillingcurrency.md
shawnalpay: As a practitioner, do we _want_ the date of the exchange rate?  Well, for invoice reconciliation, yes, it would be nice to check the work. 

Do we _need_ it?   Strictly for the purposes of converting from one to the other, no.  I could certainly entertain adding it in the next release, but I didn't hear consensus in the group on including it in this release.

@UdamDewaraja would appreciate your thoughts on this one.

specification/attributes/currency_format.md
```diff
@@ -1,24 +1,30 @@
-# Currency Code Format
+# Currency Format
 
 Columns that contain currency information in cost data following a consistent format reduce friction for FinOps practitioners who consume the data for analysis, reporting, and other use cases.
 
+A currency may be one of the following currency types:
+
+* Fiat currency (e.g. USD, EUR).
```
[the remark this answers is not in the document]
shawnalpay: I'm open to that.  Let's discuss in TF3.

supporting_content/columns/pricingdenomination.md
shawnalpay: @flanakin made a suggestion just to put `PricingCurrency`, and I sustained that suggestion because it's coming from the horse's mouth. 🐴  @ijurica lmk if you feel that's not good enough, per your comments above!

amitkinha:
Approved

amitkinha (approved):

specification/columns/pricingcurrencytobillingcurrency.md
ijurica: I think it would be useful to indicate in the name itself that this is related to **PricingToBilling** conversion.

specification/columns/listunitprice.md
```diff
@@ -1,13 +1,13 @@
 # List Unit Price
 
-The List Unit Price represents the suggested provider-published unit price for a single [Pricing Unit](#pricingunit) of the associated SKU, exclusive of any discounts. This price is denominated in the [Billing Currency](#billingcurrency). The List Unit Price is commonly used for calculating savings based on various rate optimization activities.
+The List Unit Price represents the suggested provider-published unit price for a single [Pricing Unit](#pricingunit) of the associated SKU, exclusive of any discounts. This price is denominated in the [Pricing Currency](#pricingcurrency). The List Unit Price is commonly used for calculating savings based on various rate optimization activities.
```
ijurica, line 3: **Current version of the PricingCurrency presence normative requirement:**

> * PricingCurrency presence in a [*FOCUS dataset*](#glossary:FOCUS-dataset) is defined as follows:
>    * PricingCurrency MUST be present in a FOCUS dataset when the provider presents prices in a virtual currency (e.g. credits, tokens).
>   * PricingCurrency MUST be present in a FOCUS dataset when the provider presents prices and bills in different fiat currencies (e.g. priced in USD and billed in EUR).
>   * PricingCurrency MAY be present in a FOCUS dataset in all other cases.

Given the current PricingCurrency and PricingCurrencyToBillingCurrency presence conditions, the following statement is not entirely correct:

> _... This price is denominated in the [Pricing Currency](#pricingcurrency). ..._

In some cases (datasets), the unit price would be denominated in PricingCurrency, while in others, it would be denominated in BillingCurrency, as PricingCurrency may not even be present.

specification/columns/pricingdenomination.md
ijurica: it is still applicable so I've created a new comment. resolving this conversation.

supporting_content/columns/pricingdenomination.md
ijurica: that's it! 
resolving this conversation.

specification/columns/listunitprice.md
```diff
@@ -1,13 +1,13 @@
...
 
 The ListUnitPrice column adheres to the following requirements:
 
 * The ListUnitPrice column MUST be present in a [*FOCUS dataset*](#glossary:FOCUS-dataset) when the provider publishes unit prices exclusive of discounts.
 * This column MUST be a Decimal within the range of non-negative decimal values, MUST conform to [Numeric Format](#numericformat) requirements, and be denominated in the BillingCurrency.
```
ijurica, line 8:
```suggestion
* This column MUST be a Decimal within the range of non-negative decimal values, MUST conform to [Numeric Format](#numericformat) requirements, and be denominated in the PricingCurrency.
```

or a more accurate alternative (see [this link](url)) : 

> ... be denominated in PricingCurrency when PricingCurrency is present, and in BillingCurrency when PricingCurrency is not present.

specification/attributes/currency_format.md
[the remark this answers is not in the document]
shawnalpay: Look into ISO defines currencies.  Government-issued and the one that has bitcoin.  Use whatever they use.

ijurica:
@shawnalpay, bringing this old issue to your attention, as it deals with the same topic - issue #318 . When and if we merge this PR, we can close this issue as well.
BTW, in that earlier discussion, we introduced an almost identical suggestion - two new columns: PricingCurrency and ExchangeRate. 😄

shawnalpay:
> @shawnalpay, bringing this old issue to your attention, as it deals with the same topic - issue #318 . When and if we merge this PR, we can close this issue as well. BTW, in that earlier discussion, we introduced an almost identical suggestion - two new columns: PricingCurrency and ExchangeRate. 😄

Interesting!  I will link this PR to that old issue, and if we merge this PR with the current columns, then we'll close that one too.  :)

specification/attributes/currency_format.md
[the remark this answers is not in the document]
shawnalpay: Here are some findings, per the action item I took in TF3 to look into this more.

[ISO 4217](https://en.wikipedia.org/wiki/ISO_4217) defines three-digit codes for all currencies.  It was written before the rise of cryptocurrency, but the Wiki page uses the term `national currency`, and I think that's more self-documenting than `fiat currency`.

[ISO 24165](https://www.iso.org/standard/80601.html) defines "digital token identifiers", or DTIs.  These are controlled by the [Digital Token Identifier Foundation](https://anna-web.org/digital-assets-1/).  This is the list that maps Bitcoin to BTC and Etherum to ETH.  A list of codes is [here](https://anna-web.org/digital-assets-1/).  All of that said, we can't lean into this term too much because it's not inclusive of things like Atlas credits.  [Virtual currency](https://en.wikipedia.org/wiki/Virtual_currency) still appears to me to be the best term for that. 

Here is an image from the [wiki entry for digital currency](https://en.wikipedia.org/wiki/Digital_currency).

image

Ultimately, I doubt that we can totally future-proof these terms.  There may come a day where the lines between national currency and virtual currency blur further.  But for now, I think that's as good as delineation as any.

So, all of that said, I think this is the path forward:

```suggestion
* National currency (e.g. USD, EUR).
```

specification/columns/contractedcost.md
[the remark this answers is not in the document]
shawnalpay: We discussed this in TF3, and I issued a poll in Slack to discuss further.  Cost columns will remain in Billing Currency; the crux of the debate here is whether Unit Prices would change to Pricing Currency (as proposed) or remain in Billing Currency.  More to come.

specification/columns/pricingcurrencytobillingcurrency.md
shawnalpay: @ijurica, are you saying you would prefer `PricingToBillingExchangeRate` over `BillingExchangeRate`?  That it feels important to say what we're going from and to?  If so, I agree.  @flanakin what do you think?

specification/attributes/currency_format.md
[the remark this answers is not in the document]
shawnalpay: (As a reminder to myself: if/when we change this, I would need to search/replace for a few other spots I used the term "fiat".)

specification/columns/pricingcurrency.md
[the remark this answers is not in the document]
shawnalpay: I'm committing this change.  @flanakin lmk if this isn't the right term.

specification/columns/pricingcurrencytobillingcurrency.md
UdamDewaraja: IMO PricingToBillingExchangeRate | PricingToBillingConversion | PricingToBillingConversionRate seems like good middle ground options here.

specification/columns/pricingcurrencytobillingcurrency.md
UdamDewaraja: Im not sure I know of a use case where this would be needed on a row-level.

For the non-currency -> national currency type scenarios, the rate is typically stable (based on contractual terms) and even if it weren't, it wouldn't be due to a date but instead due to which batches of credits were available.

For national currency -> national currency scenarios, I would think its dependent on the currency conversion logic of the provider and it feels like it wouldn't be something that would be useful to know on a row by row basis. For example, if a provider uses the start of the day rate or the end of the day rate, a user would know based on the charge/billing period what date the rate was converted

flanakin: How many providers offer data in billing currency today? How many include the exchange rate? What do they do?

I can tell you what one of them does 😁 From that perspective, maybe we should ask practitioners what they expect. To be clear, I'll include the date as I already do today, but I do think it's better if it's defined by the spec. Even if that's optional and not required. There's no reason for it to be a custom column.

shawnalpay:
Per further discussion, this pull request has now been replaced by:

- #848 

That PR sustains the following suggestions:

- Rename of `Pricing Currency to Billing Currency` to `Pricing To Billing Exchange Rate`
- Use of the term `national` instead of `fiat` to describe government-issued currencies (e.g. USD, EUR)

For more details, please see the top comment in that PR.  I am closing this PR.

Issue #765, opened 2025-02-19
[AI]  Supported Feature: Resource Usage in PR


rileyjenk:
### Detailed Description

We need to author a supported feature for the Resource Usage Feature, see: https://docs.google.com/spreadsheets/d/1y5zeLZXLaEwK9Y9MT3khAHOy2c2hpnSrx4YCutZCQ0A/edit?gid=1123144016#gid=1123144016

Branch the current branch were we are working on these supported features https://github.com/FinOps-Open-Cost-and-Usage-Spec/FOCUS_Spec/tree/supported_features. Use one of the existing supported features as a template. When you are done submit a PR

When done link your PR to this AI.

shawnalpay:
#809 is merged.  Closing.

Issue #643, opened 2024-11-07
[Work_Item] Define change management principles to determine backwards compatibility


AWS-ZachErdman:
### Problem Statement

FinOps practitioners who adopt FOCUS will need to upgrade to new versions over time as they are released in order to be able to receive the new features provided. However, upgrading to a new version from an old version becomes challenging if, between the new and previous version, there are changes requirement dictating how data must appear in a given column. This would mean the new specification is not "backwards compatible" and this change would be a "breaking change".  

Breaking changes may require practitioners who upgrade FOCUS versions to either (1) update their existing SQL queries that they used to run on the previous version in order to generate the desired results in the new version or (2) modify a FOCUS export such that the data of an existing column in a new export version matches the requirements of the previous version so that they do not need to change their queries.

### Use Case
NA

### Objective

The FOCUS working group recognizes that breaking changes make it more difficult for FinOps practitioners to upgrade to new FOCUS versions and thus, should be avoided when possible. However, breaking changes may be required in certain circumstances when the benefit of the change outweighs the cost it presents to practitioners to adopt.

This work item should result in clear rules for how the FOCUS working group handles breaking changes when formulating new FOCUS versions to make sure that practitioners don't have to deal with breaking changes that don't add tangible value to their FinOps practices. In order to do this, we will define:

1. What is a "breaking change" and "backwards compatibility" according to FOCUS?
2. How do we evaluate the benefit and cost to a practitioner of a breaking change?
3. How can we change our specification development process to ensure we are properly evaluating breaking changes?

### Supporting Documentation

- https://github.com/FinOps-Open-Cost-and-Usage-Spec/FOCUS_Spec/issues/578

### Proposed Solution / Approach

We will add guidelines in our FOCUS github repo that clearly define what a breaking change is and how they should be evaluated. We may also update some of our issue and work item templates to include references to breaking changes.

### Epic or Theme Association

Epic: [Backwards Compatibility]

### Stakeholders

Primary Stakeholder: Zach Erdman (AWS)

jpradocueva:
### Summary from the Maintainers' call on Nov 25:
 Context:
The aim is to document and formalize principles that ensure backwards compatibility across FOCUS versions, balancing innovation with stability for users.
This WI needs to be defined as soon as possible as the rest of WIs may have a dependency with the finding from this work.
Maintainers Assigned:
Zach, Riley
Task Force Assigned:
[...]

PR #809, opened 2025-02-27
WI #643: Add supported feature for resource usage

narlaadbe:
Related to #765 

@rileyjenk, I skipped adding the "Use Case ID" field as outlined in the [template](https://github.com/FinOps-Open-Cost-and-Usage-Spec/FOCUS_Spec/blob/supported_feature_resource_usage/specification/supported_features/supported_feature_template.md?plain=1#L5) because I didn't see it being used in the other supported features that were already drafted.

### DIFF
--- a/specification/supported_features/resource_usage.md
+++ b/specification/supported_features/resource_usage.md
@@ -4,21 +4,20 @@
 1.0
 
 ## Description
-FOCUS enables tracking of resource consumption through billing data. By recording what resources were used, in what quantities, and with what units of measure, practitioners can identify anomalies in usage patterns and optimize workloads.
+FOCUS enables tracking of resource consumption by providing information about which resources were used, in what quantities, and with what units of measure.
 
 ## Directly Dependent Columns
+* ConsumedQuantity
+* ConsumedUnit
+* ResourceId
+* SkuId
+
+## Supporting Columns
 * ChargeCategory
 * ChargePeriodEnd
 * ChargePeriodStart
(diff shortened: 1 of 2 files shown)


narlaadbe:
@rileyjenk, I skipped adding the "Use Case ID" field as outlined in the [template](https://github.com/FinOps-Open-Cost-and-Usage-Spec/FOCUS_Spec/blob/supported_feature_resource_usage/specification/supported_features/supported_feature_template.md?plain=1#L5) because I didn't see it being used in the other supported features that were already drafted.

specification/supported_features/resource_usage.md
```diff
@@ -0,0 +1,41 @@
+# Resource Usage
+
+## Introduced Version
+1.0
+
+## Description
+FOCUS enables tracking of resource consumption through billing data. By recording what resources were used, in what quantities, and with what units of measure, practitioners can identify anomalies in usage patterns and optimize workloads.
```
rileyjenk, line 7:
```suggestion
FOCUS enables tracking of resource consumption by providing what resources were used, in what quantities, and with what units of measure.
```

rileyjenk: This needs more refinement but I put a suggestion to show delta. The problem with including things like anomalies is that the spec itself doesn't support the designation of anomalies or patterns. However that is something that should and is in the Use Cases on the Finops foundation site.

rileyjenk (approved):

PR #812, opened 2025-02-28
WI #643: Add supported feature for service categorization

narlaadbe:


specification/supported_features/service_categorization.md
```diff
@@ -0,0 +1,43 @@
+# Service Categorization
+
+## Introduced Version
+1.1
+
+## Description
+FOCUS provides a structure for categorizing services based on their core functions. By classifying services into high-level categories and more granular subcategories, practitioners can analyze costs according to functional areas, identify spending patterns, discover optimization opportunities, and support more effective procurement decisions. Service categorization also enables organizations to track migration progress by monitoring how spending shifts across service categories over time as workloads move between environments.
+
+## Directly Dependent Columns
+* BilledCost
```
rileyjenk, line 10: I would remove the following from the dependent columns (these are in the query but the columns themselves to not facilitate classification.  We have gone back and forth on this this so if you feel strongly lets discuss. 
 * BilledCost
* BillingCurrency
* BillingPeriodEnd
* BillingPeriodStart
* ProviderName

narlaadbe: If that's our guidance then ProviderName [Here](https://github.com/FinOps-Open-Cost-and-Usage-Spec/FOCUS_Spec/pull/809/files#diff-03c1eb64e8cad03f65f297e8c3756fc433b988643b1064a9f2bdedf8ad54a065R15) needs to go as well. 

I know we discussed removing the "Indirectly Dependent Columns" section.

What if we rename it to "Supporting Columns"? This would help distinguish columns needed for implementation but not directly related to categorization, setting them apart from the core columns in the "Directly Dependent Columns" section.

rileyjenk: Yeah I really like that. @qquigley's pr would also benefit from that structure

beau-nelford-anglepoint: I agree - that is better wording!

rileyjenk (approved):

PR #825, opened 2025-03-06
WI #619: `SKU Family` (previously part of #683)

### LINKED ISSUE #823: [AI] Create two PRs for SKU Family/Size
### Detailed Description

_No response_

kk09v:
In TF2 today, we agreed to split the new column definition for SkuFamily from PR #683


kk09v:
CONTEXT:
- This was split out from #683 (along with #826) so that it could be discussed independently.
- There was a vote in the #workinggroup channel which was inconclusive on whether this should be its own column or whether it should be included in the SkuPriceDetails column as a predefined key.
- There was discussion on what this means for non-IaaS charges
  - E.g. for a PaaS DB where you have the ability to select underlying hardware, would this contain the hardware selected or something
- There was discussion (related to above bullet) on whether this is the right name for this concept is "SKU Family"
  - Some alternative names mentioned (with the context of SkuPriceDetails predefined keys) were "HardwareFamily" or "InfrastructureFamily"

specification/columns/skufamily.md
```diff
@@ -0,0 +1,38 @@
+# SKU Family
+
+The SKU Family column represents the common name for the family or version of the [*SKU*](#glossary:SKU).
+
+It is oftentimes the case that within a [Service Subcategory](#servicesubcategory) a [Provider](#provider) offers multiple different families of versions of SKUs. The SKU Family column provides this information to group and aggregate cost and usage data across SKUs of the same family or version.
```
cnharris10, line 5:
```suggestion
Typically, a [Service Subcategory](#servicesubcategory) a [Provider](#provider) offers multiple different families of versions of SKUs. The SKU Family column provides this information to group and aggregate cost and usage data across SKUs of the same family or version.
```

supporting_content/columns/skufamily.md
```diff
@@ -0,0 +1,29 @@
+# Column: SKU Family
+
+## Example provider mappings
+
+Current column mappings found in available data sets:
+
+| Provider | Data set                   | Column |
+| -------- | -------------------------- | ------ |
+| AWS      | CUR                        |        |
```
cnharris10, line 9: For [AWS](https://docs.aws.amazon.com/cur/latest/userguide/product-columns.html#product-details-I-instancetypefamily), `product/instanceTypeFamily` is close if not identical, and I think a good example to add in the supporting content.

specification/columns/skufamily.md
```diff
@@ -0,0 +1,38 @@
...
+
+The SkuFamily column adheres to the following requirements:
+
+* The SkuFamily column is RECOMMENDED be present in a [*FOCUS dataset*](#glossary:FOCUS-dataset).
+* This column MUST be of type String and MAY contain null values.
+* SkuFamily values SHOULD follow a common convention used by the provider and SHOULD be consistent across SkuId and SkuPriceId records.
+* SkuFamily MUST contain null values when SkuId is null or is not associated with a family or version.
+
```
cnharris10, line 13: It seems like there needs to be another normative clause around the bounds of `SkuFamily` to ensure that it *only* encapsulates a 1:N relationship over similar SKUs.

These 2 examples (attempt to) demonstrate this difference:
- Correct: `m5` --> {`m5.small`, `m5.medium`, `m5.large`, ...} ✅ 
- Incorrect: `Compute` --> {`m5.large`, `db.m5.large`, `cache.m5.large`, ...} ❌ 

Would suggest adding something like ...

```
SkuFamily MUST contain 1 or more related SkuID's encapsulated by a single *ResourceType* and *ServiceName*.
```

cnharris10 (changes requested):

rileyjenk:
I commented in the SKU Size PR: https://github.com/FinOps-Open-Cost-and-Usage-Spec/FOCUS_Spec/pull/826#issuecomment-2715714210
Same concerns.

specification/columns/skufamily.md
flanakin: I don't think we can bind SKU to ServiceSubcategory. VM bandwidth is not tied to the VM service. It's a networking charge. I'd also like to flip this around a bit...
```suggestion
SKU Family groups multiple [SKU Sizes](#skusize) that provide similar functionality. SKU Family is commonly used to group or compare prices, cost, and usage data across SKU Sizes within the same family.
```

specification/columns/skufamily.md
```diff
@@ -0,0 +1,38 @@
...
+The SKU Family column represents the common name for the family or version of the [*SKU*](#glossary:SKU).
+
+It is oftentimes the case that within a [Service Subcategory](#servicesubcategory) a [Provider](#provider) offers multiple different families of versions of SKUs. The SKU Family column provides this information to group and aggregate cost and usage data across SKUs of the same family or version.
+
+The SkuFamily column adheres to the following requirements:
+
+* The SkuFamily column is RECOMMENDED be present in a [*FOCUS dataset*](#glossary:FOCUS-dataset).
+* This column MUST be of type String and MAY contain null values.
+* SkuFamily values SHOULD follow a common convention used by the provider and SHOULD be consistent across SkuId and SkuPriceId records.
+* SkuFamily MUST contain null values when SkuId is null or is not associated with a family or version.
```
flanakin, line 12: Formatting only...
```suggestion
* SkuFamily is RECOMMENDED be present in a [*FOCUS dataset*](#glossary:FOCUS-dataset).
* SkuFamily MUST be of type String.
* SkuFamily values SHOULD follow a common convention used by the provider.
* SkuFamily SHOULD be consistent across SkuId and SkuPriceId records.
* SkuFamily MUST be null when SkuId is null or is not associated with a family or version.
* SkuFamily MAY contain null values.
```

specification/columns/skufamily.md
flanakin, line 13: Should we have a line that explains when it should not be null?
```suggestion
* SkuFamily SHOULD NOT be null when SkuId is not null and is associated with a logical group of SkuSize values.
```
Or do we want to go all in and always require a value based on SkuSize?
```suggestion
* SkuFamily MUST NOT be null when SkuId is not null and SkuSize is not null.
```
I'm sure it's somewhere in between, but it feels like a gap right now.

specification/columns/skufamily.md
```diff
@@ -0,0 +1,38 @@
+# SKU Family
+
+The SKU Family column represents the common name for the family or version of the [*SKU*](#glossary:SKU).
```
flanakin, line 3: "Version" sounds like size to me. Can we remove that?
```suggestion
The SKU Family column represents the common name for the family of the [*SKU*](#glossary:SKU).
```

specification/columns/skufamily.md
flanakin, line 1: Am I reading this correctly to understand that we are declaring SkuFamily as an attribute of a SkuId that groups SkuSize values? SkuFamily is explicitly **_not_** a parent to the SkuId that groups related SkuId values (e.g., "paper products" for our beloved notebook example).

supporting_content/columns/skufamily.md
```diff
@@ -0,0 +1,29 @@
...
+| Azure    | Cost details export or API |        |
```
flanakin, line 10:
```suggestion
| Azure    | Cost details export or API | MeterSubcategory |
```

flanakin (approved):

rileyjenk (changes requested):
Unless we declare that this column and value are also a key and a value in the skupricedetails thing, we will end up with differentiator attributes in one and not the other.  This will be a pain to use for practitioners.

daviddinhgcp:
tagging myself to add comment

specification/columns/skufamily.md
```diff
@@ -0,0 +1,38 @@
+# SKU Family
```
UdamDewaraja, line 1: Again, late to the party so feel free to ignore if the group feels strongly that this is helpful. Based on the description below, I would have no idea what I should put in this. It's vague. 

IMO, this is partly because we're trying to force a common name/property across a set of disparate things. Over the past couple of years of discussing these topics, its clear to me that when something doesn't naturally fit, we shouldn't force it to be a common property and instead use the actual set of terms that describes the things we want to capture. 

If the things we're trying to capture are Storage Class and Instance Families and some other things for other services/providers, rather than trying to group all that into a single column that might not have a sensible name, those individual properties would be better as SkuProperties / SKUDetails so that they can have a reasonable name and a clear value people can count on.

shawnalpay:
Removing milestone 1.2.  This PR will be folded into #683 by @kk09v and then closed.

PR #826, opened 2025-03-06
WI #619: `SKU Size` (previously part of #683)

### LINKED ISSUE #823: [AI] Create two PRs for SKU Family/Size

kk09v:
In TF2 today, we agreed to split the new column definition for SkuSize from PR #683

### DIFF
--- a/specification/columns/columns.mdpp
+++ b/specification/columns/columns.mdpp
@@ -51,6 +51,7 @@ The FOCUS specification defines a group of columns that provide qualitative valu
 !INCLUDE "skumeter.md",1
 !INCLUDE "skupricedetails.md",1
 !INCLUDE "skupriceid.md",1
+!INCLUDE "skusize.md",1
 !INCLUDE "subaccountid.md",1
 !INCLUDE "subaccountname.md",1
 !INCLUDE "subaccounttype.md",1
--- a/specification/columns/skusize.md
+++ b/specification/columns/skusize.md
@@ -1,0 +1,38 @@
+# SKU Size
+
+The SKU Size column represents the common name for the size and shape of the [*SKU*](#glossary:SKU), where applicable.
+
+It is oftentimes the case that a SKU has a common name to refer to its size and shape. This size may be common across multiple [SKU IDs](#skuid) and [SKU Price IDs](#skupriceid), which are differentiated based on properties other than their size (e.g. [Region Name](#regionname), [Pricing Category](#pricingcategory), etc.). The SKU Size column provides a way to group and aggregate cost and usage data across like sizes.
+
+The SkuSize column adheres to the following requirements:
+
+* The SkuSize column is RECOMMENDED be present in a [*FOCUS dataset*](#glossary:FOCUS-dataset).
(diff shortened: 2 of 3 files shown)


kk09v:
CONTEXT:
- This was split out from #683 (along with #826) so that it could be discussed independently.
- There was a vote in the #workinggroup channel which was inconclusive on whether this should be its own column or whether it should be included in the SkuPriceDetails column as a predefined key.
- There was discussion on what this means for non-IaaS charges
  - E.g. for a PaaS DB where you have the ability to select underlying hardware, would this contain the hardware selected or something
- There was discussion (related to above bullet) on whether this is the right name for this concept is "SKU Family"
  - Some alternative names mentioned (with the context of SkuPriceDetails predefined keys) were "HardwareFamily" or "InfrastructureFamily"

flanakin (approved):

rileyjenk:
@flanakin @ijurica @kk09v are we saying that we won't have it in a both the dedicated column and in the json of the skupriceid/details thing?

rileyjenk:
Also SKUs don't necessarily have sizes, neither do they have family. 
Server/instance/vm do have family/class/type do have family, generations, and sizes. I am concerned that in the implementation providers and data generators may interperate that the family is "memory Optimized" not the r5 instance family. 

On AWS and other providers applicability of size flexibility and other constructs is based on instance family. So 

| Instance Type | Family            | Generation | Additional Scope | Size    | Size Flex Applicability |
|---------------|-------------------|------------|------------------|---------|-------------------------|
| r4.xlarge     | Memory Optimized  | 4          |                  | xlarge  | r4                      |
| r5.xlarge     | Memory Optimized  | 5          |                  | xlarge  | r5                      |
| r5.2xlarge    | Memory Optimized  | 5          |                  | 2xlarge | r5                      |
| r6g.8xlarge   | Memory Optimized  | 6          | local disk       | 8xlarge | r6g                     |
| r6gd.8xlarge  | Memory Optimized  | 6          | local disk       | 8xlarge | r6gd                    |
| c6g.8xlarge   | Compute Optimized | 6          |                  | 8xlarge | c6g                     |

cnharris10:
Added this comment to the old pr ... https://github.com/FinOps-Open-Cost-and-Usage-Spec/FOCUS_Spec/pull/825#discussion_r1989161748

mike-finopsorg (approved):

ljadvey (approved):

rileyjenk (changes requested):
Unless we require that the column and value are also in skupricedetails, practitioners are going to have to use many attributes to differentiate in scenarios like size flexibility. Size needs to be in the skupricedetails along with other attributes, I would argue more than it needs to be in a separate column.  Unless you are going to also have a column that declares things like multi-az, zonal vs regional, etc...

daviddinhgcp:
tagging myself to this to add comments for sku taxonomy

shawnalpay:
Removing milestone 1.2.  This PR will be folded into #683 by @kk09v and then closed.

PR #688, opened 2025-01-24
WI #616: Initial Example Section For SAAS Vendors.

rileyjenk:

PR #830, opened 2025-03-07
WI #616: Initial Example Section - cleaned up data and spec content

UdamDewaraja:
Updating the structure of the SaaS examples, Adding spend requirements section, build and markup fixes

This is a subset of the work done in https://github.com/FinOps-Open-Cost-and-Usage-Spec/FOCUS_Spec/pull/688. There's too many things to clean up and push at once in that PR - so breaking it down to a smaller PR. 

@rileyjenk has agreed to this new formatting.


shawnalpay:
@UdamDewaraja to coordinate with @gparker-at-sf and @rileyjenk to get commits from #688 moved over to here.

shawnalpay:
Marking this PR as Draft until #831 is complete.

specification/appendix/saas_examples/simple_agreements.md
```diff
@@ -0,0 +1,76 @@
...
+
+ACME Corp allows its customers to purchase their service for a term (in this case, a year) for a &dollar;10,000. ACME provides AwesomeCorp with a single invoice for their usage. ACME does not provide detailed cost and usage reports to AwesomeCorp throughout the Charge Period after the initial purchase.
+
+Given that ACME does not charge based on or track usage, its usage details are irrelevant to this scenario.
+
+**Example FOCUS dataset:** [Simple invoice up-front for service](https://docs.google.com/spreadsheets/d/1kQTDK3Sk9BnNcn6Ovyaa37T1aMaXfHaDahsuk1Notn4/edit?gid=694253803#gid=694253803&range=A8)
+
+Note the following details in the dataset:
+
+* The Charge Period is April 1st 2025 - April 1st 2026. The Billing Period is the month of April 2025 (when the licenses were ordered) and therefore will appear in the April invoice.
```
qquigley, line 17: Should the Charge Period not be April 1st 2025 - March 31st 2026?

specification/appendix/saas_examples/simple_agreements.md
```diff
@@ -0,0 +1,76 @@
...
+* A single charge representing the total payment for the 12-month agreement (&dollar;10,000) is charged in the first invoice. BilledCost and EffectiveCost are realized in the same record since detailed usage records will not be provided during the 12-month period to realize amortized portions of this up-front payment.
+* The single charge record does not include a List Unit Price, Pricing Quantity, or SKU-related information. Alternatively, the Pricing Quantity could have been set to 1, and the List Unit Price could be the same as the total charge.
+  
+## Scenario A2: Invoice Up-front for a Quantity of a Service
+
+ACME Corp offers its customer the ability to purchase a fixed quantity of licenses for their service. ACME provides AwesomeCorp with a single invoice for their usage. ACME does not provide detailed cost and usage reports to AwesomeCorp throughout the Charge Period after the initial purchase.
+
+On April 1st, 2025, ACME executes a contract and invoices AwesomeCorp &dollar;50,000 (Billed Cost) for a Charge Period of April 1st 2025 to April 1st 2026. As there is no negotiated discount, List Cost of the purchase is also &dollar;50,000.
```
qquigley, line 25: Same as above, should the Charge Period not be April 1st 2925 through March 31st 2026?

specification/appendix/saas_examples/spend_agreements.md
```diff
@@ -0,0 +1,89 @@
...
+The scenarios described below illustrate how a FOCUS-compliant dataset should look for various spend agreement scenarios.
+
+## Baseline Scenario
+
+The following baseline conditions apply to the scenarios described below:
+
+* AwesomeCorp has signed an agreement with SaaS provider Acme Co to use their database services
+* On April 1 2025, AwesomeCorp agrees to spend &dollar;1200 (post-discounts) in the upcoming 12-months
+* AwesomeCorp receives a 20% negotiated discount in return for the commitment
+* Acme co calculates the spend counted against the agreements after discounts (like the negotiated discounts). Other providers may use the cost after discounts i.e., using List Cost for calculating the spend commitment.
```
qquigley, line 14: Capitalization on `co`

specification/appendix/saas_examples/spend_agreements.md
```diff
@@ -0,0 +1,89 @@
...
+
+* A single charge representing the total unused amount from the 12-month agreement is charged during the final month of the 12-month term
+
+### Scenario A2: Billed in arrears with a minimum spend requirement per month
+
+The spend agreement with Acme requires the customer to spend a minimum amount in each Billing Period (monthly). Unused fees are charged per Billing Period when the consumption is below this level (use-it or lose-it). For this scenario, contract terms additionally include the following terms:
+
+* A minimum of &dollar;60 needs to be spent each month
+
+AwesomeCorps consumption looks like this:
```
qquigley, line 46: Apostrophe AwesomeCorp's

specification/appendix/saas_examples/simple_agreements.md
UdamDewaraja: Good question.. In FOCUS the start date is inclusive, and the end date is exclusive. This applies to both BillingPeriod and ChargePeriod.

specification/appendix/saas_examples/simple_agreements.md
UdamDewaraja: Same as above. The *End dates are exclusive.

specification/appendix/saas_examples/spend_agreements.md
UdamDewaraja: Good call. In this PR we unfortunately have to have two people's changes and we're hoping to preserve commit history. I will keep these open and add them as a single change at the end (along with any other comments and feedback people provide)

specification/appendix/saas_examples/spend_agreements.md
UdamDewaraja: Same as above.

specification/appendix/saas_examples/spend_agreements.md
```diff
@@ -0,0 +1,89 @@
...
+
+* All charges will be billed in arrears at a monthly frequency
+
+### Scenario A1: Billed in arrears with no minimum spend requirement per month
+
+For this scenario, contract terms additionally include the following terms:
+
+* Committed spend can be used anytime within the 1-year term
+
+AwesomeCorps consumption looks like this:
```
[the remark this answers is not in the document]
UdamDewaraja: Same as above.

specification/appendix/saas_examples/spend_agreements.md
```diff
@@ -0,0 +1,89 @@
...
+
+* In the first month, AwesomeCorp uses &dollar;48 of services (4 server hours). This usage has a List Cost of &dollar;60 (before discounts)
+* In the following 2 months, AwesomeCorp has some more usage
+* For the final 9 months, AwesomeCorp does not use Acme services
+
+**Example FOCUS dataset:** [Scenario A1 entries](https://docs.google.com/spreadsheets/d/1H69HmngVv-mKpFR-sveY-fSAdR49syqQd7AKOPPQ64A/edit?gid=1740845023#gid=1740845023&range=A15)
```
ijurica, line 34: ## Oversights

### PricingQuantity column

In one of the charges PricingQuantity set to 10 and it should be 4?

---

## Suggestions:

### PricingCategory

> * PricingCategory MUST be present in a [*FOCUS dataset*](#glossary:FOCUS-dataset) when the provider supports more than one pricing category across all SKUs.
> * PricingCategory MUST NOT be null when ChargeCategory is "Usage" or "Purchase" and [ChargeClass](#chargeclass) is not "Correction".

I understand (not happy about it, but I understand) that this column is mandatory only if the provider supports more than one pricing category across all SKUs, and from that perspective, this dataset is valid. Still, **I was wondering if we could fill it out.** 😄 

_**Note:** The same suggestion applies to all other SaaS Spend Agreement scenarios._

### SKU-related columns

Considering we are providing unit prices in these sample datasets, I believe we should also **fill out the SKU-related columns**.

Listing a couple of relevant normative requirements from the FOCUS 1.1 version:

> * SkuId column MUST be present in a FOCUS dataset when the provider publishes a SKU list.
> * SkuId MUST NOT be null when ChargeClass is not "Correction" and ChargeCategory is "Usage" or "Purchase".

> * SkuPriceId MUST be present in a [*FOCUS dataset*](#glossary:FOCUS-dataset) when the provider publishes a SKU price list...
> * SkuPriceId MUST NOT be null when ChargeClass is not "Correction" and ChargeCategory is "Usage" or "Purchase",

> * Etc.

The requirements listed above are somewhat unclear, and one might interpret that these columns are mandatory only if SKUs and SKU prices are publicly published—i.e., if public price sheets exist.
In TF-2, we discussed this and agreed that **SkuId, SkuPriceId, etc., are relevant and therefore mandatory** not only when they are publicly available but also when a provider offers **custom price sheets** - since the consumer must be able to **verify unit prices**.

Even though the PR addressing the clarification of SkuId and SkuPriceId is still pending approval, I would argue that the existing wording of these normative requirements at least implies that these columns should be included in the sample datasets.

_**Note:** The same suggestion applies to all other scenarios (both Simple SaaS Agreements and SaaS Spend Agreement ._

---

## Discussion Topics

### List Cost in case of `End of contract unused fee`

I've been struggling with the **ListCost for Unused Usage** in the case of CSP's non-negotiated Commitment Discounts for a while, and I'll use this opportunity to ask the question because the **same concern arises here**.

In this use case, AwesomeCorp committed to spending $1,200 in exchange for a discounted unit price. In the end, they **overpaid** for that service, right? If they hadn't committed to $1,200, they would have spent $285 (19 * $15) and not $1,200.

I would expect **SUM(ListCost) - SUM(BilledCost)** to show that **waste of $915** ($1,200 - $285), but based on the sample data, it seems like they saved $57.

If we set ListCost for the 'End of contract unused fee' to 0, comparing the aggregated ListCost and BilledCost would outline the waste?

ijurica: #### List Cost for Unused  

When it comes to **List Cost for unused, I spoke too soon** - I’ve changed my mind in the meantime. 😄  

Instead of setting it to **0**, I believe we should fill out **ListUnitPrice**, **ContractedUnitPrice**, and **PricingQuantity** in all rows and stick to the formula:  
**ListUnitPrice × PricingQuantity = ListCost**.  

If we follow this approach, the **ListCost** value doesn’t add up in a series of records (usually the unused fee and pre-payment purchase). I’m not sure if that’s an oversight or something else?

I started adding a bunch of comments to the **[1.2 SaaS Examples - Data samples from AIs](https://docs.google.com/spreadsheets/d/1H69HmngVv-mKpFR-sveY-fSAdR49syqQd7AKOPPQ64A/edit?gid=1740845023#gid=1740845023)** sheet based on this assumption - maybe that's a waste of time?  

@UdamDewaraja @rileyjenk  – are these oversights, or is there a specific reason why **ListCost = ContractedCost** applies to those records?

ijurica: ### Scenarios and Accompanying Sample FOCUS Datasets Completeness  

I believe it would be very useful to provide fully valid datasets with all charge records relevant to each scenario.  

#### SKU-related and all Other FOCUS columns

In other words, we should include all FOCUS columns (except perhaps those related to Capacity Reservation, assuming Acme Co doesn't support that concept) and populate them in a way that meets all normative requirements. And yes, I’m aware that we would need to invent values for some columns - for example, SKU-related columns.  

#### Implied charge records

I’ve noticed that in some scenarios (or datasets), we omit usage or purchase records if they already appear in a dataset associated with another, previously mentioned scenario, relying on the end-user to recognize which records should also be present in the current scenario. I believe the UX would be better if each dataset included all corresponding **Purchase**, **Usage**, and **Unused Usage** records. 

#### Purpose

Such completed datasets could be loaded alongside the sample dataset from the **[Sandbox](https://github.com/FinOps-Open-Cost-and-Usage-Spec/FOCUS-Sample-Data)**, allowing users to run **Supported Features Queries** and **Use Case Queries**. It would also be useful for us… (Putting the FOCUS spec to the test!)

---

That being said, **I don’t think this is a show-stopper for merging this PR** into `working_draft`. We can refine this during the **Consistency Review** phase once we have a finalized set of columns for **FOCUS 1.2**. 🙂

specification/appendix/saas_examples/simple_agreements.md
```diff
@@ -0,0 +1,76 @@
...
+
+Similar to Scenario A above, ACME Corp offers its customer the ability to purchase their service with a fixed quantity of licenses. However, in Scenario B, ACME issues the invoice at the end of the usage period.
+
+On April 1st, 2026, ACME invoices AwesomeCorp &dollar;50,000 (Billed Cost) for the Charge Period of April 1st 2025 to April 1st 2026. As there is no negotiated discount, List Cost of the purchase is also &dollar;50,000.
+
+**Example FOCUS dataset:** [Billed in arrears entries](https://docs.google.com/spreadsheets/d/1kQTDK3Sk9BnNcn6Ovyaa37T1aMaXfHaDahsuk1Notn4/edit?gid=694253803#gid=694253803&range=A17)
+
+Note the following additional details in the dataset:
+
+* The Charge Period is April 1st 2025 to April 1st 2026. The Billing Period is now the month of March 2026 (since this charge is invoiced as of the last month of the Charge Period).
```
ijurica, line 55: When I compare the description and the dataset, I see a mismatch in BillingPeriodStart and BillingPeriodEnd.

If we say "The Billing Period is now the month of March 2026 (since this charge is invoiced as of the last month of the Charge Period)", I believe the data in the dataset needs to be adjusted.

**Current:**

- BillingPeriodStart: 4/1/2026
- BillingPeriodEnd: 5/1/2026

**Suggested fix:**

- BillingPeriodStart: 3/1/2026
- BillingPeriodEnd: 4/1/2026

specification/appendix/saas_examples/simple_agreements.md
```diff
@@ -0,0 +1,76 @@
...
+
+Given that ACME does not charge based on or track usage, its usage details are irrelevant to this scenario.
+
+**Example FOCUS dataset:** [Simple invoice up-front for service](https://docs.google.com/spreadsheets/d/1kQTDK3Sk9BnNcn6Ovyaa37T1aMaXfHaDahsuk1Notn4/edit?gid=694253803#gid=694253803&range=A8)
+
+Note the following details in the dataset:
+
+* The Charge Period is April 1st 2025 - April 1st 2026. The Billing Period is the month of April 2025 (when the licenses were ordered) and therefore will appear in the April invoice.
+* A single charge representing the total payment for the 12-month agreement (&dollar;10,000) is charged in the first invoice. BilledCost and EffectiveCost are realized in the same record since detailed usage records will not be provided during the 12-month period to realize amortized portions of this up-front payment.
+* The single charge record does not include a List Unit Price, Pricing Quantity, or SKU-related information. Alternatively, the Pricing Quantity could have been set to 1, and the List Unit Price could be the same as the total charge.
```
ijurica, line 19: Considering this is a Purchase charge record and the provider obviously supports unit pricing, I believe we must provide Unit Prices, PricingQuantity, PricingUnit, etc.

* [List|Contracted]UnitPrice MUST NOT be null when [ChargeClass](https://github.com/FinOps-Open-Cost-and-Usage-Spec/FOCUS_Spec/blob/working_draft/specification/columns/contractedunitprice.md#chargeclass) is not "Correction" and [ChargeCategory](https://github.com/FinOps-Open-Cost-and-Usage-Spec/FOCUS_Spec/blob/working_draft/specification/columns/contractedunitprice.md#chargecategory) is "Usage" or "Purchase"

specification/appendix/saas_examples/simple_agreements.md
```diff
@@ -0,0 +1,76 @@
+# Simple SaaS Agreements
+
+Many SaaS providers provide simple contract terms, therefore don't need to support complex scenarios like spend commitments or pricing strategies in their billing data.
+
+The scenarios described below illustrate how a FOCUS-compliant dataset should look for simple SaaS agreement scenarios (these scenarios may not be specific to SaaS agreements only).
+
+## Scenario A1: Invoice Up-front for a Purchase of a Service
+
+ACME Corp allows its customers to purchase their service for a term (in this case, a year) for a &dollar;10,000. ACME provides AwesomeCorp with a single invoice for their usage. ACME does not provide detailed cost and usage reports to AwesomeCorp throughout the Charge Period after the initial purchase.
```
ijurica, line 9: I noticed that **BilledCost is set to 10,000**, while **EffectiveCost is set to 50,000**, the same as ContractedCost and ListCost.
I assumed that the same discounts should also be applied to EffectiveCost, meaning it should match BilledCost, not ListCost and ContractedCost.

specification/appendix/saas_examples/spend_agreements.md
UdamDewaraja: **List Cost and Unused**

@ijurica I've added comments in the spreadsheet directly explaining why I don't think unit prices and negotiated discounts apply to unused charges of a spending commitment. Curious to get other peoples thoughts as well.

**Implied charge records**

Where's the Implied charge records? We need to fix that if thats the case. Would be helpful to get what specific places. I assume you mean scenarios [A1 and A2](https://docs.google.com/spreadsheets/d/1kQTDK3Sk9BnNcn6Ovyaa37T1aMaXfHaDahsuk1Notn4/edit?gid=1987749478#gid=1987749478&range=A11) in spend commitments examples? 

In this case, its billed in arrears, and there's no up-front purchase - just an agreement to spend that amount over the following year. It's not that dissimilar to a no-up-front Savings Plan. Happy to update if that's not the correct representation.

**Data completeness in examples / SKU etc.**

I think having FOCUS datasets that are complete is a different from targeted examples that are trying to show a particular scenario. Otherwise examples are going to be filled with a lot of useless details that make it hard for people to understand what things are relevant and what they should pay attention to. This isn't about putting a value in SKU or any other column. When we put values, it should tie into the example. If it doesn't, and is not necessary for explaining the example, my opinion is its better kept out of the example (both the example description / requirements and also the dataset).

specification/appendix/saas_examples/simple_agreements.md
UdamDewaraja: Argh, sorry copy paste issue. Fixed. Thanks for closely looking!

specification/appendix/saas_examples/simple_agreements.md
UdamDewaraja: I think I was trying to model the simplest/most basic scenario here.. where a provider has a charge for a service where it's just a price. There's not a unit price that was published, just an agreed-upon price that a vendor and a customer agree to. As unlikely as it might seem, smaller vendors work this way - They just have a 'contact us' as the price for the service and based on some factors, determine a price for the 'thing' you're buying.

specification/appendix/saas_examples/simple_agreements.md
shawnalpay: @ijurica Please review and resolve if the data looks good to you now!

specification/appendix/saas_examples/spend_agreements.md
ijurica: **Resolved:**

- **Implied charge records:** My mistake. I clicked 'Remove duplicate records' to remove the empty rows before I marked which scenario each row belongs to.
- **Data completeness in examples / SKU etc.: ** @UdamDewaraja I hear you. This is out-of-scope for this PR.
  - We can later use these CSVs to prepare the "completed datasets" so that we can run/test Supported Features Queries and Use Case Queries, etc. However, that is unrelated to this PR.

**Can be addressed/resolved in the Consistency Review:**

- **List Cost for Unused:** I believe we can and should discuss and address this in the Consistency review. We could either adjust some of the normative requirements to accommodate what @UdamDewaraja is suggesting, or adjust the samples to ensure they are fully aligned with the requirements.
  - BTW, @UdamDewaraja did you add the comments to the old sheet? 😄 

---

@UdamDewaraja feel free to resolve this conversation

ijurica (approved):

specification/appendix/saas_examples/simple_agreements.md
ijurica: I think this also falls into the category of "can and should be discussed and addressed in the Consistency review" (either by adjusting some of the normative requirements or by modifying the sample data).

@UdamDewaraja feel free to resolve this conversation.

specification/appendix/saas_examples/spend_agreements.md
UdamDewaraja: Fixed

specification/appendix/saas_examples/spend_agreements.md
UdamDewaraja: Fixed.

specification/appendix/saas_examples/spend_agreements.md
UdamDewaraja: @ijurica which comments? I deleted the old sheet and updated links to use only the new sheet

rileyjenk (approved):

ljadvey (approved):

jpradocueva:
As per Udam's advice, this PR—approved by the Members in the March 27 call—is merged by creating a `commit` rather than `squash & merge`. This is to preserve George's work.

jpradocueva (approved):

Issue #299, opened 2024-01-24
Invoice Issuer concerns


flanakin:
### Type
> Dimension: InvoiceIssuerName
> _An Invoice Issuer is an entity responsible for invoicing for the resources or services consumed. It is commonly used for cost analysis and reporting scenarios._

### Description
> There are scenarios where the InvoiceIssuerName may not be the organization that generates the invoice for the end customer because the provider doesn't know about the multiple layers of intermediaries (e.g., distributors, providers, resellers) between the originating provider and the customer. The originating provider will generate the data that the end customer may see, but the end customer may not even know about the upstream organization that was the direct partner of the originating provider. For example, I may get cloud services from and am invoiced by Breezy Bytes LLC, but it's actually Azure under the covers. I don't know who Breezy Bytes is working with. I just know it's Azure and I get my cost data from Microsoft. Let's assume the following set of relationships:
>
>     Me > Breezy Bytes LLC > Cloudy Clusters Inc > Sunny Systems Gmbh > Microsoft
>
> As a customer, I only know Breezy Bytes and Microsoft may only know about Sunny Systems. Neither of us may know about the extra intermediary. Legally, I'm not sure if Microsoft can even disclose upstream intermediaries. (I haven't checked yet.)
>
> Also, during internal discussions, several people questioned what "invoice issuer" meant and felt it wasn't very clear. They were asking if this is a standard term across the industry and, if not, if there is a standard term.

### Definition of done
- [x] Rationalize vendor-neutral, cross-cloud naming
- [x] Complete spec template and include naming (code name, display name), constraints, guidelines, compatibility with major providers etc.

<br>

### Give it a 👍 below ↴

If you can discuss and help finalize the change, add yourself as an assignee ↗

Comments are welcome and encouraged!

mike-finopsorg:
As far as I understood the InvoiceIssuer will always match who sent you the invoice, so it changes as you go back up your stack

**Me > Breezy Bytes LLC > Cloudy Clusters Inc > Sunny Systems Gmbh > Microsoft**

Microsoft will send an invoice to Sunny Systems Gmbh with the InvoiceIssuer as Microsoft
Sunny Systems Gmbh will send an invoice to Cloudy Clusters Inc with InvoiceIssuer as Sunny Systems Gmbh
Cloudy Clusters Inc will send an invoice to Breezy Bytes LLC with InvoiceIssuer as Cloudy Clusters Inc
[...]

PR #834, opened 2025-03-09
WI #616: Provider, Publisher, Invoice Issuer, Datagenerator Consistency Refactor

rileyjenk:
As we have reviewed the changes to Host Provider I have observed the following:

- The spec is not internally consistant when using the word provider and the definition is confusing comparing it to Publisher
- The metadata section for ProviderVersion is problematic because it is not specific
- The introduction of Host Provider by naming may appear to be a sub class of provider, no hierarchy is intended, it is more specifically the version published by the data provider. 

In addition to this there was also an issue along the same lines submitted: https://github.com/FinOps-Open-Cost-and-Usage-Spec/FOCUS_Spec/issues/299


rileyjenk:
I believe that this will go a long way in resolving concerns brought up in https://github.com/FinOps-Open-Cost-and-Usage-Spec/FOCUS_Spec/issues/299.

rileyjenk:
@GrinsteadJ I cherry-picked (hopefully correctly) your original commits form your PR and then added more tweaks.

shawnalpay:
> The introduction of Host Provider by naming may appear to be a sub class of provider, no hierarchy is intended, it is more specifically the version published by the data provider.

This will end up being a many-to-many.  For example, a practitioner's Snowflake usage may sit on both AWS and GCP, as might Databricks.  Whether we call that out explicitly in the spec, we can discuss.

specification/columns/invoiceid.md
```diff
@@ -1,6 +1,6 @@
 # Invoice ID
 
-An Invoice ID is a provider-assigned identifier for an invoice encapsulating some or all charges in the corresponding [*billing period*](#glossary:billing-period) for a given [*billing account*](#glossary:billing-account). Invoices are commonly used for scenarios like tracking billing transactions, facilitating payment processes and for performing invoice reconciliation between charges and billing periods.
+An Invoice ID is a invoice Issuer assigned identifier for an invoice encapsulating some or all charges in the corresponding [*billing period*](#glossary:billing-period) for a given [*billing account*](#glossary:billing-account). Invoices are commonly used for scenarios like tracking billing transactions, facilitating payment processes and for performing invoice reconciliation between charges and billing periods.
```
ljadvey, line 3:
```suggestion
An Invoice ID is an Invoice Issuer assigned identifier for an invoice encapsulating some or all charges in the corresponding [*billing period*](#glossary:billing-period) for a given [*billing account*](#glossary:billing-account). Invoices are commonly used for scenarios like tracking billing transactions, facilitating payment processes and for performing invoice reconciliation between charges and billing periods.
```

ljadvey (commented):
1 minor grammatical fix - all other items look good.

specification/columns/invoiceissuer.md
```diff
@@ -1,15 +1,13 @@
...
+An Invoice Issuer is the entity responsible for invoicing and receiving payment from the customer for the consumed [*resources*](#glossary:resource) or [*services*](#glossary:service), including resellers, marketplaces, or billing intermediaries. Commonly used in cost analysis and reporting.
 
 The InvoiceIssuer column adheres to the following requirements:
 
 * The InvoiceIssuer column MUST be present in a [*FOCUS dataset*](#glossary:FOCUS-dataset).
 * This column MUST be of type String and MUST NOT contain null values.
 
-See [Appendix: Origination of cost data](#originationofcostdata) section for examples of [Provider](#provider), [Publisher](#publisher) and
-Invoice Issuer values that can be used for various purchasing scenarios.
+See [Appendix: Origination of cost data](#originationofcostdata) section for examples of Invoice Issuer values that can be used for various purchasing scenarios.
```
ijurica, line 10: - **Editorial:**  The reference to the appendix (both the title and the URL) needs to be updated
- **Note:** The examples are no longer limited to purchasing scenarios?  

```suggestion
See [Appendix: Entity Identification Examples](#entityidentification) section for examples of Invoice Issuer values for various use case scenarios.
```

specification/columns/serviceprovider.md
```diff
@@ -1,26 +1,25 @@
...
-The Provider column adheres to the following requirements:
+The Service Provider column adheres to the following requirements:
 
-* The Provider column MUST be present in a [*FOCUS dataset*](#glossary:FOCUS-dataset).
+* The Service Provider column MUST be present in a [*FOCUS dataset*](#glossary:FOCUS-dataset).
 * This column MUST be of type String and MUST NOT contain null values.
 
-See [Appendix: Origination of cost data](#originationofcostdata) section for examples of Provider, Publisher and
-Invoice Issuer values that can be used for various purchasing scenarios.
+See [Appendix: Origination of cost data](#originationofcostdata) section for examples of Service Provider values that can be used for various purchasing scenarios.
```
ijurica, line 10: - **Editorial:**  The reference to the appendix (both the title and the URL) needs to be updated
- **Note:** The examples are no longer limited to purchasing scenarios?  

```suggestion
See [Appendix: Entity Identification Examples](#entityidentification) section for examples of Service Provider values for various use case scenarios.
```

ijurica:
I’d like to suggest that we use this opportunity to also review the Display Names for these columns:
- Service Provider (ServiceProviderName or ServiceProvider?)
- Invoice Issuer (InvoiceIssuerName)
- Host Provider (HostProviderName or HostProvider?)

When comparing column names and column IDs, there is an inconsistency compared to other columns.

Even though this inconsistency is aligned with our column naming rule/convention, I've noticed in the past that some practitioners find this a bit confusing.

specification/appendix/entity_identification.md
ijurica, line 1: A few things to consider:

- Replace the wording `not running directly on your cloud infrastructure.` Perhaps use `not running directly on Service Provider's cloud infrastructure.`
  - Example: *Purchase record for a cloud marketplace offering that is not running directly on your cloud infrastructure (e.g,. SaaS product, Professional Services)...*

- Replace `N/A` with `Not applicable` (potentially use or `Not available` as well)

- Add Charge Category column to the table overview of Scenarios

GrinsteadJ, line 1: This file suggests that there would only be one Data Generator for every scenario, but I don't think that's true.

In 3.1 for example, I could pull FOCUS data from the Marketplace or the Service Provider. Both of these entities will be making cost and usage data available. 

Whenever the Service Provider, Invoice Issuer, and Host Provider don't all match there's the potential to be multiple Data Generators.

It may be best to exclude Data Generator from the table.

specification/columns/hostprovider.md
```diff
@@ -0,0 +1,38 @@
+# Host Provider
+
+A Host Provider is an entity whose [*resources*](#glossary:resource) are used by the Service Provider to make their [*resources*](#glossary:resource) or [*services*](#glossary:service) available.
+
+The HostProvider column adheres to the following requirements:
+
+* The HostProvider column MUST be present in a [*FOCUS dataset*](#glossary:FOCUS-dataset) and represents the service provider where [*resources*](#glossary:resource) or [*services*](#glossary:service) are deployed on.
```
GrinsteadJ, line 7: As discussed in the other PR, I don't think this should be a required column: https://github.com/FinOps-Open-Cost-and-Usage-Spec/FOCUS_Spec/pull/774/files#r1974289696

In the "for Service Providers that DO NOT allow their customer to select or know the host provider" this column doesn't seem to be redundant.

The column would also be redundant for the host providers themselves.

PR #835, opened 2025-03-10
AI #764 Adding commit usage and under usage supported feature

beau-nelford-anglepoint:
A few things I noticed:

- [ ] This supports analysis by ServiceCategory (Compute), but I don't include it in an example query. 
- [ ] I also had to include two different queries since commits are of two kinds in the spec. 
- [ ] I added "Supporting Columns", instead of "Indirectly dependent" columns, but I didn't add anything. 
- [ ] I'm not sure when this was introduced, which version - I guessed for the 1.0 spec. 
- [ ] Should we simplify the queries a bit?

### DIFF
--- a/specification/supported_features/commit_usage_and_under_usage.md
+++ b/specification/supported_features/commit_usage_and_under_usage.md
@@ -1,0 +1,59 @@
+# Commit Usage and Under Usage
+
+## Introduced Version
+1.0
+
+## Description
+The FOCUS spec supports the tracking of committed use discounts usage and underusage, which can come in the form of commitment discounts, or capacity reservations. 
+
+## Directly Dependent Columns
+* Provider
+* BillingAccountID
+* CommitmentDiscountID / CapacityReservationID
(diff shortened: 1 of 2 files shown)


specification/supported_features/commit_usage_and_under_usage.md
```diff
@@ -0,0 +1,59 @@
+# Commit Usage and Under Usage
+
+## Introduced Version
+1.0
+
+## Description
+The FOCUS spec supports the tracking of committed use discounts usage and underusage, which can come in the form of commitment discounts, or capacity reservations. 
+
+## Directly Dependent Columns
+* Provider
```
rileyjenk, line 10: This looks like supporting column

specification/supported_features/commit_usage_and_under_usage.md
```diff
@@ -0,0 +1,59 @@
...
+* BillingAccountID
```
rileyjenk, line 11: This looks like a supporting column

PR #836, opened 2025-03-10
AI #766 Adding the charge-categorization supported feature

beau-nelford-anglepoint:
- [ ] Since this feature supports at least 5 of the use cases from F2, this was hard to narrow down. I'd love for some help narrowing down the queries.


rileyjenk (approved):
This is good enough to merge into the supported features branch. Lets get more feedback.

PR #837, opened 2025-03-10
AI #806 Adding cost-comparison supported feature

beau-nelford-anglepoint:
This one was a little harder, and I want to make sure
- [ ]  my Introduced version is correct
- [ ] Example query is correct 
- [ ] I'm not missing any features in cost comparison.


rileyjenk (approved):
Lets get more feedback but this is fine for merging.

Issue #617, opened 2024-10-23
[Work_Item] Revise glossary definition for FOCUS dataset to include provider columns


ijurica:
## 1. **Problem Statement** *
> Describe the problem, issue, use case, or opportunity that this work item addresses.
> Include practitioner quotes illustrating real examples a) of questions being asked by practitioners and b) value unlocked by answering these questions, if available.
> - What is the problem?: Explain the context and why it needs resolution.
> - Impact: Describe how the problem affects users, systems, or the project.

The current FOCUS dataset specification does not clarify whether custom/native provider-specific columns (prefixed with x_) that contain essential information not covered by standard FOCUS columns should be included in the FOCUS dataset. 

Due to this omission, some providers who support FOCUS might continue to provide key data only in separate provider-specific cost and usage datasets. If that happens, practitioners will encounter challenges, as there is no reliable way to correlate charge records between the FOCUS dataset and native datasets.

## **Use Case**
Follow instructions on [how to craft a use case](https://docs.google.com/document/d/1oMV78K7cGmQ5cjcAv_aZBe-sQzV9zatvJVTEsE7aL7M/edit?tab=t.0) 

## 2. **Objective** *
> State the objective of this work item. What outcome is expected?
> * Success Criteria: Define how success will be measured (e.g. metrics and KPIs).

The objective is to revise the FOCUS dataset specification to clarify that custom/native columns (prefixed with x_), which provide crucial information unavailable in standard FOCUS columns, should be included in the FOCUS dataset. This will ensure consistency across providers' FOCUS datasets.

Success Criteria:
- Clear specification: The FOCUS dataset glossary will be updated to explicitly state that such columns should be included where they add value beyond standard FOCUS columns.
- Inclusion of custom columns: Custom provider-specific columns (x_ prefixed) that offer essential unavailable in standard FOCUS columns data will be integrated into the FOCUS dataset produced by all providers.

## 3. **Supporting Documentation** *
> Include links to supporting documents such as:
> - Data Examples: [Link to data or relevant files; DO NOT share proprietary information]
> - Related Use Cases or Discussion Documents: [Link to discussion]
> - PRs or Other References: [Link to relevant references]

[...]

PR #838, opened 2025-03-12
WI #617: Proposal for data included in a FOCUS dataset

flanakin:
This PR adds a new requirement that data generators MUST include custom columns for anything they provide that isn't yet covered in FOCUS. Please pay special attention to the wording. That will be the only sticky point.

I also cleaned up the requirements to align to our normative requirements goals.


specification/attributes/column_handling.md
```diff
@@ -0,0 +1,66 @@
...
+
+* Who is responsible for incurring or delivering the service.
+* What the charge is for.
+* When the charge was incurred.
+* Where the service was delivered.
+* Why the charge was incurred for a specific price.
+* How much the charge is and how that cost is calculated.
+
+Given the broad range of ever-evolving providers and service offerings, FOCUS as a specification may not define all the columns needed to comprehensively describe every provider's unique cost and usage charges. Providers and data generators are responsible for including additional information required to *accurately* and *completely* describe the charges included within a *FOCUS dataset*. Rows in a *FOCUS dataset* may be aggregated or split differently than non-FOCUS datasets to meet the requirements defined by FOCUS columns and attributes (e.g., [Discount Handling](#discounthandling)). Data generators are responsible for ensuring the accuracy of dimensions and metrics when aggregating or splitting rows, especially summable metrics like costs and quantities.
```
UdamDewaraja, line 12: The last two sentences seem useful.  Feels like in the first part, we're saying 'provide good data' but I'm not sure this provides a lot of value. 

I'm not seeing anything in the requirements to describe how this can be enforced. If there's not a strong reason, I would remove these things that sound like 'requirements' if they're not actually requirements that are enforceable or are expected.

specification/attributes/column_handling.md
```diff
@@ -0,0 +1,66 @@
+# Column Handling
```
UdamDewaraja, line 1: Could we make necessary changes by renaming the previous file (if its needed), which preserves the git history/blame. Otherwise we make it seem like this is something new introduced in 1.2 when much of these requirements are there from 0.5. It also narrows down the exact change so its easier to see what we're reviewing.

For example, the change log says:
- `Column Naming and Ordering` attribute includes the following new provisions:
  - Providers must include custom columns or any concepts not covered by FOCUS columns when they publish a non-FOCUS cost and usage dataset.

I don't see where we're enforcing anything of this sort here in this file.

CHANGELOG.md
```diff
@@ -263,61 +266,61 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
...
-| SkuId                        | 1.0-preview             |                   |
-| SkuMeter                     | 1.1                     |                   |
-| SkuPriceDetails              | 1.1                     |                   |
-| SkuPriceId                   | 1.0-preview             |                   |
-| SubAccountId                 | 0.5                     |                   |
-| SubAccountName               | 0.5                     |                   |
-| Tags                         | 1.0-preview             |                   |
-| UsageQuantity                | 1.0-preview             | Renamed to ConsumedQuantity in v1.0 |
-| UsageUnit                    | 1.0-preview             | Renamed to ConsumedUnit in v1.0 |
+| **Column ID**              | **Revision Introduced** | **Status**                                 |
```
UdamDewaraja, line 269: nit: could we please avoid mass whitespace-only changes to narrow down the actual change to this content? I don't think I see a change here. Let's not we lose the git blame for things that were done over different releases for just formatting that doesn't result in a change to the spec.

CHANGELOG.md
```diff
@@ -19,6 +19,9 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 
 **Changed:**
 
+- `Column Naming and Ordering` attribute includes the following new provisions:
```
UdamDewaraja, line 22: I could be wrong but trying to find the changes in the content below that is enforcing anything like this and not coming up with anything.

specification/attributes/column_handling.md
```diff
@@ -0,0 +1,66 @@
...
+
+A [*FOCUS dataset*](#glossary:FOCUS-dataset) consists of a set of columns that convey information about the charges incurred with a provider. Each column describes an aspect of the charge, including but not limited to:
+
+* Who is responsible for incurring or delivering the service.
+* What the charge is for.
+* When the charge was incurred.
+* Where the service was delivered.
+* Why the charge was incurred for a specific price.
+* How much the charge is and how that cost is calculated.
```
UdamDewaraja, line 10: nit: Attributes are generally used to enforce something about the data. This seems like great content up at the top of the spec to help people understand this or maybe in the glossary? Having this buried here means very few will learn this about FOCUS datasets

specification/glossary.md
```diff
@@ -94,7 +94,7 @@ An open-source specification that defines requirements for billing data.
...
+A structured collection of cost and usage data that meets or exceeds the "MUST" and "SHOULD" criteria defined by FOCUS. In addition to FOCUS columns, the dataset should include custom provider columns (prefixed with `x_`) when these columns provide additional information not captured by the existing FOCUS columns. If introducing a custom column could result in splitting original charge records into multiple entries, the data generator is responsible for ensuring that the FOCUS dataset fully conforms to all aggregation-related requirements for metric columns, particularly those concerning costs and quantities.
```
UdamDewaraja, line 97:
```suggestion
A structured collection of cost and usage data that meets the [BCP14](https://tools.ietf.org/html/bcp14) criteria defined by FOCUS. In addition to FOCUS columns, the dataset should include custom provider columns (prefixed with `x_`) when these columns provide additional information not captured by the existing FOCUS columns. If introducing a custom column could result in splitting original charge records into multiple entries, the data generator is responsible for ensuring that the FOCUS dataset fully conforms to all aggregation-related requirements for metric columns, particularly those concerning costs and quantities.
```

CHANGELOG.md
```diff
@@ -19,6 +19,9 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
...
+  - Providers must include custom columns or any concepts not covered by FOCUS columns when they publish a non-FOCUS cost and usage dataset.
```
gparker-at-sf, line 23: If the intent is as defined in the conversation: "This PR adds a new requirement that data generators MUST include custom columns for anything they provide that isn't yet covered in FOCUS." Then, 

I feel like the intent of this could come through more clearly in the wording: 

Suggest 
- When publishing a FOCUS cost and usage dataset, providers must include custom columns for any concepts not covered by FOCUS columns.

specification/attributes/column_handling.md
```diff
@@ -0,0 +1,66 @@
...
+
+## Description
+
+Naming, ordering, and inclusion criteria for columns appearing in a *FOCUS dataset*.
+
+## Requirements
+
+### Column Inclusion
+
+When the provider publishes a non-FOCUS cost and usage dataset, the following applies:
```
gparker-at-sf, line 34: >When the provider publishes a non-FOCUS cost and usage dataset, the following applies:

Why are we writing this requirement for when they are not planning to be FOCUS compliant? This confuses me and appears misaligned to the intent in the conversation: "This PR adds a new requirement that data generators MUST include custom columns for anything they provide that isn't yet covered in FOCUS. "

CHANGELOG.md
flanakin: We spoke about this on the maintainers call today. Udam missed the fact that there was one new section above the naming section that introduced this. I'll resolve this for now. This will be better when we have IDs per requirement, so we can call out each requirement that changed.

CHANGELOG.md
flanakin: @gparker-at-sf This is just the summary of what was changed. Can you look at the first requirement in the column_handling.md file to see if that conveys the requirement accurately enough for you? Please leave a comment there if you think any clarity is needed. I'll close this comment here. Feel free to unresolve if you think we need both. I just want to make sure that one is prioritized since this is just a call out that people should look at the requirement.

specification/attributes/column_handling.md
flanakin: I did rename. Looks like GitHub doesn't show that 🙁 I felt the same when I saw it. I can rename it back and we can rename the file later as part of the consistency review if you want. I'll resolve this for now, but let me know if you want me to rename it back for the review.

specification/attributes/column_handling.md
flanakin: The discussion we had in TF2 last week was that the glossary wasn't discoverable enough, so the ask was to call it out here so it's closer to the requirements. I'm fine wherever we land it. We can also move it as part of the consistency review. I'll leave this open to get others' thoughts.

specification/attributes/column_handling.md
flanakin: It's effectively saying, "abide by all requirements". This was explicitly in the glossary and was requested to be added here. I felt it was fine with or without, so I just added it as it doesn't hurt anything to double-down the point. Happy to go whatever way the group prefers.

specification/attributes/column_handling.md
flanakin: This is the conditionality for when the next line is required. I believe this is aligned with the normative requirements guidance but let me know if I missed anything. Happy to adjust. Will resolve this for now.

specification/attributes/column_handling.md
flanakin: Turns out it actually didn't go in as a rename. Not sure why since I confirmed it was a rename when I pushed 😕 Either way, I fixed it. Thanks for calling this out. Hopefully, the comments will line up 🤞

flanakin: It didn't 😒 Not sure what to do. It looks like changing it at this point won't fix it, even if the history shows it as a rename.

specification/attributes/column_naming_and_ordering.md
```diff
@@ -1,36 +1,54 @@
...
+Column Handling
 
 ## Description
 
-Naming and ordering convention for columns appearing in a [*FOCUS dataset*](#glossary:FOCUS-dataset).
+Naming, ordering, and inclusion criteria for columns appearing in a *FOCUS dataset*.
 
 ## Requirements
 
+### Column Inclusion
```
flanakin, line 32: This section is the primary purpose of the PR. All other changes are purely editorial cleanup that happened as we discussed where and how to add this requirement..

specification/attributes/column_naming_and_ordering.md
```diff
@@ -1,36 +1,54 @@
...
+
+When the provider publishes a non-FOCUS cost and usage dataset, the following applies:
+
+* Custom columns MUST be included for all information not covered by FOCUS columns that exists in the latest version of non-FOCUS cost and usage datasets.
+
```
flanakin, line 37: Adding this for clarity based on the discussion in the maintainer meeting...
```suggestion
* Data generators MAY allow practitioners to select a subset of FOCUS or custom columns.
```

specification/attributes/column_naming_and_ordering.md
```diff
@@ -1,36 +1,54 @@
...
+* Who is responsible for incurring or delivering the service.
+* What the charge is for.
+* When the charge was incurred.
+* Where the service was delivered.
+* Why the charge was incurred for a specific price.
+* How much the charge is and how that cost is calculated.
+
+Given the broad range of ever-evolving providers and service offerings, FOCUS as a specification may not define all the columns needed to comprehensively describe every provider's unique cost and usage charges. Providers and data generators are responsible for including additional information required to *accurately* and *completely* describe the charges included within a *FOCUS dataset*. Rows in a *FOCUS dataset* may be aggregated or split differently than non-FOCUS datasets to meet the requirements defined by FOCUS columns and attributes (e.g., [Discount Handling](#discounthandling)). Data generators are responsible for ensuring the accuracy of dimensions and metrics when aggregating or splitting rows, especially summable metrics like costs and quantities.
```
thecloudman, line 12:
```suggestion
While FOCUS establishes the core structure and standardizes columns for consistent reporting of cost and usage data, the diverse and evolving landscape of providers and service offerings may require providers and data generators to include supplemental columns in the FOCUS dataset. These additional columns may enable deeper analysis and provide more detailed descriptions of usage that may not be fully captured by standard FOCUS dataset columns.
In such cases, providers and data generators are responsible for ensuring that their usage and cost data is accurately and comprehensively represented by including necessary supplemental columns without duplicating data in FOCUS columns. Rows in a FOCUS dataset may be aggregated or split differently than non-FOCUS datasets to align with FOCUS requirements (e.g., Discount Handling), while enriching the dataset, providers and data generators must maintain the integrity of FOCUS-defined dimensions and metrics. Providers and data generators must ensure the accuracy of all dimensions and metrics, particularly summable values such as costs and quantities, when performing these transformations.
Additionally, any extra columns added by providers and data generators should be optional for consumers, allowing them to include only the additional data they specifically need.
```

thecloudman (changes requested):
I didnt like the wording provided for why custom columns must be added. It could have led providers to believe FOCUS is incomplete and could inhibit uptake, additionally, pushing for custom columns in the current wording gave providers an out for not adopting FOCUS altogether. I have provided updated wording which I believe is cleaner and still achieves what was originally proposed. I also added a line to make custom columns selectable by consumers as by adding every column will significantly bloat the size of the dataset, this will lead to increased cost of 3rd party platforms to users as well as lead to increased costs and complexity for users with homegrown toolsets.

specification/attributes/column_naming_and_ordering.md
```diff
@@ -1,36 +1,54 @@
...
+Naming, ordering, and inclusion criteria for columns appearing in a *FOCUS dataset*.
 
 ## Requirements
 
+### Column Inclusion
+
+When the provider publishes a non-FOCUS cost and usage dataset, the following applies:
+
+* Custom columns MUST be included for all information not covered by FOCUS columns that exists in the latest version of non-FOCUS cost and usage datasets.
+* Data generators MAY allow practitioners to select a subset of FOCUS or custom columns.
```
shawnalpay, line 37: I believe these two lines form the crux of the quasi-philosophical conversation the working group needs to have on this PR.  A couple points to consider:

- Shall we _require_ providers to put all of their non-FOCUS columns into a FOCUS dataset?
- Shall FOCUS _require_ providers to allow practitioners to select a subset of columns in a FOCUS dataset?

My personal opinion is that either A) both lines be MUST, or B) both lines be SHOULD.  Forcing a practitioner to consume every column under the sun without giving them a way to filter down the list isn't very kind of us.

Further, we must decide whether this section shall be A) "north star" requirements that many providers likely will not meet (at least at first), or B) weaker requirements that will not implicitly encourage providers to "level up" and meet.  On this point, I am more ambivalent.

specification/attributes/column_naming_and_ordering.md
```diff
@@ -1,36 +1,54 @@
...
-Naming and ordering convention for columns appearing in a [*FOCUS dataset*](#glossary:FOCUS-dataset).
+Naming, ordering, and inclusion criteria for columns appearing in a *FOCUS dataset*.
 
 ## Requirements
 
+### Column Inclusion
+
+When the provider publishes a non-FOCUS cost and usage dataset, the following applies:
+
+* Custom columns MUST be included for all information not covered by FOCUS columns that exists in the latest version of non-FOCUS cost and usage datasets.
```
thecloudman, line 36:
```suggestion
*Custom columns MUST be available for export in a FOCUS Dataset export for all information not covered by FOCUS columns.
```

specification/attributes/column_naming_and_ordering.md
thecloudman: My opinion is that columns should be available to be selected by practitioners during the export process or during import process. Including every available column by default would impact dataset sizes rather significantly. It would also have the potential to increase the cost of 3rd party tools and increase complexity and cost of home-grown tools. It could also create an environment where providers don't need to do the hard yards to become more FOCUS orientated (compliant) and instead opt to just include all their data in their custom columns.

gparker-at-sf: Putting myself in the practitioner shoes, I'd love it to be A(Must). Putting myself in the vendor shoes, I'd love it to be B(Should). 

If I want to to drive adoption and compliance of the standard, that also seems to gravitate towards B.

specification/glossary.md
```diff
@@ -94,7 +94,7 @@ An open-source specification that defines requirements for billing data.
...
+A structured collection of cost and usage data that meets the [BCP14](https://tools.ietf.org/html/bcp14) criteria defined by FOCUS. In addition to FOCUS columns, the dataset should include custom provider columns (prefixed with `x_`) when these columns provide additional information not captured by the existing FOCUS columns. If introducing a custom column could result in splitting original charge records into multiple entries, the data generator is responsible for ensuring that the FOCUS dataset fully conforms to all aggregation-related requirements for metric columns, particularly those concerning costs and quantities.
```
flanakin, line 97: @ijurica Does this cover what you were thinking? Not sure if we need to put "must" given it's not normative, but fine either way.
```suggestion
A structured collection of cost and usage data that meets the [BCP14](https://tools.ietf.org/html/bcp14) criteria defined by FOCUS. In addition to FOCUS columns, the dataset includes custom provider columns (prefixed with `x_`) when these columns provide additional information not captured by the existing FOCUS columns. If introducing a custom column could result in splitting original charge records into multiple entries, the data generator is responsible for ensuring that the FOCUS dataset fully conforms to all aggregation-related requirements for metric columns, particularly those concerning costs and quantities.
```

specification/attributes/column_naming_and_ordering.md
flanakin: +1 to @gparker-at-sf. But as a reminder, we're building the spec for practitioners to get what they need.

specification/attributes/column_naming_and_ordering.md
flanakin: People are pushing back on making selection a requirement. Let's keep that as a separate requirement for now. I'm personally fine with MUST or SHOULD, but let's keep it separate either way. I'll propose that change for you.

specification/attributes/column_naming_and_ordering.md
flanakin, line 37: Based on Graham's suggestion and the conversation in TF1, proposing making this a strong suggestion but not a requirement since it's a big functional change which also has some hidden requirements. This gives us room to make it required in the future.
```suggestion
* Data generators SHOULD allow practitioners to select a subset of FOCUS or custom columns.
```

flanakin, line 37: Adding missing space
```suggestion
* Data generators MAY allow practitioners to select a subset of FOCUS or custom columns.

```

specification/attributes/column_naming_and_ordering.md
flanakin: Resolving this since I added a separate suggestion to make selection a SHOULD. If anyone feels strongly for or against that (or making a MUST), please comment on the next line.

specification/attributes/column_naming_and_ordering.md
```diff
@@ -1,36 +1,55 @@
...
+* Data generators MAY allow practitioners to select a subset of FOCUS or custom columns.
+
```
flanakin, line 38: As discussed in the TF1 call on Mar 18, I'm proposing we add a statement about conformance. Given all the conversation around MUSTs and SHOULDs around inclusion of columns, this would give data generators a clear way to convey what is and isn't supported so practitioners will know how to handle each case. I'm making it a SHOULD since we're not providing guidance for how to add conformance documentation. I do believe it should be a MUST in 1.3.
```suggestion
* Data generators SHOULD provide conformance documentation that indicates whether the *FOCUS dataset* fully, partially, or does not conform to each requirement.
  * Data generators SHOULD include an explanation of what does not conform when a requirement is not fully conformant so practitioners will know they need to handle the difference.

```

aws-letianfeng (changes requested):
I am concerned about the following language. I would propose to exclude the following language from the specification. I am equally concerned if this is a SHOULD requirement instead of a MUST requirement. 

> * Custom columns MUST be included for all information not covered by FOCUS columns that exists in the latest version of non-FOCUS cost and usage datasets.

This will introduce two issues.
1. AWS provides cost and usage data that provides additional rows and columns. Just including more columns is not meaningful for certain features. For example, AWS provides Split Cost Allocation Data for EKS and ECS. If we just introduce split cost columns without having additional rows, the additional columns are not meaningful against existing FOCUS rows because tasks and pods are not billable resources and they don’t show up in FOCUS datasets. If we want to also include these rows, the FOCUS specification is not yet designed to handle the definition of split cost rows. For example, the charge category of tasks and pods cannot be any of the allowed values in FOCUS 1.2 spec. There’s no SKU for task and pod resources, so the SkuId would be undefined. Forcing new data when the detailed specification of the new data is undefined does not help customers.
2. When a column is first introduced as a vendor extension but later get standardized, it creates a backward compatibility issue for AWS customers. A customer may initially code against vendor extension columns, but later have to change their code when the column becomes standardized in later version of FOCUS spec (given there’s no guarantee later FOCUS specification will adopt the specific format of vendor extensions). If we keep the same information in both vendor extension and standardized columns, it leads to duplicated data, which adds the processing cost of both our customers and AWS.


FOCUS specification is a data schema specification, not a functional specification of a service. The following language imposes requirements on the service that generates FOCUS datasets. I'm concerned that this is an over-reach of the FOCUS spec.

> * Data generators MAY allow practitioners to select a subset of FOCUS or custom columns.

rileyjenk:
This change is problematic on two fronts and I believe merits further discussion. For 1.2 I don't think we have enough time to evaluate the impact from such a change. More time needs to be spent on what the core issues that we are trying to solve.  My concerns:

1. The FOCUS spec is a data specification and not an application or feature specification. For the FOCUS spec to say that the Data Generator MUST provide a feature of column selection in their product seems to be an overreach for a data specification and is a different tier of requirement. 
  - The specification itself do not specify which columns are "safe" to be unselected from delivery, that is different likely different than columns that the spec currently defines as MUST be present
2. There are deficiencies in the spec that need to improved, we all know this. I don't think we can say at this time that we have a sound enough data structure that any other data that is being provided in a provider specific cost and usage dataset can be provided. Do I want this yes....can we force it....well no. I actually think that if we put this we would hinder adoption because the bar is substantially higher. 

On conformance I think there is more discussion to be had there. I would love to say that you MUST provide a document that stats any deltas in your implementation of the spec. Schema we have, logic and values we do not. 

About backward compatibility: In general our sector of the tech world has be negatively impacted by poor communication and documentation when changes happen to reporting. The belief that practitioners can't handle/refuse to adapt to changing data structures is a legitimate concern but is also a major symptom of the core problem. This is why many cost reporting structures have an extreme amount of columns and often have columns were little to no one understand why they are there or how to properly use them.  Thus our drive for proper metadata needs to accompany any data delivery.

rileyjenk:
Also, if you are a provider and have existing reports that have things in them that the spec doesn't handle. We need that sort of signal so that we can prioritize. This is actually an area where SaaS is far ahead of the CSPs because most SaaS vendors have more data structures than just cost and usage. They have contract information, renewal dates, burn down, etc...

If you don't include in it things because....well...you didn't, your customers can and will compare the spec to the data they are getting and start asking when you will.

ijurica:
#### Normative requirements:

> When the provider publishes a non-FOCUS cost and usage dataset, the following applies:
> * Custom columns MUST be included for all information not covered by FOCUS columns that exists in the latest version of non-FOCUS cost and usage datasets.

FOCUS columns enable a uniform way to support certain FinOps capabilities. However, for information not covered by FOCUS columns, if it is not available through x_columns in the FOCUS dataset, we still need to inject and process Custom Cost and Usage datasets. If these are separate datasets, in addition to processing two datasets, there is also the added challenge of joining them - especially if the Data Generator does not support unique charge record identifiers.

A few weeks ago, we approved the use of SHOULD in the FOCUS dataset glossary term definition, but it’s important to clarify this in the requirement as well. Until we address some concerns about the implications of including custom (x_) columns, I assume SHOULD is good enough for now. In a future iteration i.e., once we better define/clarify cost and usage scope and double counting handling in the specification - this **should become MUST** (same applies to some other SHOULD requirements in the spec 😄 ).

> * Data generators MAY/SHOULD allow practitioners to select a subset of FOCUS or custom columns.

This requirement indicates that, once a Data Generator enables exporting FOCUS and custom (x_) columns, the consumer has the option to exclude certain columns from the export to reduce dataset size. Exporting a FOCUS dataset that conforms to the initial requirements remains possible, while consumers can still choose to export only a subset of columns.

Even if a provider or Data Generator does not offer the ability to exclude specific columns from export, the consumer can still decide not to import certain columns - so MAY or SHOULD is sufficient.

#### Considerations for Custom (x_) Columns:

- **FOCUS dataset size** – If not reduced during export, it can still be minimized during import. However, consumers interested in information provided in custom datasets (but not in custom (x_) columns within the FOCUS dataset) would still need to process two datasets.

- **Metrics and double counting** – Introducing custom (x_) columns can, in some cases, lead to double counting, which can be quite challenging to prevent. It’s important to explicitly highlight this as a consideration when adding custom columns. In the glossary, we attempted to address this with the following: _"If introducing a custom column could result in splitting original charge records into multiple entries, the data generator is responsible for ensuring that the FOCUS dataset fully conforms to all aggregation-related requirements for metric columns, particularly those concerning costs and quantities."_

- **Scope of the FOCUS Cost and Usage dataset** – As Riley mentioned, "cost and usage" is a broad term, and we should probably refine that before imposing MUST in the first requirement. In particular, we may need to more clearly define the scope of cost and usage data versus, for instance, invoice data scope, contractual data scope, etc..

timwright2000:
I share at least two of the concerns put forward by @aws-letianfeng here (quoted below).

a) introducing reliance on provider specific columns, leading to backwards compatibility and /or duplicated data issues
b) overreach of spec if we start getting into defining functional requirements

I am wondering if instead of vendor specific columns (which can lead to the backward compatibility issue) we should instead have a provider specific join key that enables joining between a focus row and the corresponding row(s) in the provider's native exports until such time as the spec has evolved sufficiently to accommodate the super set of what's available in all of the providers' native exports.


> I am concerned about the following language. I would propose to exclude the following language from the specification. I am equally concerned if this is a SHOULD requirement instead of a MUST requirement.
> 
> > * Custom columns MUST be included for all information not covered by FOCUS columns that exists in the latest version of non-FOCUS cost and usage datasets.
> 
> This will introduce two issues.
> 
> 1. AWS provides cost and usage data that provides additional rows and columns. Just including more columns is not meaningful for certain features. For example, AWS provides Split Cost Allocation Data for EKS and ECS. If we just introduce split cost columns without having additional rows, the additional columns are not meaningful against existing FOCUS rows because tasks and pods are not billable resources and they don’t show up in FOCUS datasets. If we want to also include these rows, the FOCUS specification is not yet designed to handle the definition of split cost rows. For example, the charge category of tasks and pods cannot be any of the allowed values in FOCUS 1.2 spec. There’s no SKU for task and pod resources, so the SkuId would be undefined. Forcing new data when the detailed specification of the new data is undefined does not help customers.
> 2. When a column is first introduced as a vendor extension but later get standardized, it creates a backward compatibility issue for AWS customers. A customer may initially code against vendor extension columns, but later have to change their code when the column becomes standardized in later version of FOCUS spec (given there’s no guarantee later FOCUS specification will adopt the specific format of vendor extensions). If we keep the same information in both vendor extension and standardized columns, it leads to duplicated data, which adds the processing cost of both our customers and AWS.
> 
> FOCUS specification is a data schema specification, not a functional specification of a service. The following language imposes requirements on the service that generates FOCUS datasets. I'm concerned that this is an over-reach of the FOCUS spec.
> 
> > * Data generators MAY allow practitioners to select a subset of FOCUS or custom columns.

rileyjenk (changes requested):
See earlier comments.

specification/attributes/column_handling.md
flanakin: @UdamDewaraja can we consider this feedback something that would be editorial that we could change in the consistency review? 

If there are no other comments, I'm inclined to keep it as-is since that's what came out of the TF discussion. But as I mentioned, I'm fine either way. The intent was to make this more discoverable than solely being in the glossary.

flanakin:
We couldn't get closure on this in the steering committee call and discussed pushing this to 1.3 to give us time to collect more explicit feedback from practitioners about whether the lack of data is why they're not adopting FOCUS.

I would love to bring in the other updates, even if we don't cover column inclusion, but we can discuss that in TF1.

specification/attributes/column_naming_and_ordering.md
```diff
@@ -1,36 +1,57 @@
...
+Naming, ordering, and inclusion criteria for columns appearing in a *FOCUS dataset*.
 
 ## Requirements
 
+### Column Inclusion
+
+When the provider publishes a non-FOCUS cost and usage dataset, the following applies:
+
+* Custom columns MUST be included for all information not covered by FOCUS columns that exists in the latest version of non-FOCUS cost and usage datasets.
+* Data generators SHOULD allow practitioners to select a subset of FOCUS or custom columns.
```
shawnalpay, line 39: @flanakin 
As discussed in the Thu Mar 20 SteerCo and Mon Mar 24 Maintainer meetings, this PR does not appear to have a path forward with these two lines included in any normative variant.  Please remove these lines, and we can discuss the content of the rest possibly moving forward in TF1 tomorrow.
```suggestion
```

shawnalpay, line 39: @flanakin 
As discussed in the Thu Mar 20 SteerCo and Mon Mar 24 Maintainer meetings, this PR does not appear to have a path forward with these two lines included in any normative variant.  Please remove these lines, and we can discuss the content of the rest possibly moving forward in TF1 tomorrow.
```suggestion
```

shawnalpay (changes requested):

specification/attributes/column_naming_and_ordering.md
flanakin: Resolving this since we decided to remove the column inclusion requirement for 1.2

CHANGELOG.md
flanakin, line 23:
```suggestion
```

specification/attributes/column_naming_and_ordering.md
```diff
@@ -1,36 +1,57 @@
...
+
+While FOCUS establishes the core structure and standardizes columns for consistent reporting of cost and usage data, the diverse and evolving landscape of providers and service offerings may require providers and data generators to include supplemental columns in the FOCUS dataset. These additional columns may enable deeper analysis and provide more detailed descriptions of usage that may not be fully captured by standard FOCUS dataset columns.
+In such cases, providers and data generators are responsible for ensuring that their usage and cost data is accurately and comprehensively represented by including necessary supplemental columns without duplicating data in FOCUS columns. Rows in a FOCUS dataset may be aggregated or split differently than non-FOCUS datasets to align with FOCUS requirements (e.g., Discount Handling), while enriching the dataset, providers and data generators must maintain the integrity of FOCUS-defined dimensions and metrics. Providers and data generators must ensure the accuracy of all dimensions and metrics, particularly summable values such as costs and quantities, when performing these transformations.
+
```
flanakin, line 14:
```suggestion

In such cases, providers and data generators are responsible for ensuring that their usage and cost data is accurately and comprehensively represented by including necessary supplemental columns without duplicating data in FOCUS columns. Rows in a FOCUS dataset may be aggregated or split differently than non-FOCUS datasets to align with FOCUS requirements (e.g., Discount Handling), while enriching the dataset, providers and data generators must maintain the integrity of FOCUS-defined dimensions and metrics. Providers and data generators must ensure the accuracy of all dimensions and metrics, particularly summable values such as costs and quantities, when performing these transformations.
```

specification/attributes/column_naming_and_ordering.md
```diff
@@ -1,36 +1,57 @@
...
 
 ## Attribute Name
 
-Column Naming and Ordering
+Column Handling
 
 ## Description
 
-Naming and ordering convention for columns appearing in a [*FOCUS dataset*](#glossary:FOCUS-dataset).
+Naming, ordering, and inclusion criteria for columns appearing in a *FOCUS dataset*.
```
flanakin, line 30:
```suggestion
Naming and ordering convention for columns appearing in a *FOCUS dataset*.
```

supporting_content/attributes/column_handling.md
```diff
@@ -9,7 +9,7 @@ Examples of column names found in provider datasets:
 | AWS       | CUR                      | bill_invoicing_entity |
 | GCP       | Big Query Billing Export | resource.global_name  |
 | Microsoft | Cost details             | ResourceId            |
-| OCI       | Cost reports             | product/resourceId    | 
+| OCI       | Cost reports             | product/resourceId    |
 
 ## Discussion / Scratch space
```
flanakin, line 15:
```suggestion

* The following was a proposal to require providers to include columns in 1.2. This was removed to discuss further in a future release:
  > When the provider publishes a non-FOCUS cost and usage dataset, the following applies:
  > * Custom columns MUST be included for all information not covered by FOCUS columns that exists in the latest version of non-FOCUS cost and usage datasets.
  > * Data generators SHOULD allow practitioners to select a subset of FOCUS or custom columns.
  > * Data generators SHOULD provide conformance documentation that indicates whether the *FOCUS dataset* fully, partially, or does not conform to each requirement.
  >   * Data generators SHOULD include an explanation of what does not conform when a requirement is not fully conformant so practitioners will know they need to handle the difference.
```

specification/attributes/column_naming_and_ordering.md
flanakin: Removed for 1.2. Will re-evaluate later.

specification/glossary.md
```diff
@@ -94,7 +94,7 @@ An open-source specification that defines requirements for billing data.
...
+A structured collection of cost and usage data that meets the [BCP14](https://tools.ietf.org/html/bcp14) criteria defined by FOCUS. In addition to FOCUS columns, the dataset includes custom provider columns (prefixed with `x_`) when these columns provide additional information not captured by the existing FOCUS columns. If introducing a custom column could result in splitting original charge records into multiple entries, the data generator is responsible for ensuring that the FOCUS dataset fully conforms to all aggregation-related requirements for metric columns, particularly those concerning costs and quantities.
```
shawnalpay, line 97: As discussed in TF1 Mar 25.

```suggestion
A structured collection of cost and usage data that meets the [BCP14](https://tools.ietf.org/html/bcp14) criteria defined by FOCUS. In addition to FOCUS columns, the dataset should include custom provider columns (prefixed with `x_`) when these columns provide additional information not captured by the existing FOCUS columns. If introducing a custom column could result in splitting original charge records into multiple entries, the data generator is responsible for ensuring that the FOCUS dataset fully conforms to all aggregation-related requirements for metric columns, particularly those concerning costs and quantities.
```

specification/attributes/column_handling.md
shawnalpay: @UdamDewaraja Discussed in TF1 Mar 25.  Weren't sure the path forward.  Do we need to make a specific change here, and if so, what?  Would be ideal to resolve before the Members meeting tomorrow.

specification/attributes/column_handling.md
rileyjenk: Another NIT, this datasets is not just for cost....cost and usage.

shawnalpay: Discussed in TF1 Mar 25.  Will leave this here with no action other than to apply @rileyjenk's nit suggestion in the most recent commit.

specification/attributes/column_naming_and_ordering.md
```diff
@@ -1,36 +1,51 @@
-# Column Naming and Ordering
+# Column Handling
 
-Column IDs provided in cost data following a consistent naming and ordering convention reduce friction for FinOps practitioners who consume the data for analysis, reporting, and other use cases.
+A [*FOCUS dataset*](#glossary:FOCUS-dataset) consists of a set of columns that convey information about the charges incurred with a provider. Each column describes an aspect of the charge, including but not limited to:
```
shawnalpay, line 3: Per @rileyjenk:
```suggestion
A [*FOCUS dataset*](#glossary:FOCUS-dataset) consists of a set of columns that convey information about the cost and usage incurred with a provider. Each column describes an aspect of the charge, including but not limited to:
```

shawnalpay: Discussed further in TF1 Mar 25.  Decided to leave as charge, and need to discuss the term for a row in a FOCUS dataset (row, line item, charge, charge record, something else).

shawnalpay:
@rileyjenk, @aws-letianfeng, and @thecloudman: your re-review has been requested on this PR two days ago, per changes @flanakin has performed.  Most notably, all requirements around non-FOCUS columns have been removed, to be discussed in a future release.  The remaining content is more procedural in nature.

Please review and either approve or request further changes; otherwise, your stale reviews will be dismissed in 48 hours.

thecloudman (approved):

aws-letianfeng (approved):

ijurica (approved):

specification/attributes/column_naming_and_ordering.md
```diff
@@ -24,13 +24,15 @@ Naming and ordering convention for columns appearing in a [*FOCUS dataset*](#glo
...
-* Columns that have an ID and a Name MUST have the `Id` or `Name` suffix in the Column ID. Display Name for a Column MAY avoid the `Name` suffix if it is considered superfluous.
-* Columns with the `Category` suffix MUST be normalized.
-* Custom (e.g., provider-defined) columns SHOULD follow the same rules listed above for FOCUS columns.
+  * Columns that have an ID and a Name MUST have the `Id` or `Name` suffix in the Column ID.
+  * Column display names MAY avoid the `Name` suffix if there are no other columns with the same name prefix.
+  * Columns with the `Category` suffix MUST be normalized.
+* Custom (e.g., provider-defined) columns that are not defined by FOCUS but included in a *FOCUS dataset* MUST follow the following rules:
+  * Custom columns MUST be prefixed with a consistent `x_` prefix to identify them as external, custom columns and distinguish them from FOCUS columns to avoid conflicts in future releases.
+  * Custom columns MUST be included for all unique concepts not defined by or that map to FOCUS columns when the provider publishes a non-FOCUS cost and usage dataset.
```
sarahcloudpm: Custom columns should not be MUST but SHOULD provided to avoid the implementation of the spec from being diluted/larger than it should for the intended audience.

For example should this include all sub resource information (e.g. Kubernetes pod level costs as well?). This is why we have chosen to have a standard and a detailed billing export -- not everyone wants everything in one export. Additionally this also increases costs for customers to manage and store this data that we should be mindful of.

aws-letianfeng: Is this the latest version of the diff? I thought this clause had been deferred for later discussions.

ijurica: @aws-letianfeng you're right - that part is excluded from [the most recent version](https://github.com/FinOps-Open-Cost-and-Usage-Spec/FOCUS_Spec/pull/838/files), i.e., it was deferred for later discussions

specification/attributes/column_naming_and_ordering.md
```diff
@@ -1,36 +1,51 @@
...
+
+While FOCUS establishes the core structure and standardizes columns for consistent reporting of cost and usage data, the diverse and evolving landscape of providers and service offerings may require providers and data generators to include supplemental columns in the FOCUS dataset. These additional columns may enable deeper analysis and provide more detailed descriptions of usage that may not be fully captured by standard FOCUS dataset columns.
+
+In such cases, providers and data generators are responsible for ensuring that their usage and cost data is accurately and comprehensively represented by including necessary supplemental columns without duplicating data in FOCUS columns. Rows in a FOCUS dataset may be aggregated or split differently than non-FOCUS datasets to align with FOCUS requirements (e.g., Discount Handling), while enriching the dataset, providers and data generators must maintain the integrity of FOCUS-defined dimensions and metrics. Providers and data generators must ensure the accuracy of all dimensions and metrics, particularly summable values such as costs and quantities, when performing these transformations.
```
jpickardgreen, line 14:
```suggestion
In such cases, providers and data generators are responsible for ensuring that their usage and cost data is accurately and comprehensively represented by including necessary supplemental columns without duplicating data in FOCUS columns. Rows in a FOCUS dataset may be aggregated or split differently than non-FOCUS datasets to align with FOCUS requirements (e.g. Discount Handling), while enriching the dataset, providers and data generators must maintain the integrity of FOCUS-defined dimensions and metrics. When performing these transformations, providers and data generators must ensure the accuracy of all dimensions and metrics, particularly summable values such as costs and quantities.
```

specification/attributes/column_naming_and_ordering.md
```diff
@@ -1,36 +1,51 @@
...
+
+Columns within FOCUS include an ID and a display name. Column IDs are used in files and database tables and display names can be used in report output and other descriptive content, like documentation. Column IDs provided in a *FOCUS dataset* follow consistent naming and ordering conventions to reduce friction for FinOps practitioners who consume the data for analysis, reporting, and other use cases.
```
jpickardgreen, line 16: Do we need to mention the 'reducing friction for FinOps practitioners piece? Could we change it to ' Column IDs provided in a *FOCUS dataset* MUST follow consistent naming and ordering conventions in order to streamline the ingestion of data for analysis, reporting and other use case.

specification/attributes/column_naming_and_ordering.md
```diff
@@ -1,36 +1,51 @@
...
   * Column IDs SHOULD NOT use acronyms.
   * Column IDs SHOULD NOT exceed 50 characters to accommodate column length restrictions of various data repositories.
-* All custom columns MUST be prefixed with a consistent `x_` prefix to identify them as external, custom columns and distinguish them from FOCUS columns to avoid conflicts in future releases.
-* Columns that have an ID and a Name MUST have the `Id` or `Name` suffix in the Column ID. Display Name for a Column MAY avoid the `Name` suffix if it is considered superfluous.
-* Columns with the `Category` suffix MUST be normalized.
-* Custom (e.g., provider-defined) columns SHOULD follow the same rules listed above for FOCUS columns.
+  * Columns that have an ID and a Name MUST have the `Id` or `Name` suffix in the Column ID.
+  * Column display names MAY avoid the `Name` suffix if there are no other columns with the same name prefix.
+  * Columns with the `Category` suffix MUST be normalized.
+* Custom (e.g., provider-defined) columns that are not defined by FOCUS but included in a *FOCUS dataset* MUST follow the following rules:
```
jpickardgreen, line 46:
```suggestion
* Custom (e.g. provider-defined) columns that are not defined by FOCUS but included in a *FOCUS dataset* MUST follow the following rules:
```

supporting_content/attributes/column_handling.md
```diff
@@ -9,14 +9,20 @@ Examples of column names found in provider datasets:
...
 
 * Discussed the need for naming convention to apply to FOCUS columns
 * Group wanted to get naming convention adopted broadly for provider columns as well, to simplify reporting and analysis by practitioners
 * Different conventions like camel case, snake case, and title case were considered
 * Discussion included different columns naming conventions used by databases
+* The following was a proposal to require providers to include columns in 1.2. This was removed to discuss further in a future release:
+  > When the provider publishes a non-FOCUS cost and usage dataset, the following applies:
+  > * Custom columns MUST be included for all information not covered by FOCUS columns that exists in the latest version of non-FOCUS cost and usage datasets.
+  > * Data generators SHOULD allow practitioners to select a subset of FOCUS or custom columns.
+  > * Data generators SHOULD provide conformance documentation that indicates whether the *FOCUS dataset* fully, partially, or does not conform to each requirement.
```
jpickardgreen, line 24: I'm struggling with wording for Row 24 so the data generator would need to provide documentation indicating that something is fully compliant/partially compliant or not compliant to each requirement is that the correct interpretation of what is being said. ?

jpickardgreen (commented):
Reviewed with some comments.

specification/attributes/column_handling.md
UdamDewaraja: columns and attributes sections in the spec are normative sections, meaning they contain requirements that are described with BCP 14 language. Then there are sections like 1. Introduction and 7. Appendix (should be there in a few more sections) that are not normative (they start with a *This section is non-normative.* statement). 

The content added here to the column handling attribute is not normative. It sounds like suggestions/requirements for providers to follow but these aren't enforced. It's just confusing to a reader who may miss that these don't have SHOULD/MUST/MUST NOT, but it just sounds like it's implying that.

Example:
>  Data generators are responsible for ensuring the accuracy of dimensions and metrics when aggregating or splitting rows, especially summable metrics like costs and quantities.

1. this isn't anything to do with column handling
2. this isn't enforced, it is merely a suggestion. To a normal reader, this might seem like a requirement.

My suggestion is to either:
1. remove the content
4. move it to the 1. Introduction section where we talk about what FOCUS is and the responsibilities of data producers etc.

Given where we're at, reducing 'nice to have' type content to get PRs down to required/necessary content would help get these in more easily.

If the group feels strongly that this is not misleading and adding ambiguity to the spec, I'm happy to go with the groups opinion.

specification/attributes/column_naming_and_ordering.md
shawnalpay: Agree with @jpickardgreen's suggestion, which I edited to add back the comma after e.g.

specification/attributes/column_naming_and_ordering.md
shawnalpay: Commas belong after e.g. (see [here](https://english.stackexchange.com/questions/16172/should-i-always-use-a-comma-after-e-g-or-i-e) for more).  Resolving.

specification/attributes/column_handling.md
flanakin: TL;DR: This adds positive value, is more discoverable here, and doesn't hurt the spec. We should leave it and resolve this thread.

The full story...

The first section in attributes and columns is there to describe and, to a certain degree, justify the normative requirements. In general, the intro provides additional context when normative requirements aren't clear enough. In this case, we don't have a normative section covering the entire spec, but we were using this attribute to cover inclusion of data, so the TF felt it was supporting that requirement. While we removed the inclusion criteria for now, the content is still accurate, so we left it.

The TF discussed placement and agreed putting it into an attribute is the best place because of its proximity to normative requirements. Simply put, none of the other sections are required to be read. The logic behind this is:

- Many people will likely implement FOCUS without reading the spec at all.
- Hopefully, most will read the spec, but most will only look at the normative requirements and will not read it end-to-end.
- When the normative requirements aren't clear, they will look at the intro text.
- If still confused, they *may* look up very specific terms in the glossary if those terms aren't clear, but that will only be one-off and they're unlikely to look up "FOCUS dataset" because that is fairly self-explanatory. But in general, most people don't look up word definitions.
- Very, very few people will be curious enough or so confused they eventually discover the appendix. This is very unlikely, however.
- Almost nobody will look at the supporting content in the repo.

The TF did discuss the idea of potentially creating a new section for this and column inclusion but decided that was something editorial that could be moved later, so we agreed to keep it here for now and move it later if we have a better place.

All that said, unless the content is wrong, impedes on the goals of the spec, or has a more discoverable home, I would keep it here. As far as I can tell, it adds positive value and supports the ultimate goals of column handling. I don't get the sense that there is a negative to putting this here, so I would suggest keeping it and filing a work item to find a better place that aligns with normative requirements, based on the TF discussions.

jpradocueva: It was discussed in TF-2 on April 2nd. The group agreed this PR is ready for approval. Editorial changes suggested in this conversation can be discussed and applied during the consistency review

jpradocueva:
This PR was approved during the Members’ meeting on April 3rd.

PR #839, opened 2025-03-12
WI #616: `Host Provider`, with Refined Datageneration table, Refined 

rileyjenk:


specification/appendix/origination_of_cost_data.md
```diff
@@ -5,21 +5,28 @@ Cost data presented in the billing datasets originates from various sources depe
 * [Provider](#provider): The entity that made the [*resources*](#glossary:resource) or [*services*](#glossary:service) available for purchase.
 * [Publisher](#publisher): The entity that produced the *resources* or *services* that were purchased.
 * [Invoice Issuer](#invoiceissuer): The entity responsible for invoicing for the *resources* or *services* consumed.
+* [Host Provider](#hostprovider): The entity whose *resources* or *services* are used by the Service Provider to make their *resources* or *services* available.
```
GrinsteadJ, line 8: This definition refers to "service provider" but this version of the PR has dropped Service Provider as an entity definition.

specification/columns/hostprovider.md
```diff
@@ -0,0 +1,38 @@
+# Host Provider
+
+A Host Provider is an entity whose [*resources*](#glossary:resource) are used by the Service Provider to make their [*resources*](#glossary:resource) or [*services*](#glossary:service) available.
+
+The HostProvider column adheres to the following requirements:
+
+* The HostProvider column MUST be present in a [*FOCUS dataset*](#glossary:FOCUS-dataset) and represents the service provider where [*resources*](#glossary:resource) or [*services*](#glossary:service) are deployed on.
```
GrinsteadJ, line 7: Moving my comment over from #774 

I think MUST is too strict for this column and will be redundant or NULL in a lot of datasets. It's likely better to take the approach that #799  took with its new columns. In these cases, the new columns woul only be required in certain cases.

* HostProvider presence in a [*FOCUS dataset*](#glossary:FOCUS-dataset) is defined as follows:
  * HostProvider MUST be present in a FOCUS dataset when the provider allows the customer to know or select the host provider.
   * HostProvider MAY be present in a FOCUS dataset in all other cases.

specification/appendix/origination_of_cost_data.md
rileyjenk: Unintentional, fixed.

specification/columns/hostprovider.md
rileyjenk: I do not agree because the column is answering a distinct question, if we allow it to not be included, the user has to conditionally use the column. While the value may be redundant the question it is answering is specific and distinct.  The user would have to derive the host provider logically rather than have it directly in the data.

specification/columns/invoiceid.md
```diff
@@ -1,6 +1,6 @@
 # Invoice ID
 
-An Invoice ID is a provider-assigned identifier for an invoice encapsulating some or all charges in the corresponding [*billing period*](#glossary:billing-period) for a given [*billing account*](#glossary:billing-account). Invoices are commonly used for scenarios like tracking billing transactions, facilitating payment processes and for performing invoice reconciliation between charges and billing periods.
+An Invoice ID is a invoice Issuer assigned identifier for an invoice encapsulating some or all charges in the corresponding [*billing period*](#glossary:billing-period) for a given [*billing account*](#glossary:billing-account). Invoices are commonly used for scenarios like tracking billing transactions, facilitating payment processes and for performing invoice reconciliation between charges and billing periods.
```
ljadvey, line 3:
```suggestion
An Invoice ID is an Invoice Issuer assigned identifier for an invoice encapsulating some or all charges in the corresponding [*billing period*](#glossary:billing-period) for a given [*billing account*](#glossary:billing-account). Invoices are commonly used for scenarios like tracking billing transactions, facilitating payment processes and for performing invoice reconciliation between charges and billing periods.
```

specification/columns/hostprovider.md
```diff
@@ -0,0 +1,38 @@
+# Host Provider
+
+A Host Provider is an entity whose [*resources*](#glossary:resource) are used by the Service Provider to make their [*resources*](#glossary:resource) or [*services*](#glossary:service) available.
```
GrinsteadJ, line 3: There are a few references to Service Provider which I think are from an earlier version of this PR that recommended a broader definition change. I'll flag the ones I've not.

specification/columns/hostprovider.md
```diff
@@ -0,0 +1,38 @@
...
+
+The HostProvider column adheres to the following requirements:
+
+* The HostProvider column MUST be present in a [*FOCUS dataset*](#glossary:FOCUS-dataset) and represents the service provider where [*resources*](#glossary:resource) or [*services*](#glossary:service) are deployed on.
+* For Service Providers that allow their customer to select or know the host provider where resources or services are deployed, this SHOULD be the name of that selected or visible host provider
+* For Service Providers that DO NOT allow their customer to select or know the host provider where resources or services are deployed, this SHOULD be the name of the [Service Provider](#serviceprovider)
```
GrinsteadJ, line 9: I think Service Provider is supposed to be Provider in all of these rows.

specification/columns/hostprovider.md
```diff
@@ -0,0 +1,38 @@
...
+
+HostProvider
+
+## Display Name
+
+Host Provider
+
+## Description
+
+The name of the entity whose *resources* are used by the Service Provider to make their [*resources*](#glossary:resource) or [*services*](#glossary:service) available.
```
GrinsteadJ, line 24: Service Provider should be Provider?

specification/columns/invoiceid.md
```diff
@@ -26,7 +26,7 @@ Invoice Id
 
 ## Description
 
-The provider-assigned identifier for an invoice encapsulating some or all charges in the corresponding billing period for a given billing account.
+The service provider-assigned identifier for an invoice encapsulating some or all charges in the corresponding billing period for a given billing account.
```
GrinsteadJ, line 29: Service Provider should be Provider?

specification/columns/hostprovider.md
GrinsteadJ: If the field is left null by a large number of publishers, won't users be using the column conditionally anyway? If publishers don't want to or can't reveal this information then we are requiring them to support a pointless (for their data) field in their API response in order to be FOCUS compliant?

rileyjenk: If they don't elect to or don't want to it is not null but is the name of the provider. The null is only relevant when the line item is not hosted (like a purchase record for atlas credits, these aren't specific to a host provider).

specification/columns/hostprovider.md
rileyjenk: I will do another pass and clean them up...the git surgery was not entirely clean.

rileyjenk: Done

specification/columns/hostprovider.md
GrinsteadJ: Right, but I think the issues remain.

You stated "While the value may be redundant the question it is answering is specific and distinct. The user would have to derive the host provider logically rather than have it directly in the data." But if we force Data Generators to fill in this column, then users still have to apply some logic to figure out the meaning of the column. Is the Host Provider the actual Host Provider? Or is just the Provider because the Host Provider isn't known? How do users figure out the answer to this?

Additionally, for a large number (the majority?) of Data Generators, the Provider and Host Provider will be identical. Either because the Provider is the actual Host Provider or because we've made the Data Generator indicate it as such. Are all currently compliant FOCUS datasets going to become non-compliant because they don't have the redundant column?

specification/columns/hostprovider.md
```diff
@@ -0,0 +1,38 @@
+# Host Provider
+
+A Host Provider is an entity whose [*resources*](#glossary:resource) are used by the Provider to make their [*resources*](#glossary:resource) or [*services*](#glossary:service) available.
+
+The HostProvider column adheres to the following requirements:
+
```
GrinsteadJ, line 6: One final "service provider" in line 7.

specification/columns/hostprovider.md
```diff
@@ -0,0 +1,38 @@
+# Host Provider
```
flanakin, line 1: What other names have we explored? Reusing "provider" will lead to confusion. I am not in favor of this.

specification/columns/hostprovider.md
```diff
@@ -0,0 +1,38 @@
...
+
+A Host Provider is an entity whose [*resources*](#glossary:resource) are used by the Provider to make their [*resources*](#glossary:resource) or [*services*](#glossary:service) available.
+
+The HostProvider column adheres to the following requirements:
+
+* The HostProvider column MUST be present in a [*FOCUS dataset*](#glossary:FOCUS-dataset) and represents the service provider where [*resources*](#glossary:resource) or [*services*](#glossary:service) are deployed on.
+* For Providers that allow their customer to select or know the host provider where resources or services are deployed, this SHOULD be the name of that selected or visible host provider
```
flanakin, line 8: nit: Do we use the term "customer"? I thought we use "practitioner"? Also applies on other lines.

specification/appendix/origination_of_cost_data.md
```diff
@@ -5,21 +5,28 @@ Cost data presented in the billing datasets originates from various sources depe
 * [Provider](#provider): The entity that made the [*resources*](#glossary:resource) or [*services*](#glossary:service) available for purchase.
 * [Publisher](#publisher): The entity that produced the *resources* or *services* that were purchased.
 * [Invoice Issuer](#invoiceissuer): The entity responsible for invoicing for the *resources* or *services* consumed.
+* [Host Provider](#hostprovider): The entity whose *resources* or *services* are used by the Provider to make their *resources* or *services* available.
+* [Data Generator](#datagenerator): The entity that generated the FOCUS dataset.
```
flanakin, line 9: What happens when someone is a proxy for the data? Does that count as a data generator? What if they tweak data from another generator? What if they only tweak one row?

specification/appendix/origination_of_cost_data.md
```diff
@@ -5,21 +5,28 @@ Cost data presented in the billing datasets originates from various sources depe
...
+| 2.3 | Purchasing labor services from managed service provider                                                                                                                                                                                           | Managed Service Provider     | Managed Service Provider     | Managed Service Provider                                   | N/A                                                        | Managed Service Provider                                   |
+| 3.1 | Purchasing a cloud marketplace offering that runs on the cloud provider                                                                                                                                                                           | Marketplace seller           | Marketplace seller           | Cloud service provider                                     | Cloud service provider                                     | Cloud service provider                                     |
```
flanakin, line 21: This feels wrong. The cloud provider is making the service available by providing a marketplace. I get the publisher adds it to the marketplace but they wouldn't be able to do that if the marketplace wasn't provided in the first place. (At least not in this dataset.)

specification/appendix/origination_of_cost_data.md
```diff
@@ -5,21 +5,28 @@ Cost data presented in the billing datasets originates from various sources depe
 * [Provider](#provider): The entity that made the [*resources*](#glossary:resource) or [*services*](#glossary:service) available for purchase.
 * [Publisher](#publisher): The entity that produced the *resources* or *services* that were purchased.
 * [Invoice Issuer](#invoiceissuer): The entity responsible for invoicing for the *resources* or *services* consumed.
+* [Host Provider](#hostprovider): The entity whose *resources* or *services* are used by the Provider to make their *resources* or *services* available.
```
flanakin, line 8: Unless I'm missing something, every example shows Provider = Publisher. It seems like we're trying to make Provider be what Publisher is today and create a new HostProvider that serves the purpose of what Provider does today. While I understand the ambiguity of "provider", it doesn't seem worth removing value from what we've already defined just to add confusion with an overlapping term. Provider seems to be adequate, despite the historical ambiguity.

flanakin (changes requested):
This seems to be a half-implememted attempt to rename Publisher to Provider and Provider to HostProvider, leaving Publisher as is, making it needlessly duplicated. I don't see what value this adds, but it will require people to change how they use Provider today. This feels like it's not ready. I would rather see this get baked more and come in 1.3 when it's been fully thought through.

specification/appendix/origination_of_cost_data.md
GrinsteadJ: I don't think Data Generator should be included in this table, since the Scenario doesn't indicate where the user is accessing the data.

E.g. for 3.1 FOCUS data could be generated by the Marketplace Seller and the Cloud Service Provider. The data that is generated would be different, but there could still be two different FOCUS datasets. For example, the Marketplace Seller's dataset would have billed cost = 0, and the CSP's dataset would have a billed cost but would likely be less granular than the dataset generated by the Marketplace Seller.

shawnalpay:
> This feels like it's not ready. I would rather see this get baked more and come in 1.3 when it's been fully thought through.

@flanakin 
We will bring your conern to TF3 on Friday to discuss.  Something to consider: we already know that multiple providers will need a column containing this information as part of a FOCUS implementation.  If we don't define it centrally, they will each individually stand it up as a custom column, with varying names and normative requirements.  Thus, I would argue that postponing this column may introduce more confusion than it alleviates.

specification/appendix/origination_of_cost_data.md
rileyjenk: Yes that is true because provider and publisher definitions are not specific enough. The original PR was to deprecate publisher because it is not specific enough.

specification/columns/hostprovider.md
rileyjenk: this is EXACTLY! why the original PR was a rename and a deprecation because "provider" with a column called Host Provider seems to indicate hierarchy...which is not what we are trying to imply. The definition of publisher and provider are so close an non specific it leads to confusion.

specification/appendix/origination_of_cost_data.md
rileyjenk: Even so whoever delivers the data is the DataGenerator.

specification/appendix/origination_of_cost_data.md
rileyjenk: No I don't agree. Azure doesn't make Datadog sell anything in their marketplace, its the other way around. Also who is the customer contracting with, Datadog, through the Azure marketplace. 

@GrinsteadJ the name of the section is Origination of Cost Data. It needs to be included somewhere, do you have a suggestion where it should be put?

specification/columns/hostprovider.md
rileyjenk: Here is the core of what I am saying.

A column with two different and distinct meanings is problematic because the minute the two questions have a different answer you are stuck. Thus why this column is being added. 

If we a make this column optional, it means that practitioners will be faced with the problem above because for some FOCUS datasets they will have one column that answers both questions and for others they will have both columns.  This complicates querying and analysis because you will have to do go to different places to answer the  "where is this hosted?" question.  With a distinct column that has a single meaning and that column is always  present the user simply uses that single column to know where it is hosted. If it is the same as the provider column, they then have the data to know if they are the same.  With one column they do not have that information, they are just missing a column. 

If we make this optional, then would have to change the definition of provider to handle both definitions....this is problematic.

specification/appendix/origination_of_cost_data.md
GrinsteadJ: @rileyjenk if the Data Generator columns remains perhaps we indicate the options for what the Data Generator could be. E.g. for 3.1 it could be "Marketplace Seller or Cloud Service Provider"

specification/appendix/origination_of_cost_data.md
flanakin: Doesn't that mean that ever vendor tool would be the data generator for what their customers see? Especially if they manipulate data in any way (e.g., adding metadata or cost allocation).

If no data is changing, that feels wrong. If one row changes, then should all data be flagged as "generated" by the vendor? Even if one column is added, it seems like a sledgehammer approach to say the vendor "generated" all other columns.

I'm thinking of the data generator as "where do I file a support request when something looks wrong." In all the examples I mentioned, we wouldn't want it to be on the vendor if the problem isn't in something they touched.

specification/appendix/origination_of_cost_data.md
flanakin: @rileyjenk I acknowledged that. Maybe I'm thinking about it more as the org that directly provides the service to the customer. I totally expect this to be the SaaS company when working directly with them, but I can't see this being anything but the cloud company when something's purchased thru the cloud marketplace because it's that company that allowed the customer to purchase it.

specification/columns/hostprovider.md
flanakin: Fwiw, I'm not even sure why we need host provider. But if we're adding it, we should explore all possible options to avoid adding confusion. Either that or wait to make a complete change.

shawnalpay:
As discussed in Mar 21 TF3 and Mar 24 Maintainers, we have agreed to postpone the inclusion of `Host Provider` to 1.3, as it does ultimately require a refactor of some combination of `Provider`, `Publisher`, and `Invoice Issuer` as initially proposed by @rileyjenk in PR #834.  

I am removing this PR from milestone `v1.2`.

shawnalpay:
Putting this PR back into draft until it has either been cleaned up or replaced, per AI #1122.

shawnalpay:
Closing this PR in favor of a refactored #1139 as authored by @rileyjenk.

Issue #854, opened 2025-03-25
[Editorial] 1.2 Consistency Review


shawnalpay:
A list of things we need to tackle as part of the 1.2 consistency review.

shawnalpay:
Need to clean up dashes vs underscores for the guidelines files.

https://github.com/FinOps-Open-Cost-and-Usage-Spec/FOCUS_Spec/tree/working_draft/guidelines

shawnalpay:
Need to create CSV section in repo for examples in these PRs:

- https://github.com/FinOps-Open-Cost-and-Usage-Spec/FOCUS_Spec/pull/830
- #535

shawnalpay:
Consider renaming this file and putting it in the `guidelines` folder.

https://github.com/FinOps-Open-Cost-and-Usage-Spec/FOCUS_Spec/blob/working_draft/spec-development-workflow.md

And/or perhaps overhauling/replacing this file with the "way we work" document as being developed by @mike-finopsorg per AI #814.

ijurica:
**Candidate Action Items for FOCUS  1.2 consistency review:**

- [[AI] Discuss List Cost for Unused Usage per Simple & Spend SaaS Agreements](https://github.com/FinOps-Open-Cost-and-Usage-Spec/FOCUS_Spec/issues/857)

- Review newly added columns and those updated in recent PRs against the Column Normative Requirements and apply adjustments as needed. Changes should be limited to non-material updates.

- Review the spec and apply Editorial Guidelines (e.g., adding missing links to glossary terms, etc.).
  Clarify what is meant by a "section" in the guidelines - does it refer to an entire column specification or to a specific part of it (e.g., column intro section, normative requirements section, description section, etc.)?
  - The first occurrence of a term within a section should be linked to that section.
  - The first occurrence of a term within a section should be linked to the glossary.

shawnalpay:
Per @UdamDewaraja [here](https://github.com/FinOps-Open-Cost-and-Usage-Spec/FOCUS_Spec/pull/848#discussion_r2024220629): do a pass on Contracted Unit Price to check for proper placement of asterisks.

shawnalpay:
As discussed [here](https://github.com/FinOps-Open-Cost-and-Usage-Spec/FOCUS_Spec/pull/838#discussion_r1992498329): determine whether some of the content in the Column Handling attribute belongs elsewhere (e.g., the intro).

[...]

PR #117, opened 2023-06-15
#92 - BillingCurrency attribute

SanjnaSrivatsaVMware:
Initial commit for billing currency

PR #891, opened 2025-04-10
WI #854: Billing currency normative requirement change

thecloudman:
Change billing Currency normative requirement of currency matching invoice currency from MUST to SHOULD as AWS billing currency is USD only in FOCUS and CUR exports but invoices may be in any other currency of the customers choosing. The Description also insists that the column represents the currency the charge was billed in and therefore we shouldn't be enforcing the requirement to match currency in the dataset and the invoice.


shawnalpay:
@thecloudman The use of MUST in this normative requirement goes all the way back to the 0.5 release (see PR https://github.com/FinOps-Open-Cost-and-Usage-Spec/FOCUS_Spec/pull/117).  I do respect your desire to discuss this further, and we can certainly do so for a future release.  That said, we cannot sustain the change of a normative requirement during consistency review, especially for a requirement that has existed unchanged in the spec for a significant amount of time.

I recommend that you author a Work Item that proposes this change, and we can consider that work for 1.3.

thecloudman:
@shawnalpay I will add it as an item for 1.3 but isn't a consistency review to determine consistency in wording, structure, spelling etc. While we can agree this has been here for some time, it isn't formed properly as providers are not exactly meeting this correctly, does this get called out anywhere if they are not meeting it correctly?

shawnalpay:
@thecloudman 
To your point, FOCUS does currently lack the formal definition of a conformance program; some providers generate their own conformance gap report, and others do not.  That conformance program will eventually come together, but until then, the onus is on the provider to self-certify.  

I can also say that the Steering Committee is working on a Mission / Vision / Goals document that will, amongst other things, take a stance on to what extent the specification should run idealistically ahead of how providers provide their data today.

I will remove the 1.2 milestone from this PR, and we can consider in a future release, per the work item you shall create.

shawnalpay:
Discussed in Jul 28 Maintainers.  Agreed to close until we decide to consider this in a future release, or an active contributor in the working group is willing to advocate for its inclusion.

Issue #902, opened 2025-04-16
[AI] Revise and align SKU and SKU Price related columns post-merge


ijurica:
### Detailed Description

Revise SKU and SKU Price-related columns to resolve any post-merge inconsistencies.

PR #905, opened 2025-04-16
AI #902: Revise and align SKU and SKU Price related columns post-merge

ijurica:
This PR addressed any post-merge inconsistencies in SKU and SKU Price-related columns.


shawnalpay (changes requested):
@ijurica 
Of the three changes in this PR:

 * I agree with striking the duplicate listing of `Property value MUST represent the value for a single [PricingUnit](#pricingunit) when the property holds a numeric value` -- that was an error, as we discussed earlier today, and that was the genesis for this PR.
 * ~~I'm on the fence on whether we should add the two normative requirements you propose.  Please present your reasoning as to how this is a valid change during consistency review.~~
 * ~~I'm opposed to the revision of the the column description.  You've changed it to the extent that that should have happened during spec development, as it requires gathering consensus -- and it feels out of bounds for changing during consistency review.  I'm certainly open to you proposing we revise that section in a future release, but it feels too late now.  Feel free to provide your reasoning if you feel otherwise.~~

EDIT: Never mind!  I see now that this came from #683, and a few items were dropped due to merge conflicts.  Reversing my review to approval!

shawnalpay (approved):

specification/columns/skupricedetails.md
```diff
@@ -17,13 +17,14 @@ The SkuPriceDetails column adheres to the following requirements:
 * When SkuPriceDetails is not null, SkuPriceDetails adheres to the following additional requirements:
   * SkuPriceDetails MUST be associated with a given SkuPriceId.
   * SkuPriceDetails MUST NOT include properties that are not applicable to the corresponding SkuPriceId.
+  * SkuPriceDetails SHOULD include all FOCUS-defined SKU Price properties listed below that are applicable to the corresponding SkuPriceId.
+  * SkuPriceDetails MUST include the FOCUS-defined SKU Price property when an equivalent property is included as a Provider-defined property.
```
ijurica, line 21: @kk09v @flanakin  @shawnalpay - not sure the original wording clearly conveys the intended behavior.  It might be worth replacing it with something like this to clarify that FOCUS-defined should take precedence over provider-defined properties.
Thoughts? Alternatives?

```suggestion
  * SkuPriceDetails SHOULD include all applicable Provider-defined SKU Price properties for the corresponding SkuPriceId when an equivalent FOCUS-defined property is not included.
  * SkuPriceDetails MUST use FOCUS-defined SKU Price properties in place of any equivalent Provider-defined property.
```

shawnalpay: I don't think consistency review is the time to try rewording normative requirements...

ijurica: Moved this suggestion to a separate PR (#910 ) since it might fall outside the consistency review scope.
Resolving this conversation.

ijurica:
Additional suggestions in this PR are **limited to editorial adjustments** (best effort), including adding links to glossary terms, etc., and aligning formatting with editorial guidelines. 

For SkuPriceDetails, these changes are addressed here instead of in a dedicated PR to **minimize merge conflicts**.

jpradocueva:
Approved by the Members’ during April 17 call.

jpradocueva (approved):

Issue #863, opened 2025-04-03
[AI] Create CSV section for examples


shawnalpay:
### Detailed Description

Need to create CSV section in repo for examples in these PRs:

- https://github.com/FinOps-Open-Cost-and-Usage-Spec/FOCUS_Spec/pull/830
- #535 

This would also apply to the CD appendix we introduced in 1.1.

shawnalpay:
Blocked by #857.  Once that is complete, I can proceed.

PR #896, opened 2025-04-13
AI #863: Convert appendix examples to CSV: Commitment Discounts

shawnalpay:
Convert appendix examples from JSON to CSV for the following sections:

 * Examples: Commitment Discount Scenarios
 * Examples: Commitment Discount Flexibility

SaaS examples are tackled in PR #908.

We will leave the Metadata section in JSON.

PR #908, opened 2025-04-16
AI #863: Convert appendix examples to CSV: SaaS

UdamDewaraja:


shawnalpay (approved):
While I'm a bit concerned about the order of the columns (they are _somewhat_ in alpha order, but not totally; as a spec reader, I would expect either 1) the most important columns to be at the beginning, or 2) full alpha sort) -- **I'm willing to accept these the way they are.**  (That's certainly also true with the examples I added in #896.) My main priority here has been for us to get all the examples in; applying consistency feels like a nice-to-have that we can go back and scrub in the future.

@UdamDewaraja (and others): agree/disagree?

ljadvey:
@shawnalpay This likely needs a larger conversation BUT for simplicity sake - alpha ordering.

I'd be happy to discuss what is more important vs. others on a call in the future.

UdamDewaraja:
> While I'm a bit concerned about the order of the columns (they are _somewhat_ in alpha order, but not totally; as a spec reader, I would expect either 1) the most important columns to be at the beginning, or 2) full alpha sort) -- **I'm willing to accept these the way they are.** (That's certainly also true with the examples I added in #896.) My main priority here has been for us to get all the examples in; applying consistency feels like a nice-to-have that we can go back and scrub in the future.
> 
> @UdamDewaraja (and others): agree/disagree?

I agree that first pref when reading data is not alpha order but instead at least move some of the related things like cost metrics next to each other as theyre often compared. However, there were some votes for alpha when we talked about this last. If we ever do end up 'suggesting' an order, it would probably need to allow alpha sort in addition to what we suggest.

For this, I can just fix it to be alpha for now which is the simplest approach though it sucks when billed cost is all the way at the start and end dates are before start dates (to name a few glaring readability issues).

shawnalpay:
> For this, I can just fix it to be alpha for now which is the simplest approach though it sucks when billed cost is all the way at the start and end dates are before start dates (to name a few glaring readability issues).

Cool.  FWIW, like I said, I'm totally fine with leaving it as-is and addressing it in the future (not least because, as I said, I didn't alpha sort the examples in #896, and we're now low on time).  But if you want to revise them, I'm in support!

UdamDewaraja:
> > For this, I can just fix it to be alpha for now which is the simplest approach though it sucks when billed cost is all the way at the start and end dates are before start dates (to name a few glaring readability issues).
> 
> Cool. FWIW, like I said, I'm totally fine with leaving it as-is and addressing it in the future (not least because, as I said, I didn't alpha sort the examples in #896, and we're now low on time). But if you want to revise them, I'm in support!

I just checked and its all sorted as far as excel sorting can validate (unless im messing something up). Do you have an example of a case that's not sorted?

shawnalpay:
> I just checked and its all sorted as far as excel sorting can validate (unless im messing something up). Do you have an example of a case that's not sorted?

I retract my statement.  They all look great.  My apologies.

jpradocueva:
Approved by the Members’ during April 17 call.

PR #910, opened 2025-04-16
AI #909: SkuPriceDetails req. fix -  prefer FOCUS-defined over Provider-defined properties

### LINKED ISSUE #909: [AI] Clarify inclusion logic for SkuPriceDetails: prefer FOCUS-defined over Provider-defined
### Detailed Description

One of the current requirements for SkuPriceDetails does not clearly define the precedence between FOCUS-defined and Provider-defined properties. This could lead to ambiguity or duplication when both exist. 

Clarify that FOCUS-defined properties should be preferred and used instead of equivalent Provider-defined ones.

### LINKED ISSUE #1024: [FR] Clarify guidance on including provider SKU price properties
###  Problem Statement
The following SkuPriceDetails normative requirement does not clearly define the precedence between FOCUS-defined and Provider-defined SKU Price properties:

> SkuPriceDetails MUST include the FOCUS-defined SKU Price property when an equivalent property is included as a Provider-defined property.

The current wording implies that both the FOCUS-defined and Provider-defined properties must be included when an equivalent exists, rather than indicating a preference for the FOCUS-defined property over the Provider-defined one.

Additionally, the following requirement only covers FOCUS-defined properties and does not address Provider-defined properties or specify that all applicable properties (not just FOCUS-defined) should be included:

> SkuPriceDetails SHOULD include all FOCUS-defined SKU Price properties listed below that are applicable to the corresponding SkuPriceId.

---
###  Use Case
- Related PR #910

---
###  Desired Outcome / Practitioner Impact
When generating SkuPriceDetails, providers need clear guidance on which properties must or should be included, and whether to retain Provider-defined properties alongside the equivalent FOCUS-defined properties, or include only the FOCUS-defined ones when equivalent properties exist.

---
###  Type of Request
- [x] Refinement of an existing FOCUS attribute

---
### ️Organizations Requesting or Supporting This Feature

- Neos
- Microsoft

---
###  FinOps Scope Alignment (Optional)

---
###  Impacted Parties

---
###  Level of Ambiguity
How clearly defined is the request? Rate from 1 to 5:
- 1 = very well-defined, low complexity
- 3 = moderately scoped, some ambiguity
- 5 = vague, high complexity or conceptual

1

---
###  Supporting Documentation
The SKU Price Details normative requirements should clearly specify that:

- All applicable SKU Price properties (not just FOCUS-defined) should be included.
- FOCUS-defined properties are preferred over equivalent Provider-defined properties, and duplication must (or should) be avoided.

---
###  Proposed Solution / Approach
The SKU Price Details normative requirements should clearly specify that:

- All applicable SKU Price properties (not just FOCUS-defined) should be included.
- FOCUS-defined properties are preferred over equivalent Provider-defined properties, and duplication must (or should) be avoided.

#### Subset of existing requirements relevant to this discussion:

***Note:** Two highlighted requirements are proposed for revision.*
 
> * When SkuPriceDetails is not null, SkuPriceDetails adheres to the following additional requirements:
>   * SkuPriceDetails MUST be associated with a given SkuPriceId.
>   * SkuPriceDetails MUST NOT include properties that are not applicable to the corresponding SkuPriceId.
>   * **SkuPriceDetails SHOULD include all FOCUS-defined SKU Price properties listed below that are applicable to the corresponding SkuPriceId.**
>   * **SkuPriceDetails MUST include the FOCUS-defined SKU Price property when an equivalent property is included as a Provider-defined property.**
>   * SkuPriceDetails MAY include properties that are already captured in other dedicated columns.

#### Proposed revised version:

***Note:** Two revised requirements are highlighted.*

> * When SkuPriceDetails is not null, SkuPriceDetails adheres to the following additional requirements:
>   * SkuPriceDetails MUST be associated with a given SkuPriceId.
>   * SkuPriceDetails MUST NOT include properties that are not applicable to the corresponding SkuPriceId.
>   * **SkuPriceDetails SHOULD include all applicable SKU Price properties for the corresponding SkuPriceId.**
>   * **SkuPriceDetails MUST include FOCUS-defined SKU Price properties instead of any equivalent Provider-defined properties.**
>   * SkuPriceDetails MAY include properties that are already captured in other dedicated columns.

#### What these changes achieve:

- `"SkuPriceDetails SHOULD include all applicable SKU Price properties"` — This requirement covers all relevant properties, regardless of whether they are FOCUS-defined or Provider-defined.

- `"SkuPriceDetails MUST include FOCUS-defined SKU Price properties instead of any equivalent Provider-defined properties."` — This makes it explicit that, when an equivalent FOCUS-defined property exists, it must be included instead of the Provider-defined one. This avoids duplication and establishes FOCUS-defined properties as the preferred.

---
###  Community Support (Add Your Voice)

  - Your **organization**

ijurica:
The following requirement does not clearly define the precedence between FOCUS-defined and Provider-defined properties. This could lead to ambiguity or duplication when both exist.

> * SkuPriceDetails MUST include the FOCUS-defined SKU Price property when an equivalent property is included as a Provider-defined property.

The goal of this PR is to clarify that FOCUS-defined properties should be preferred and used instead of equivalent Provider-defined ones.


ljadvey (approved):

shawnalpay:
I would appreciate some context as to why this might be presented as a bugfix vs a material and/or net-new change to normative requirements.  Our remit for revisions during a consistency review is narrow, and this is decidedly not a time for making changes that might require the gathering of consensus.

ijurica:
@shawnalpay @kk09v @flanakin 

This PR was created against the working_draft branch instead of the 902-sku-and-skuprice-review branch, where we are addressing post-merge conflicts. That is why the issue isn't immediately obvious.

That said, here is a subset of the requirements from 902-sku-and-skuprice-review, including the requirement in question.

> * When SkuPriceDetails is not null, SkuPriceDetails adheres to the following additional requirements:
>   * SkuPriceDetails MUST be associated with a given SkuPriceId.
>   * SkuPriceDetails MUST NOT include properties that are not applicable to the corresponding SkuPriceId.
>   * SkuPriceDetails SHOULD include all FOCUS-defined SKU Price properties listed below that are applicable to the corresponding SkuPriceId.
>   * **SkuPriceDetails MUST include the FOCUS-defined SKU Price property when an equivalent property is included as a Provider-defined property.**
>   * SkuPriceDetails MAY include properties that are already captured in other dedicated columns.

### BUG :
This requirement:

> * SkuPriceDetails MUST include the FOCUS-defined SKU Price property when an equivalent property is included as a Provider-defined property.

suggests that if a Provider-defined property is present, the equivalent FOCUS-defined property must also be included. It does not explicitly say that the Provider-defined property should be excluded or replaced. 

Based on the current wording, the expected result would be **that both properties are included** - the original Provider-defined one and the equivalent FOCUS-defined one.
If the intent is to use only the FOCUS-defined property in such cases, the wording would need to be adjusted to reflect that more clearly.

### SUGGESTED FIX:

Replace that requirement with these two:

>   * SkuPriceDetails SHOULD include all applicable Provider-defined SKU Price properties for the corresponding SkuPriceId when an equivalent FOCUS-defined property is not included.
>   * SkuPriceDetails MUST use FOCUS-defined SKU Price properties in place of any equivalent Provider-defined property.

kk09v:
Here's how I read this:
- If you're already providing a property and a new FOCUS-defined property is introduced, you MUST include the new FOCUS-defined property
  - `Existing SkuPriceDetails properties SHOULD remain consistent over time.` indicates that they SHOULD NOT remove their property when they introduce the FOCUS-defined property.
- If, for business or other reasons, a provider decides to include their own property where there's an equivalent FOCUS-defined property, the provider must also include the FOCUS-defined property
  - E.g. for marketing reasons, a provider wants to include the maximum value for a property of burstable SKU instead of the sustained/baseline value, they would do that as a provider-defined property. They're welcome to do so, but they MUST also include the FOCUS-defined property which reflects the sustained/baseline value.

How would you interpret the second suggested clause in the context of the above?


> SkuPriceDetails SHOULD include all applicable Provider-defined SKU Price properties for the corresponding SkuPriceId when an equivalent FOCUS-defined property is not included.

I don't love this clause. To me (by itself), it implies that there's a situation where a provider should include their own property if they don't include the FOCUS-defined property. If we want something like this I'd suggest something along the lines of this:
`SkuPriceDetails SHOULD include all applicable Provider-defined SKU Price properties for the corresponding SkuPriceId when there is no equivalent FOCUS-defined property.`

shawnalpay:
If some combination of @ijurica, @kk09v, and @flanakin can align on this (both the content, and the fact that this is truly a consistency refinement) before the Members meeting (e.g., in the next 90 minutes), we can put it forward for consideration.  But we will not have time to discuss and refine this in the meeting itself.

shawnalpay:
@ijurica There is a merge conflict on this PR.  Per AI #1213, you plan to incorporate feedback from others.

Putting this PR back to draft.

ijurica:
@kk09v I ended up going with your slightly rephrased suggestion — thank you for that.

In addition, I changed SHOULD to MUST in the other requirement to encourage Providers to start adopting FOCUS-defined properties (I assume you agree that was our intended goal). Strictly speaking, this is a material change as well, but the previous version already included a requirement stating “MUST include the FOCUS-defined SKU Price property when an equivalent property is included”, so FOCUS-defined properties were already treated as mandatory to some extent. 😃 

Anyway, @kk09v @shawnalpay the PR is ready for your review. When/if you give it the green light, we can proceed with presenting it to the members.

specification/columns/skupricedetails.md
```diff
@@ -17,8 +17,8 @@ The SkuPriceDetails column adheres to the following requirements:
 * When SkuPriceDetails is not null, SkuPriceDetails adheres to the following additional requirements:
   * SkuPriceDetails MUST be associated with a given SkuPriceId.
   * SkuPriceDetails MUST NOT include properties that are not applicable to the corresponding SkuPriceId.
-  * SkuPriceDetails SHOULD include all FOCUS-defined SKU Price properties listed below that are applicable to the corresponding SkuPriceId.
-  * SkuPriceDetails MUST include the FOCUS-defined SKU Price property when an equivalent property is included as a Provider-defined property.
+  * SkuPriceDetails MUST include all FOCUS-defined SKU Price properties listed below that are applicable to the corresponding SkuPriceId.
```
kk09v: This is a very significant change because it goes from being a "best effort" for providers to fill in properties to being mandatory.

ijurica:
```suggestion
  * SkuPriceDetails SHOULD include all FOCUS-defined SKU Price properties listed below that are applicable to the corresponding SkuPriceId.
```

ijurica: I agree. While MUST is our north star, we're not there yet — reverting back to SHOULD.

ijurica:
While MUST might reflect our long-term goal, we’re not there yet. Therefore, the previously proposed change from SHOULD to MUST in the requirement below has been reverted to preserve the original SHOULD wording.

> * SkuPriceDetails SHOULD include all FOCUS-defined SKU Price properties listed below that are applicable to the corresponding SkuPriceId.

shawnalpay:
> When/if you give it the green light, we can proceed with presenting it to the members.

@ijurica, before I will give this time in a future Members meeting, I will require you to 1) bring this PR out of draft, and 2) obtain explicit PR approval from at least one other working group member (ideally more).  Let me know if you have any questions.

kk09v (approved):

specification/columns/skupricedetails.md
```diff
@@ -18,7 +18,7 @@ The SkuPriceDetails column adheres to the following requirements:
   * SkuPriceDetails MUST be associated with a given SkuPriceId.
   * SkuPriceDetails MUST NOT include properties that are not applicable to the corresponding SkuPriceId.
   * SkuPriceDetails SHOULD include all FOCUS-defined SKU Price properties listed below that are applicable to the corresponding SkuPriceId.
-  * SkuPriceDetails MUST include the FOCUS-defined SKU Price property when an equivalent property is included as a Provider-defined property.
+  * SkuPriceDetails SHOULD include all Provider-defined SKU Price properties that are applicable to the corresponding SkuPriceId when there is no equivalent FOCUS-defined property.
```
flanakin: This is a different requirement than the original version. The original is saying that the FOCUS property has to be there, even if a provider has a custom property. The new states that all custom properties should be added.

I don't have a problem with the new requirement, but I think we need to keep the old one as well.

I changed "provider-defined" to "custom" to account for data generators adding custom columns in addition to providers. 
```suggestion
* SkuPriceDetails MUST include the FOCUS-defined SKU Price property when an equivalent property is included as a custom property.
  * SkuPriceDetails SHOULD include all custom SKU Price properties that are applicable to the corresponding SkuPriceId when there is no equivalent FOCUS-defined property.
```

flanakin (changes requested):

specification/columns/skupricedetails.md
ijurica: Good catch on the custom vs. Provider-defined distinction. I’ll update line 69 too, where Provider-defined keys are mentioned.

Regarding the normative requirements, I’ll post a follow-up question/suggestion.

ljadvey: I like Michael's suggestion here - keeping it the same but clarifying for the custom property for generators.

specification/columns/skupricedetails.md
```diff
@@ -18,7 +18,8 @@ The SkuPriceDetails column adheres to the following requirements:
   * SkuPriceDetails MUST be associated with a given SkuPriceId.
   * SkuPriceDetails MUST NOT include properties that are not applicable to the corresponding SkuPriceId.
   * SkuPriceDetails SHOULD include all FOCUS-defined SKU Price properties listed below that are applicable to the corresponding SkuPriceId.
-  * SkuPriceDetails MUST include the FOCUS-defined SKU Price property when an equivalent property is included as a Provider-defined property.
+  * SkuPriceDetails MUST include the FOCUS-defined SKU Price property when an equivalent property is included as a custom property.
```
ijurica: @flanakin @kk09v @rileyjenk Can you remind me why we specified SHOULD instead of MUST here?

> * SkuPriceDetails SHOULD include all FOCUS-defined SKU Price properties listed below that are applicable to the corresponding SkuPriceId.

I suggest we consider specifying that **any FOCUS-defined property MUST be included if applicable, rather than relying on whether it has already been included as a Provider-defined property**.

In other words, I propose replacing this:

> * ~~SkuPriceDetails SHOULD include all FOCUS-defined SKU Price properties listed below that are applicable to the corresponding SkuPriceId.~~
>  * ~~SkuPriceDetails MUST include the FOCUS-defined SKU Price property when an equivalent property is included as a custom property.~~

with this:

>  * SkuPriceDetails MUST include all FOCUS-defined SKU Price properties listed below that are applicable to the corresponding SkuPriceId.

shawnalpay:
@ijurica If this PR is fully authored, as I believe it is, then can you please pull it out of draft?

rileyjenk (approved):

shawnalpay:
Shared this PR in the Aug 25 Maintainers meeting -- no pushback on putting it forward for Member review and approval this coming Thu.

specification/columns/skupricedetails.md
shawnalpay: @ijurica Please do resolve this comment one way or the other before Member review on Thursday!

ijurica: Resolving due to lack of feedback; presenting PR in current state (two approvals) at Members meeting (Aug-28). 
Follow-up PR may be created.

flanakin (approved):

UdamDewaraja (approved):

shawnalpay (approved):

ljadvey (approved):

jpradocueva:
PR was approved during the Members Meeting on September 4.

Issue #1090, opened 2025-05-23
[FR] Add Structured "BI" schema to FOCUS


robmartin33:
### Problem Statement

There are elements of the FOCUS specification which would be better organized using a more structured data schema to support easier access by Business Intelligence tooling. Additionally, many of the more detailed elements which might be included in future enhancements might be better provided in related datasets which have a defined schema relationship to the primary "rows" in the current FOCUS dataset, and which could be more granular without bloating the current dataset. 

### Use Case (Optional)

Establishing a schema of related tables would provide easier entity relationships to be reported upon in cases where organizations were building their own dashboards using Tableau, PowerBI, Apache Superset, or similar.

A structured schema with an established dashboard could provide an easier path to onboarding for new FOCUS users.

### Desired Outcome / Practitioner Impact

Publish a schema which reflects the entity relationships between current (and future) FOCUS columns, and which creates more efficient dataset sizes, allows for further granularity or detail to be added in specific instances, and which is expressed as would be required for use in a BI tool.

### Type of Request

Net-new concept (not currently supported by providers or FOCUS)

### Organizations Requesting or Supporting This Feature

Wayfair asked about this

### FinOps Scope Alignment (Optional)

- [x] Public Cloud – e.g., AWS, Azure, GCP, OCI
- [x] Software-as-a-Service (SaaS) – e.g., Salesforce, Snowflake
- [x] Data Center – on-prem compute and infrastructure
- [x] Licensing – subscription or usage-based licensing *(under development)*
- [x] AI – cost and usage for AI models and platforms *(under development)*
- [x] Custom – internal tooling, specialized infra *(under development)*

### Impacted Parties

- [x] FinOps Practitioner – end users who analyze or act on the data
- [x] FOCUS Data Generator – providers generating aligned output
- [ ] Vendor Supporting FOCUS – tools or platforms using the spec
- [ ] Other (please explain in comments)

### Level of Ambiguity (1–5)

5

### Proposed Solution / Approach

Creating a vanilla dashboard using an open source BI tool could be a first step to provide organizations something to build from or use as an example. 

It is notable here that most of the clouds announcing support for FOCUS 1.0 also provided an example dashboard in their native tools, without restructuring the dataset into a consistent format.

shawnalpay:
[...]

PR #1149, opened 2025-06-30
FR #1090: New table demo - Option 1

rileyjenk:


specification/tables/cost_and_usage/columns/availabilityzone.md
flanakin, line 1: I would keep the columns where they are today. New datasets should pull from the global set of columns rather than duplicate column definitions. This will be important when we reuse concepts.

specification/tables/cost_and_usage/purpose.md
```diff
@@ -0,0 +1,2 @@
+The Cost and Usage table is a record of the costs and usage of resources or services in a given billing period. It is used to track and analyze the financial aspects of resource consumption.
```
flanakin: Can we refer to these as "datasets" rather than "tables"?
```suggestion
The Cost and Usage dataset is a record of the costs and usage of resources or services in a given billing period. It is used to track and analyze the financial aspects of resource consumption.
```

specification/tables/invoices/columns/invoiceid.md
flanakin, line 1: This is the perfect example of a duplicated column. If we maintain columns at a spec level, then we can reuse the definition. And they should ideally maintain the definition across datasets or it would be confusing.

That said, we will need to look at how cross-column requirements are applied. I'm thinking we want to avoid duplicating those, however.

specification/tables/invoices/columns/invoiceid.md
```diff
@@ -0,0 +1,42 @@
+# Invoice ID
+
+An Invoice ID is a provider-assigned identifier for an invoice encapsulating some or all [*charges*](#glossary:charge) in the corresponding [*billing period*](#glossary:billing-period) for a given [*billing account*](#glossary:billing-account). Invoices are commonly used for scenarios like tracking billing transactions, facilitating payment processes and for performing invoice reconciliation between *charges* and billing periods.
+
+The InvoiceId column adheres to the following requirements:
+
+* InvoiceId is RECOMMENDED to be present in a [*FOCUS dataset*](#glossary:FOCUS-dataset).
```
flanakin: I would expect to see a table of columns that should be in each dataset with their required level of inclusion. InvoiceId may be RECOMMENDED in cost dataset, but would be a MUST in the invoice dataset.

specification/tables/cost_and_usage/columns/availabilityzone.md
mike-finopsorg: @flanakin I agree with the reuse, I just wanted to build out your approach

* All columns for all datasets live in the same folder. 
* We use MDPP includes to select from columns to include in the relevant sections of the document. Allowing the document to have the primary FOCUS dataset definition, followed by adjacent datasets selecting the relevant columns (including some of those in the primary FOCUS dataset)
* Only one set of attributes to be included in the document covering attributes used in all datasets


does this match what you are saying?

mike-finopsorg: If that's right I have a couple of questions for the group

1. Are we ok for the structure of the document to not be reflective of the structure of the filesystem/files. 
2. Inside the document (not files in git): Do we duplicate the definition of a column in the FOCUS dataset when it's used in an adjacent dataset or do we link/reference it back to the definition used in the FOCUS dataset?

flanakin: I was thinking there'd be a single data dictionary with columns and attributes, then separate datasets that have a single table listing applicable columns and attributes.

Hopefully, we won't need to duplicate content in the spec. We just need to confirm whether cross-column requirements make sense in this model or if they'd need to change.

mike-finopsorg: I guess this is a step away from what we do today which the list of columns is the FOCUS data spec. We would instead move to have the list of columns cover all data columns generically and then a section of the specification to state which columns go to which dataset (FOCUS or otherwise). 

I think this works ok, but would require us to all agree and understand this approach.

shawnalpay:
@rileyjenk What are the next steps on this PR?  The PDF builder is broken, which I would expect, given that this substantively changes the nature of the PDF generation.  What, if anything, do you need from the group to move this forward?  Or has it served its purpose and will not be merged?

shawnalpay:
We discussed this PR in Jul 28 Maintainers.  While we understand the utility of this PR, it would be helpful for there to be a summary of the two possible paths forward.  @mike-finopsorg took AI #1225 to bring that back to the Maintainers next week.

mike-finopsorg:
I have created https://github.com/FinOps-Open-Cost-and-Usage-Spec/FOCUS_Spec/pull/1229 which demonstrates how a single column library with FOCUS dataset definitions would work as an alternative option to this PR for review of the maintainers.

mike-finopsorg:
I have pushed a heap of changes to this PR to make it build correctly, including some patches to the vendored package (markdown-pp) and build scripts.

shawnalpay:
> I have pushed a heap of changes to this PR to make it build correctly, including some patches to the vendored package (markdown-pp) and build scripts.


Thanks for pulling this together @mike-finopsorg!  This looks really good.  For quick reference for everyone, **[here](https://github.com/user-attachments/files/21623189/spec.pdf)** is the PDF showing the output.  Notice that the columns are listed underneath the dataset names in the Table of Contents, but Mike has handled for suppressing that same depth for other ToC sections.

mike-finopsorg:
Closed with https://github.com/FinOps-Open-Cost-and-Usage-Spec/FOCUS_Spec/pull/1337 superseding

Issue #1016, opened 2025-05-10
[AI] Create github guide document


ljadvey:
### Detailed Description

Create a github guide document on how to interact with the FOCUS repository on github

### Definition of Done

Document contains instructions on the following:
- Login to github
- Configure notifications
- Describe specification layout
- Instructions on how to submit an issue
- Instructions on how to interact (review, comment, participate in, etc.) an issue
- Instructions on how to clone the FOCUS repository using github desktop
- Instructions on how to submit a PR using github desktop and vscode
- Instructions on how to submit changes to an an existing PR using github desktop and vscode

ljadvey:
Updating PR now

ljadvey:
PR #1088 submitted, please review

shawnalpay:
Closing this issue, as PR #1088 is now out of draft and approaching final approval in the Members meeting.

PR #1088, opened 2025-05-22
AI #1016: Create github guide document

ljadvey:
**Description**
From feedback and our FOCUS survey, we determined a gap and need in how to use GitHub, tooling, pull requests,  working with issues, etc.  This PR is focused on filling this gap.

**Related to Issues**
For #1016 

**Ask for FOCUS Members**
Please review the guideline document within and provide comments, suggestions, and feedback.

PR #1196, opened 2025-07-21
AI #1016: Adding development container and documentation

eyzy31:
This pull request adds a development container and updates GitHub guidelines on how to use GitHub Codespaces via Github UI.

To test this out:
1. Go to the Branch: https://github.com/FinOps-Open-Cost-and-Usage-Spec/FOCUS_Spec/tree/1016/github-add-devcontainer-and-instructions
2. Click the Green Code Button
3. Click on the next following Green button to open a Codespace on this branch

Looking to confirm:
1. It works on your end
2. Any additional things you want in the setup

The development container contains the following installation
1. Git
2. Github CLI

VS Code Extensions incude:
1. Gitlens
2. Github-vscode-pull-request

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -1,0 +1,29 @@
+{
+  "name": "FOCUS_Spec_development",
+  "image": "mcr.microsoft.com/devcontainers/base:ubuntu", 
+  "features": {
+    "ghcr.io/devcontainers/features/git:1": {},
+    "ghcr.io/devcontainers/features/github-cli:1": {}
+  },
+  "customizations": {
+    "vscode": {
+      "extensions": [
+        "GitHub.vscode-pull-request-github",
+        "eamodio.gitlens"
+      ],
+      "settings": {
+        "git.enableSmartCommit": true,
+        "git.confirmSync": false,
+        "githubPullRequests.remotes": ["origin"],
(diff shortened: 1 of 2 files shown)


shawnalpay:
@eyzy31 I certainly appreciate your willingness to contribute!

That said, I wonder if it is premature for us to include a mention of codespaces in a best practices guide like this, for two reasons:

- The FOCUS repo is predominantly a collection of English words, not code -- so the typical FOCUS contributor will not have a need for codespaces.
- To the best of my understanding, no one has been using codespaces on the FOCUS project -- and I would like this document to detail our current best practices instead of suggesting a future best practice (which this very well could become in time!).

@ljadvey: as the author of #1088, I'm curious to get your take!  Do you agree or disagree with my assessment?

eyzy31:
Hey @shawnalpay, happy to defer to what you and @ljadvey  have in mind. I think this might be better as a subsequent task compared to what you/Larry already have in the first iteration. Rather we align first before going to any deeper into this.

ljadvey:
I like the idea @eyzy31, but agree with Shawn that we're not using codespaces today, and I do not expect we will be using them, so no need to add that into the document at this time.

shawnalpay:
Thanks @eyzy31 and @ljadvey!  With that, then, I'm going to close this PR.  It'll be here for reference in the future if and when we do start using Codespaces!

PR #1229, opened 2025-07-29
FR #1090: New table demo - Option 2

mike-finopsorg:
This is (Option 2) in the proposal for including adjacent datasets into FOCUS, which defines tables as a collection of columns from the column definitions inside the FOCUS document, leaving a single column dictionary and a selection of dataset definitions in the FOCUS specification.


mike-finopsorg:
[spec.pdf](https://github.com/user-attachments/files/21480644/spec.pdf)

To see how this looks in the actual document build I have attached a built version.

mike-finopsorg:
Closed in preference for the following https://github.com/FinOps-Open-Cost-and-Usage-Spec/FOCUS_Spec/pull/1337

PR #1321, opened 2025-08-06
FR #1082: Clarify requirements to cover all invoiced charges in FOCUS datasets

### LINKED ISSUE #1082: [FR] FOCUS datasets should cover all invoiced charges
### Summary

Spec does not clearly state that all costs for an invoice need to be included in FOCUS

### Which area does this issue relate to?

B: Existing FOCUS specification content

### Detailed Description

Right now the FOCUS definition of InvoiceId does not clearly state that a billing generator needs to provide billing lines for all costs in an invoice. This is really important for Invoice Reconciliation processes, if some costs are missing from the FOCUS data then the total BilledCost will not equal that which a practitioner sees on their invoice.

### What use cases, FinOps or others, can't be performed with the existing specification unless this issue is addressed?

Invoice Reconciliation 

### Which FinOps personas does this issue relate to?

Practitioner, Finance

### Which provider groups or specific providers does this issue relate to?

ALL

### On a scale of 1 - 4, how critical is this for your organization?

3: Important for adoption in the next 6-12 months

### Objective

Add normative text to ensure the requirement is clear that all charges for an invoice need to be covered in the FOCUS dataset

### Data Examples

_No response_

### Issues, PRs, or Other References

_No response_

---

### 🧪 Use Case

TBD

---

TBD

---

### 📨 Type of Request

- [ ] Supporting Content (e.g., appendices, use cases)

---

TBD

---

---

### 📊 Impacted Parties

---

### 🌫️ Level of Ambiguity

TBD

---

https://github.com/FinOps-Open-Cost-and-Usage-Spec/FOCUS_Spec/issues/944

---

TBD

---

TBD

### LINKED ISSUE #1240: [AI] Review PR for invoice-handling attribute with members
### Detailed Description

We don't have an existing attribute that captures the scope of invoiced charges and their relationship with the FOCUS dataset. This PR will provide a draft of that attribute that we can use to align on the expected behavior of data providers. 

### Definition of Done

Review PR with FOCUS members

Matt-Cowsert:
Introducing an invoice handling attribute that incorporates feedback given on draft [doc](https://docs.google.com/document/d/1PSqgNk9guQ3vRVK3iCcPUG8CKMmkLLNMwUa1AahMYPE/edit?tab=t.0#heading=h.nkq0gl2tpvwd) associated with [FR] Clarify requirements to cover all invoiced charges in FOCUS datasets. Also added invoice_handling.md to the attributes.mdpp file in alphanumeric order. 
Fixes #1082


shawnalpay (changes requested):
@Matt-Cowsert 
For the sake of demonstration, it would be helpful for you to apply this attribute via normative requirement to at least one column (e.g. `Effective Cost`).  That will clarify how you may need to revise the language in this attribute to be more applicable from one column to the next, and it will also better demonstrate to the working group where the rubber meets the road on this change.

shawnalpay:
I would also expect this PR (or a related one) to address the description of [Invoice ID](https://github.com/FinOps-Open-Cost-and-Usage-Spec/FOCUS_Spec/blob/working_draft/specification/columns/invoiceid.md), which includes (emphasis mine):

> An Invoice ID is a provider-assigned identifier for an invoice encapsulating **some or all** charges in the corresponding billing period...

I would think this should change to **all** charges, even if we're framing the requirement as a SHOULD.

Matt-Cowsert:
Thanks @shawnalpay I've added normative text to the columns I expect to be affected by the Invoice handling attribute and tightened up the definition for InvoiceId.

specification/attributes/invoice_handling.md
```diff
@@ -0,0 +1,34 @@
...
+
+## Description
+
+Indicates how invoice-level monetary charges, including those not directly tied to usage, SHOULD be represented in a FOCUS dataset. 
+
+## Requirements
+
+* All costs that appear on an invoice SHOULD be present in the FOCUS dataset.
+* Monetary transactions MAY be expressed as individual rows or as part of aggregated rows, provided they remain distinguishable using existing FOCUS dimensions and metrics.
+* ChargeCategory SHOULD be used to differentiate types of non-usage charges (e.g., “Tax”, “Credit”, “Purchase”, “Support”).
```
ijurica: Note: 
- Among the listed items, "Support" is not a valid allowed value for Charge Category.
- Valid Charge Categories include: "Usage", "Purchase", "Credit", "Tax", and "Adjustment".

```suggestion
* ChargeCategory SHOULD be used to differentiate types of non-usage charges (e.g., "Tax", "Credit", "Purchase", "Adjustment").
```

specification/attributes/invoice_handling.md
```diff
@@ -0,0 +1,34 @@
...
+* If a monetary transaction cannot be expressed using existing FOCUS fields, providers SHOULD include supplemental columns (e.g., x_ChargeSubType) to capture this information.
```
ijurica: nit:

```suggestion
* If a monetary transaction cannot be expressed using existing FOCUS columns, providers SHOULD include supplemental columns (e.g., x_ChargeSubType) to capture this information.
```

BTW, Providers vs Data Generators is a separate discussion topic 😄

specification/columns/invoiceid.md
```diff
@@ -1,12 +1,13 @@
...
+An Invoice ID is a provider-assigned identifier for an invoice encapsulating all [*charges*](#glossary:charge) in the corresponding [*billing period*](#glossary:billing-period) for a given [*billing account*](#glossary:billing-account). Invoices are commonly used for scenarios like tracking billing transactions, facilitating payment processes and for performing invoice reconciliation between *charges* and billing periods.
 
 The InvoiceId column adheres to the following requirements:
 
 * InvoiceId is RECOMMENDED to be present in a [*FOCUS dataset*](#glossary:FOCUS-dataset).
 * InvoiceId MUST be of type String.
 * InvoiceId MUST conform to [StringHandling](#stringhandling) requirements.
+* The sum of BilledCost values for all rows sharing the same InvoiceId MUST match the total payable amount on the corresponding invoice, unless exceptions defined in the InvoiceHandling attribute apply.
```
ijurica:
```suggestion
* The sum of BilledCost for a given [InvoiceId](#invoiceid) MUST match the sum of the payable amount provided in the corresponding invoice with the same id generated by the [InvoiceIssuer](#InvoiceIssuer).
```

Personally, I prefer the original version as specified in section 1.2 under the BilledCost column for several reasons:

- While unlikely, it's still possible for two different InvoiceIssuers to generate invoices with the same Invoice ID. Therefore, I think it's valuable to keep the phrasing: "with the same id generated by the InvoiceIssuer."

- Phrasing differences: In other normative requirements, we haven't explicitly mentioned "values" (e.g., BilledCost values), as it's generally implied that the ColumnId refers to the value itself.

- Question: Do the exceptions listed in the attribute apply to the entire invoice, even the one related to the withheld charges? If that's the case, than there is no exception in this requirement.

Positioning of the requirement: This requirement should be placed last — anyway not before the nullability-related requirements.

specification/attributes/invoice_handling.md
```diff
@@ -0,0 +1,34 @@
...
+## Description
+
+Indicates how invoice-level monetary charges, including those not directly tied to usage, SHOULD be represented in a FOCUS dataset. 
+
+## Requirements
+
+* All costs that appear on an invoice SHOULD be present in the FOCUS dataset.
+* Monetary transactions MAY be expressed as individual rows or as part of aggregated rows, provided they remain distinguishable using existing FOCUS dimensions and metrics.
+* ChargeCategory SHOULD be used to differentiate types of non-usage charges (e.g., "Tax", "Credit", "Purchase", "Adjustment").
+* If a monetary transaction cannot be expressed using existing FOCUS columns, providers SHOULD include supplemental columns (e.g., x_ChargeSubType) to capture this information.
```
UdamDewaraja: This to me reads as if we're expecting invoice issuers to provide FOCUS terminology in the invoice which seems not practical IMO. Invoices due to the way they're commonly provided (PDFs that are roughly A4 paper size) will not be able to provide highly tablular data. This means commonly things like taxes and discounts are handled in an overall manner (typically at the bottom of a simple tabular dataset.

specification/attributes/invoice_handling.md
```diff
@@ -0,0 +1,34 @@
+# Invoice Handling
+
+FinOps practitioners must be able to reconcile structured billing data with the invoices they receive from providers. In practice, this means ensuring that all monetary charges that appear on an invoice — including those not tied to metered usage — are represented in the FOCUS dataset. Without this alignment, it becomes difficult to perform accurate invoice reconciliation, financial reporting, and chargeback.
```
UdamDewaraja:
```suggestion
FinOps practitioners must be able to reconcile FOCUS datasets with the corresponding invoices and usage statements they receive from providers. In practice, this means ensuring that all monetary charges that appear on an invoice or usage statement — including those not tied to metered usage — are represented in the FOCUS dataset. Without this alignment, it becomes difficult to perform accurate invoice reconciliation, financial reporting, and chargeback.
```

specification/attributes/invoice_handling.md
```diff
@@ -0,0 +1,34 @@
...
+
+This attribute introduces requirements for how non-usage charges such as taxes, credits, refunds, support, training, and marketplace transactions SHOULD be captured and categorized. It also defines expectations around the completeness and consistency of invoice-level totals within the dataset, enabling FOCUS to serve as a reliable system of record for all invoiced costs.
```
UdamDewaraja:
```suggestion
This attribute introduces requirements for how charges such as usage, taxes, credits, refunds, etc, inclusive of support, training, and marketplace transactions, and any other type of charge SHOULD be captured and categorized. It also defines expectations around the completeness and consistency of invoice-level totals within the dataset, enabling FOCUS to serve as a reliable system of record for all invoiced costs.
```

UdamDewaraja (commented):
I'd like to propose that this is inclusive of the invoice and the 'usage statements' that providers commonly provide in the case of pre-paid scenarios (invoice is the pre payment, usage statement is the monthly statement that helps reconcile the burned down usage). Otherwise reconciliation can only be done on a cash basis but not usage-basis (which are both critically important).

specification/attributes/invoice_handling.md
```diff
@@ -0,0 +1,34 @@
...
+
+## Description
+
+Indicates how invoice-level monetary charges, including those not directly tied to usage, SHOULD be represented in a FOCUS dataset. 
+
+## Requirements
+
+* All costs that appear on an invoice SHOULD be present in the FOCUS dataset.
+* Monetary transactions MAY be expressed as individual rows or as part of aggregated rows, provided they remain distinguishable using existing FOCUS dimensions and metrics.
+* ChargeCategory SHOULD be used to differentiate types of non-usage charges (e.g., "Tax", "Credit", "Purchase", "Adjustment").
```
shawnalpay: Is this the right place to put a requirement that applies only to one column?  (Genuine question; I'm not sure the answer.)

An alternative might be to put this normative requirement directly within `ChargeCategory`, given that it's applicable only to that one column as currently written.

specification/attributes/invoice_handling.md
```diff
@@ -0,0 +1,34 @@
...
+
+InvoiceHandling
+
+## Attribute Name
+
+Invoice Handling
+
+## Description
+
+Indicates how invoice-level monetary charges, including those not directly tied to usage, SHOULD be represented in a FOCUS dataset.
```
shawnalpay: SHOULD doesn't need capitalization here.  All caps are only used in normative requirements.
```suggestion
Indicates how invoice-level monetary charges, including those not directly tied to usage, should be represented in a FOCUS dataset. 
```

specification/attributes/invoice_handling.md
```diff
@@ -0,0 +1,34 @@
...
+* Monetary transactions MAY be expressed as individual rows or as part of aggregated rows, provided they remain distinguishable using existing FOCUS dimensions and metrics.
+* ChargeCategory SHOULD be used to differentiate types of non-usage charges (e.g., "Tax", "Credit", "Purchase", "Adjustment").
+* If a monetary transaction cannot be expressed using existing FOCUS columns, providers SHOULD include supplemental columns (e.g., x_ChargeSubType) to capture this information.
+* The total cost associated with a given Invoice ID in the FOCUS dataset SHOULD match the total cost presented on the associated invoice.
+
+## Exceptions
+
+* Invoice corrections or adjustments applied outside the represented billing period MAY be excluded from the current dataset if they are expected to appear in a subsequent period’s FOCUS dataset.
+* Charges withheld due to legal, regulatory, or contractual restrictions MAY be omitted if they cannot be shared within structured billing data. In such cases, the data generator SHOULD document the reason for exclusion and MAY include a redacted or summarized charge entry with explanatory metadata.
```
shawnalpay: @ijurica How should we should structure this section of multiple normative requirements with bullets?

shawnalpay (changes requested):
Good start @Matt-Cowsert!  I left a few more comments.  I'd also expect every impacted column to be addressed before we bring this out of draft.  Do you yet know that list?

specification/attributes/invoice_handling.md
Matt-Cowsert: we aligned with removing this exception for the time being, as we don't have an example to lean on and view it as a two way door that we can add later if needed.

specification/attributes/invoice_handling.md
Matt-Cowsert: I'm fine with either approach. I'll bring it to TF-3 for discussion and will align with the TF on where we should place it.

specification/attributes/invoice_handling.md
Matt-Cowsert: The intent of this requirement isn't to dictate the format of the invoice, rather to ensure the transactions captured on the invoice are present within the FOCUS dataset. I'll re-write the requirement to be more specific. Thanks for calling out the ambiguity.

specification/columns/invoiceid.md
```diff
@@ -1,12 +1,13 @@
 # Invoice ID
 
-An Invoice ID is a provider-assigned identifier for an invoice encapsulating some or all [*charges*](#glossary:charge) in the corresponding [*billing period*](#glossary:billing-period) for a given [*billing account*](#glossary:billing-account). Invoices are commonly used for scenarios like tracking billing transactions, facilitating payment processes and for performing invoice reconciliation between *charges* and billing periods.
+An Invoice ID is a provider-assigned identifier for an invoice encapsulating all [*charges*](#glossary:charge) in the corresponding [*billing period*](#glossary:billing-period) for a given [*billing account*](#glossary:billing-account). Invoices are commonly used for scenarios like tracking billing transactions, facilitating payment processes and for performing invoice reconciliation between *charges* and billing periods.
```
UdamDewaraja:
```suggestion
An Invoice ID is a provider-assigned identifier for an invoice encapsulating all [*charges*](#glossary:charge) in the corresponding [*billing period*](#glossary:billing-period) for a given [*billing account*](#glossary:billing-account). Invoices are commonly used for scenarios like tracking billing transactions, facilitating payment processes and for performing invoice reconciliation between *charges* and billing periods.
```

Putting 'all' here feels like the wrong messaging @thecloudman @rileyjenk @shawnalpay. Providers can provide multiple invoices for the same billing period and billing account for various reasons. In some cases, AWS Support, for example, comes in a separate invoice, even though you have an original invoice for the billing account and period. There are possibly other cases like purchases or corrections, where you have multiple invoices/invoice ids.

A possible improvement would be to say the sum of all charges provided by invoice issuers that provide invoice ids for a given BillingAccount and BillingPeriod should match the sum of all the invoices associated with the same scope.

specification/attributes/invoice_handling.md
```diff
@@ -0,0 +1,33 @@
+# Invoice Handling
+
+FinOps practitioners must be able to reconcile FOCUS datasets with the corresponding invoices and usage statements they receive from providers. In practice, this means ensuring that all monetary charges that appear on an invoice or usage statement — including those not tied to metered usage — are represented in the FOCUS dataset. Without this alignment, it becomes difficult to perform accurate invoice reconciliation, financial reporting, and chargeback.
+
+This attribute introduces requirements for how charges such as usage, taxes, credits, refunds, etc, inclusive of support, training, and marketplace transactions, and any other type of charge SHOULD be captured and categorized. It also defines expectations around the completeness and consistency of invoice-level totals within the dataset, enabling FOCUS to serve as a reliable system of record for all invoiced costs.
```
UdamDewaraja:
```suggestion
This attribute introduces requirements for how charges such as usage, taxes, credits, refunds, etc, inclusive of support, training, and marketplace transactions, and any other type of charge should be captured and categorized. It also defines expectations around the completeness and consistency of invoice-level totals within the dataset, enabling FOCUS datasets to be used in a system of record for all invoiced costs.
```
I think you want to put the normative requirements in the Requirements section.

specification/attributes/invoice_handling.md
```diff
@@ -0,0 +1,33 @@
+# Invoice Handling
+
+FinOps practitioners must be able to reconcile FOCUS datasets with the corresponding invoices and usage statements they receive from providers. In practice, this means ensuring that all monetary charges that appear on an invoice or usage statement — including those not tied to metered usage — are represented in the FOCUS dataset. Without this alignment, it becomes difficult to perform accurate invoice reconciliation, financial reporting, and chargeback.
```
UdamDewaraja:
```suggestion
FinOps practitioners must be able to reconcile FOCUS datasets with the corresponding invoices and usage statements they receive from [*Invoice Issuers*](#glossary:InvoiceIssuer). In practice, this means ensuring that all monetary [*charges*](#glossary:charge) that appear on an invoice or usage statement — including those not tied to metered usage — are represented in the [*FOCUS dataset*](#glossary:FOCUS-dataset). Without this alignment, it becomes difficult to perform accurate invoice reconciliation, financial reporting, and chargeback.
```

specification/attributes/invoice_handling.md
```diff
@@ -0,0 +1,33 @@
...
+
+InvoiceHandling
+
+## Attribute Name
+
+Invoice Handling
+
+## Description
+
+Indicates how invoice-level transactions, including those not directly tied to usage, should be represented in a FOCUS dataset.
```
UdamDewaraja:
```suggestion
Indicates how invoice-level *charges*, including those not directly tied to usage, should be represented in a FOCUS dataset. 
```

Seems like rather than introduce a new concept called 'transaction' we should stick to charges.

specification/attributes/invoice_handling.md
```diff
@@ -0,0 +1,33 @@
...
+
+## Description
+
+Indicates how invoice-level transactions, including those not directly tied to usage, should be represented in a FOCUS dataset. 
+
+## Requirements
+
+* All costs that appear on an invoice SHOULD be present in the FOCUS dataset.
+* Invoice-level transactions MAY be expressed as individual rows or as part of aggregated rows, provided they remain distinguishable using existing FOCUS dimensions and metrics.
+* ChargeCategory SHOULD be used to differentiate types of non-usage charges (e.g., "Tax", "Credit", "Purchase", "Adjustment").
```
UdamDewaraja:
```suggestion
* ChargeCategory SHOULD be used to differentiate types of non-usage *charges* (e.g., "Tax", "Credit", "Purchase", "Adjustment").
```

specification/attributes/invoice_handling.md
```diff
@@ -0,0 +1,33 @@
...
+Invoice Handling
+
+## Description
+
+Indicates how invoice-level transactions, including those not directly tied to usage, should be represented in a FOCUS dataset. 
+
+## Requirements
+
+* All costs that appear on an invoice SHOULD be present in the FOCUS dataset.
+* Invoice-level transactions MAY be expressed as individual rows or as part of aggregated rows, provided they remain distinguishable using existing FOCUS dimensions and metrics.
```
UdamDewaraja:
```suggestion
* Invoice-level *charges* MAY use or closely align with the FOCUS schema.
```

specification/attributes/invoice_handling.md
```diff
@@ -0,0 +1,33 @@
...
+* ChargeCategory SHOULD be used to differentiate types of non-usage charges (e.g., "Tax", "Credit", "Purchase", "Adjustment").
+* If an invoice-level transaction appears on a customer’s invoice but cannot be fully represented using existing FOCUS columns, providers SHOULD include supplemental columns (e.g., x_ChargeSubType) to ensure that the invoice charge is captured in the FOCUS dataset. 
+* The total cost associated with a given Invoice ID in the FOCUS dataset SHOULD match the total cost presented on the associated invoice.
+
```
UdamDewaraja:
```suggestion
* The total cost associated with a given Invoice ID in a *FOCUS dataset* MUST match the total cost presented on the associated invoice.

```

This really seems like the core requirement here and likely belongs in the InvoiceId.

specification/attributes/invoice_handling.md
```diff
@@ -0,0 +1,33 @@
...
+## Exceptions
+
+* Invoice corrections or adjustments applied outside the represented billing period MAY be excluded from the current dataset if they are expected to appear in a subsequent period’s FOCUS dataset.
```
UdamDewaraja: Not sure if this is needed. This would be true anyway, you have charges coming in a different billing period, they would have their own invoice id that corresponds to the invoice(s) that map to that billing period.

Matt-Cowsert:
Flipping this back to draft until I'm able to incorporate the charge category normative text additions

specification/attributes/invoice_handling.md
Matt-Cowsert: Fully agree with replacing transactions with charges and I can simplify the FOCUS schema. However, I don't want to remove the individual vs. aggregate row call out, as I think it's important to acknowledge there isn't always a 1:1 relationship between invoice-level charges and what's represented within a FOCUS dataset.

specification/attributes/invoice_handling.md
Matt-Cowsert: agreed, removing

specification/columns/invoiceid.md
Matt-Cowsert: agreed, reverting the language to give space for multiple invoices within a single billing period.

specification/attributes/invoice_handling.md
Matt-Cowsert: I agree with the suggestion and will represent this requirement within the Invoice ID normative requirements, rather than the invoice-handling section.

specification/attributes/invoice_handling.md
Matt-Cowsert: I've moved this to the ChargeCategory normative text. We mentioned in the TF-3 call to reference the normative text within the invoice-handling section as well, but it wasn't clear to me that was necessary (or how) to reference a single requirement from another column.

specification/attributes/invoice_handling.md
```diff
@@ -0,0 +1,32 @@
...
+
+## Description
+
+Indicates how invoice-level *charges*, including those not directly tied to usage, should be represented in a FOCUS dataset. 
+
+## Requirements
+
+* All costs that appear on an invoice SHOULD be present in the FOCUS dataset.
+* Invoice-level charges MAY be expressed as individual rows or as part of aggregated rows, provided they remain distinguishable and use or closely align with the FOCUS schema.
+* If an invoice-level transaction appears on a customer’s invoice but cannot be fully represented using existing FOCUS columns, providers SHOULD include supplemental columns (e.g., x_ChargeSubType) to ensure that the invoice charge is captured in the FOCUS dataset.
```
UdamDewaraja:
```suggestion
* If an invoice-level *charge* appears on a customer’s invoice but cannot be fully represented using existing FOCUS columns, providers SHOULD include supplemental columns (e.g., x_ChargeSubType) to ensure that the details of the invoice *charge* are captured in the *FOCUS dataset*. 
```

specification/attributes/invoice_handling.md
```diff
@@ -0,0 +1,32 @@
...
+Invoice Handling
+
+## Description
+
+Indicates how invoice-level *charges*, including those not directly tied to usage, should be represented in a FOCUS dataset. 
+
+## Requirements
+
+* All costs that appear on an invoice SHOULD be present in the FOCUS dataset.
+* Invoice-level charges MAY be expressed as individual rows or as part of aggregated rows, provided they remain distinguishable and use or closely align with the FOCUS schema.
```
shawnalpay: Per discussion in Aug 28 Members: this line currently reads like an FYI to me.  If that's true, it should be struck as a requirement and rephrased in the description section above.

However, @ijurica expressed the belief that there is a "hidden MUST" in here, and thus it should be rephrased here to reflect that.  I will leave it to her to mention what that should be.

specification/attributes/invoice_handling.md
```diff
@@ -0,0 +1,32 @@
...
+
+Invoice Handling
+
+## Description
+
+Indicates how invoice-level *charges*, including those not directly tied to usage, should be represented in a FOCUS dataset. 
+
+## Requirements
+
+* All costs that appear on an invoice SHOULD be present in the FOCUS dataset.
```
UdamDewaraja:
```suggestion
* All costs that appear on an invoice SHOULD be present in the *FOCUS dataset*.
```

specification/columns/billedcost.md
```diff
@@ -7,6 +7,7 @@ The BilledCost column adheres to the following requirements:
 * BilledCost MUST be present in a [*FOCUS dataset*](#glossary:FOCUS-dataset).
 * BilledCost MUST be of type Decimal.
 * BilledCost MUST conform to [NumericFormat](#numericformat) requirements.
+* BilledCost MUST conform to [InvoiceHandling](#invoicehandling) requirements.
```
UdamDewaraja:
```suggestion
```
I get what we're going for here, but there's not really a requirement (or even a mention of BilledCost in InvoiceHandling_.

UdamDewaraja (approved):

specification/attributes/invoice_handling.md
```diff
@@ -0,0 +1,32 @@
...
+Invoice Handling
+
+## Description
+
+Indicates how invoice-level *charges*, including those not directly tied to usage, should be represented in a FOCUS dataset. 
+
+## Requirements
+
+* All costs that appear on an invoice SHOULD be present in the *FOCUS dataset*.
+* Invoice-level charges MAY be expressed as individual rows or as part of aggregated rows, provided they remain distinguishable and use or closely align with the FOCUS schema.
```
ijurica: ### Discussion Topic

We should decide on a general approach regarding whether normative requirements that are already specified at the column level should also be repeated (in full or in reference form) within overarching attribute sections (e.g., Invoice Handling, Discount Handling, etc.).

Options:

- **Do not duplicate:** Keep normative requirements only at the column level, and avoid repeating them in overarching attribute sections.

- **Repeat with reference:** Reiterate key normative requirements in the overarching section, but include a note that they are defined at the column level.
Example:

> The sum of the BilledCost for a given InvoiceId MUST match the sum of the payable amount **denominated in the Billing Currency** provided in the corresponding invoice with the same id generated by the InvoiceIssuer, **as defined in the BilledCost column requirements**.

- **Add a general reference:** Include a general statement in the overarching section (i.e., invoice handling attribute that points to the normative requirements defined at the column level. (Karl may have already done something similar for Split Cost)

---

Note: I plan to provide some additional feedback on this particular  requirement:

> * Invoice-level charges MAY be expressed as individual rows or as part of aggregated rows, provided they remain distinguishable and use or closely align with the FOCUS schema.

shawnalpay (changes requested):
@Matt-Cowsert please do the following before we bring this PR back to Members for approval:

 - [ ] Address and resolve all outstanding comments.
 - [x] Resolve the outstanding merge conflicts (three at the time of this writing).

specification/attributes/invoice_handling.md
Matt-Cowsert: we aligned during TF-3 that the simplest approach is to strike this as a normative requirement. Even putting it in the description is an unnecessary elaboration.

specification/attributes/invoice_handling.md
Matt-Cowsert: I've removed this, but let me know @ijurica if we still need to discuss the meta points in your options

specification/columns/billedcost.md
Matt-Cowsert: I'm okay with striking this, as we already captured this requirement within Billed Cost normative requirements:
_last requirement in 3.2 Billed Cost:_
- The sum of the BilledCost for a given [InvoiceId](https://focus.finops.org/focus-specification/v1-2/#invoiceid) MUST match the sum of the payable amount provided in the corresponding invoice with the same id generated by the [InvoiceIssuer](https://focus.finops.org/focus-specification/v1-2/#InvoiceIssuer).

We also touch on this within our supported features section:

**Section 2.2. Billed Cost and Invoice Alignment**
 2.2.1. Description
FOCUS data should be consistent with the costs indicated on payable invoices. This is relevant to the total cost of the invoice, as well as the period of time the invoice covers.

specification/attributes/invoice_handling.md
ijurica: @Matt-Cowsert , I agree that the question I raised here is really part of a broader discussion.

I mentioned that I thought this requirement should be phrased differently, but the more I think about it, the more I believe what I had in mind is already covered by the existing requirement on BilledCost.

Resolving this conversation.

specification/attributes/invoice_handling.md
```diff
@@ -0,0 +1,30 @@
...
+
+Invoice Handling
+
+## Description
+
+Indicates how invoice-level *charges*, including those not directly tied to usage, should be represented in a FOCUS dataset. 
+
+## Requirements
+
+* All costs that appear on an invoice SHOULD be present in the *FOCUS dataset*.
```
ljadvey:
```suggestion
* All costs that appear on an invoice MUST be present in the *FOCUS dataset*.
```

ljadvey: This line conflicts with newly added line 16; suggesting changing this to MUST to align
```
The sum of BilledCost for a given [InvoiceId](#invoiceid) MUST match the sum of the payable amount provided in the corresponding invoice with the same id generated by the [InvoiceIssuer](#InvoiceIssuer).
```

Matt-Cowsert: This is a good callout Larry, thanks for pointing this out. I'm going to accept the suggestion and socialize with TF-3 tomorrow before going back to member review / approval.

specification/attributes/invoice_handling.md
```diff
@@ -0,0 +1,30 @@
...
+
+Invoice Handling
+
+## Description
+
+Indicates how invoice-level *charges*, including those not directly tied to usage, should be represented in a FOCUS dataset. 
+
+## Requirements
+
+* All costs that appear on an invoice MUST be present in the *FOCUS dataset*.
```
flanakin: Shouldn't this be "FOCUS Cost and Usage dataset"?

flanakin: Does this imply that all charges invoiced by a provider must be covered by FOCUS cost? It seems to imply that. But where's the line?

specification/attributes/invoice_handling.md
```diff
@@ -0,0 +1,30 @@
...
+* If an invoice-level *charge* appears on a customer’s invoice but cannot be fully represented using existing FOCUS columns, providers SHOULD include supplemental columns (e.g., x_ChargeSubType) to ensure that the details of the invoice *charge* are captured in the *FOCUS dataset*.
```
flanakin: What does "fully represented" mean? I feel like this may get misunderstood and generally missed. It sounds like it's asking for custom columns to be added for anything necessary to explain the row, but it's doing it in a way that has no teeth, as far as I can tell. Maybe it's just me tho.

specification/columns/chargecategory.md
```diff
@@ -8,6 +8,7 @@ The ChargeCategory column adheres to the following requirements:
 * ChargeCategory MUST be of type String.
 * ChargeCategory MUST NOT be null.
 * ChargeCategory MUST be one of the allowed values.
+* ChargeCategory SHOULD be used to differentiate types of non-usage charges (e.g., "Tax", "Credit", "Purchase", "Adjustment").
```
flanakin: Isn't this implied by having the different values? I'm not sure this "requirement" adds any value. We already require they select the proper value.

specification/columns/invoiceid.md
```diff
@@ -25,7 +26,7 @@ Invoice ID
 
 ## Description
 
-The provider-assigned identifier for an invoice encapsulating some or all *charges* in the corresponding billing period for a given billing account.
+The provider-assigned identifier for an invoice encapsulating all *charges* in the corresponding billing period for a given billing account.
```
flanakin: I don't think we can make this change. Billing periods may be covered by multiple invoices. Or at least, we'd need to say something like, "...that were billed on the invoice identified by the invoice ID." (But in a much more graceful way 🙃)

Matt-Cowsert:
after much back and forth, I've address the merge conflict issues, now on to addressing the actual comments in the PR

specification/columns/invoiceid.md
Matt-Cowsert: I reverted back to some or all, but gave the additional clarification that the full set of charges (can span across multiple invoices) need to be represented within the FOCUS dataset. Let me know what you think

specification/columns/chargecategory.md
Matt-Cowsert: If we don't think it's necessary, or it's duplicative, I'm okay with removing it

specification/attributes/invoice_handling.md
Matt-Cowsert: Yeah, that's a good point. Let me sync with Shawn, as I don't think we have a glossary term (yet) for the FOCUS Cost and Usage dataset drafted. It may be something we tackle during consistency review, but I agree it's something we should be more specific on.

Matt-Cowsert: It doesn't imply it, it mandates it explicitly. If you have a suggestion on how to improve the language, I'm all ears.

specification/attributes/invoice_handling.md
Matt-Cowsert: Let me try to sharpen the language and let me know what you think of it.

specification/attributes/invoice_handling.md
shawnalpay: To my mind, the verbiage as stated here is fine because, going forward, we are going to explicitly declare the association of attributes with datasets (per PR #1431, which is scheduled to be approved by the Members tomorrow).  _That_ is where the association happens, not here.

I don't think we need to change this line.

specification/columns/invoiceid.md
shawnalpay: I struggle to reconcile the "some or all" part here with line 21 that states:

> All costs that appear on an invoice MUST be present in the *FOCUS dataset*.

In any case, the Description section should strive to be as short as possible.  Any nuance you add like this should be captured elsewhere in the markdown.  Thus, here, I would suggest watering it down further to:

```suggestion
The provider-assigned identifier for an invoice encapsulating *charges* in the corresponding billing period for a given billing account.
```

...And then adding that nuance either up at the top of the document, or in the normative requirements.

specification/attributes/invoice_handling.md
flanakin: But a SKU catalog will eventually be a "FOCUS dataset" and this statement doesn't apply to a SKU catalog, right? I think this requirement only applies to the cost and usage dataset.

flanakin: I can read this as "costs that appear in any invoice that you would ever get from the provider" or "costs that appear in a single invoice". Both of those need to be handled differently.

We already have requirements on BilledCost, I believe, stating that cost and individual invoices should align. That's fine. My concern is the other perspective, which I don't feel is very clear.

This could be interpreted as *EVERY* invoice that comes from a company must be included in a single cost and usage dataset. I don't know how feasible that is from every provider. I know Microsoft has legacy systems that have invoices and will never be in the FOCUS data. Based on conversations with Gartner, that seems to be very common across the industry.

I would argue this loose implication (which is not explicit) should be clearer about all charges in covered invoices should be in the cost and usage dataset. If we say all invoices across all things a company could ever do, then that probably needs to be a "SHOULD".

specification/columns/invoiceid.md
flanakin: I like this. Simple and safe.

specification/attributes/invoice_handling.md
shawnalpay: You're right that it only applies to the Cost and Usage dataset (for now; maybe another dataset comes along in the future for which this is also applicable).  And the explicit association of attributes with datasets per #1431 is where we should make that clear.

Once #1431 has been approved (in theory, that's tomorrow, alongside this one), it should carry the line:

> * CostAndUsage MUST conform to [InvoiceHandling](#invoicehandling) requirements.

We could make a note on #1431 that we will add that line once #1321 gets approved, or delay the approval of #1321 one more week so that @Matt-Cowsert can make use of #1431 to add that line to this PR.

specification/columns/invoiceid.md
Matt-Cowsert: I've applied these recommended changes in the latest commit, though Github wouldn't let me commit the suggestion directly. as it's related to an outdated comment.

specification/attributes/invoice_handling.md
Matt-Cowsert: As this is a minor follow-on change and we have number of PRs to review/seek approval with members, I'd prefer to make the note on #1431, approve #1321 as is, and make a separate PR to add the improved language once it's been approved. @shawnalpay @flanakin

shawnalpay: I have now recorded that note on #1431 [here](https://github.com/FinOps-Open-Cost-and-Usage-Spec/FOCUS_Spec/pull/1431/files#r2338474379).

I defer to @flanakin to decide whether that addresses his original concern.

specification/attributes/invoice_handling.md
```diff
@@ -0,0 +1,30 @@
...
+
+Invoice Handling
+
+## Description
+
+Indicates how invoice-level *charges*, including those not directly tied to usage, should be represented in a FOCUS dataset.
+
+## Requirements
+
+* When an InvoiceId appears in the *FOCUS dataset*, all costs from that invoice MUST be included in the dataset.
```
ijurica: - Regarding the requirement tied to a **specific InvoiceId**, as @flanakin mentioned, **payable costs** are already covered through a combination of definitions and existing requirements:
  - **Invoice Issuer definition:** *An Invoice Issuer is an entity responsible for issuing payable invoices for the resources or services consumed.*
    - *(out of scope of this PR)* We may want to revisit whether "consumed" is sufficient here, or if it should be "used or purchased" to better cover purchase scenarios.
    - *(out of scope of this PR)* We're also discussing whether the Invoice Issuer should be limited to entities issuing only payable invoices.
  - **BilledCost requirement:** *The sum of the BilledCost for a given InvoiceId MUST match the sum of the payable amount provided in the corresponding invoice with the same id generated by the InvoiceIssuerName.*
  - **BillingCurrency requirement:** *BillingCurrency MUST match the currency used in the invoice generated by the Invoice Issuer.*

- Regarding the broader interpretation - that all costs from all invoices ever issued must (or should) be present in the FOCUS dataset - I agree this represents our **north star** goal.
  While it's unlikely that all providers or Invoice Issuers will be able to fully meet this (especially due to legacy systems), it's still the direction we want to move toward, at least for **payable costs** and **issued invoices**.
  - The key question is: **Do we want the requirement to reflect our ideal target (MUST), or do we want to accommodate provider, i.e., invoice issuer limitations (SHOULD)?**  
    *I believe I've heard this question before* 😄

```suggestion
* All payable costs that appear on an invoice MUST be present in a *FOCUS dataset*.
* All non-payable costs that appear on an invoice SHOULD be present in a *FOCUS dataset*.
```

ijurica (approved):

specification/attributes/invoice_handling.md
Matt-Cowsert: We agreed during the TF-3 call on 9/12 to revert to the language in the original PR, which is simpler

specification/attributes/invoice_handling.md
Matt-Cowsert: Hey @flanakin . We discussed this specific point on the call with TF-3 last Friday. The consensus we reached on that call was to state the requirement as a MUST and recognize that not all providers will be able to meet this requirement. From a practical standpoint, I expect this requirement to sit outside of the "essential" list of requirements that need to be met for conformance certification. Notes from the call captured [here](https://github.com/FinOps-Open-Cost-and-Usage-Spec/FOCUS_Spec/issues/1082#issuecomment-3293526565)

flanakin: If it's not essential, why is it a MUST?

shawnalpay: Discussed in Sep 18 Members.  While "SHOULD" or "MUST" are perhaps both acceptable for both @kk09v and @flanakin, what is more at issue is the scope of what "all" means.

shawnalpay: Per @flanakin: maybe there is utility to saying "you SHOULD include X" and "you MUST include Y".

specification/attributes/invoice_handling.md
```diff
@@ -0,0 +1,30 @@
...
+
+Invoice Handling
+
+## Description
+
+Indicates how invoice-level *charges*, including those not directly tied to usage, should be represented in a FOCUS Cost and Usage dataset.
+
+## Requirements
+
+* All costs that appear on an invoice MUST be included in the FOCUS Cost and Usage dataset.
```
ijurica: perhaps?

```suggestion
* All costs that appear on an invoice related to Information Technology MUST be included in the FOCUS Cost and Usage dataset.
```

shawnalpay: Perhaps... but even then, this PR would need a section defining what _that_ means.  Per feedback in the Members meeting today, I don't think there's a way to escape some scope of charges defined, either in this attribute or in the glossary, and then reference that.

specification/attributes/invoice_handling.md
shawnalpay: Discussed in Sep 19 TF3.  We all like the idea @kk09v offered to @ijurica yesterday around tying this to BillingAccountId.  I will coordinate with him to get a suggestion put forward.

shawnalpay:
Good discussion on this PR in TF3 today!  We will loop in @kk09v on how to incorporate Billing Account ID in as a normative requirement for when to include costs, and then we can likely bring this to the Members for final approval next Thu.

specification/attributes/invoice_handling.md
Matt-Cowsert: I took a stab at the scope, tying the scope of charges to the billing account (thanks for that recommendation @kk09v ). 

I also added some additional details in the exceptions section for informational/supplemental details that sometimes show up on the invoice of certain providers. If we don't think the clarification is needed at this time, I'm fine with striking it, but we'll want to revisit this in future releases.

specification/attributes/invoice_handling.md
```diff
@@ -18,10 +18,16 @@ Indicates how invoice-level *charges*, including those not directly tied to usag
 
 ## Requirements
 
-* All costs that appear on an invoice MUST be included in the FOCUS Cost and Usage dataset.
+* All costs that appear on an invoice issued to a [*Billing Account*](#glossary:BillingAccount) MUST be included in the FOCUS Cost and Usage dataset.
```
shawnalpay:
```suggestion
* All costs that appear on an invoice issued to a [*BillingAccountId*](#billingaccountid) MUST be included in the FOCUS Cost and Usage dataset.
```

specification/attributes/invoice_handling.md
```diff
@@ -0,0 +1,37 @@
...
+
+Invoice Handling
+
+## Description
+
+Indicates how invoice-level *charges*, including those not directly tied to usage, should be represented in a FOCUS Cost and Usage dataset.
+
+## Requirements
+
+* All costs that appear on an invoice issued to a [*BillingAccountId*](#billingaccountid) MUST be included in the FOCUS Cost and Usage dataset.
```
ijurica: Assuming our intent is to ensure that **all invoices related to a billing account** are reflected in the Cost and Usage dataset I suggest:

```suggestion
* All costs that appear on any invoice issued to a [*BillingAccountId*](#billingaccountid) MUST be included in the FOCUS Cost and Usage dataset.
```

specification/attributes/invoice_handling.md
```diff
@@ -0,0 +1,37 @@
...
+* If an invoice-level *charge* appears on a customer invoice but cannot be expressed using existing FOCUS columns, providers MUST include provider-defined columns (e.g., x_ChargeSubType) to capture the non-FOCUS-defined details needed to support invoice *charges* reconciliation using the FOCUS Cost and Usage dataset.
+
+## Exceptions
+
+* Informational line items that have zero monetary impact and are included solely for transparency MAY be excluded. Examples include:
+  * Free tier usage tracking with $0.00 charges
+  * Breakdowns of bundled services where only the bundle total is billed
```
ijurica: I believe these information should (even must) be included in the Cost and Usage dataset artifacts, as the dataset is intended to reflect both cost and usage, not just monetary impact.

specification/attributes/invoice_handling.md
shawnalpay: Agree.

specification/attributes/invoice_handling.md
shawnalpay: Agree.  @ijurica Are you suggesting we strike these lines?

Matt-Cowsert: Understood, will strike the bundles service line. @ijurica are you fine with the other expections listed?

specification/attributes/invoice_handling.md
Matt-Cowsert: resolving this as I believe we've landed on the scope we're envisioning

specification/attributes/invoice_handling.md
shawnalpay: @Matt-Cowsert, you have now addressed this via the addition of verbiage for BillingAccountId. 

I believe you can now resolve this comment, and leave it to @flanakin to reopen if necessary.

specification/attributes/invoice_handling.md
shawnalpay: @Matt-Cowsert This is taking issue with both of the highlighted lines, which includes `Free tier usage tracking with $0.00 charges`.  I agree that that is not a valid exception, as that is meaningful information.

Both of those lines should be struck, and then I believe you can resolve this.

Matt-Cowsert: understood, thanks for the clarification @shawnalpay making the adjustment now

shawnalpay:
All comments resolved!  Nice work @Matt-Cowsert.  Marking as ready for approval!

jpradocueva:
PR was approved during the Members’ Meeting on September 25.

Issue #1311, opened 2025-08-05
[AI] Draft PR for Contract columns in Cost and Usage dataset


shawnalpay:
### Detailed Description

A strawman data model presenting various options was completed as part of AI #1227.  The first part of this now needs to be moved forward in a pull request.

### Definition of Done

Author a pull request including column additions for the Cost and Usage dataset.

shawnalpay:
I created the PR for this last week, and we will review in TF1 tomorrow!  Closing.

Issue #1070, opened 2025-05-21
[FR] Clarify commitment concepts and glossary entries


Matt-Cowsert:
### 🧠 Problem Statement

Terminology aside, providers offer various Standard and Negotiated, Burn-Down and Build-Up Commitment-based discounts.

Since "Burn-Down" and "Build-Up" are not commonly used terms, I am including a brief description of these terms to clarify what I mean by them:

- Burndown (Consumption-based):
  - You purchase a commitment that you then consume (or "burn down") over time. At the end of the Commitment Cycle Interval, any unused commitment expires, and you may receive informative charge records for the unused commitment.
  - various SaaS commitments,  (non-funded) OCI Universal Credits, etc.

- Build-Up (Contribution-based):
  - You commit to a spend that you build up over time to fulfill the requirement. Your expenditures (billed costs) accumulate to meet the required commitment (contributing to fulfilling your overall commitment). If you do not meet the commitment, you should expect a true-up payment charge record.
  - Microsoft MACC, AWS EDP

Considering that FOCUS 1.1 is limited to Standard Commitment-based Discounts (primarily comes down to RIs and Savings Plans), it no longer specifies what’s expected regarding Negotiated Burn-down Commitments (such as most SaaS commitments, non-funded OCI Universal Credits, etc.)

**Concerns**

The FOCUS specification uses the umbrella term *Commitment Discount* in a way that:

- excludes negotiated agreements 
- and assumes a pre-purchased, burn-down style of commitment (see attached glossary definition).

While this may work well — or perhaps it's simply what we're used to — in the context of public cloud, **it becomes problematic when applied more broadly**, such as to the SaaS scope, because:

- Negotiated discounts also involve customer commitments.
- Negotiated discounts may functionally behave like commitment discounts -  they are not limited to build-up models (e.g., AWS EDP, Microsoft MACC), but can also include burn-down models (e.g. OCI Universal Credits).

The fact that, throughout the specification, we repeatedly add **explicit clarifications** that *Commitment Discounts* do **not** refer to *Negotiated Discounts* indicates that the current terminology does not provide enough clarity on its own. For example:

> *...The CommitmentDiscountId column is only applicable to [commitment discounts](#glossary:commitment-discount) and not [negotiated discounts](#glossary:negotiated-discount)...*

*(Note: Similar disclaimers appear in multiple commitment-discount-related column specifications.)*

---

### 🧪 Use Case

[...]

Issue #1227, opened 2025-07-29
[AI] Generate strawman of data model for TF consideration


shawnalpay:
### Definition of Done

A proposal of how to model the data to facilitate negotiated commitments, with an initial presentation given to the Task Force.

shawnalpay:
[Here](https://docs.google.com/spreadsheets/d/1l42uPa_XhHTgnaOW1tFHQBGL7dEIss3-onsSuY5qVuc/edit?gid=0#gid=0) is a sketch of possible data model options.  I have included the contract line items even though they are out of scope for 1.3 because I believe it important to have a vision for where the data model will go over time, and it would be ideal not to paint ourselves into a corner.

- `1` Separate dimension tables for Contract and Contract Commitment
- `2` Combined dimension tables for Contract and Contract Commitment
- `3` Single table with properties as separate fields
  -  Not a real option IMO, but included in order to demonstrate why
- `4` Single table with properties as JSON
- `5` Single table with only IDs (no properties)

These examples assume application of multiple commitments to a single Cost and Usage row.

I have also made plenty of choices around column names.  These options are meant more to demonstrate how we might model the data; column names are still up for debate, but I hope these placeholders at least convey the meaning.

shawnalpay:
Reviewed the options with the attendees of Aug 5 TF1.

 - Agreement that **Option 1 is preferred**, with **Option 5 being a starting point**.
 - Option 4 would follow the pattern established by `SkuPriceDetails`, but it would not move us forward in the establishment of multiple FOCUS datasets, so **Option 4 is not preferred**.
 - **Options 2 and 3 are not feasible.**

This AI is now complete, and I can proceed with an AI drafting a PR for TF review in next week's meeting!

shawnalpay:
Reopening this design issue, as I have sustained enough comments in Aug 12 TF1 and on PR #1327 for me to believe **I needed to rethink the design**.

There are a few issues that folks have pointed out with Option 1 since I presented yesterday (y'all, I'm paraphrasing -- please correct me if I misunderstood your points):

- @ijurica: If we model the value of the contract at the Contract level, then how do we know which of the Contract Commitment cost/quantity rows should match the Contract cost/quantity?  These are potentially inherently different, which is confusing.
- @flanakin: usage of the contract needs to happen at the Commitment level, not the Contract level.
[...]

PR #1327, opened 2025-08-07
AI #1311: Add Contract columns to Cost and Usage dataset

shawnalpay:
This PR adds the following columns to the Cost and Usage dataset:

 * `Contract ID`: the unique identifier of the contract representing one or more commitments associated with discounts negotiated between the provider and the customer
 * `Contract Applied Cost`: the cost of the charge applied to the contract
 * `Contract Applied Quantity`: the quantity of the charge applied to the contract

The scope of FR #1070 also includes the creation of a separate dataset representing Contracts.  That shall be surfaced as a separate pull request, as it is not yet clear whether we shall proceed with the addition of multiple datasets in FOCUS 1.3.  The full vision is described in [this spreadsheet](https://docs.google.com/spreadsheets/d/1l42uPa_XhHTgnaOW1tFHQBGL7dEIss3-onsSuY5qVuc/edit?gid=1641662850#gid=1641662850).

### DIFF
--- a/specification/columns/contractappliedcost.md
+++ b/specification/columns/contractappliedcost.md
@@ -1,0 +1,39 @@
+# Contract Applied Cost
+
+Contract Applied Cost represents the cost of the charge applied to the contract.  Contract Applied Cost is applied to the contract via [Contract ID](#contractid).  Contract Applied Cost is commonly used for monitoring the progress towards fulfilling contractual commitments that facilitate discounts for [*resources*](#glossary:resource) or [*services*](#glossary:service) as negotiated between a provider and a customer.
+
+The ContractAppliedCost column adheres to the following requirements:
+
+* ContractAppliedCost MUST be present in a [*FOCUS dataset*](#glossary:FOCUS-dataset) when the provider supports *negotiated discounts*.
+* ContractAppliedCost MUST be of type Decimal.
+* ContractAppliedCost MUST conform to [NumericFormat](#numericformat) requirements.
+* ContractAppliedCost MUST NOT be null when [Contract ID](#contractid) is not null.
+* ContractAppliedCost MUST be a valid decimal value.
+* ContractAppliedCost MUST be denominated in the BillingCurrency.
+
+## Column ID
+
+ContractAppliedCost
+
+## Display Name
+
+Contract Applied Cost
+
+## Description
+
+Cost calculated by multiplying *contracted unit price* and the corresponding Pricing Quantity.
+
+## Content Constraints
+
+| Constraint      | Value                   |
+|:----------------|:------------------------|
+| Column type     | Metric                  |
+| Feature level   | Conditional             |
+| Allows nulls    | True                    |
+| Data type       | Decimal                 |
+| Value format    | [Numeric Format](#numericformat) |
+| Number range    | Any valid decimal value |
+
+## Introduced (version)
+
+1.3
--- a/specification/columns/contractappliedcost.md
+++ b/specification/columns/contractappliedcost.md
@@ -7,7 +7,9 @@ The ContractAppliedCost column adheres to the following requirements:
 * ContractAppliedCost MUST be present in a [*FOCUS dataset*](#glossary:FOCUS-dataset) when the provider supports *negotiated discounts*.
 * ContractAppliedCost MUST be of type Decimal.
 * ContractAppliedCost MUST conform to [NumericFormat](#numericformat) requirements.
-* ContractAppliedCost MUST NOT be null when [Contract ID](#contractid) is not null.
+* ContractAppliedCost nullability is defined as follows:
+  * ContractAppliedCost MUST be null when [Contract ID](#contractid) is null.
+  * ContractAppliedCost MUST NOT be null when [Contract ID](#contractid) is not null.
 * ContractAppliedCost MUST be a valid decimal value.
 * ContractAppliedCost MUST be denominated in the BillingCurrency.
 
@@ -21,7 +23,7 @@ Contract Applied Cost
 
 ## Description
 
-Cost calculated by multiplying *contracted unit price* and the corresponding Pricing Quantity.
+The cost of the *charge* applied to the contract.
 
 ## Content Constraints
 
--- a/specification/columns/contractappliedquantity.md
+++ b/specification/columns/contractappliedquantity.md
@@ -1,0 +1,40 @@
+# Contract Applied Quantity
+
+Contract Applied Quantity represents the quantity of the charge applied to the contract.  Contract Applied Quantity is applied to the contract via [Contract ID](#contractid).  Contract Applied Quantity is commonly used for monitoring the progress towards fulfilling contractual commitments that facilitate discounts for [*resources*](#glossary:resource) or [*services*](#glossary:service) as negotiated between a provider and a customer.
+
+The ContractAppliedQuantity column adheres to the following requirements:
+
+* ContractAppliedQuantity MUST be present in a [*FOCUS dataset*](#glossary:FOCUS-dataset) when the provider supports *negotiated discounts*.
+* ContractAppliedQuantity MUST be of type Decimal.
+* ContractAppliedQuantity MUST conform to [NumericFormat](#numericformat) requirements.
+* ContractAppliedQuantity nullability is defined as follows:
+  * ContractAppliedQuantity MUST be null when [Contract ID](#contractid) is null.
+  * ContractAppliedQuantity MUST NOT be null when [Contract ID](#contractid) is not null.
+* ContractAppliedQuantity MUST be a valid decimal value.
+
+## Column ID
+
+ContractAppliedQuantity
+
+## Display Name
+
+Contract Applied Quantity
+
+## Description
+
+The quantity of the *charge* applied to the contract.
+
+## Content Constraints
+
+| Constraint      | Value                   |
+|:----------------|:------------------------|
+| Column type     | Metric                  |
+| Feature level   | Conditional             |
+| Allows nulls    | True                    |
+| Data type       | Decimal                 |
+| Value format    | [Numeric Format](#numericformat) |
+| Number range    | Any valid decimal value |
+
+## Introduced (version)
+
+1.3
--- a/specification/columns/contractid.md
+++ b/specification/columns/contractid.md
@@ -1,6 +1,6 @@
 # Contract ID
 
-A Contract ID is a provider-assigned identifier for a contract describing the terms of one or more [*negotiated discounts*](#glossary:negotiated-discount) between a provider and a customer. Such discounts are typically the result of a customer committing to a certain amount of spend or usage over an agreed period of time.
+A Contract ID is a provider-assigned identifier for a contract describing the agreed terms of one or more [*negotiated discounts*](#glossary:negotiated-discount) between a provider and a customer. Such discounts are typically the result of a customer committing to a certain amount of spend or usage over an agreed period of time.
 
 The ContractId column adheres to the following requirements:
 
--- a/specification/columns/contractid.md
+++ b/specification/columns/contractid.md
@@ -1,6 +1,6 @@
 # Contract ID
 
-A Contract ID is a provider-assigned identifier for a contract describing the agreed terms of one or more [*negotiated discounts*](#glossary:negotiated-discount) between a provider and a customer. Such discounts are typically the result of a customer committing to a certain amount of spend or usage over an agreed period of time.
+A Contract ID is a provider-assigned identifier for a contract describing the agreed terms between a provider and a customer.  Contracts often describe one or more [*negotiated discounts*](#glossary:negotiated-discount). Contracts can include commitment to a certain amount of spend or usage over an agreed period of time.
 
 The ContractId column adheres to the following requirements:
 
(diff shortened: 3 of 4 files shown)


specification/columns/contractid.md
```diff
@@ -0,0 +1,41 @@
+# Contract ID
+
+A Contract ID is a provider-assigned identifier for a contract describing the agreed terms of one or more [*negotiated discounts*](#glossary:negotiated-discount) between a provider and a customer. Such discounts are typically the result of a customer committing to a certain amount of spend or usage over an agreed period of time.
```
rileyjenk: Contracts do not require negotiated discounts. This implies that contracts must include negotiated discounts, when that is not the case.

shawnalpay: Fair point.  I would love a suggestion, if you have it!  How about this:

```suggestion
A Contract ID is a provider-assigned identifier for a contract describing the agreed terms between a provider and a customer.  Contracts often describe one or more [*negotiated discounts*](#glossary:negotiated-discount), obtained through the customer's commitment to a certain amount of spend or usage over an agreed period of time.
```

rileyjenk: Minor change as contracts can include neither spend or usage commits.  You could go even further and remove the spend or usage commit entirely. 

```suggestion
A Contract ID is a provider-assigned identifier for a contract describing the agreed terms between a provider and a customer.  Contracts often describe one or more [*negotiated discounts*](#glossary:negotiated-discount). Contracts can include commitment to a certain amount of spend or usage over an agreed period of time.
```

shawnalpay: I'd like to keep that reference because, while you are correct that they may not include such commitments, they represent the bulk of what we're considering here. 

Accepting your suggestion!

specification/columns/contractid.md
```diff
@@ -0,0 +1,41 @@
+# Contract ID
+
+A Contract ID is a provider-assigned identifier for a contract describing the agreed terms between a provider and a customer.  Contracts often describe one or more [*negotiated discounts*](#glossary:negotiated-discount). Contracts can include commitment to a certain amount of spend or usage over an agreed period of time.
```
Matt-Cowsert, line 3:
```suggestion
A Contract ID is a provider-assigned identifier for a contract describing the agreed terms between a provider and a customer. Contracts often describe one or more [*negotiated discounts*](#glossary:negotiated-discount). Contracts can include time-bound spend or usage commitments.
```

specification/columns/contractid.md
```diff
@@ -0,0 +1,41 @@
...
+
+The ContractId column adheres to the following requirements:
+
+* ContractId MUST be present in a [*FOCUS dataset*](#glossary:FOCUS-dataset) when the provider supports *negotiated discounts*.
```
shawnalpay, line 7: We may need to revisit the conditional nature of this requirement, per discussion in Aug 12 TF1.

@marc-perreaut and @rileyjenk will produce suggestions.

shawnalpay: @SatyaVandrangi1 is concerned that if we make this more globally applicable in terms of presence, then we are not necessarily facilitating just an MVP of simple drawdown contracts in 1.3.

marc-perreaut: I would tend to think the Contract ID should be always present, as there should be always a contract when charging cost.
Indeed, the Contracted Cost column must be always present.

specification/columns/contractappliedcost.md
```diff
@@ -0,0 +1,41 @@
...
+Contract Applied Cost represents the cost of the charge applied to the contract.  Contract Applied Cost is applied to the contract via [Contract ID](#contractid).  Contract Applied Cost is commonly used for monitoring the progress towards fulfilling contractual commitments that facilitate discounts for [*resources*](#glossary:resource) or [*services*](#glossary:service) as negotiated between a provider and a customer.
+
+The ContractAppliedCost column adheres to the following requirements:
+
+* ContractAppliedCost MUST be present in a [*FOCUS dataset*](#glossary:FOCUS-dataset) when the provider supports *negotiated discounts*.
+* ContractAppliedCost MUST be of type Decimal.
+* ContractAppliedCost MUST conform to [NumericFormat](#numericformat) requirements.
+* ContractAppliedCost nullability is defined as follows:
+  * ContractAppliedCost MUST be null when [Contract ID](#contractid) is null.
+  * ContractAppliedCost MUST NOT be null when [Contract ID](#contractid) is not null.
```
shawnalpay, line 12: @ijurica points out that all of our cost columns in FOCUS so far must be present and null is not allowed.  This would set a precedent of allowing nulls in a cost column, which could introduce an issue for practitioners that are aggregating whole datasets that include nulls.

specification/columns/contractappliedcost.md
```diff
@@ -0,0 +1,41 @@
+# Contract Applied Cost
```
ijurica, line 1: We expect multiple Contract Commitment IDs, and the associated Contract Commitments Applied Cost values may vary. If that’s the case (i.e., one size doesn’t fit all coexisting commitments), what should we expect in Contract Applied Cost and Contract Applied Quantity?

shawnalpay: The application of varying costs and quantities to the Commitment level would need to be handled in a separate column -- likely JSON, per my current thinking -- that would be a list of Contract Commitment IDs and the applicable costs/quantities.  `Contract Applied Cost` and `Contract Applied Quantity` would be the application at the Contract level.  [Here](https://docs.google.com/spreadsheets/d/1l42uPa_XhHTgnaOW1tFHQBGL7dEIss3-onsSuY5qVuc/edit?gid=1641662850#gid=1641662850) is the conceptual sketch I did for these ideas.

specification/columns/contractappliedcost.md
```diff
@@ -0,0 +1,41 @@
...
+
+Contract Applied Cost represents the cost of the charge applied to the contract.  Contract Applied Cost is applied to the contract via [Contract ID](#contractid).  Contract Applied Cost is commonly used for monitoring the progress towards fulfilling contractual commitments that facilitate discounts for [*resources*](#glossary:resource) or [*services*](#glossary:service) as negotiated between a provider and a customer.
```
flanakin, line 3: nit: "Contract Applied Cost is applied to the contract..." presents itself as self-defining, but I'm not sure it is. I think I understand what this is covering... One possible idea:
```suggestion
Contract Applied Cost is the portion of a charge counted toward a specific contract, identified by its [Contract ID](#contractid). Contract Applied Cost is commonly used for monitoring the progress towards fulfilling contractual commitments that facilitate discounts for [*resources*](#glossary:resource) or [*services*](#glossary:service) as negotiated between a provider and a customer.
```

Fwiw, it's not really applied to the contract. It's applied to a commitment.

specification/columns/contractappliedcost.md
flanakin, line 1: Is this really applied to the contract? Isn't it applied to a commitment, which is separate from the contract? For Microsoft it is, at least. For instance, MACC is separate from the contract itself. MACC is not tied to a discount. (Although many think it is.) MACC is a negotiating tool sales teams use to establish a commitment, which makes them more comfortable with higher discounts. But the MACC is not the same as the contract.

specification/columns/contractappliedcost.md
```diff
@@ -0,0 +1,41 @@
...
+
+The ContractAppliedCost column adheres to the following requirements:
+
+* ContractAppliedCost MUST be present in a [*FOCUS dataset*](#glossary:FOCUS-dataset) when the provider supports *negotiated discounts*.
```
flanakin, line 7: Shouldn't it only be required when there's a commitment-tracking mechanism? Someone may offer negotiated discounts where everything applies and there's no need to track what was or wasn't applied, right?

specification/columns/contractappliedcost.md
flanakin: Fwiw, I think nulls are inevitable. If we don't allow nulls, you're going to get a `0` when that is factually incorrect. That also goes against our null handling requirements, which say you can't use placeholder values when there is no value. `0` would be a placeholder value in this case, which sends the wrong message.

FYI: Microsoft uses `0` in place of a real price/cost when they don't have the data. This means practitioners can't tell the difference between something being free and the data not being available.

specification/columns/contractappliedquantity.md
```diff
@@ -0,0 +1,40 @@
+# Contract Applied Quantity
```
flanakin, line 1: Is this intended to cover a scenario where someone commits to a specific quantity of usage, rather than cost? Is this required that's not the case? Is ContractAppliedCost required when quantity is committed? I'm not sure how a quantity can be applied to a cost commitment, so this feels confusing unless it's null when not applicable. Same for the ContractAppliedCost column when there's a quantity commitment.

specification/columns/contractappliedquantity.md
```diff
@@ -0,0 +1,40 @@
...
+
+Contract Applied Quantity represents the quantity of the charge applied to the contract.  Contract Applied Quantity is applied to the contract via [Contract ID](#contractid).  Contract Applied Quantity is commonly used for monitoring the progress towards fulfilling contractual commitments that facilitate discounts for [*resources*](#glossary:resource) or [*services*](#glossary:service) as negotiated between a provider and a customer.
+
+The ContractAppliedQuantity column adheres to the following requirements:
+
+* ContractAppliedQuantity MUST be present in a [*FOCUS dataset*](#glossary:FOCUS-dataset) when the provider supports *negotiated discounts*.
```
flanakin, line 7: I feel like this should only be when there are quantity-based commitments.

specification/columns/contractid.md
flanakin: I would agree there's always a contract, but that is separate from the commitment, which I believe is what this is trying to track. I don't think we can merge those concepts.

specification/columns/contractappliedcost.md
ijurica: I think I didn’t phrase my question very well.

Let’s imagine this scenario:

* A contract has multiple build-up commitments that can be linked to a specific charge.
* Each commitment has its own Contract Commitments Applied Cost.

In that case, what should we put in Contract Applied Cost?

Based on the following two requirements, **it MUST NOT be null**, but **there is no single value that applies to all associated commitments**:

> * ContractId MUST NOT be null when a *charge* is related to a *negotiated discount*.
> * ContractAppliedCost MUST NOT be null when [Contract ID](#contractid) is not null.

shawnalpay: @flanakin 
The idea is to eventually model _both_ the Contract _and_ the Commitments that are described in the Contract.  [Here](https://docs.google.com/spreadsheets/d/1l42uPa_XhHTgnaOW1tFHQBGL7dEIss3-onsSuY5qVuc/edit?gid=1641662850#gid=1641662850) is the working spreadsheet I've presented in the last few TF1 calls that describes the approach.  Contracts are modeled in 1.3; Contract Commitments are modeled in 1.4.  The TF1 attendees have agreed with this approach, namely: @ijurica, @rileyjenk, @deejacruz, @cnharris10, @marc-perreaut, and @Matt-Cowsert (though I believe @rileyjenk would like to model it all at once to increase value).

In the Microsoft world, are you saying that the contract and the MACC have different IDs, and that the MACC is a child of the contract?  Or are they wholly separate entities?  I'd be curious for you have a look at that spreadsheet I linked and tell me how those items would fit (or not) into the proposed schema.

I'd be happy to go over this with you 1:1 between now and next TF1 -- or if you want to come to TF1 next week, we can discuss further.

specification/columns/contractappliedcost.md
shawnalpay: A change is likely needed here, yes.  Per [this comment](https://github.com/FinOps-Open-Cost-and-Usage-Spec/FOCUS_Spec/pull/1327#discussion_r2270323894) on the related presence requirements for Contract ID, both @rileyjenk and @marc-perreaut expressed a concern in today's TF1 about this condition (or indeed _any_ condition), and they are going to produce a suggested revision (which would apply to this line too).

specification/columns/contractappliedcost.md
shawnalpay: It's a good question.  Maybe it means that we need to model burndown and build-up in the same level of Contract Commitment, and jettison application at the Contract level...?

I've really wanted to keep build-up use cases out of scope, but maybe we have to consider it in order to avoid refactoring later...

specification/columns/contractappliedcost.md
shawnalpay: @flanakin I agree.  However, we have a precedent set of keeping all cost columns null -- so this would introduce a difference that we would have to make sure we explain clearly, either here within the spec, or as part of our supporting content.

specification/columns/contractappliedquantity.md
shawnalpay: I think you're right.  I'll revise.

shawnalpay:
Putting this PR back into draft, based on feedback so far, and based on the need to rethink the design per AI #1227.

We will discuss the design, and from there, I will either retool this PR or create a new one

shawnalpay:
I am closing this PR in favor of #1349.  Thank you @ijurica @rileyjenk @flanakin @marc-perreaut @mike-finopsorg for your feedback!  It was very helpful in surfacing the second draft.

PR #1337, opened 2025-08-12
FR #1090: Add multi dataset support to FOCUS spec layout

### LINKED ISSUE #1090: [FR] Add Structured "BI" schema to FOCUS

mike-finopsorg:
This pull request updates the repo layout to support multiple FOCUS datasets


specification/datasets/focus/focus.mdpp
shawnalpay, line 1: As I've been socializing for the last couple of weeks: I propose calling this the `Cost and Usage` dataset.

If agreed, this would become `costandusage.mdpp`.

specification/datasets/focus/focus.mdpp
```diff
@@ -0,0 +1,7 @@
+# FOCUS
+
+The FOCUS dataset is the primary dataset for FOCUS cost and usage data.
+
+The FOCUS specification for FOCUS datasets defines a group of columns that provide qualitative values (such as dates, resource, and provider information) categorized as "dimensions" and quantitative values (numeric values) categorized as "metrics" that can be used for performing various [FinOps capabilities][FODOFC]. Metrics are commonly used for aggregations (sum, multiplication, averaging etc.) and statistical operations within the dataset. Dimensions are commonly used to categorize, filter, and reveal details in your data when combined with metrics. The Columns are presented in Alphabetical order.
```
shawnalpay:
```suggestion
# Cost and Usage

The Cost and Usage dataset is the primary dataset for FOCUS cost and usage data.

The specification for the Cost and Usage dataset defines a group of columns that provide qualitative values (such as dates, resource, and provider information) categorized as "dimensions" and quantitative values (numeric values) categorized as "metrics" that can be used for performing various [FinOps capabilities][FODOFC]. Metrics are commonly used for aggregations (sum, multiplication, averaging etc.) and statistical operations within the dataset. Dimensions are commonly used to categorize, filter, and reveal details in your data when combined with metrics. The columns are presented in alphabetical order.
```

specification/Makefile
```diff
@@ -40,7 +40,7 @@ spec.md: $(SPEC_SOURCE_FILES)
 	  cp versions/${STYLE}.md version.md; \
 	fi
 	./validate_includes.py supported_features
-	./validate_includes.py columns
+	./validate_includes.py datasets/focus/columns
```
shawnalpay: We would need to change this to `datasets/costandusage/columns`.

specification/datasets/datasets.mdpp
```diff
@@ -0,0 +1,5 @@
+# Datasets
+
+FOCUS defines many individual datasets made up of a selected set of Columns which abide by the Attributes outlined in this FOCUS Specification.
+
+!INCLUDE "focus/focus.mdpp",1
```
shawnalpay:
```suggestion
!INCLUDE "costandusage/costandusage.mdpp",1
```

specification/glossary.md
```diff
@@ -94,7 +94,19 @@ An open-source specification that defines requirements for billing data.
 
 <a name="glossary:FOCUS-dataset"><b>FOCUS Dataset</b></a>
```
shawnalpay, line 95: I think the definition of `FOCUS Dataset` needs to continue being _any_ of the datasets within FOCUS.  From there, we do need two other terms to delineate between the two types.

I would propose `fact` and `dimension`, but that doesn't align with our existing FOCUS terminology.

Thus, perhaps we use `transaction` and `reference` to refer to tables like `Cost and Usage` and `Contract`, respectively.  Or terms along those lines.

shawnalpay (changes requested):
I would like to avoid referring to our "main" dataset as the `FOCUS` dataset.  That will cause a lot of confusion -- and I think that _any_ of the datasets defined in FOCUS should be a "FOCUS dataset".  Therefore, I propose calling our current, single dataset `Cost and Usage`.  The changes requested in this review make this terminology change.

specification/glossary.md
mike-finopsorg: I am open to suggestions here on what glossary terms we should include in the PR and the definitions they should have.

caligo-erik:
I'd distinguish between _human readable_ and _machine readable_.
What about:
_FOCUS Cost and Usage Dataset_ (human readable)
`FOCUS.CostAndUsage` (machine readable) // copying this idea from namespaces, objects, variables

So, in theory this could lead to a type/schema/interface definition where we can get the type
`FOCUS.CostAndUsage.BillingPeriod`
@rileyjenk thought of the same positive side effect of this choice. :smiley_cat:

specification/glossary.md
rileyjenk: I would not use the terms fact and dimension terms as they aren't exactly a total fit. I would refer to it as FOCUS includes datasets:
- cost and usage dataset
- invoice dataset
- contract dataset

rileyjenk:
Need to add the dataset construct in metadata requirements...not too hard just need to do it.

specification/glossary.md
shawnalpay: I think there's benefit down the road of having people understand we have a "main" transactional dataset, and some "supporting" reference tables.  But for simplicity and for now, skipping that is fine with.  So that means we would just strike the "FOCUS adjacent" section.

specification/Makefile
```diff
@@ -40,7 +40,7 @@ spec.md: $(SPEC_SOURCE_FILES)
 	  cp versions/${STYLE}.md version.md; \
 	fi
 	./validate_includes.py supported_features
-	./validate_includes.py columns
+	./validate_includes.py datasets/cost_and_usage/columns
```
flanakin, line 43: This approach of duplicating columns across each dataset won't scale if we need to copy the same columns everywhere. It would be much simpler to have a single library of columns and simply reference those columns from each individual dataset. Each dataset may add additional requirements for inter-relationships, but the core set of requirements for each construct should be applied wherever the column surfaces.

I think I brought this up before, but I think that was before I was out for a while, so not sure where the discussion went.

specification/datasets/data_model.md
```diff
@@ -0,0 +1,3 @@
+This document is where we declare the overal data model of the various tables in FOCUS.
```
flanakin:
```suggestion
This document is where we declare the overall data model of the various tables in FOCUS.
```

specification/datasets/datasets.mdpp
```diff
@@ -0,0 +1,5 @@
+# Datasets
+
+FOCUS defines many individual datasets made up of a selected set of Columns which abide by the Attributes outlined in this FOCUS Specification.
```
flanakin: Shouldn't "columns" be lowercased?
```suggestion
FOCUS defines many individual datasets made up of a selected set of columns which abide by the Attributes outlined in this FOCUS Specification.
```

specification/glossary.md
flanakin: I would prefer avoiding "primary" vs. "adjacent" datasets. They're all FOCUS datasets. We just need to clarify which one. Some people may only care about an adjacent one for specific efforts, so I don't think those labels add any value.

specification/spec.mdpp
```diff
@@ -21,7 +21,7 @@ FOCUS is an open specification for billing data. It defines a common schema for
 
 ## Table of Contents
 
-!TOC
+!TOC 2 2 3 2 2 2 2 2
```
flanakin, line 24: Is this intentional? Just checking.

specification/Makefile
shawnalpay: We discussed this point at length in Aug 4 Maintainers and closed the loop with consensus on Aug 11.  @mike-finopsorg mocked up the option that you describe per PR #1229. Mike and I were definitely in your camp until @rileyjenk made the excellent point that the same column will have different normative requirements across datasets.

For example: consider `Invoice ID`.  It would have a uniqueness requirement in an `Invoice` dataset that it would not carry in the `Cost and Usage` dataset.  Therefore, the markdown files would need to stay functionally separate.

From a data modeling perspective, if we have to "copy the same columns everywhere", we have done something fundamentally wrong.  The only fields that will overlap should be identifier columns that tie one dataset to the next.

specification/datasets/datasets.mdpp
shawnalpay: I'd say yes -- in which case, attributes also needs lower-casing.

specification/spec.mdpp
shawnalpay: Yes!  @mike-finopsorg tweaked the PDF builder to dynamically specify the level down to show in the table of contents.  So now you see three levels down in the Datasets section (3 Datasets -> 3.1 Cost and Usage -> 3.1.1 Availability Zone), whereas the other sections only go down two levels, as before.

specification/datasets/datasets.mdpp
shawnalpay:
```suggestion
FOCUS defines many individual datasets made up of a selected set of columns which abide by the attributes outlined in this FOCUS Specification.
```

specification/glossary.md
```diff
@@ -94,7 +94,19 @@ An open-source specification that defines requirements for billing data.
...
+
+<a name="glossary:FOCUS-adjactent-dataset"><b>FOCUS Adjacent Dataset</b></a>
+
+A FOCUS Adjacent Dataset is any dataset defined within the FOCUS specification that supports or augments the primary FOCUS dataset, but does not itself contain normalized cost and usage records.
+
+These datasets are designed to provide additional context, metadata, mapping, or enrichment information that enhances the interpretability or completeness of the FOCUS dataset. Examples include datasets for pricing reference data, commitment discount mapping, or region definitions.
+
+FOCUS Adjacent Datasets follow the same principles of structure and consistency defined by the specification, including column naming conventions and formatting expectations.
```
shawnalpay: Need to strike usage of "adjacent".  Just have an entry for FOCUS dataset.

mike-finopsorg:
```suggestion

A FOCUS dataset is a structured collection of records that conforms to the BCP14 criteria established by FOCUS. All columns included must be defined in the FOCUS Columns section of the specification.

In addition to these standardized columns, data generators may include custom provider-specific columns (prefixed with `x_`) where additional context is needed beyond what is captured in the defined FOCUS columns. If such custom columns introduce record-splitting (i.e., a single original charge results in multiple rows), the data generator is responsible for ensuring that all cost and quantity metrics still meet the aggregation and consistency rules required by the specification.

The collection of datasets are designed to provide billing insight, additional context, metadata, mapping, or enrichment information that enhances the interpretability or completeness.
```

specification/glossary.md
shawnalpay: Issued separate commit to address this.  Resolving.

specification/spec.mdpp
mike-finopsorg: What is probably not clear is what it's doing, it is setting the heading depth for each H2 level in the TOC.

shawnalpay: @mike-finopsorg will grab an AI to create a markdown guidelines on how mdpp works for the repo.  Resolving.

shawnalpay:
Reviewed in Aug 20 TF2.  Moving to Member review!

shawnalpay (approved):

ijurica (approved):

jpradocueva:
This PR was approved by the group without objections or abstentions on the Members' call on Aug 28.

PR #1349, opened 2025-08-14
AI #1311: Add Contract columns to Cost and Usage dataset (second draft)

shawnalpay:
This PR adds the following columns to the Cost and Usage dataset:

 * `Contract ID`: the unique identifier of the contract representing one or more commitments associated with discounts negotiated between the provider and the customer
 * `Contract Applied`: a set of four datapoints that associate a charge to one or more contract commitments, denoted as key-value pairs in JSON format
   * `Contract Commitment ID`: The unique identifier representing the contract line item.
   * `Contract Commitment Applied Cost`: The value of the charge applied to the contract line item.
   * `Contract Commitment Applied Quantity`: The usage of the charge applied to the contract line item.
   * `Contract Commitment Applied Unit`: The unit of measure for the usage of the charge applied to the contract line item.

This PR reflects the design decisions presented in [this modeling spreadsheet](https://docs.google.com/spreadsheets/d/1l42uPa_XhHTgnaOW1tFHQBGL7dEIss3-onsSuY5qVuc/edit?gid=1324762603#gid=1324762603).

The scope of FR #1070 also includes the creation of a separate dataset representing Contracts.  That shall be surfaced as a separate pull request, as it is not yet clear whether we shall proceed with the addition of multiple datasets in FOCUS 1.3.

This PR is an iteration of a first draft, PR #1327.


specification/columns/contractid.md
shawnalpay, line 1: Based on the data model design, Contract ID is not strictly necessary to include in the `Cost and Usage` dataset.  However, it is necessary for each charge to be applicable to one and only one Contract ID, and this makes that exceedingly clear.  It also solves for the use case of easy aggregation of activity of cost and usage by Contract ID and Contract Commitment ID without having to refer out to a separate dataset.

So, my current draft includes it here.  Open to feedback on that.

specification/columns/contractapplied.md
shawnalpay, line 1: Suuuuuper not sure how to best structure the sections of this column definition, given that the JSON column is, essentially, the combination of four columns, each of which carry separate normative requirements.

Definitely would love to hear opinions.

cnharris10: This data contains scalar, deterministic data.  Why is a JSON object preferred over 4 dedicated columns?

specification/columns/contractapplied.md
```diff
@@ -0,0 +1,112 @@
+# Contract Applied
+
+Contract Applied is a set of four datapoints that associate a charge to one or more [*contract commitments*](#glossary:contract-commitment), denoted as key-value pairs in JSON format.  Contract Applied allows the practitioner to track the progress of the commitments to which they have agreed with a provider.
+
+The datapoints are:
+
+* `Contract Commitment ID`: The unique identifier representing the contract line item.
+* `Contract Commitment Applied Cost`: The value of the charge applied to the contract line item.
```
rileyjenk, line 8: Is line item too nebulous? A line item may be interpreted as a specific discount, allocation, or feature requirement and not specifically what we would term as a commitment. Should we turn commitment into Commitment and have it be a reserved term in FOCUS land?

specification/columns/contractapplied.md
```diff
@@ -0,0 +1,112 @@
...
+A set of four datapoints that associate a charge to one or more [*contract commitments*](#glossary:contract-commitment).
+
+## Content Constraints
+
+| Constraint    | Value                              |
+| :------------ | :--------------------------------- |
+| Column type   | Dimension and Metric               |
+| Feature level | Conditional                        |
+| Allows nulls  | True                               |
+| Data type     | JSON                               |
```
rileyjenk, line 115: I think we might run into a problem here due to the restrictions on our kev-value format. Reason being is that you could have two commits that are applied to a charge. If we supported array value then it could be provided...except our key-value format doesn't allow sub objects or arrays.

rileyjenk: For example: I have a commitment to spend x amount of $ on RDS but have a more detailed commitment to use 1.5M r6g.xlarge Aurora compute hours. They can both apply. 

It could be shown:
```
ContractApplied: {
 [
     {
          "ContractCommitmentID": "12346", 
           "ContractCommitmentAppliedCost": 100.00, 
           "ContractCommitmentAppliedQuantity" : 0, 
           "ContractCommitmentAppliedUnit" : "compute_hours"
      },
      {
          "ContractCommitmentID": "23456", 
           "ContractCommitmentAppliedCost": 0, 
           "ContractCommitmentAppliedQuantity" : 10, 
           "ContractCommitmentAppliedUnit" : "compute_hours"
      }
]
```

specification/columns/contractid.md
rileyjenk: @shawnalpay I played around with this for a bit....I really like your approach coupled with the Contract Applied column.

specification/columns/contractapplied.md
shawnalpay: That's worth discussing.  It's possible that a 0 cost, 0 quantity line item that may be on a contract (think "free box of cigars every year").  It's not a commitment _per se_, but it's a benefit that results from the other commitments.  It's not directly applicable to cost and usage, though, so I could see the argument of throwing it out.  Open question in my mind right now.  What do you think?

specification/columns/contractapplied.md
shawnalpay: Met with @cnharris10 and discussed the above.  He and I are are aligned now.

specification/columns/contractapplied.md
```diff
@@ -0,0 +1,120 @@
...
+  * ContractCommitmentAppliedUnit MUST be null when ContractCommitmentAppliedQuantity is null.
+  * ContractCommitmentAppliedUnit MUST NOT be null when ContractCommitmentAppliedQuantity is not null.
+
+## Examples
+
+```json
+{
+    <<Need example here>>
+}
+```
```
shawnalpay, line 94: Per discussion in Aug 18 Maintainers -- would need a couple of examples to show how this would work.  I've put an `Examples` section here, following the precedent in `SkuPriceDetails`.  I will put together an example or two and add them here.

specification/columns/contractapplied.md
shawnalpay: @rileyjenk Totally agree that we will have to make a change to our key-value format restrictions in order to handle for this.  The example you've assembled is how I've been thinking about it conceptually.

specification/columns/contractapplied.md
shawnalpay: Applying suggestion from @rileyjenk per [this comment](https://github.com/FinOps-Open-Cost-and-Usage-Spec/FOCUS_Spec/pull/1349/files#r2278192323).

@rileyjenk Would we allow a provider to skip specifying the Unit if the Quantity is 0?

```suggestion
A single charge has two contractual commitments applied:  

1) Spend a certain amount of money on the relevant service. 
2) Consume a certain number of compute hours on the relevant resource type.

```json
{
ContractApplied: {
 [
     {
          "ContractCommitmentID": "12346", 
           "ContractCommitmentAppliedCost": 100.00, 
           "ContractCommitmentAppliedQuantity" : 0, 
           "ContractCommitmentAppliedUnit" : "compute_hours"
      },
      {
          "ContractCommitmentID": "23456", 
           "ContractCommitmentAppliedCost": 0, 
           "ContractCommitmentAppliedQuantity" : 10, 
           "ContractCommitmentAppliedUnit" : "compute_hours"
      }
]
}
```
```

shawnalpay:
I am closing this PR in favor of #1350, which more holistically describes the Contract Commitment approach.

I will reopen this PR if/when we reject the addition of a separate dataset in FOCUS 1.3.

PR #1350, opened 2025-08-14
FR #976: Add support for contract commitments

shawnalpay:
## Summary

This PR adds:
- A **new column called `Contract Applied`** to the `Cost and Usage` dataset.  This allows practitioners to see how their cost and usage applies to the progress of their various contract commitments.
- A **net-new dataset called `Contract Commitment`**. This allows practitioners to:
  - Analyze and aggregate their cost and usage by various contract commitment properties.
  - Easily see the values remaining on their contract commitments.

## Diagram
(screenshot from [this modeling spreadsheet](https://docs.google.com/spreadsheets/d/1l42uPa_XhHTgnaOW1tFHQBGL7dEIss3-onsSuY5qVuc/edit?gid=695370188#gid=695370188))

Screenshot 2025-09-19 at 9 30 49 PM

## Cost and Usage

 * `Contract Applied`: A set of five properties that associate a charge to one or more contract commitments, denoted in JSON.
   * `Contract ID`: A provider-assigned identifier for a contract describing the agreed terms between a provider and a customer.
   * `Contract Commitment ID`: The unique identifier representing the contract line item.
   * `Contract Commitment Applied Cost`: The value of the charge applied to the contract line item.
   * `Contract Commitment Applied Quantity`: The usage of the charge applied to the contract line item.
   * `Contract Commitment Applied Unit`: The unit of measure for the usage of the charge applied to the contract line item.
   * Zero or more custom key-value pairs, provided by the data generator.

### Columns

- `Contract Commitment ID`: A provider-assigned identifier describing an agreement negotiated between a provider and a customer.
- `Contract ID`: A provider-assigned identifier for a contract describing the agreed terms between a provider and a customer.
- `Contract Period Start`: The *inclusive start bound* of a *contract period*.
- `Contract Period End`: The *exclusive end bound* of a *contract period*.
- `Contract Commitment Period Start`: The *inclusive start bound* of a *contract commitment period*.
- `Contract Commitment Period End`: The *exclusive end bound* of a *contract commitment period*.
- `Contract Commitment Description`: The self-contained summary of the *contract commitment's* terms.
- `Contract Commitment Cost`: The monetary value of the *contract commitment*.
- `Contract Commitment Quantity`: The amount associated with the *contract commitment*.
- `Contract Commitment Unit`: A provider-specified measurement unit for the amount declared in Contract Commitment Quantity.
- `Contract Commitment Type`: A provider-assigned name to identify the type of *contract commitment*.
[...]

PR #1383, opened 2025-08-22
FR #1056: Create json_object_format.md

### LINKED ISSUE #1056: [FR] Add JSON data type and strategy for structured JSON columns
### 🧠 Problem Statement
The current spec does not have a clear definition for JSON data format in the focus Spec. Besides this, the FOCUS spec to date has attempted to solve taxonomical and structural data issues within our current Key Value structure. We have received feedback that this approach is limiting because of naming collisions. There are also limited ability to support complex data structures available in modern database technologies. This is of particular use when considering complex relationships or additional metadata that corresponds to a charge record.

We have multiple columns that use the key value data type, but the data type documentation is not clear from a ingestion perspective (JSON vs Serialized JSON STRING).

---
### 🧪 Use Case

As a FinOps Practitioner, I want to know the capabilities of the JSON data types so I can avoid naming collisions.  

---

- We have a more clear JSON Structure
- We decide if FOCUS support complex object structure
- Decide if we need to continue with prefix naming convention vs object structure in key value
- Decide on FOCUS identification vs data provider in the data.

- Support complex object structures in JSON (perhaps bounded)

---
### 📨 Type of Request
- [x] Refinement of an existing FOCUS attribute

---

- Domo

---

---
### 📊 Impacted Parties

---
### 🌫️ Level of Ambiguity

4

---

- Tag Issue: https://github.com/FinOps-Open-Cost-and-Usage-Spec/FOCUS_Spec/issues/540

---
- We have a more clear JSON Structure
- We decide if FOCUS support complex object structure
- Decide if we need to continue with prefix naming convention vs object structure in key value
- Decide on FOCUS identification vs data provider in the data.

---

  - Your **organization**

kk09v:
Reviving the Object format previously proposed by @flanakin because multiple FRs have had considerations to use a complex JSON object, which is not compatible with key-value format.

Changes from prior proposals:
- ~~Object -> JSON Object~~  Reverted because this is redundant with how FOCUS communicated data format.
  - Data Type will be JSON, so value format should just be Object Format since JSON is already specified.
  - Ref:
https://github.com/FinOps-Open-Cost-and-Usage-Spec/FOCUS_Spec/blob/05ce51a4690c9d458a3ad007348f5eafe091aa73/specification/columns/tags.md?plain=1#L85
https://github.com/FinOps-Open-Cost-and-Usage-Spec/FOCUS_Spec/blob/05ce51a4690c9d458a3ad007348f5eafe091aa73/specification/columns/tags.md?plain=1#L86
- Added a clarification that the JSON objects can be further specified or restricted by the column definition.
  - This is somewhat redundant because the columns restricting/specifying will enforce such restrictions with MUST clauses. 
  - I believe this clause eliminates possible ambiguity, and ECMA-404 similarly specifies that standards using JSON may add further restrictions (as we did with the key-value format): `It is expected that other standards will refer to this one, strictly adhering to the JSON syntax, while imposing semantics interpretation and restrictions on various encoding details.`


specification/attributes/json_object_format.md
```diff
@@ -0,0 +1,37 @@
...
+
+* JSON Object Format columns MUST contain a serialized JSON string, consistent with the [ECMA 404](https://www.ecma-international.org/wp-content/uploads/ECMA-404_2nd_edition_december_2017.pdf) definition of an object.
+* Object keys MUST be unique within an object.
+* Object values MUST be one of the following types: number, string, `true`, `false`, array, object, or `null`.
+* Object array elements MUST be one of the following types: number, string, or object.
+* Object array elements MUST all use the same, consistent type.
+* Object array elements MUST NOT be repeated.
+* Object array elements MUST NOT be null.
+* Objects MUST conform to all requirements of the corresponding column definition, which may specify or restrict the shape or contents of the JSON Object.
+* Objects are RECOMMENDED to not exceed 3 levels of nesting.
```
shawnalpay: So, I definitely need an attribute like this for the contract commitment work per #1350, but I think the most contentious part about it will be how we define nesting.  Would we really let a provider nest to hog heaven?  (That's a technical term.)

For my purposes in PR #1350, the TF would want to restrict the `Contract Applied` JSON nesting to one level only.  Should we require the spec author to specify the number of allowed levels at the column level, and mention that here as a note in the attribute normative requirements?

kk09v:
This PR should still define a way to communicate specific JSON object shapes, defined keys, defined values, and normative requirements.

[RFC 8927](https://www.ietf.org/rfc/rfc8927.html) provides the first, which is great for Conformance Requirements, but I believe the human readable element is missing.

kk09v:
Changed JSON Object Format to Object Format because JSON is redundant with how FOCUS communicates data format.
  - Data Type will be JSON, so value format should just be "Object Format" since JSON is already specified.
  - Example with key-value format (**not** named JSON key-value format)
https://github.com/FinOps-Open-Cost-and-Usage-Spec/FOCUS_Spec/blob/05ce51a4690c9d458a3ad007348f5eafe091aa73/specification/columns/tags.md?plain=1#L85
https://github.com/FinOps-Open-Cost-and-Usage-Spec/FOCUS_Spec/blob/05ce51a4690c9d458a3ad007348f5eafe091aa73/specification/columns/tags.md?plain=1#L86

kk09v:
In addition to creating an object format, we also need to standardize an approach to communicating normative requirements around the shape and contents of the JSON object.

- Prior attempts:
  -  https://github.com/FinOps-Open-Cost-and-Usage-Spec/FOCUS_Spec/blob/6af90c19a87348dbf38d14e4e342eef879f03cdc/specification/datasets/cost_and_usage/columns/contractapplied.md
  - https://github.com/FinOps-Open-Cost-and-Usage-Spec/FOCUS_Spec/blob/working_draft/specification/columns/skupricedetails.md#focus-defined-properties

- RFC 8927 is great for describing the overall shape for the Conformance Requirements, but I don't find it to be the most human-readable and I don't see our list of normative requirements for an object fitting easily into this.

I'm spitballing ideas, but here's the direction my brain is going so far:

For SkuPriceDetails 

_Normative requirements for the column_

| Parent | Key | Type | Extensible |
|---|---|---|---|
| null | null | Key-Value | TRUE |

_Description of object goes here_
- _Normative requirements for overall object go here_
- e.g. Provider-defined keys will be prefixed with "x_"

| Parent | Key | Type | Extensible |
|---|---|---|---|
| root | CoreCount | Number | FALSE |

FOCUS-defined property capturing the number of physical or virtual CPUs available
- CoreCount MUST be measured in quantity of cores

***
For ContractApplied

| Parent | Key | Type | Extensible |
|---|---|---|---|
| null | null | Object | FALSE |

_Description of object goes here_
- _Normative requirements for overall object go here_

| Parent | Key | Type | Extensible |
|---|---|---|---|
| root | ContractApplied | Key-Value | FALSE |

_Description of ContractApplied (object) goes here_
- Object is repeatable

| Parent | Key | Type | Extensible |
|---|---|---|---|
| ContractApplied | ContractCommitmentID | Number | FALSE |

A Contract Commitment ID is a provider-assigned identifier describing an agreement negotiated between a provider and a customer. Contracts can include commitment to a certain amount of spend or usage over an agreed period of time.

The ContractCommitmentID column adheres to the following requirements:

- ContractCommitmentID MUST be present in a FOCUS dataset when the provider supports contract commitments.

specification/attributes/object_format.md
```diff
@@ -0,0 +1,37 @@
...
+## Attribute Name
+
+Object Format
+
+## Description
+
+Rules and formatting requirements for columns appearing in a [*FOCUS dataset*](#glossary:FOCUS-dataset) that convey data as complex, hierarchical objects.
+
+## Requirements
+
```
shawnalpay: Discussed in Aug 25 Maintainers.  @ijurica suggests we should have a requirement that, for custom provider JSON Object columns, they MUST provide public documentation and/or schema metadata definition.

specification/attributes/object_format.md
```diff
@@ -0,0 +1,37 @@
...
+* Object Format columns MUST contain a serialized JSON string, consistent with the [ECMA 404](https://www.ecma-international.org/wp-content/uploads/ECMA-404_2nd_edition_december_2017.pdf) definition of an object.
+* Object keys MUST be unique within an object.
+* Object values MUST be one of the following types: number, string, `true`, `false`, array, object, or `null`.
+* Object array elements MUST be one of the following types: number, string, or object.
+* Object array elements MUST all use the same, consistent type.
+* Object array elements MUST NOT be repeated.
+* Object array elements MUST NOT be null.
+* Objects MUST conform to all requirements of the corresponding column definition, which may specify or restrict the shape or contents of the Object.
+* Objects are RECOMMENDED to not exceed 3 levels of nesting.
```
shawnalpay: Lots of conversation on nesting in Aug 25 Maintainers.  Need to align on nesting limit for this attribute.

specification/attributes/object_format.md
kk09v: Added a requirement for this

specification/attributes/object_format.md
```diff
@@ -0,0 +1,38 @@
+# Object Format
```
shawnalpay: Should we call this `JSON Object` instead of `Object` to be more clear?  Or is `Object` truly sufficient?

kk09v: I originally had JSON Object, but changed it to Object. I put the reasoning in the change comments.
https://github.com/FinOps-Open-Cost-and-Usage-Spec/FOCUS_Spec/pull/1383#issuecomment-3217693454

specification/attributes/json_object_format.md
shawnalpay: This comment is waaaaay out of date, given all the discussions we've had in the last two weeks.  Resolving in deference to other comments!

specification/attributes/object_format.md
```diff
@@ -0,0 +1,38 @@
...
+
+Columns that provide complex, hierarchical information can be used in place of separate columns to support more dynamic structures that may change across rows, but where the simpler [*Key-Value Format*](#key-valueformat) may not be sufficient. Objects extend key-value pairs to add support for complex data types like arrays and nested key-value pairs. Objects are also referred to as maps, trees, or hashtables.
```
ijurica: The current version describes Object format as an alternative to separate columns, but it doesn’t really highlight cases where it’s the only viable option (for example, in negotiated build-up commitments where splitting records isn’t applicable). 

Here is a suggestion, though I’m sure it can be phrased better 😄 

```suggestion
Objects extend [*Key-Value Format*](#key-valueformat) to support complex data types such as arrays and nested key-value pairs. This format is used when additional replacement columns or the simpler Key-Value Format are not sufficient to represent the complex data structure accurately. It can also be used when grouping related data within a single record is preferred over splitting records. Objects are also referred to as maps, trees, or hashtables.
```

specification/attributes/object_format.md
```diff
@@ -0,0 +1,38 @@
...
+* Object Format columns MUST contain a serialized JSON string, consistent with the [ECMA 404](https://www.ecma-international.org/wp-content/uploads/ECMA-404_2nd_edition_december_2017.pdf) definition of an object.
+* Object keys MUST be unique within an object.
+* Object values MUST be one of the following types: number, string, `true`, `false`, array, object, or `null`.
+* Object array elements MUST be one of the following types: number, string, or object.
+* Object array elements MUST all use the same, consistent type.
+* Object array elements MUST NOT be repeated.
+* Object array elements MUST NOT be null.
+* Objects MUST conform to all requirements of the corresponding column definition, which may specify or restrict the shape or contents of the Object.
+* Objects included in provider-defined [custom columns](#column_handling:custom-column) MUST have their object schema documented by the provider.
+* Objects are RECOMMENDED to not exceed 3 levels of nesting.
```
ijurica: Throughout the specification, RECOMMENDED is used only in presence-related requirements. All other requirements consistently use SHOULD and SHOULD NOT.

```suggestion
* Objects SHOULD NOT exceed 3 levels of nesting.
```

specification/attributes/json_object_format.md
```diff
@@ -0,0 +1,39 @@
...
+JSON Object Format
+
+## Description
+
+Rules and formatting requirements for columns appearing in a [*FOCUS dataset*](#glossary:FOCUS-dataset) that convey data as complex, hierarchical objects.
+
+## Requirements
+
+* JSONObjectFormat columns MUST contain a serialized JSON string, consistent with the [ECMA 404](https://www.ecma-international.org/wp-content/uploads/ECMA-404_2nd_edition_december_2017.pdf) definition of an object.
+* Objects used within ObjectFormat adhere to the following additional requirements:
```
shawnalpay:
```suggestion
* Objects used within JsonObjectFormat adhere to the following additional requirements:
```

specification/attributes/json_object_format.md
```diff
@@ -0,0 +1,39 @@
...
+  * Object keys MUST be unique within an object.
+  * Object values MUST be one of the following types: number, string, `true`, `false`, array, object, or `null`.
+* Arrays used within JSONObjectFormat adhere to the following additional requirements:
+  * Array elements MUST all use the same, consistent type.
```
shawnalpay, line 26: What does this mean?  Does `type` have a meaning here in JSON world?

specification/attributes/json_object_format.md
```diff
@@ -0,0 +1,39 @@
...
+  * Array elements MUST NOT be repeated.
+  * Array elements MUST NOT be null.
+* JSONObjectFormat columns MUST conform to all requirements of the corresponding column definition, which may specify or restrict the shape or contents of the Object.
+* Provider-defined [custom columns](#column_handling:custom-column) whose contents contain a JSON object MUST have their object schema documented by the provider.
+* Objects SHOULD NOT exceed 3 levels of nesting.
```
shawnalpay:
```suggestion
* JsonObjectFormat objects SHOULD NOT exceed 3 levels of nesting.
```

guidelines/normative-requirements-guidelines.md
```diff
@@ -69,7 +69,58 @@
 
 * **Default to Singular Form**: Column references should be singular, with the understanding that the requirement applies to all values in the column.
 
-### Additional Guidelines for Columns in Simplified JSON Format
+### Additional Guidelines for Columns in JSON Format
```
shawnalpay, line 79: Discussed with @kk09v.  He will scrub this content slightly to align with our decisions.

specification/attributes/object_format.md
shawnalpay: Discussed and resolved.

specification/attributes/object_format.md
shawnalpay: Committed and resolved.

guidelines/normative-requirements-guidelines.md
```diff
@@ -69,7 +69,58 @@
...
+      "RecommendedKey3" : 0.5
+    }, {
+      "RequiredKey1" : 0.1,
+      "RecommendedKey3" : 4,
+      "ProviderDefinedKey4": "SomeString"
+    } ]
+  }
+  ```
+
+  **Example JTD**
```
shawnalpay, line 112: @kk09v will take a straw poll in Sep 10 TF2 as to whether to retain JTD directly within the column definition vs putting it elsewhere (i.e., supporting content) vs cutting it completely.

guidelines/normative-requirements-guidelines.md
```diff
@@ -69,7 +69,67 @@
...
+  
+  **Example Pattern 2**
+
+  ```markdown
+  * <CostColumnId> MUST equal the product of <UnitPriceColumnId> and PricingQuantity when <UnitPriceColumnId> is not null and PricingQuantity is not null.
+  ```  
+
+### Additional Guidelines for Columns in JSON Format
+
+#### Column Definition Structure
```
ijurica, line 81: Would be great to add one of these two columns - ContractApplied or AllocatedMethodDetails - to the *Examples* section at the bottom of the guidelines.

I guess we’ll approve this PR first and only then the ones introducing those columns, so maybe we can add the Example to the guidelines in either #1350 or #1394.

BTW, great job!

shawnalpay:
Repaired remaining merge conflict.  Flagging this to start Member approval tomorrow!

guidelines/normative-requirements-guidelines.md
```diff
@@ -76,7 +76,59 @@
...
+      "RecommendedKey3" : 0.5
+    }, {
+      "RequiredKey1" : 0.1,
+      "RecommendedKey3" : 4,
+      "ProviderDefinedKey4": "SomeString"
+    } ]
+  }
+  ```
+
+* **Include a [JSON Type Definition](https://www.rfc-editor.org/rfc/rfc8927) (JTD) as an approximation of the expected schema, but clarify that normative requirements take precedence when there is a discrepancy**: JSON Type Definition is a convenient way to visualize the expected shape of JSON data, but it often cannot replicate the JSON schema normative requirements of FOCUS. E.g. [NumericFormat](#numericformat) allows for multiple numeric data types and precisions, but JDT requires both to be specified.
```
qquigley: Nit: Should "JDT" be "JTD" here?

guidelines/normative-requirements-guidelines.md
```diff
@@ -76,7 +76,59 @@
...
+  * Column normative requirements specify requirements of the column such as nullability.
+  * JSON schema normative requirements specify the shape of the JSON.
+  * Contents normative requirements usually specify the expected Keys, the format of the Values, and the expected contents of the Values.
+
+#### JSON Schema
+
+* **Omit JSON schema normative requirements for Key-Value Format columns**: The Key-Value Format definition is sufficient to define the expected JSON schema.
+
+* **Include JSON schema normative requirements for JSON Object Format columns**: The JSON Object Format specifies that the format is subject to the requirement of the column and that provider-defined columns must have documented schema.
+  * A recommended pattern is an Object containing a collection whose key is "Elements" containing one or more objects in the Key-Value format.
```
qquigley: Is this a pattern we recommend, or a pattern we would want to require? If we're recommending it, we should maybe add a short sentence explaining why it's being recommended.

The reason I bring up maybe requiring it is because this new column type might be less "scary" to practitioners and providers alike if we're controlling the complexity of the object right out of the gate. It's easier to adjust this requirement with future updates than it would be to put the genie back in the bottle later.

guidelines/normative-requirements-guidelines.md
```diff
@@ -90,9 +142,11 @@
...
   * In **SkuPriceDetails**, refer to **property key** when addressing only the key, and **property value** when addressing only the value.
   * When linking a key to its value, use **corresponding value**.
-  
+
 * **First Mention and Context**: In the case of SkuPriceDetails property key, the first mention explicitly uses "SkuPriceDetails property key" to establish the context. Subsequent references to "property key" and "property value" omit "SkuPriceDetails" as the context is already understood. In contrast, for Tags, this is not necessary, as the context is inherently clear from the column name.
 
+* **Put references to a specific key in double quotes**: In the case of AllocatedMethodDetails, normative requirements are applied to specific keys. To delineate for example that the object with the key "Elements" is being referred, the key should be used in its exact casing inside of double quotation marks `"`. This may result in situations that are traditionally gramatically incorrect where the period resides outside the qutations when the sentence ends in a reference to a specific key.
```
qquigley: Suggestion: I don't mean to start A Thing™, but putting a period inside or outside of a sentence is situational with different regional rules.

I think this sentence might be redundant anyway and could be struck.

guidelines/normative-requirements-guidelines.md
shawnalpay: Yes, agree!

guidelines/normative-requirements-guidelines.md
shawnalpay: While we could make this attribute more stringent and make the Object always include an array of Objects -- I believe @kk09v's vision was for this to handle not only for that use case, but for any complex JSON use case, leaving the definitions to each individual column to lay out.

> we should maybe add a short sentence explaining why it's being recommended.

What might you suggest?

guidelines/normative-requirements-guidelines.md
shawnalpay: 😆 I'm laughing because @kk09v really hemmed and hawed about this exact point.

In any case, I don't disagree that we could likely strike the last sentence.

```suggestion
* **Put references to a specific key in double quotes**: In the case of AllocatedMethodDetails, normative requirements are applied to specific keys. To delineate for example that the object with the key "Elements" is being referred, the key should be used in its exact casing inside of double quotation marks `"`.
```

guidelines/normative-requirements-guidelines.md
qquigley: I don’t have a suggestion because I don’t know why it’s being recommended. Hence the suggestion of adding an explanation. 😅

guidelines/normative-requirements-guidelines.md
```diff
@@ -76,7 +76,59 @@
...
+
+* **Separate normative requirements into sections for column, JSON schema, and contents**: Communicating the normative requirements for a column, JSON schema, and the contents can be convoluted. Separating these requirements provides better clarity.
+  * Column normative requirements specify requirements of the column such as nullability.
+  * JSON schema normative requirements specify the shape of the JSON.
+  * Contents normative requirements usually specify the expected Keys, the format of the Values, and the expected contents of the Values.
+
+#### JSON Schema
+
+* **Omit JSON schema normative requirements for Key-Value Format columns**: The Key-Value Format definition is sufficient to define the expected JSON schema.
+
```
shawnalpay, line 91: This empty line is causing the bullets to be spaced weirdly in markdown.
```suggestion
```

guidelines/normative-requirements-guidelines.md
shawnalpay: Zooming back out, I see what you mean!  I think this sentence would be not a _recommendation_, but more of an _example_ of one possible pattern.  So I think maybe this sentence gets moved down to the **Example JSON** section and gets changed to:

```suggestion
One possible pattern is an Object containing a collection whose key is "Elements" containing one or more objects in the Key-Value format.
```

guidelines/normative-requirements-guidelines.md
ijurica: I’d vote to retain JTD directly within the column definition.

Regardless of the "disclaimer", I believe JTD adds meaningful value to the JSON object-formatted column definition, especially in terms of clarity

qquigley: I'm also team JTD directly within the column definition. I like everything I need wrapped up in a single package, or in this case, doc file.

shawnalpay: Sounds good!  I don't know if you ended up taking that straw poll in TF2, @kk09v, but that's enough signal for me to keep it in there.  You can resolve!

guidelines/normative-requirements-guidelines.md
kk09v: This is a weird one. When rendered, it adds an visual space between lines 47/48 as well as here between lines 92/93, but it doesn't between lines 83/84. I tried making a few changes like this and previewing, but I still can't get it to look right.

guidelines/normative-requirements-guidelines.md
kk09v: I changed the verbiage to refer to the columns using this as examples, without making a recommendation. Wasn't sure where to move this to, so it's in the same place for now.

guidelines/normative-requirements-guidelines.md
kk09v: I was hoping to prevent ambiguity about that, but if neither of you think it's necessary, I'll strike it.

guidelines/normative-requirements-guidelines.md
kk09v:
```suggestion
* **Include a [JSON Type Definition](https://www.rfc-editor.org/rfc/rfc8927) (JTD) as an approximation of the expected schema, but clarify that normative requirements take precedence when there is a discrepancy**: JSON Type Definition is a convenient way to visualize the expected shape of JSON data, but it often cannot replicate the JSON schema normative requirements of FOCUS. E.g. [NumericFormat](#numericformat) allows for multiple numeric data types and precisions, but JTD requires both to be specified.
```
Co-authored-by: Andrew "Q" Quigley <qquigley@users.noreply.github.com>

guidelines/normative-requirements-guidelines.md
kk09v: Added verbiage as suggested in TF2 this week

guidelines/normative-requirements-guidelines.md
kk09v: I added a hyperlink (presumptive) to those columns, but we can keep eyes on these 3 and incorporate the the second once one is merged.

specification/attributes/object_format.md
kk09v: I took this feedback, but ended up going a different direction with the rephrasing. Let me know what you think.

guidelines/normative-requirements-guidelines.md
shawnalpay: Let's commit this and resolve -- we can address further during consistency review if necessary.

guidelines/normative-requirements-guidelines.md
shawnalpay: I think we can resolve this thread now, yes?

specification/attributes/object_format.md
shawnalpay: This has been addressed.  Resolving!

specification/attributes/object_format.md
shawnalpay: I think we can resolve this one...

guidelines/normative-requirements-guidelines.md
shawnalpay: Resolving this without resolution -- we can address during consistency review if necessary!

ijurica (approved):

shawnalpay (approved):
Looks good to me.  Thanks for making this happen @kk09v!

qquigley (approved):

jpradocueva:
PR was approved during the Members’ Meeting on September 18.

PR #1209, opened 2025-07-23
FR #1054: Initial commit with Requirements Model Data for FOCUS 1.2

mike-finopsorg:
For more information on the content of this pull request, please review the [readme.md](https://github.com/FinOps-Open-Cost-and-Usage-Spec/FOCUS_Spec/blob/5fa616570eb21b5806eda820247061c062a435ad/specification/conformance/README.md) file.

Issue #1055, opened 2025-05-16
[FR] Clarify roles of Provider, Publisher, and Invoice Issuer


Matt-Cowsert:
### 🧠 Problem Statement
The columns for publisher and provider are ambiguous and not internally consistent within the spec itself. Lack of clarity on the definitions of the columns has lead to user feedback that the columns are confusing. In 1.2 it was decided that we not make any aggressive corrective changes, in favor of addressing in 1.3.

---
### 🧪 Use Case

- As a FinOps Practitioner, I want to be able to distinguish between the entities involved when I'm purchasing software from a provider marketplace so I understand who I'm purchasing from and who to contact when I have questions or billing disputes. 

---
### 🎯 Desired Outcome / Practitioner Impact

- Users clearly know who they are purchasing from. 
- Users know what entities were involved in a purchase in the event of a purchase through a marketplace.

---
### 📨 Type of Request
- [x] Refinement of an existing FOCUS attribute
- [ ] Addition of a provider-supported field not yet in FOCUS
- [ ] Net-new concept (not currently supported by providers or FOCUS)

---
### 🏛️ Organizations Requesting or Supporting This Feature

- Domo
- Microsoft

---
### 🌐 FinOps Scope Alignment (Optional)
- [x] Public Cloud – e.g., AWS, Azure, GCP, OCI
- [x] Software-as-a-Service (SaaS) – e.g., Salesforce, Snowflake
- [ ] Data Center – on-prem compute and infrastructure
- [ ] Licensing – subscription or usage-based licensing models *(under development)*
- [ ] AI – cost and usage for AI models and platforms *(under development)*
- [ ] Custom – internal tooling, specialized infra *(under development)*

---
### 📊 Impacted Parties
- [x] FinOps Practitioner – end users who analyze or act on the data
- [x] FOCUS Data Generator – providers generating output aligned to the spec
- [x] Vendor Supporting FOCUS – vendors or tools ingesting the spec or using the spec language in their UI
- [ ] Other (please explain in comments)

---
### 🌫️ Level of Ambiguity

3

---
### 📂 Supporting Documentation

use cases and address concerns raised in WI https://github.com/FinOps-Open-Cost-and-Usage-Spec/FOCUS_Spec/issues/985

---
### 🛠️ Proposed Solution / Approach

---
### 💬 Community Support (Add Your Voice)
If your organization supports this request or has a similar use case:

- Add a **comment** below including:
[...]

Issue #1054, opened 2025-05-16
[FR] Creation of Static Conformance Requirements (SCR)


Matt-Cowsert:
### 🧠 Problem Statement
As the FOCUS specification grows in scope and adoption, the absence of structured, enforceable conformance requirements creates **ambiguity** in **how implementations interpret and apply the specification**. This may lead to inconsistent data exports, difficulty validating compliance, and confusion among practitioners and vendors regarding what constitutes a "conformant" implementation.
**Impact**
- Vendors struggle to **prove conformance** or **claim support** without a baseline framework.
- Without clear **conformance criteria**, comparisons across tools and platforms are unreliable.
- The FOCUS ecosystem lacks a foundation for **automation**, **certification**, or **machine-based** validation.

---
### 🧪 Use Case
- Create a Static Conformance Requirements **(SCR) Implementation Guide**
- FinOps FOCUS v1.2 Columns and groups
- AI #969 Creation of Static Conformance Requirements (SCR)

---
### 🎯 Desired Outcome / Practitioner Impact
Enable vendors, tooling providers, and practitioners to consistently **declare, verify, and validate conformance** to the FOCUS specification using a **machine-readable**, structured checklist that defines which features (columns) are mandatory, optional, or conditional for various deployment contexts.

---
### 📨 Type of Request
- [ ] Refinement of an existing FOCUS attribute
- [ ] Addition of a provider-supported field not yet in FOCUS
- [ ] Net-new concept (not currently supported by providers or FOCUS)

---
### 🏛️ Organizations Requesting or Supporting This Feature
TBD

---
### 🌐 FinOps Scope Alignment (Optional)
- [ ] Public Cloud – e.g., AWS, Azure, GCP, OCI
- [ ] Software-as-a-Service (SaaS) – e.g., Salesforce, Snowflake
- [ ] Data Center – on-prem compute and infrastructure
- [ ] Licensing – subscription or usage-based licensing models *(under development)*
- [ ] AI – cost and usage for AI models and platforms *(under development)*
- [ ] Custom – internal tooling, specialized infra *(under development)*

---
### 📊 Impacted Parties
- [ ] FinOps Practitioner – end users who analyze or act on the data
- [ ] FOCUS Data Generator – providers generating output aligned to the spec
- [ ] Vendor Supporting FOCUS – vendors or tools ingesting the spec or using the spec language in their UI
- [ ] Other (please explain in comments)

---
[...]

PR #1412, opened 2025-08-29
AI #1290: Create conformance rules for ProviderName column

### LINKED ISSUE #1290: [AI] Author PR for Conformance definition for column: provider
Review [provider_cr.md](https://github.com/FinOps-Open-Cost-and-Usage-Spec/FOCUS_Spec/blob/1121-ai-align-on-approach-for-scrs/specification/conformance/cr-stage1/provider_cr.md) and craft the list of ConformanceRules based on the guides in [writing-conformance-rules](https://github.com/FinOps-Open-Cost-and-Usage-Spec/FOCUS_Spec/blob/1121-ai-align-on-approach-for-scrs/guidelines/writing-conformance-rules.md.

flanakin:
## Summary
- Creates comprehensive conformance rules for the ProviderName column based on CR specification
- Implements 4 validation rules covering presence, data type, format, and nullability requirements
- References StringHandling cross-attribute dependency for format validation

## Changes
- **Added**: `specification/conformance/conformance_rules/columns/providername.json`
  - ProviderName-C-000-M: Main composite rule
  - ProviderName-C-001-M: Presence validation  
  - ProviderName-C-002-M: String data type requirement
  - ProviderName-C-003-M: StringHandling format compliance
  - ProviderName-C-004-M: Non-null requirement

## Test plan
- [ ] Validate JSON schema compliance
- [ ] Verify rule references and dependencies
- [ ] Confirm alignment with CR specification requirements

Resolves #1290

🤖 Generated with [Claude Code](https://claude.ai/code)


shawnalpay:
@mike-finopsorg @jpradocueva 
Are we using the column ID or the display name for conformance rules?

* If the former, @flanakin needs to change this to use `Provider` instead of `ProviderName`.
* If the latter, you all need to change #1209 to use `ProviderName` instead of `Provider`.

My opinion?  I think we should be using the column ID -- so it should be `ProviderName`.

caligo-erik (approved):

flanakin:
> @mike-finopsorg @jpradocueva Are we using the column ID or the display name for conformance rules?
> 
> * If the former, @flanakin needs to change this to use `Provider` instead of `ProviderName`.
> * If the latter, you all need to change [FR #1054: Initial commit with Conformance Data structure #1209](https://github.com/FinOps-Open-Cost-and-Usage-Spec/FOCUS_Spec/pull/1209) to use `ProviderName` instead of `Provider`.
> 
> My opinion? I think we should be using the column ID -- so it should be `ProviderName`.

@shawnalpay The lack of "Name" in Provider and I think Publisher has always been a point of confusion. We've talked about changing this in the past. Are you okay if I submit a PR for that?

shawnalpay:
> @shawnalpay The lack of "Name" in Provider and I think Publisher has always been a point of confusion. We've talked about changing this in the past. Are you okay if I submit a PR for that?

@flanakin I see you've created PR #1446 for this.  That works for me, as I'd argue that's part of #1055 -- but we'll need to therefore discuss it further in TF3.

That doesn't change what needs doing here.  This initial conformance work is against FOCUS v1.2, which for better or for worse is cast in amber -- so this PR needs to align with the verbiage as stated in that version.

@mike-finopsorg please consider this when approving this PR.

mike-finopsorg:
Manually imported to the base branch with fixes for unit tests to pass

PR #1373, opened 2025-08-20
Data recency metadata

rileyjenk:

Issue #1051, opened 2025-05-16
[FR] Establish glossary-usage guidelines


ijurica:
### Problem Statement

**Lack of Guidelines for Glossary Usage**

FOCUS columns represent metadata that describes, identifies, or quantifies FinOps business terms and concepts.

However, the absence of clear guidance on when to create a glossary entry for these business terms — versus relying on descriptions in column intro sections — has led to inconsistencies, duplication, and reduced documentation quality. 

Some business terms are defined as glossary entries with links from column specification intro sections; others are described only within column specifications, and some are documented in both (or even multiple) places.

This inconsistency makes the documentation harder to read, maintain, and contribute to, and increases the risk of conflicting or fragmented definitions.

***Note:** A similar issue might also exist in certain FOCUS attributes that describe behavioral rules related to specific FinOps terms,*

To address this, we propose developing clear and structured guidelines that include, but are not limited to, the following:

- When a glossary entry should be created versus when (or if) an inline definition is sufficient.
- How glossary entries should be phrased and formatted.
- How glossary terms should be linked:
  - From column specification intro sections to glossary entries
  - Between glossary terms (interlinking related concepts)
- How and if glossary terms should link back to relevant columns, attributes, or other specification sections (e.g., Appendix).

Standardizing these practices would improve documentation readability, maintainability, and conceptual clarity — while also enabling more effective collaboration and empowering contributors to make consistent, high-quality additions.

**(Optional) Beyond Glossary: Capturing Full Conceptual Context**

Additionally, if some business terms or concepts require detailed explanations that go beyond concise definitions, the glossary is not the appropriate place for these extended definitions, as glossary entries should remain short and focused. The FOCUS Attributes section typically addresses behavioral rules related to certain business concepts, so it may not be suitable for this purpose either. This may indicate the need for a separate section (e.g., Business Concepts) where broader business logic and context can be appropriately described.

---

### Use Case / User Story

- As a FOCUS contributor, I want clear guidelines on when and how to define glossary entries versus using inline business term descriptions, and when and how to link glossary entries from and to the rest of the specification, so that I can avoid unnecessary duplication, maintain consistency, and make the documentation easier to navigate for all readers.

---

### Success Criteria

- [ ] Contributors can determine whether a new business term warrants a glossary entry or an inline definition by following documented decision criteria
- [ ] Spec readers can navigate from any column or attribute reference to the authoritative definition of a business term via consistent linking patterns
[...]

PR #1431, opened 2025-09-01
FR #1090: Associate attributes with datasets

### LINKED ISSUE #1090: [FR] Add Structured "BI" schema to FOCUS

### LINKED ISSUE #1439: [AI] Associate attributes explicitly with datasets
### Detailed Description

Our single dataset has had all attributes applied up until now.

Now that we have multiple datasets, we need to apply specific attributes to specific datasets.

### Definition of Done

Complete a PR that associates specific attributes with specific datasets via normative requirements.

shawnalpay:
We need to finesse the spec to associate specific attributes with specific datasets.  

Current verbiage is:

> Attributes are requirements that apply across all FOCUS datasets instead of an individual column level.

1) That's not true for all attributes: we apply some at the column level.
2) This means that every attribute is associated with every dataset, which will be problematic/inaccurate as our number of datasets grows beyond one.  For example, Discount Handling has nothing to do with Contracts.

This PR therefore associates the `Cost and Usage` dataset to all the current attributes.

FYI that PR #1350 performs the same association for `Contract Commitment` (or rather, it will, once we get agreement on this PR).

### DIFF
--- a/specification/datasets/cost_and_usage/dataset.md
+++ b/specification/datasets/cost_and_usage/dataset.md
@@ -68,11 +68,7 @@ The specification for the Cost and Usage dataset defines a group of columns that
 
 <div class='h4-nonindex'>Relationships</div>
 
-The Cost and Usage dataset can be joined to the Contract Commitment dataset through the use of the Contract Commitment ID field.
-
-| Dataset A           | Dataset A Column       | Dataset B           | Dataset B Column       |
-| ------------------- | ---------------------- | ------------------- | ---------------------- |
-| Cost and Usage      | Contract Commitment ID | Contract Commitment | Contract Commitment ID |
+(none)
 
 <div class='h4-nonindex'>Requirements</div>
 
--- a/specification/datasets/cost_and_usage/dataset.md
+++ b/specification/datasets/cost_and_usage/dataset.md
@@ -6,7 +6,65 @@ The specification for the Cost and Usage dataset defines a group of columns that
 
 <div class='h4-nonindex'>Columns</div>
 
-<<column list TBD>>
+| Column                                                                        | Column Type | Feature Level | Allows Nulls | Data Type | Required |
+| ----------------------------------------------------------------------------- | ----------- | ------------- | ------------ | --------- | -------- |
+| [Billed Cost](#billedcost)                                                    | Metric      | Mandatory     | False        | Decimal   | TRUE     |
+| [Billing Account ID](#billingaccountid)                                       | Dimension   | Mandatory     | False        | String    | TRUE     |
+| [Billing Account Type](#billingaccounttype)                                   | Dimension   | Conditional   | False        | String    | FALSE    |
+| [Billing Currency](#billingcurrency)                                          | Dimension   | Mandatory     | False        | String    | TRUE     |
+| [Billing Period End](#billingperiodend)                                       | Dimension   | Mandatory     | False        | Date/Time | TRUE     |
+| [Billing Period Start](#billingperiodstart)                                   | Dimension   | Mandatory     | False        | Date/Time | TRUE     |
+| [Charge Category](#charge category)                                           | Dimension   | Mandatory     | False        | String    | TRUE     |
+| [Charge Frequency](#chargefrequency)                                          | Dimension   | Recommended   | False        | String    | TRUE     |
+| [Charge Period End](#chargeperiodend)                                         | Dimension   | Mandatory     | False        | Date/Time | TRUE     |
+| [Charge Period Start](#chargeperiodstart)                                     | Dimension   | Mandatory     | False        | Date/Time | TRUE     |
+| [Contracted Cost](#contractedcost)                                            | Metric      | Mandatory     | False        | Decimal   | TRUE     |
+| [Effective Cost](#effectivecost)                                              | Metric      | Mandatory     | False        | Decimal   | TRUE     |
+| [Invoice Issuer](#invoiceissuername)                                          | Dimension   | Mandatory     | False        | String    | TRUE     |
+| [List Cost](#listcost)                                                        | Metric      | Mandatory     | False        | Decimal   | TRUE     |
+| [Provider](#providername)                                                     | Dimension   | Mandatory     | False        | String    | TRUE     |
+| [Publisher](#publishername)                                                   | Dimension   | Mandatory     | False        | String    | TRUE     |
+| [Service Category](#servicecategory)                                          | Dimension   | Mandatory     | False        | String    | TRUE     |
+| [Service Name](#servicename)                                                  | Dimension   | Mandatory     | False        | String    | TRUE     |
+| [Service Subcategory](#servicesubcategory)                                    | Dimension   | Recommended   | False        | String    | FALSE    |
+| [Availability Zone](#availabilityzone)                                        | Dimension   | Recommended   | True         | String    | FALSE    |
+| [Billing Account Name](#billingaccountname)                                   | Dimension   | Mandatory     | True         | String    | TRUE     |
+| [Capacity Reservation ID](#capacityreservationid)                             | Dimension   | Conditional   | True         | String    | TRUE     |
+| [Capacity Reservation Status](#capacityreservationstatus)                     | Dimension   | Conditional   | True         | String    | TRUE     |
+| [Charge Class](#charge class)                                                 | Dimension   | Mandatory     | True         | String    | TRUE     |
+| [Charge Description](#chargedescription)                                      | Dimension   | Mandatory     | True         | String    | TRUE     |
+| [Commitment Discount Category](#commitmentdiscountcategory)                   | Dimension   | Conditional   | True         | String    | TRUE     |
+| [Commitment Discount ID](#commitmentdiscountid)                               | Dimension   | Conditional   | True         | String    | TRUE     |
+| [Commitment Discount Name](#commitmentdiscountname)                           | Dimension   | Conditional   | True         | String    | TRUE     |
+| [Commitment Discount Quantity](#commitmentdiscountquantity)                   | Metric      | Conditional   | True         | Decimal   | TRUE     |
+| [Commitment Discount Status](#commitmentdiscountstatus)                       | Dimension   | Conditional   | True         | String    | FALSE    |
+| [Commitment Discount Type](#commitmentdiscounttype)                           | Dimension   | Conditional   | True         | String    | TRUE     |
+| [Commitment Discount Unit](#commitmentdiscountunit)                           | Dimension   | Conditional   | True         | String    | TRUE     |
+| [Consumed Quantity](#consumedquantity)                                        | Metric      | Conditional   | True         | Decimal   | TRUE     |
+| [Consumed Unit](#consumedunit)                                                | Dimension   | Conditional   | True         | String    | TRUE     |
+| [Contracted Unit Price](#contractedunitprice)                                 | Metric      | Conditional   | True         | Decimal   | FALSE    |
+| [Invoice ID](#invoiceid)                                                      | Dimension   | Recommended   | True         | String    | FALSE    |
+| [List Unit Price](#listunitprice)                                             | Metric      | Conditional   | True         | Decimal   | TRUE     |
+| [Pricing Category](#pricingcategory)                                          | Dimension   | Conditional   | True         | String    | TRUE     |
+| [Pricing Currency](#pricingcurrency)                                          | Dimension   | Conditional   | True         | String    | TRUE     |
+| [Pricing Currency Contracted Unit Price](#pricingcurrencycontractedunitprice) | Metric      | Conditional   | True         | Decimal   | TRUE     |
+| [Pricing Currency Effective Cost](#pricingcurrencyeffectivecost)              | Metric      | Conditional   | True         | Decimal   | TRUE     |
+| [Pricing Currency List Unit Price](#pricingcurrencylistunitprice)             | Metric      | Conditional   | True         | Decimal   | TRUE     |
+| [Pricing Quantity](#pricingquantity)                                          | Metric      | Mandatory     | True         | Decimal   | TRUE     |
+| [Pricing Unit](#pricingunit)                                                  | Dimension   | Mandatory     | True         | String    | TRUE     |
+| [Region ID](#regionid)                                                        | Dimension   | Conditional   | True         | String    | TRUE     |
+| [Region Name](#regionname)                                                    | Dimension   | Conditional   | True         | String    | FALSE    |
+| [Resource ID](#resourceid)                                                    | Dimension   | Conditional   | True         | String    | TRUE     |
+| [Resource Name](#resourcename)                                                | Dimension   | Conditional   | True         | String    | TRUE     |
+| [Resource Type](#resourcetype)                                                | Dimension   | Conditional   | True         | String    | TRUE     |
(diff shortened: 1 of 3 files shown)


shawnalpay:
Received good feedback from the group during Sep 3 TF2.  I have now incorporated that feedback!

specification/datasets/cost_and_usage/dataset.md
```diff
@@ -0,0 +1,96 @@
...
+| [SKU ID](#skuid)                                                              | Dimension   | Conditional   | True         | String    |
+| [SKU Meter](#skumeter)                                                        | Dimension   | Conditional   | True         | String    |
+| [SKU Price Details](#skupricedetails)                                         | Dimension   | Conditional   | True         | JSON      |
+| [SKU Price ID](#skupriceid)                                                   | Dimension   | Conditional   | True         | String    |
+| [Sub Account ID](#subaccountid)                                               | Dimension   | Conditional   | True         | String    |
+| [Sub Account Name](#subaccountname)                                           | Dimension   | Conditional   | True         | String    |
+| [Sub Account Type](#subaccounttype)                                           | Dimension   | Conditional   | True         | String    |
+| [Tags](#tags)                                                                 | Dimension   | Conditional   | True         | JSON      |
+
+<div class='h4-nonindex'>Relationships</div>
```
shawnalpay, line 69: These `<div>` headers aren't rendering the way we set it up, @mike-finopsorg. 

See this PDF output (go to Section 3.1): [spec.pdf](https://github.com/user-attachments/files/22127586/spec.pdf) (screenshot below also)

Any clue why this might be?

Image

specification/datasets/cost_and_usage/dataset.md
```diff
@@ -0,0 +1,96 @@
...
+
+(none)
+
+<div class='h4-nonindex'>Requirements</div>
+
+The CostAndUsage dataset adheres to the following requirements:
+
+* CostAndUsage MUST be present.
```
flanakin: This doesn't seem to be needed, right? I'm not sure what it means. If they have the dataset, these rules apply. If they don't have the dataset, then these rules don't apply because they're not even looking at cost and usage data. I'm not sure what this is really conveying as a requirement.

Unless it's saying that the CostAndUsage dataset is required for them to be able to claim FOCUS support? But if that's it, what are they "certifying" is FOCUS-compatible? A dataset or a tool? I believe we can only certify datasets, so I'm not sure this line is needed.

flanakin (approved):

specification/datasets/cost_and_usage/dataset.md
```diff
@@ -0,0 +1,96 @@
+# Cost and Usage
+
+The Cost and Usage dataset is the primary dataset for FOCUS cost and usage data.
```
ijurica, line 3: I suggest we consider expanding the glossary with additional dataset-related terms, either as part of this PR or during the Consistency review phase. 

Currently, we only have a definition for a FOCUS dataset, and in all columns in the Cost and Usage dataset, presence-related normative requirements link to this single glossary term.

Patterns for column presence normative requirements:

```markdown
* <ColumnId> MUST be present in a [*FOCUS dataset*](#glossary:FOCUS-dataset).
* <CoumnId> MUST be present in a [*FOCUS dataset*](#glossary:FOCUS-dataset) when <Condition>.
* <ColumnId> is RECOMMENDED to be present in a [*FOCUS dataset*](#glossary:FOCUS-dataset) when <Condition>.
```

Current FOCUS dataset glossary term definition:

> *A FOCUS dataset is a structured collection of records that conforms to the BCP14 criteria established by FOCUS. All columns included must be defined in the FOCUS Columns section of the specification.*
> 
> *In addition to these standardized columns, data generators may include custom provider-specific columns (prefixed with `x_`) where additional context is needed beyond what is captured in the defined FOCUS columns. If such custom columns introduce record-splitting (i.e., a single original charge results in multiple rows), the data generator is responsible for ensuring that all cost and quantity metrics still meet the aggregation and consistency rules required by the specification.*
> 
> *The collection of datasets are designed to provide billing insight, additional context, metadata, mapping, or enrichment information that enhances the interpretability or completeness.*

specification/datasets/cost_and_usage/dataset.md
```diff
@@ -0,0 +1,96 @@
...
+
+CostAndUsage
+
+<div class='h4-nonindex'>Display Name</div>
+
+Cost and Usage
+
+<div class='h4-nonindex'>Description</div>
+
+Describes the cost and usage incurred by consuming a provider's resources and services.
```
ijurica: I'm not sure if this wording sufficiently covers purchase scenarios.

specification/datasets/cost_and_usage/dataset.md
```diff
@@ -0,0 +1,96 @@
...
+| [SKU Price Details](#skupricedetails)                                         | Dimension   | Conditional   | True         | JSON      |
+| [SKU Price ID](#skupriceid)                                                   | Dimension   | Conditional   | True         | String    |
+| [Sub Account ID](#subaccountid)                                               | Dimension   | Conditional   | True         | String    |
+| [Sub Account Name](#subaccountname)                                           | Dimension   | Conditional   | True         | String    |
+| [Sub Account Type](#subaccounttype)                                           | Dimension   | Conditional   | True         | String    |
+| [Tags](#tags)                                                                 | Dimension   | Conditional   | True         | JSON      |
+
+<div class='h4-nonindex'>Relationships</div>
+
+(none)
```
ijurica: nit: Here it says **`(none)`**, while for Exceptions in Attribute definitions we used **`None`** - should we align?

ijurica (approved):

specification/datasets/cost_and_usage/dataset.md
shawnalpay: @flanakin My point here was, a Cost and Usage dataset is required in all circumstances.  Some of the supporting datasets may be required conditionally; for example, generating the Contract Commitment dataset as described in #1350 is only applicable if the data generator supports contract commitments.  

I suspect what you're getting at is: this isn't a _dataset_ requirement, but more of a _global_ requirement.  Kind of like how, per our conformance requirements work via #1209, the presence of a column is a dataset requirement, and not a column requirement.  _That_, I could agree with.  And maybe we do need a global requirement space.  But I'd probably put that out of scope for the time being and just strike this line for now, if so.

Let me know what you think about that.

specification/datasets/cost_and_usage/dataset.md
shawnalpay: @ijurica Feel free to suggest an alternative!

specification/datasets/cost_and_usage/dataset.md
shawnalpay: Happy to stay consistent!

```suggestion
None
```

specification/datasets/cost_and_usage/dataset.md
shawnalpay: @ijurica What are some examples of terms you would want to add?

specification/datasets/cost_and_usage/dataset.md
flanakin: If we have this, we're saying that tools are FOCUS compliant or not. And that compliance defines what datasets you MUST and SHOULD provide. 

I was assuming tools would have the ability to say they have a "FOCUS compliant cost and usage" -OR- "FOCUS compliant contract commitments" -OR- "FOCUS compliant SKU catalog", etc. 

Was there a decision about the level at which we are measuring compliance and enforcing requirements? I'm fine either way, I suspect I just came at it with a different assumption. I see pros/cons with both approaches.

mike-finopsorg: This line surfaces the next entity which we discovered in the CR work. What you are alluding to here is that there is a "Data Generator" entity and it has requirements of "CostAndUsage MUST be provided by the Data Generator". I would like the group to consider skipping this normative text for this release and to put a FR in to scope requirements of the Data Generator entity in order to avoid the same issue we have today with Column requirements that should be relocated to the Dataset entity.

shawnalpay: Great points, both -- thank you @flanakin and @mike-finopsorg.  I'll strike it!  I agree that we should add an FR to scope normative requirements at the "global" (i.e., data generator) level; I'll coordinate with @Matt-Cowsert.

```suggestion
```

specification/datasets/cost_and_usage/dataset.md
shawnalpay: Discussed with @mike-finopsorg.  This doesn't work because this CSS was implemented via PR #1350 and hasn't yet hit `working_draft` 😅   My mistake!  

Working as intended, and the use of these divs will work fine once #1350 is approved.  Resolving!

specification/datasets/cost_and_usage/dataset.md
ijurica: I feel like consumption is a bit more oriented toward usage-related scenarios. In some other places, we’ve used wording like:

> *…associated with a resource or service used or purchased...*
> 
> *…associated with a used or purchased resource or service...*

Here is my attempt to include purchase scenarios as well:

```suggestion
Describes the cost and usage incurred through using or purchasing a provider’s [*resources*](#glossary:resource) or [*services*](#glossary:service).
```

specification/datasets/cost_and_usage/dataset.md
ijurica: ***Note:***
*No need to rush this! I just noticed that we could be a bit more precise in these normative requirements*

---

#### Option A:

Reference the specific dataset in presence-related normative requirements, rather than the existing FOCUS dataset glossary term.

#### Option B:

Introduce glossary terms for individual datasets and reference them in presence-related normative requirements, instead of the existing FOCUS dataset glossary term.

Examples of glossary terms we could add:

- Cost and Usage dataset
- Contract Commitment dataset (scope of PR #1350)
.

ijurica (approved):

specification/datasets/cost_and_usage/dataset.md
shawnalpay: When we approved #1337, we agreed not to have separate glossary entries for each dataset (@mike-finopsorg, please chime in if you remember it differently).

I think Option A is the safer and less duplicative play.  I propose that we tackle that as part of work in FR #1051 in the next release.

Let me know if this works for you, and I will resolve this comment.

specification/datasets/cost_and_usage/dataset.md
shawnalpay: This works for me!  Accepting suggestion.

specification/datasets/cost_and_usage/dataset.md
```diff
@@ -0,0 +1,95 @@
...
+
+CostAndUsage
+
+<div class='h4-nonindex'>Display Name</div>
+
+Cost and Usage
+
+<div class='h4-nonindex'>Description</div>
+
+Describes the cost and usage incurred through using or purchasing a provider’s [*resources*](#glossary:resource) or [*services*](#glossary:service).
```
shawnalpay: Striking smart character to fix the linter.

```suggestion
Describes the cost and usage incurred through using or purchasing a provider's [*resources*](#glossary:resource) or [*services*](#glossary:service).
```

specification/datasets/cost_and_usage/dataset.md
ijurica: It was just a note... 
resolving this conversation.

shawnalpay:
All comments resolved!  Flagging as Ready for Approval, to be considered for approval in tomorrow's Members meeting.

specification/datasets/cost_and_usage/dataset.md
```diff
@@ -0,0 +1,95 @@
...
+None
+
+<div class='h4-nonindex'>Requirements</div>
+
+The CostAndUsage dataset adheres to the following requirements:
+
+* CostAndUsage MUST conform to [ColumnHandling](#columnhandling) requirements.
+* CostAndUsage MUST conform to [DiscountHandling](#discounthandling) requirements.
+* CostAndUsage MUST conform to [NullHandling](#nullhandling) requirements.
+
```
shawnalpay, line 80: Per [this thread](https://github.com/FinOps-Open-Cost-and-Usage-Spec/FOCUS_Spec/pull/1321#discussion_r2337300515) on PR #1321, we will add the following line if/when #1321 is approved:

```suggestion
CostAndUsage MUST conform to [InvoiceHandling](#invoicehandling) requirements.

```

specification/datasets/cost_and_usage/dataset.md
```diff
@@ -0,0 +1,95 @@
...
+<div class='h4-nonindex'>Dataset ID</div>
```
rileyjenk, line 81: I ran into a problem in the recency PR that I want to bring up here @shawnalpay. I am going to author a change there but here is the key points of the problem that we likely need to coordinate:

- In metadata, we refer to datasets by a single key called "dataset" and i examples that is a guid
- - Because Data generators may have multiple reports using the same dataset we likely need a id reference to the FOCUS dataset, and the report that uses a focus dataset
- -  noticed that we call it Dataset ID here....which was my initial thought in the metadata section. What do you think about dataset_instance_id
- We should include a name in the metdata which is the name of the dataset in FOCUS

shawnalpay: Discussed with @rileyjenk.  He will iterate on PR #1373, and we can resolve this.

jpradocueva:
PR was approved during the Members’ Meeting on September 11.

specification/datasets/cost_and_usage/dataset.md
shawnalpay: This line will be added to #1321, now that this PR has been approved.  Resolving.

Issue #1059, opened 2025-05-19
[FR] Establish FOCUS attribute-specification guidelines


ijurica:
### Problem Statement

Lack of formal guidelines for structuring and writing FOCUS attribute specifications — especially the normative requirements — including how attributes reference relevant columns or are referenced by columns, increases the risk of inconsistent, unclear, or even incorrect attribute definitions. This can lead to incorrect implementation, misaligned expectations between producers and consumers, and reduced overall data quality.

To address this issue, the following is suggested:

- Establish FOCUS attribute guidelines for structuring attribute specifications, with an emphasis on normative requirements and including clear rules on if and how attributes should reference relevant columns and if and how columns should reference applicable attributes.  
  - Account for the diverse nature of FOCUS attributes — for instance, some define format or validation requirements directly linked to individual columns (e.g., String Format, Numeric Format), while others specify requirements spanning multiple columns across various use cases.

- Review existing attributes and apply previously established guidelines for clarity, consistency, and completeness. 

- Identify and flag discrepancies, overlaps, or ambiguous phrasing.

- Update where possible, while capturing larger issues for future iterations.  
  - Although this effort does not intend to introduce material changes to existing attribute behavior, we acknowledge that during the review process, we may identify deeper discrepancies or discussion topics that require more substantial effort or cross-team alignment. In such cases, the priority will be to capture those as separate issues in the backlog to be addressed in future releases.

### Use Cases

- As a FOCUS contributor, I want clear guidelines to consistently add new attributes and efficiently review existing specifications to identify any inconsistencies or ambiguities.
- As a FinOps Practitioner, Provider, or Vendor, I want clear and unambiguous attribute normative requirements to ensure correct implementation and interpretation.

### Desired Outcome / Practitioner Impact

- **Establish FOCUS attribute guidelines**  
  - Clear, structured guidelines will:
    - support the refinement of existing FOCUS attribute specifications  
    - assist contributors in adding new attributes by ensuring consistency and reducing the risk of ambiguity or misalignment early on

- **Identify and address issues in existing attribute specifications**  
  - The review process will help surface discrepancies, overlaps, and unclear or ambiguous normative requirements in current attributes.  
  - *Note:* Wherever possible, issues should be addressed as part of this effort. More complex or broader issues will be captured and tracked as separate feature requests for future releases.

- **Support Static Conformance Requirements and future conformance tooling efforts**  
[...]

Issue #1060, opened 2025-05-19
[FR] Establish FOCUS metadata-specification guidelines


ijurica:
### Problem Statement

Lack of formal guidelines for structuring and writing the FOCUS Metadata Specification (especially the normative requirements of individual elements) increases the risk of inconsistencies, ambiguities, and misalignment across implementations. This can result in varied interpretations, incomplete metadata, or reduced usability for downstream systems and users.

To address this issue, the following is suggested:

- Establish clear and consistent guidelines for structuring and maintaining the FOCUS Metadata Specification, with an emphasis on the normative requirements of individual elements.
-  Review existing FOCUS Metadata element specifications and align them with the newly established guidelines.
- Identify inconsistencies, redundant or missing elements, and areas requiring clarification.
- Update the existing FOCUS Metadata Specification where possible, and document broader issues for future iterations.

***Note:** Assuming this effort includes adding glossary entries for terms like "FOCUS Metadata" and "FOCUS Metadata Schema", it's worth reviewing where these terms already appear in the specification (e.g., in the ChargeDescription column normative requirements) and updating those references to point to the newly added glossary definitions.*

### Desired Outcome

- **Established FOCUS Metadata Specification Guidelines**, supporting both the refinement of existing FOCUS Metadata content and enabling consistent, high-quality contributions going forward.
  
- **Identification and resolution of issues in the existing FOCUS Metadata Specification**  
  - *Note:* Broader or more complex issues that cannot be addressed immediately will be captured as follow-up Feature Requests.

- **Support for preparation of Static Conformance Requirements and conformance tooling**  
  - Similar to the refinement of FOCUS Columns and Attributes, refining the Metadata Specification is a prerequisite for defining Static Conformance Requirements, which are required for the implementation of conformance tooling (see pending PR).

**- Applied to existing FOCUS metadata**
        - Once the guidelines exist, we need to implement them for existing FOCUS metadata. 

#### Not in scope
The guidelines we have in place today are not normative (enforceable) as they live outside of the specification. We will need to stand up a new feature request to evaluate which of our guidelines will be refit to be enforced through traditional normative requirement guidelines. 

### Type of Request

Refinement of an existing FOCUS attribute

### Organizations Requesting or Supporting This Feature

- Neos
- FinOps Foundation
- Domo

### FinOps Scope Alignment (Optional)

- [ ] Public Cloud – e.g., AWS, Azure, GCP, OCI
[...]

PR #1435, opened 2025-09-02
FR #1059: Update Column Handling Attribute

jpradocueva:
This PR responds to the comment on #1059, which suggested updating the normative requirements for Attributes. The goal is to align the expression of Attribute requirements with the approach used for Columns. Specifically, we propose creating a composite requirement to group custom features and separating multiple normative requirements within a single statement into distinct, single-statement rules.

### DIFF
--- a/specification/attributes/column_handling.md
+++ b/specification/attributes/column_handling.md
@@ -49,8 +49,11 @@ Naming and ordering convention for columns appearing in a *FOCUS dataset*.
 ### Column Order
 
 * All FOCUS columns SHOULD be first in the provided dataset.
-* Custom columns SHOULD be listed after all FOCUS columns and SHOULD NOT be intermixed.
-* Columns MAY be sorted alphabetically, but custom columns SHOULD be after all FOCUS columns.
+* For custom columns the following column order requirements apply:
+  * Custom columns SHOULD be listed after all FOCUS columns.
+  * Custom columns SHOULD NOT be intermixed with FOCUS columns.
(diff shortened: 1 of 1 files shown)


shawnalpay:
What I came to realize upon discussion with @mike-finopsorg  yesterday is that, while this work is very important, it **doesn’t actually help with the conformance rules work right now** because that effort is against 1.2, which is cast in stone — and any changes we make in 1.3 don’t make a difference.

Further, I feel a bit uncomfortable attempting to do this work without having actually established the guidelines per FR [#1059](https://github.com/FinOps-Open-Cost-and-Usage-Spec/FOCUS_Spec/issues/1059).  I think it’s very important that @ijurica  head up that work, and her time is currently fully spoken for.

Therefore, I propose that we:
 * Prioritize bringing FR #1059 into scope for 1.4.
   * Prioritize assignment to Irena (likely alongside FR [#1060](https://github.com/FinOps-Open-Cost-and-Usage-Spec/FOCUS_Spec/issues/1060) (metadata normative requirements guidelines) and [#1051](https://github.com/FinOps-Open-Cost-and-Usage-Spec/FOCUS_Spec/issues/1051) (glossary guidelines).
If we close on Irena’s 1.3 tasks early (haha :sweat_smile:), we can bring it forward ahead of time.
 * Close PR #1435 for now.
 
 @Matt-Cowsert @mike-finopsorg @ijurica Do you agree/disagree?

jpradocueva:
The group agreed to create the guidelines as defined in #1059. At that time this PR may be reopened.

PR #1452, opened 2025-09-09
Update github actions to build on all branches

### LINKED ISSUE #1450: [MAINTENANCE]: Branch builds are not triggered for branches with a slash in their name
### Description

It has been noticed that branches with a slash eg `matt-cowsert/1082-invoice-handling-requirements` do not get a build via GitHub actions. After some initial research it turns out a trigger pattern of '*' does not cross the / boundary.

### Definition of Done

Builds on branches including a / also run

mike-finopsorg:
Update github actions to build on all branches and avoid a / causing a build to skip.

Signed-off-by: Mike Fuller <mike@finops.org>

### DIFF
--- a/.github/workflows/working_draft.yml
+++ b/.github/workflows/working_draft.yml
@@ -2,8 +2,8 @@ name: generate_draft_spec
 on:
   push:
     branches-ignore:
-      - main
-      - candidate_release
+      - "main"
+      - "candidate_release"
 
 jobs:
   # The job that will use the container image you just pushed to ghcr.io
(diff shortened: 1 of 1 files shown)


.github/workflows/working_draft.yml
```diff
@@ -1,10 +1,9 @@
 name: generate_draft_spec
 on:
   push:
-    branches:
-      - "*"
-      - "!main"
-      - "!candidate_release"
+    branches-ignore:
+      - main
+      - candidate_release
```
shawnalpay: I know it works without the double quotes, but `main.yml` and `candidate_release.yml` use them, so it would be be good to stay consistent.
```suggestion
      - "main"
      - "candidate_release"
```

shawnalpay (approved):

PR #1461, opened 2025-09-12
match contract object layout

qquigley:
Updates to more closely align with contractapplied in PR 976.

### DIFF
--- a/specification/datasets/cost_and_usage/columns/allocatedmethoddetails.md
+++ b/specification/datasets/cost_and_usage/columns/allocatedmethoddetails.md
@@ -4,6 +4,14 @@ Allocated Method Details provides information about how resources are allocated 
 
 Allocated Resource Details consists of a valid JSON object which contains an array consisting of key-value objects describing the one or more factors that determined the split cost allocation. Each object consists of FOCUS-defined keys but can be extended to provide additional details about the allocation.
 
+The FOCUS-defined properties are:
+
+* `AllocatedRatio`: The ratio of a [*charge*](#glossary:charge) that this allocation represents.
+* `UsageUnit`: Unit being measured used to calculate this allocation.
+* `UsageQuantity`: The value of the charge applied to a single contract term.
+
+In addition to these, a provider may include one or more custom properties, also denoted as key-value pairs.
+
 ## Requirements
 
 ### Column Requirements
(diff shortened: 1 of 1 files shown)


specification/datasets/cost_and_usage/columns/allocatedmethoddetails.md
```diff
@@ -40,6 +48,12 @@ If AllocatedMethodDetails is not null, the JsonObjectFormat for AllocatedMethodD
 
 The following keys are used for allocation properties to facilitate querying data across allocations and across providers. Focus-defined keys will appear in the list below and Provider-defined keys will be prefixed with "x_" to make them easy to identify as well as prevent collisions.
 
+| Key | ValueType | Required | Description |
+| ----- | ---- | ---------- | ----------- |
+| AllocatedRatio | Numeric | True | Percentage of overall cost derived from corresponding method and metric. |
+| UsageUnit | [String](#stringhandling) | Conditional | Unit being measured used to calculate allocation. |
+| UsageQuantity | Numeric | False | Volume of UsageUnit consumed or used. |
+
```
shawnalpay, line 56: This makes me realize that I've duplicated this in PR #1350.  Should I only have one copy there (and then also here)?

specification/datasets/cost_and_usage/columns/allocatedmethoddetails.md
```diff
@@ -130,6 +144,53 @@ The `Elements` array contains one or more objects, each of which contains the fo
...
+    "UsageUnit" : "Hours",
+    "UsageQuantity" : 300
+  }
+}
+```
+### Scenario 2: Multiple "UsageUnit" values used for allocation
+
+When multiple "UsageUnit" values are used to calculate the allaction, another object is added to the "Elements" collection.
+
+```json
```
shawnalpay, line 164: Does this block show up as red JSON for you all also?  Is that a GitHub bug, or is there something wrong with the examples?
Screenshot 2025-09-12 at 3 16 13 PM

shawnalpay (approved):
The examples really help!  Please address my comments, and then merge as you see fit.

specification/datasets/cost_and_usage/columns/allocatedmethoddetails.md
qquigley: I'm hoping we can talk through this a bit at office hours. I think both places make sense, but I don't love the idea of maintaining two manually created identical static tables in the same md file. Maybe an anchor would suffice rather than showing the whole table again?

Also, there are some differences (like your's lists column, while this one has Key) that I think we should agree fit the JSON Object normative requirements, or make adjustments so they do.

specification/datasets/cost_and_usage/columns/allocatedmethoddetails.md
qquigley: Yep, same for me. I think it's a GitHub bug because it looks fine in VSCode for me.

PR #1465, opened 2025-09-15
Shift around ContractApplied normative requirements

shawnalpay:
@ijurica rightly pointed out something to me today about the location of various normative requirements on PR #1350.  Per her DM:

> Back to #1350 - this part struck me as a bit awkward and maybe even incorrect
> 
> > When ContractApplied is not null, ContractApplied adheres to the following additional requirements:
> > - ContractApplied objects MUST contain four key-value pairs, representing "ContractCommitmentID", "ContractCommitmentAppliedCost", "ContractCommitmentAppliedQuantity", and "ContractCommitmentAppliedUnit".
> > - ContractApplied objects MAY contain custom key-value pairs, representing additional datapoints provided by the data generator.
> 
> Would you agree that ContractApplied objects can be interpreted as values in the ContractApplied column?
> 
> And these requirements apply to "Elements" and not that parent object. Perhaps we could use the term contract application objects, which we've already used when describing key-value pairs? And perhaps even move these requirements to a different section, i.e., Object Schema Requirements?

Thus, I have attempted to shift around the normative requirements to seem more in-line with where they belong.

### DIFF
--- a/specification/datasets/cost_and_usage/columns/contractapplied.md
+++ b/specification/datasets/cost_and_usage/columns/contractapplied.md
@@ -32,10 +32,10 @@ Contract Applied consists of a valid JSON object which contains an array of key-
   * Each item in "Elements" MUST be an object.
   * "Elements" objects MUST conform to [KeyValueFormat](#key-valueformat) requirements.
   * "Elements" objects MUST contain key-value pairs (contract application properties).
+  * Contract application property keys SHOULD conform to [PascalCase](#glossary:pascalcase) format.
   * "Elements" objects MUST contain four key-value pairs, representing "ContractCommitmentID", "ContractCommitmentAppliedCost", "ContractCommitmentAppliedQuantity", and "ContractCommitmentAppliedUnit".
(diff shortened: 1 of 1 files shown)


specification/datasets/cost_and_usage/columns/contractapplied.md
```diff
@@ -22,29 +22,28 @@ The ContractApplied column adheres to the following requirements:
 * ContractApplied MUST conform to [JsonObjectFormat](#jsonobjectformat) requirements.
 * ContractApplied property keys SHOULD conform to [PascalCase](#glossary:pascalcase) format.
```
qquigley: I think an argument could be made that this could go in the schema requirements since it's specifically referring to keys within the JSON.

shawnalpay: @qquigley You're right.  But I think it needs to be reworded.

shawnalpay: I moved it down and reworded it.  Give it a look!

qquigley (approved):

PR #1474, opened 2025-09-17
update name to ResourceId and add Dataset requirement

jpradocueva:
Editorial update to change columnID to `ResourceId` instead of `ResourceID`, and add Dataset requirement.


jpradocueva:

PR #1493, opened 2025-09-30
FR #1070: Clarify use of "period" vs "duration" vs "term"

### LINKED ISSUE #1070: [FR] Clarify commitment concepts and glossary entries

### LINKED ISSUE #1380: [Process] 1.3 consistency review
This is a placeholder task to group the work associated with consistency review, which will occur Oct 16 to Oct 30.

shawnalpay:
When authoring the spec, we have been inconsistent in our use of the following words:

- `period`
- `duration`
- `term`

We have had a glossary entry for `term` that means "the length of a commitment discount".  However, with the introduction of Contract Commitments via PR #1350, it becomes important for the word "term" to not just mean length of time, but also any particular parameter specified on the contract.  **Thus, any legacy use of `term` in the FOCUS spec when describing time becomes potentially confusing and/or inaccurate.**

Therefore, this consistency PR performs the following changes:

- Add glossary entry for `period`, which is simply defined as "a length of time".
- **Change all previous use of `term` and `duration` to `period` (or `commitment period`, when applicable) when used in the context of *time*.**
- Strike `term` from the glossary.

As a consistency PR, this should not introduce any material change to the normative requirements of the specification.


robmartin33:
I'd suggest we leave the word "Term" in the glossary, and to define it as "A condition of a contract. References to time should use the word "Period". "

This will keep people from asking later "why don't we use the word "term" for contract term" since some providers use that word. And if someone is looking for the word Term they will be pointed explicitly to Period. 

I'd like to note, there was a great pun opportunity here and I passed it up. You're welcome.

specification/glossary.md
```diff
@@ -211,10 +215,6 @@ A metadata label assigned to a resource to provide information about it or to ca
 
 A Resource or Provider-defined construct for grouping resources and/or other Provider-defined construct that a Tag can be assigned to.
 
-<a name="glossary:term"><b>Term</b></a>
-
-A duration of a contractual agreement like with a [*commitment discount*](#glossary:commitment-discount) or [*negotiated discount*](#glossary:negotiated-discount).
```
aws-jrmarks, line 220: should we consider keeping "Term" in the glossary and adjusting the definition to move away from duration of time? It was mentioned during the weekly meeting today that term can mean provisions or clauses in an agreement, maybe we could keep all of the s/term/period and s/duration/period changes as proposed, _and_ keep Term here in the legal / contractual sense to avoid the pitfall of it getting used later.

specification/glossary.md
```diff
@@ -211,10 +215,6 @@ A metadata label assigned to a resource to provide information about it or to ca
 
 A Resource or Provider-defined construct for grouping resources and/or other Provider-defined construct that a Tag can be assigned to.
```
shawnalpay, line 222: Suggestion to keep the glossary entry "Term", per feedback from @aws-jrmarks and @robmartin33.  Let me know if this addresses the concern or if I've been too verbose:

```suggestion
<a name="glossary:term"><b>Term</b></a>

A condition of a [*contract*](#glossary:contract). (Previous FOCUS versions used the word "term" to reference time; FOCUS now uses the word [*period*](#glossary:period) for that purpose.)
```

specification/glossary.md
shawnalpay: Added suggestion [here](https://github.com/FinOps-Open-Cost-and-Usage-Spec/FOCUS_Spec/pull/1493/files#r2402464809); please lmk if that works.  (GitHub does not support suggestions on deleted lines 😞.)

specification/glossary.md
```diff
@@ -159,6 +159,10 @@ The [charge](#glossary:charge) that existed prior to an operation. This is used
 
 Pascal Case (PascalCase, also known as UpperCamelCase) is a format for identifiers which contain one or more words meaning the words are concatenated together with no delimiter and the first letter of each word is capitalized.
 
+<a name="glossary:period"><b>Period</b></a>
+
+A length of time.
```
ijurica: nit: 

I noticed that in the definitions of `Charge Period` and `Billing Period`, we use the phrasing "*The time window...*".
While "*length of time*" can be interpreted as open-ended, "*time window*" implies clearly defined **start and end boundaries**. 
So, if we want to be just a bit more precise, perhaps we should consider using this alternative.

```suggestion
A time window.
```

specification/glossary.md
ijurica: I suggest we omit the reference to the previous version.

ijurica (approved):

specification/glossary.md
shawnalpay: Understood.  But let's maybe be even a little more specific.  How about:

```suggestion
A time window, with a specifically defined start and end date/time.
```

specification/glossary.md
shawnalpay: Fair enough.  So then:

```suggestion
<a name="glossary:term"><b>Term</b></a>

A condition of a [*contract*](#glossary:contract).
```

specification/glossary.md
shawnalpay: Confirmed with @aws-jrmarks that that works for him!  Resolving.

aws-jrmarks (approved):

specification/glossary.md
```diff
@@ -214,11 +218,9 @@ A metadata label assigned to a resource to provide information about it or to ca
 <a name="glossary:tag-source"><b>Tag Source</b></a>
 
 A Resource or Provider-defined construct for grouping resources and/or other Provider-defined construct that a Tag can be assigned to.
-
 <a name="glossary:term"><b>Term</b></a>
 
-A duration of a contractual agreement like with a [*commitment discount*](#glossary:commitment-discount) or [*negotiated discount*](#glossary:negotiated-discount).
-
+A condition of a [*contract*](#glossary:contract).
```
shawnalpay: Per feedback from @SatyaVandrangi1 in Oct 7 TF1, we need to use a word other than "condition".  Per [this article](https://www.enzuzo.com/blog/difference-between-terms-and-conditions), there is a meaningful difference between "term" and "condition".  Terms are the _agreements_; _conditions_ are the expectations that must be met to fulfill those terms.

Thus, I am changing this to:

```suggestion
An agreement specified on a [*contract*](#glossary:contract).
```

ljadvey (approved):

jpradocueva:
PR was approved during the Members’ Meeting on October 9.

jpradocueva (approved):
Approved by the Members' group on the October 9th call.

Issue #1508, opened 2025-10-09
[AI] Add Requirements subheader to every column definition


shawnalpay:
### Detailed Description

There is currently no separation between the beginning section of a column description and the requirements that follow.  We started introducing this subheader in some (but not all) of the new 1.3 columns.  It would be good to be consistent about this, and also, it would be helpful for the reader to see more quickly / precisely where the requirements are.

### Definition of Done

A pull request that adds a Requirements subheader to every column across both `Cost and Usage` and `Contract Commitment`.

shawnalpay:
I have added draft PR #1509 with an example.  That pattern could be followed by any member of the working group.

shawnalpay:
Completed via the PRs we merged today!

PR #1509, opened 2025-10-09
AI #1508: Add Requirements subheader to every column definition

shawnalpay:
This PR demonstrates how to execute on AI #1508.

This pattern would need to be repeated for every column definition in the specification.

### DIFF
--- a/specification/datasets/cost_and_usage/columns/allocatedmethodid.md
+++ b/specification/datasets/cost_and_usage/columns/allocatedmethodid.md
@@ -2,7 +2,7 @@
 
 Allocated Method ID is the unique identifier for the [allocated method](#glossary:allocated-method) defined by the provider which was used for the [Provider-Calculated Split Cost Allocation](#providercalculatedsplitcosthandling). This unique identifier can be used to find how the [allocated charge](#glossary:allocated-charge) was calculated in the provider's documentation.
 
-The AllocatedMethodId column adheres to the following requirements:
+## Requirements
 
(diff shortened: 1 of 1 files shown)


specification/datasets/cost_and_usage/columns/allocatedmethodid.md
```diff
@@ -2,7 +2,7 @@
 
 Allocated Method ID is the unique identifier for the [allocated method](#glossary:allocated-method) defined by the provider which was used for the [Provider-Calculated Split Cost Allocation](#providercalculatedsplitcosthandling). This unique identifier can be used to find how the [allocated charge](#glossary:allocated-charge) was calculated in the provider's documentation.
 
-The AllocatedMethodId column adheres to the following requirements:
+## Requirements
```
ijurica, line 5: I really like the proposed change, especially since we already use the Requirements sub-header for attributes.

---

@shawnalpay Would you mind keeping the intro sentence as well?

Each column and each attribute in the conformance JSON files starts with its corresponding Composite Function, which brings all related requirements together.

I prefer the following phrasing:

> "MustSatisfy": "{ColumnOrAttributeId} [column|attribute] adheres to the following requirements:"

over a more generic version such as:

> "MustSatisfy": "Requirements"

This is particularly relevant for columns that contain requirements referring not to the column name itself, but to one of its sub-elements (see SkuPriceDetails, for example).

```suggestion
## Requirements

The AllocatedMethodId column adheres to the following requirements:
```

shawnalpay: @ijurica 
I see.  So how about:

```suggestion
## Requirements

AllocatedMethodId adheres to the following requirements:
```

PR #1536, opened 2025-10-21
AI #1508: Update Requirements headers for allocatedresourceid.md through chargeperiodend.md

davidearney:
Requirements subheader

PR #1559, opened 2025-10-28
AI #1508: Update Requirements headers for contract_commitment columns

### LINKED ISSUE #1508: [AI] Add Requirements subheader to every column definition

davidearney:
Changed to match the language in the other Requirements areas throughout the document.

### DIFF
--- a/specification/datasets/contract_commitment/columns/billingcurrency.md
+++ b/specification/datasets/contract_commitment/columns/billingcurrency.md
@@ -2,7 +2,7 @@
 
 [*Billing currency*](#glossary:billing-currency) is an identifier that represents the currency of a [*contract commitment*](#glossary:contract-commitment).
 
-The BillingCurrency column adheres to the following requirements:
+## Requirements
 
 * BillingCurrency MUST be present in a [*FOCUS dataset*](#glossary:FOCUS-dataset).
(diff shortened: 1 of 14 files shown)


shawnalpay (changes requested):
We need to retain the statement "[[X]] adheres to the following requirements:" in all columns, in addition to the Requirements subheader, as illustrated in [this comment](https://github.com/FinOps-Open-Cost-and-Usage-Spec/FOCUS_Spec/pull/1536#discussion_r2455896940) on #1536. (This is a small change from the previous pattern of "The [[X]] column adheres to the following requirements:")

Please apply this pattern to all the columns.

shawnalpay (approved):

jpradocueva (approved):
@davidearney good work!

jpradocueva:
Approved by the group without objections or abstentions during the October 30 Members' call.

PR #1650, opened 2025-11-25
AI #1630: Document recent decisions by the TF-RM on naming and numbering

mike-finopsorg:
Signed-off-by: Mike Fuller <mike@finops.org>

--

# PR Summary
(added by @shawnalpay, AI-assisted)

## Overview
Update the **FOCUS writing guidelines** to document decisions made by the Requirements Model Task Force regarding **naming conventions**, **numbering**, and **entity structure**, specifically focusing on how to handle **multi-dataset** scenarios.

## Key Changes
The primary changes are to `guidelines/writing-requirements-model-guidelines.md`:

1.  **Entity Structure & Naming**:
    * Documented decisions regarding **multi-dataset entity structure**.
    * Updated **naming and numbering** conventions for the spec.
2.  **Resource Model (RM) Keys**:
    * Added details about a new **"Order" key** in the Resource Model.
    * Added **"Dataset" keys** to the guidelines.
3.  **Diagrams**:
    * Updated the **entity diagram** to reflect structural changes.

PR #1668, opened 2025-12-08
FR #1466: Add all 1.3 spec changes to the requirements model

### LINKED ISSUE #1466: [FR] Add 1.3 Specification revisions to Requirements Model
### Problem Statement

Practitioners need to validate the output they are receiving from their provider matches the FOCUS spec for a given version (e.g., v1.3).

---

### Use Case / User Story

As a FinOps Practitioner, I want to evaluate the output of my provider's FOCUS implementation of a given version, so I can rely on the providers' output as I integrate FOCUS into my FinOps practice.

---

### Success Criteria

- [ ] Practitioners can validate a provider's FOCUS 1.3 output against the Requirements Model and receive accurate pass/fail results for column, attribute, and metadata conformance
- [ ] Conformance tooling authors can programmatically reference 1.3-specific requirements (including any new or changed columns, attributes, and metadata) from the Requirements Model
- [ ] Data generators can verify their 1.3 implementation completeness by checking their output against every normative requirement captured in the model

---

### Type of Request

- [x] **Enhancement** – Refinement or extension of an existing FOCUS column, metric, or capability
- [ ] **Net New** – A concept not yet represented in FOCUS or widely available from data generators
- [ ] **Enablement/Supporting Content** – Non-normative items like appendix explanations, examples, or supporting documentation

---

_No response_

---

_No response_

---

_No response_

---

_No response_

---

_No response_

---

---

## Maintainer Assessment

### Adoption Impact

- [ ] **Adoption Accelerator** – This feature would help organizations start using FOCUS
- [ ] **Practitioner Enhancement** – Improves experience for existing FOCUS users
- [x] **Technical Improvement** – Internal specification refinement

**Justification:**

No specific organizations are listed, but this is infrastructure work that underpins conformance validation for the entire ecosystem. Without it, the Requirements Model is incomplete for 1.3 and tooling authors cannot build version-accurate validators.

- **Gap:** Practitioners and tooling authors cannot validate provider output against 1.3-specific normative requirements because the Requirements Model does not yet include 1.3 revisions.
- **Org signals:** No organizations listed. This is an internally driven maintenance item, but all conformance tooling consumers depend on it.
- **Adoption position:** Does not block initial adoption, but undermines trust in conformance tooling. Practitioners who rely on automated validation get stale results until this is delivered.

---

**Feature Description:**

This request enhances the Schema Metadata supported feature by ensuring the machine-readable Requirements Model reflects all normative changes introduced in FOCUS 1.3. It does not introduce new columns or capabilities.

- **Current limitation:** The Requirements Model captures requirements up to a prior version. 1.3 changes to columns, attributes, and metadata are not yet represented, so conformance checks cannot distinguish pre-1.3 from 1.3 requirements.
- **Use case impact:** Conformance tooling authors, data generators verifying implementation completeness, and practitioners validating provider output all depend on version-accurate requirements.
- **Structural gap:** The Requirements Model exists but lacks 1.3-specific entries. This is a content gap, not a structural design gap.
- **Design note:** This is a mechanical update. The model's schema and format are established. The work is auditing 1.3 changes and encoding them.

---

### Implementation Scope

**MVP Definition:**

Encode all normative requirement changes from FOCUS 1.3 into the Requirements Model, including new columns, modified attributes, and updated metadata requirements. Enable conformance tools to run version-specific 1.3 validation. Satisfies Success Criteria 1, 2, and 3.

**North Star Vision:**

A Requirements Model that is updated as part of the release process for every FOCUS version, with automated diff generation between versions so data generators can quickly identify implementation deltas.

**Phasing Strategy:**

Single phase. MVP and North Star are equivalent for the 1.3 update itself. The automated release-process integration is a separate process improvement.

---

### Specification Impact

**Impacted Parties:**
- [ ] Other: [specify]

Practitioners use conformance results to trust provider data. Data generators check their 1.3 implementation against the model. Vendors building conformance tooling need version-specific requirement references.

**Level of Ambiguity:**

2

The Requirements Model schema already exists and is proven. The work is enumerating and encoding 1.3-specific changes. Minor judgment calls may arise around how to represent changed vs. new requirements, but the approach is well-established.

---

### FinOps Scope Alignment

Agree with requester. The Requirements Model applies to all providers generating FOCUS-compliant data across cloud, SaaS, and data center scopes.

## Community Support

  - Your **organization**

mike-finopsorg:


shawnalpay:
Amazing stuff @mike-finopsorg!!  Flipping this to TF Review.  Looking forward to reviewing it in TF-RM next week.

specification/requirements_model/model_rules/datasets/cost_and_usage/columns/billedcost.json
```diff
@@ -157,23 +193,29 @@
       },
       "Condition": {
         "CheckFunction": "CheckNotSameValue",
-        "ColumnAName": "ProviderName",
+        "ColumnAName": "ServiceProviderName",
```
ijurica: For a Marketplace offering purchase, the expected values are:

- ServiceProviderName: SaaS provider
- InvoiceIssuerName: CSP
- BilledCost: 123.45

Given the current definitions of participating entities, I believe this can only be handled as a dynamic rule.

We likely need an entity that reflects where the information was originally produced, regardless of which entity later transformed, enriched, and/or delivered the data.

Options: introduce an additional column or repurpose the DataGenerator (available in Metadata).

specification/requirements_model/releases/1.3/model_rules/attributes/columnhandling.json
ijurica, line 1: **Question:** were the column handling exceptions intentionally omitted (e.g. because they currently contain no normative verbs), or was this unintentional?

Attaching the following for context:

> ## Exceptions
> 
> * Identifiers will use the "Id" abbreviation since this is a standard pattern across the industry.
> * Product offerings that incur charges will use the "Sku" abbreviation because it is a well-understood term both within and outside the industry.

specification/requirements_model/model_rules/datasets/cost_and_usage/columns/invoiceid.json
```diff
@@ -1,60 +1,77 @@
 {
-  "InvoiceId-C-000-O": {
+  "CAU-InvoiceId-C-000-O": {
```
thecloudman: Change to CAU-InvoiceID-C-000-O

specification/requirements_model/model_rules/datasets/cost_and_usage/columns/allocatedmethodid.json
```diff
@@ -1,5 +1,5 @@
 {
-  "CAU-AllocatedMethodId-C-000-C": {
+  "CAU-AllocatedMethodID-C-000-C": {
```
jpradocueva: @thecloudman, the requirement model ID (RMID) is composed as follows:
<DatasetType>-<EntityId>-<Entity>-NNN-<Status>
In this case:
- <DatasetType> = CAU (Cost & Usage)
- <EntityId> = Column Id (section in specs) = AllocatedMethodId
- <Entity> = Column = C
- NNN = 000 (root leading requirement)
- <Status>= C,  M = Mandatory; O = Optional; C = Conditional
In this case, the <EntityId> matches what we have in the ## Column ID section = "AllocatedMethodId" rather than "AllocatedMethodID".

specification/requirements_model/releases/1.3/model_rules/attributes/columnhandling.json
mike-finopsorg: Intentional, no BCP14 terms makes these non-normative. We should resolve in 1.4

specification/requirements_model/model_rules/datasets/cost_and_usage/columns/billedcost.json
mike-finopsorg: Is your recommendation to make this rule dynamic today or keep the ServiceProviderName and refine in future release?

jpradocueva (commented):
@thecloudman, the requirement model ID (RMID) is composed as follows:
---NNN-
In this case:

    = CAU (Cost & Usage)
    = Column Id (section in specs) = AllocatedMethodId
    = Column = C
    NNN = 000 (root leading requirement)
    = C, M = Mandatory; O = Optional; C = Conditional
    In this case, the matches what we have in the ## Column ID section = "AllocatedMethodId" rather than "AllocatedMethodID".

specification/requirements_model/model_rules/datasets/cost_and_usage/columns/allocatedmethodid.json
mike-finopsorg: Hey @thecloudman, yeah the Id is intentional. For changes like this it would be preferred if you raise a comment on the PR first so we notice the changes coming in. I have reverted the ID to Id to clear the build errors.

thecloudman:
Yeah sorry. Figured I’d done something wrong from a notice about a build
fail but then internet access dropped off.

beau-nelford-anglepoint (changes requested):
Typo in filename: specification/requirements_model/model_rules/datasets/cost_and_usage/objects/allocatedmedthoddetail.json should be "..allocatedmethoddetail.json"

beau-nelford-anglepoint (commented):
The spec column ID is InvoiceIusserName, the compiled model uses InvoiceIusserName, but the PR json file is named specification/requirements_model/mode_rules/datasets/cost_and_usage/columns/invoiceissuer.json. Should that be "invoiceissuername.json"?

specification/requirements_model/model_rules/datasets/cost_and_usage/columns/billedcost.json
mike-finopsorg: @ijurica

ijurica: My recommendation is to make it Dynamic in 1.3

mike-finopsorg: OK I will make this Dynamic

ijurica:
# Attributes review

## 1. Column Handling

### ⚠️ Anchor Requirement
  
The intro section contains the following sentence with a BCP14 keyword. I assume this omission is intentional.

> All columns defined in the FOCUS specification MUST follow the naming and ordering requirements listed below.

### ⚠️ Exceptions
  
The Exceptions section appears to have been omitted. I assume this is intentional, since it does not contain BCP14 keywords.

> * Identifiers will use the "Id" abbreviation since this is a standard pattern across the industry.  
> * Product offerings that incur charges will use the "Sku" abbreviation because it is a well-understood term both within and outside the industry.

### ❌ Custom columns composite requirement

The requirement related to Custom columns is currently nested under "All columns defined by FOCUS". However, Custom columns are not defined by FOCUS. Therefore, this requirement should not be structurally nested under that parent requirement.

> * All columns defined by FOCUS MUST follow the following rules:
>   * ...
>   * Custom (e.g., service-provider-defined) columns that are not defined by FOCUS but included in a FOCUS dataset MUST follow the following rules:
>     * Custom columns MUST be prefixed with a consistent `x_` prefix to identify them as external, custom columns and distinguish them from FOCUS columns to avoid conflicts in future releases.
>     * Custom columns SHOULD follow the same rules listed above for FOCUS columns.

---

## 2. CurrencyFormat

### ⚠️ Anchor Requirement and Custom Columns

The intro section contains the following two sentences with BCP14 keywords:

> All columns capturing a currency value, defined in the FOCUS specification, MUST follow the requirements listed below.  
> Custom currency-related columns SHOULD also follow the same formatting requirements.

However, the RM output currently states:

> * CurrencyFormat adheres to the following requirements:
>   * ...

It appears that the Custom column-related requirement (SHOULD) is not reflected in the RM output.

Was this omission intentional? If not, should we distinguish between FOCUS column and Custom column requirements (i.e., explicitly repeat them for Custom columns using the **SHOULD** keyword)?

---

## 3. DataGeneratorCalculatedSplitCostAllocationHandling

### 📝 Anchor Requirement

The anchor requirement is not provided in the spec. The RM output contains:

> * DataGeneratorCalculatedSplitCostAllocationHandling adheres to the following requirements:
>   * ...

LGTM.

---

## 4. DateTimeFormat

### ⚠️ Anchor Requirement and Custom Columns

The intro section contains the following two sentences with BCP14 keywords:

> All columns capturing a date/time value, defined in the FOCUS specification, MUST follow the formatting requirements listed below. 
> Custom date/time-related columns SHOULD also follow the same formatting requirements.

However, the RM output currently states:

> * DateTimeFormat adheres to the following requirements:
>   * ...

Was this omission intentional? If not, should we distinguish between FOCUS column and Custom column requirements (i.e., explicitly repeat them for Custom columns using the **SHOULD** keyword)?

### 📝 Missing BCP14 keyword

BCP14 keywords appear to be missing from several nested requirements in the spec. In the RM output, they have been added explicitly.

> * ...
>   * **MUST** Include the date and ...
>   * **MUST** Use two-digit ...

---

## 5. DiscountHandling

### 📝 Overall Assessment

#### Findings

* **Anchor Requirement Difference:** There is a difference in the anchor requirement between the Spec and the RM output. I assume this change was intentional.

Spec:

> All rows defined in FOCUS MUST follow the discount handling requirements listed below.

RM output:

> The DiscountHandling attribute adheres to the following requirements:

* **Nesting Difference:** There is a nesting difference in one of the composite requirements. However, it does not appear to introduce a material change in meaning.

* **Requirement Decomposition:** In several places, a single bullet point in the Spec contains multiple normative statements, whereas the RM output breaks them down into separate bullets.

### Overall Conclusion

The DiscountHandling requirements in the Spec should be revisited (ideally in FOCUS 1.4). For **FOCUS 1.3**, however, the **differences appear tolerable**, as all requirements remain Dynamic (informational).

---

## 6. InvoiceHandling

### 📝 Anchor Requirement

The anchor requirement is not provided in the spec. The RM output contains:

> * InvoiceHandling adheres to the following requirements:
>   * ...

### ⚠️ Nesting issue in the Exception

Spec:

>  * Informational line items that have zero monetary impact and are included solely for transparency MAY be excluded. Examples include:
>    * Tax exemption notifications
>    * SLA credit details when the credit is already applied to the charged amount
>  * If such informational items are excluded, data generators MUST document this in their FOCUS implementation guide and ensure the sum of included charges still equals the invoice total.

RM output (current):

>  * Informational line items that have zero monetary impact and are included solely for transparency MAY be excluded. Examples include:
>  * Tax exemption notifications
>  * SLA credit details when the credit is already applied to the charged amount.
>  * If such informational items are excluded, data generators MUST document this in their FOCUS implementation guide and ensure the sum of included charges still equals the invoice total.

RM output (suggestion):

>  * Informational line items that have zero monetary impact and are included solely for transparency MAY be excluded. Examples include: Tax exemption notifications, SLA credit details when the credit is already applied to the charged amount.
>  * If such informational items are excluded, data generators MUST document this in their FOCUS implementation guide and ensure the sum of included charges still equals the invoice total.

---

## 7. JsonObjectFormat

LGTM.

---

## 8. KeyValueFormat

LGTM.

---

## 9. NullHandling

TODO

---

## 10. NumericFormat

TODO

---

## 11. StringHandling

TODO

---

## 12. UnitFormat

TODO

jpradocueva:
### Latest Identified Issues:
@mike-finopsorg, please review the list below:

- [x]  `CAU-PricingCategory-C-000-M` is missing ApplicabilityCriteria.
- [x] `CAU-ResourceId-C-007-M` was missing the word "service". This commit added - [commit](https://github.com/FinOps-Open-Cost-and-Usage-Spec/FOCUS_Spec/pull/1668/changes/9712e407d230b311db246a929692a7488d20d8ed)
- [x] `CAU-ContractedUnitPrice-C-009-C` and `CAU-ContractedUnitPrice-C-009-C010-C` are marked as conditional, but I believe they are Mandatory.
- [x] `"CAU-ContractAppliedObject-O-028-M"` contains incorrectly `-033-C` as a child when it should be `-032-C`
- [x] `"CAU-ContractAppliedObject-O-059-M"` contains `-C-065-C` as a child when it should be `-C-064-C`
- [x] `"CAU-ConsumedQuantity-C-003-C"` is missing a requirement that has not been recorded with an ID. Requirement: `ConsumedQuantity MUST be null when SkuPriceId is null.` is missing entirely. 
- [x] Incorrect order for `"CAU-BillingAccountId-C-001-M"`, it cannot be "20" as it is the 3rd requirement
- [x] "CAU-AllocatedMethodDetailsObject-O-008-O", `O-009-C`, `O-010-O`, `O-011-M`, `O-015-M`, and `O-019-O` are not connected to a parent, are floating requirements. Also, `O-012-M`, `O-013-M`, and `O-014-M` are duplicated. They are connected to `O-004-M` and also to `O-011-M`.
- [x] `"CCT-ContractPeriodStart-C-004-M"` contains an extra word at the end of the requirement "commitment".
- [x] `"CCT-ContractPeriodEnd-C-004-M"` same as above, it contains an extra word at the end of the requirement sentence, "commitment" that doesn't exist in the Specs.
- [x] `"StringHandling-A-001-M"`. It is defined in the model, it is not connected to any parent and it doesn't appear to be defined in the Specs

guidelines/writing-requirements-model-guidelines.md
```diff
@@ -30,7 +30,13 @@ The first stage of conversion of rules from the normative text to model rules is
...
+- `Reference` - Area of the specification this entry relates to
+- `EntityType` - The type of the entity this rule applies to (Valid types: `Dataset`, `Column`, `Object`, `Attribute`)
+- `EntityId` - The Id of the entity this rule applies to
+- `EntityName` - The Name of the entity this rule applies to
+- `DatasetType` - The three letter acronym for the dataset of the dataset entity this rule is related to
+- `DatasetId` - The Id of the dataset entity this rule is related to
+- `DatasetName` - The Name of the dataset entity this rule is related to
 - `ApplicabilityCriteria` - Specific criteria that must be true of the data generator for this rule to apply to the dataset
 - `Condition` - The definition of conditions under which this rule applies
 - `Requirement` - The definition of what is required for model
```
thecloudman: Is this missing some words to outline what the requirement really is? The definition of what is required for model

specification/requirements_model/releases/1.3/model_rules/attributes/datageneratorcalculatedsplitcostallocationhandling.json
```diff
@@ -0,0 +1,302 @@
...
+    "EntityName": "Data Generator Calculated Split Cost Allocation Handling",
+    "EntityId": "DataGeneratorCalculatedSplitCostAllocationHandling",
+    "Notes": "",
+    "ModelVersionIntroduced": "1.3",
+    "Status": "Active",
+    "ApplicabilityCriteria": [],
+    "Type": "Dynamic",
+    "Order": 40,
+    "ValidationCriteria": {
+      "MustSatisfy": "For each allocated charge records in a FOCUS dataset, all dimension columns and non-aggregatable metric columns MUST match the values of the origin charge record.",
```
thecloudman, line 165: Should this read: "For each allocated charge record" instead of "For each allocated charge records"

thecloudman (changes requested):

specification/requirements_model/releases/1.3/model_rules/attributes/datageneratorcalculatedsplitcostallocationhandling.json
jpradocueva: @ijurica, This comment should be cleaned up in Release v1.4.

mike-finopsorg:
> # Attributes review
> ## 1. Column Handling
> ### ⚠️ Anchor Requirement
> The intro section contains the following sentence with a BCP14 keyword. I assume this omission is intentional.

Yes

> ### ⚠️ Exceptions
> The Exceptions section appears to have been omitted. I assume this is intentional, since it does not contain BCP14 keywords.

Yes

> ### ❌ Custom columns composite requirement
> The requirement related to Custom columns is currently nested under "All columns defined by FOCUS". However, Custom columns are not defined by FOCUS. Therefore, this requirement should not be structurally nested under that parent requirement.

This is a modelling issue, we would then end up with two top level rules, I believe we need to either have a new top level that wraps both FOCUS defined and Custom Columns or rephrase the top level key to apply to both. Leaving this as is for 1.3 and look forward to improvements in 1.4.

> ## 2. CurrencyFormat
> ### ⚠️ Anchor Requirement and Custom Columns
> The intro section contains the following two sentences with BCP14 keywords:

Added rule

> Was this omission intentional? If not, should we distinguish between FOCUS column and Custom column requirements (i.e., explicitly repeat them for Custom columns using the **SHOULD** keyword)?
Similar to above we need to either have a top level that covers both or have nested composites for the two situations. 

> ## 3. DataGeneratorCalculatedSplitCostAllocationHandling
> ### 📝 Anchor Requirement
> The anchor requirement is not provided in the spec. The RM output contains:
> LGTM.

👍 

> ## 4. DateTimeFormat
> ### ⚠️ Anchor Requirement and Custom Columns
> The intro section contains the following two sentences with BCP14 keywords:
> 
> > All columns capturing a date/time value, defined in the FOCUS specification, MUST follow the formatting requirements listed below.
> > Custom date/time-related columns SHOULD also follow the same formatting requirements.

Added

> ### 📝 Missing BCP14 keyword
> BCP14 keywords appear to be missing from several nested requirements in the spec. In the RM output, they have been added explicitly.

I think we need to resolve these in 1.4. We will continue to carry them with the added intent in the model

> ## 5. DiscountHandling
> ### 📝 Overall Assessment
> #### Findings
> * **Anchor Requirement Difference:** There is a difference in the anchor requirement between the Spec and the RM output. I assume this change was intentional.

The model has a built in preference for "adheres to the following requirements:"

> 
> * **Nesting Difference:** There is a nesting difference in one of the composite requirements. However, it does not appear to introduce a material change in meaning.

Corrected

> * **Requirement Decomposition:** In several places, a single bullet point in the Spec contains multiple normative statements, whereas the RM output breaks them down into separate bullets.

Yes needed for modelling, hoping we fix in 1.4

> 
> ### Overall Conclusion
> The DiscountHandling requirements in the Spec should be revisited (ideally in FOCUS 1.4). For **FOCUS 1.3**, however, the **differences appear tolerable**, as all requirements remain Dynamic (informational).
> 
> ## 6. InvoiceHandling
> ### 📝 Anchor Requirement
> The anchor requirement is not provided in the spec. The RM output contains:
> 
> > * InvoiceHandling adheres to the following requirements:
> >   
> >   * ...
> 
> ### ⚠️ Nesting issue in the Exception
> Spec:

updated to suggestion, would be good to ban newlines inside normative rules.

> 
> ## 7. JsonObjectFormat
> LGTM.
> 
> ## 8. KeyValueFormat
> LGTM.


Commit incoming today with these updates.

guidelines/writing-requirements-model-guidelines.md
mike-finopsorg: Added more words to round out the description, commit incoming today with the fix.

specification/requirements_model/releases/1.3/model_rules/attributes/datageneratorcalculatedsplitcostallocationhandling.json
mike-finopsorg: Closing this out as its not relevant to 1.3 RM, I believe @ijurica has captured this for the clean up work happening in 1.4

mike-finopsorg:
> * Incorrect order for `"CAU-BillingAccountId-C-001-M"`, it cannot be "20" as it is the 3rd requirement

I don't think this is correct @jpradocueva 

CAU-BillingAccountId-C-000-M: Order 0 "BillingAccountId adheres to the following requirements:"
CAU-CostAndUsage-D-024-M: Order 10 "BillingAccountId MUST be present in a Cost and Usage FOCUS dataset ."
CAU-BillingAccountId-C-001-M: Order 20 "BillingAccountId MUST be of type String"

mike-finopsorg:
> * `"CCT-ContractPeriodStart-C-004-M"` contains an extra word at the end of the requirement "commitment".

I do not believe this is an extra word, the specification has the words: "effective period of
the contract commitment ."

jpradocueva:
@mike-finopsorg on response to your comments:

About:

> "CCT-ContractPeriodStart-C-004-M" contains an extra word at the end of the requirement "commitment".
 "CCT-ContractPeriodEnd-C-004-M" same as above, it contains an extra word at the end of the requirement sentence, "commitment" that doesn't exist in the Specs.
- I don't see the extra word in [`contractperiodend`](https://github.com/FinOps-Open-Cost-and-Usage-Spec/FOCUS_Spec/blob/v1.3/specification/datasets/contract_commitment/columns/contractperiodend.md) and [`contractperiodstart` ](https://github.com/FinOps-Open-Cost-and-Usage-Spec/FOCUS_Spec/blob/v1.3/specification/datasets/contract_commitment/columns/contractperiodstart.md)

About:
>  Incorrect order for "CAU-BillingAccountId-C-001-M", it cannot be "20" as it is the 3rd requirement.

- In the model-1.3.json `"CAU-BillingAccountId-C-001-M",` = "BillingAccountId MUST NOT be null.", but this is the third requirement, not the first.

mike-finopsorg:
> @mike-finopsorg on response to your comments:
> 
> About:
> 
> > "CCT-ContractPeriodStart-C-004-M" contains an extra word at the end of the requirement "commitment".
> > "CCT-ContractPeriodEnd-C-004-M" same as above, it contains an extra word at the end of the requirement sentence, "commitment" that doesn't exist in the Specs.
> 
> * I don't see the extra word in [`contractperiodend`](https://github.com/FinOps-Open-Cost-and-Usage-Spec/FOCUS_Spec/blob/v1.3/specification/datasets/contract_commitment/columns/contractperiodend.md) and [`contractperiodstart` ](https://github.com/FinOps-Open-Cost-and-Usage-Spec/FOCUS_Spec/blob/v1.3/specification/datasets/contract_commitment/columns/contractperiodstart.md)
> 

My bad I was looking at ContractCommitmentPeriodEnd. Fixed now.

> About:
> 
> > Incorrect order for "CAU-BillingAccountId-C-001-M", it cannot be "20" as it is the 3rd requirement.
> 
> * In the model-1.3.json `"CAU-BillingAccountId-C-001-M",` = "BillingAccountId MUST NOT be null.", but this is the third requirement, not the first.

I understand now, also fixed.

jpradocueva (approved):
Approved! Ready for WG Review & Approval

shawnalpay:
Amazing!!  Advancing to Member review.  Thank you @mike-finopsorg and @jpradocueva, really herculean work on this one!

supporting_content/requirements_model/1.2-1.3MigrationDocs/attributes/ColumnHandling.md
shawnalpay, line 1: Doesn't follow same casing / convention as the other files?

supporting_content/requirements_model/Initial_stage1_content/attributes/discounthandling.json
shawnalpay, line 1: Do we need this stage1 content?  If not, can we delete?

shawnalpay (changes requested):
Please have a look at the `supporting_content/requirements_model` section.  While this isn't spec or model content, it still makes the PR look much heavier than it really is -- and in the case of the 1.2 RM PR, we made sure to drop all the stage 1 content.  Can we jettison some or all of these new files?

supporting_content/requirements_model/1.2-1.3MigrationDocs/attributes/ColumnHandling.md
mike-finopsorg: @jpradocueva if you keep this file in your clean up can you correct

supporting_content/requirements_model/Initial_stage1_content/attributes/discounthandling.json
mike-finopsorg: @jpradocueva is going to review

thecloudman (approved):

beau-nelford-anglepoint (approved):
Filename changes I requested look good. Thank you!

shawnalpay (changes requested):
There are conflicts in the guidelines file between #1650 and #1668.  Please either remove the guidelines file from #1668, or clean up #1650 (including addressing all 14 open threads over there) and merge into #1668.  Then I can approve.

specification/requirements_model/tests/test_dependencies_not_deprecated.py
shawnalpay, line 1: This appears to be a total duplicate to `test_dependencies_not_removed.py`.  I think you can safely delete it?

specification/requirements_model/releases/1.3/model_rules/datasets/cost_and_usage/columns/skupricedetails.json
shawnalpay, line 1: In CAU-SkuPriceDetails-C-006-C, the newly added rule CAU-SkuPriceDetails-C-023-C is listed at the bottom of the Dependencies array, but it was forgotten in the Items array for the AND CheckFunction.

specification/requirements_model/model_rules/datasets/cost_and_usage/objects/contractappliedobject.json
```diff
@@ -0,0 +1,2149 @@
...
+    "ModelVersionIntroduced": "1.3",
+    "Status": "Active",
+    "ApplicabilityCriteria": [],
+    "Order": 80,
+    "DatasetType": "CAU",
+    "DatasetId": "CostAndUsage",
+    "DatasetName": "Cost And Usage",
+    "Type": "Dynamic",
+    "ValidationCriteria": {
+      "MustSatisfy": "\"Elements\" objects MAY contain custom key-value pairs, representing additional typepoints provided by the type generator.",
```
shawnalpay: I think this was a copy/paste gone awry.

```suggestion
      "MustSatisfy": "\"Elements\" objects MAY contain custom key-value pairs, representing additional datapoints provided by the data generator.",
```

guidelines/writing-requirements-model-guidelines.md
```diff
@@ -104,7 +110,8 @@ Provider -.->|Used in rule conditions| Column
...
 | `Dataset`          | Whole billing dataset                               | Structural presence, versioning, coverage | Dataset MUST include all columns required by the declared FOCUS version                                         |
 | `Row`              | Individual line item in dataset                     | Logic conditions, nullability, alignment  | Rows with `ChargeCategory = Purchase` MUST contain a `SkuId`                                                    |
-| `Column`           | Named field across rows                             | Data type, format, constraints            | Column `BillingPeriodStart` MUST be of type `DateTime`                                                          |
+| `Column`           | Named field across rows                             | Data type, format, constraints            | Each item in `AllocatedMethodDetailsObject.Elements` MUST be an object.                                                         |
+| `Object`           | JSONObject content of a column                      | Data type, format, constraints            | Column `BillingPeriodStart` MUST be of type `DateTime`                                                          |
```
shawnalpay, line 114: These look transposed.

```suggestion
| `Column`           | Named field across rows                             | Data type, format, constraints            | Column `BillingPeriodStart` MUST be of type `DateTime`                                                         |
| `Object`           | JSONObject content of a column                      | Data type, format, constraints            | Each item in `AllocatedMethodDetailsObject.Elements` MUST be an object.                                                          |
```

ruben567 (approved):

ruben567:
While there are no changes to InvoiceId rules, I discovered conflicting rules with regards to nullability:
- InvoiceId-C-004-C (MUST BE null) and InvoiceId-C-005-C (MUST NOT BE null) don't have any conditions defined, so InvoiceId rules will fail with or without a value.
- Both should have "Dynamic" as condition

mike-finopsorg:
> While there are no changes to InvoiceId rules, I discovered conflicting rules with regards to nullability:
> 
> * InvoiceId-C-004-C (MUST BE null) and InvoiceId-C-005-C (MUST NOT BE null) don't have any conditions defined, so InvoiceId rules will fail with or without a value.
> * Both should have "Dynamic" as condition

Thanks @ruben567 you are correct the condition of the normative text is not possible to model and the current requirements are actually conflicting. I will push an update to this one today.

guidelines/writing-requirements-model-guidelines.md
mike-finopsorg: Fixed in another PR.

mike-finopsorg:
> There are conflicts in the guidelines file between #1650 and #1668. Please either remove the guidelines file from #1668, or clean up #1650 (including addressing all 14 open threads over there) and merge into #1668. Then I can approve.

#1650 should be merged into this PR when reviewed.

specification/requirements_model/model_rules/datasets/cost_and_usage/objects/contractappliedobject.json
```diff
@@ -0,0 +1,2149 @@
...
+            "CheckFunction": "CheckModelRule",
+            "ModelRuleId": "CAU-ContractAppliedObject-O-039-C"
+          },
+          {
+            "CheckFunction": "CheckModelRule",
+            "ModelRuleId": "CAU-ContractAppliedObject-O-040-O"
+          }
+        ]
+      },
+      "Condition": {},
```
shawnalpay: MustSatisfy is:

> "When 'ContractCommitmentID' is not null, 'ContractCommitmentID' adheres to the following additional requirements:"

Therefore, do we want/need to check the presence of ContractCommitmentID, to avoid performing other child rules f ContractCommitmentID is null?

```suggestion
      "Condition": {
        "CheckFunction": "JSONCheckPathNotValue",
        "ColumnName": "ContractApplied",
        "Path": "$.Elements[*].ContractCommitmentID",
        "Value": null
      },
```

specification/requirements_model/check_functions.json
```diff
@@ -178,6 +177,139 @@
...
+      "Description": "Check distinct count of parent elements at JSON path equals expected count for a child path",
+      "Arguments": [
+        "ColumnName",
+        "ParentPath",
+        "ChildPath",
+        "ExpectedCount"
+      ],
+      "FormatAttributes": []
+    },
+    "FormatJSONFormat": {
```
shawnalpay: Want to confirm this is the expected function name, and not `JSONFormat`?  The other ones all start with `JSON`, and `FormatJSONFormat` reads a bit strangely.

specification/requirements_model/model_rules/datasets/cost_and_usage/columns/allocatedmethoddetails.json
```diff
@@ -0,0 +1,253 @@
...
+    "Order": 40,
+    "DatasetType": "CAU",
+    "DatasetId": "CostAndUsage",
+    "DatasetName": "Cost And Usage",
+    "ValidationCriteria": {
+      "MustSatisfy": "AllocatedMethodDetails MUST conform to JsonObjectFormat requirements.",
+      "Keyword": "MUST",
+      "Requirement": {
+        "CheckFunction": "FormatJSONFormat",
+        "ColumnName": "AllocatedMethodDetails"
```
shawnalpay: The FormatJSONFormat definition expects ColumnName and Path, so I think we have to specify Path here, yeah?

```suggestion
        "CheckFunction": "FormatJSONFormat",
        "ColumnName": "AllocatedMethodDetails",
        "Path": "$"
```

specification/requirements_model/model_rules/datasets/cost_and_usage/columns/contractapplied.json
```diff
@@ -0,0 +1,130 @@
...
+    "DatasetType": "CAU",
+    "DatasetId": "CostAndUsage",
+    "DatasetName": "Cost And Usage",
+    "Type": "Static",
+    "ValidationCriteria": {
+      "MustSatisfy": "ContractApplied MUST conform to JsonObjectFormat requirements.",
+      "Keyword": "MUST",
+      "Requirement": {
+        "CheckFunction": "FormatJSONFormat",
+        "ColumnName": "ContractApplied"
```
shawnalpay: Similar to the example in AllocatedMethodDetails, I think FormatJSONFormat is expecting a Path argument here, yes?

```suggestion
        "CheckFunction": "FormatJSONFormat",
        "ColumnName": "ContractApplied",
        "Path": "$"
```

specification/requirements_model/releases/1.3/model_rules/datasets/cost_and_usage/columns/tags.json
```diff
@@ -1,62 +1,74 @@
...
-        "Tags-C-002-O",
-        "Tags-C-003-C",
-        "Tags-C-011-M",
-        "Tags-C-014-M"
+        "CAU-CostAndUsage-D-064-C",
+        "CAU-Tags-C-001-M",
+        "CAU-Tags-C-002-O",
+        "CAU-Tags-C-003-C",
+        "CAU-Tags-C-011-M",
+        "CAU-Tags-C-014-M"
```
shawnalpay, line 54: Gemini notices that these two rules aren't nested under the null check.  That correctly reflects the normative requirements as stated in Tags.  Should we revise the Tags normative requirements upstream to be nested under the null check?

Here is how it is stated in the spec today:

> * When Tags is not null, Tags adheres to the following additional requirements:
>   * ...
> * Provider-defined tags adhere to the following additional requirements:
>   * ...
> * User-defined tags adhere to the following additional requirements:
>   * ...

specification/requirements_model/model_rules/datasets/cost_and_usage/columns/allocatedmethoddetails.json
```diff
@@ -0,0 +1,253 @@
...
+    "ModelVersionIntroduced": "1.3",
+    "Status": "Active",
+    "ApplicabilityCriteria": [
+      "DATA_GENERATOR_SPLIT_COST_ALLOCATION_SUPPORTED"
+    ],
+    "Type": "Static",
+    "Order": 0,
+    "DatasetType": "CAU",
+    "DatasetId": "CostAndUsage",
+    "DatasetName": "Cost And Usage",
```
shawnalpay: Official display name of this dataset is "Cost and Usage" (`And` not capitalized).  

The model is inconsistent on this point.  Some columns (e.g., AvailabilityZone) have it lower case.  Others (e.g., ContractAppliedObject) have it proper case.  A simple, global regex find/replace should solve it.

```suggestion
    "DatasetName": "Cost and Usage",
```

specification/requirements_model/model_rules/datasets/cost_and_usage/costandusage.json
```diff
@@ -2110,5 +2606,225 @@
...
+    "Notes": "",
+    "ModelVersionIntroduced": "1.3",
+    "Status": "Active",
+    "ApplicabilityCriteria": [
+      "CONTRACT_COMMITMENTS_SUPPORTED"
+    ],
+    "Order": 10,
+    "Type": "Static",
+    "ValidationCriteria": {
+      "MustSatisfy": "ContractApplied MUST be present in a Cost and Usage *FOCUS dataset* when the service provider supports contractcommitments.",
```
shawnalpay:
```suggestion
      "MustSatisfy": "ContractApplied MUST be present in a Cost and Usage *FOCUS dataset* when the service provider supports contract commitments.",
```

specification/requirements_model/tests/test_reference_prefix_for_columns.py
shawnalpay, line 1: This seems to have the exact same code as the newly-created `test_reference_prefix_for_columns_and_objects.py`.  I think we can delete this file?

specification/requirements_model/releases/1.3/model_rules/datasets/cost_and_usage/objects/contractappliedobject.json
```diff
@@ -0,0 +1,2149 @@
...
+    "Type": "Static",
+    "ValidationCriteria": {
+      "MustSatisfy": "\"ContractCommitmentAppliedCost\" MUST be a valid decimal value.",
+      "Keyword": "MUST",
+      "Requirement": {
+        "CheckFunction": "JSONCheckPathNotValue",
+        "ColumnName": "ContractApplied",
+        "Path": "$.Elements[*].ContractCommitmentAppliedCost",
+        "Value": null
+      },
```
shawnalpay, line 1575: Is this the right check function?  MustSatisfy is looking for a valid decimal value, but the checkfunction is checking for value presence.  Do we perhaps need a `JSONCheckPathDecimalValue`?

specification/requirements_model/releases/1.3/model_rules/datasets/cost_and_usage/objects/contractappliedobject.json
```diff
@@ -0,0 +1,2149 @@
...
+    "Type": "Static",
+    "ValidationCriteria": {
+      "MustSatisfy": "\"ContractCommitmentAppliedQuantity\" MUST be a valid decimal value.",
+      "Keyword": "MUST",
+      "Requirement": {
+        "CheckFunction": "JSONCheckPathNotValue",
+        "ColumnName": "ContractApplied",
+        "Path": "$.Elements[*].ContractCommitmentAppliedQuantity",
+        "Value": null
+      },
```
shawnalpay, line 1860: Same point as made [here](https://github.com/FinOps-Open-Cost-and-Usage-Spec/FOCUS_Spec/pull/1668/files#r2885429688).  Do we need a new/different function?

specification/requirements_model/model_rules/datasets/cost_and_usage/objects/contractappliedobject.json
```diff
@@ -0,0 +1,2149 @@
...
+    "DatasetName": "Cost And Usage",
+    "Type": "Static",
+    "ValidationCriteria": {
+      "MustSatisfy": "\"ContractCommitmentID\" MUST be equal to ResourceID when ChargeCategory is \"Purchase\".",
+      "Keyword": "MUST",
+      "Requirement": {
+        "CheckFunction": "JSONCheckPathSameValue",
+        "ColumnAName": "ContractApplied",
+        "PathA": "$.Elements[*].ContractCommitmentID",
+        "ColumnBName": "ResourceID",
```
shawnalpay:
```suggestion
        "ColumnBName": "ResourceId",
```

specification/requirements_model/releases/1.3/model_rules/datasets/cost_and_usage/objects/contractappliedobject.json
```diff
@@ -0,0 +1,2149 @@
...
+    "ModelVersionIntroduced": "1.3",
+    "Status": "Active",
+    "ApplicabilityCriteria": [],
+    "Order": 170,
+    "DatasetType": "CAU",
+    "DatasetId": "CostAndUsage",
+    "DatasetName": "Cost And Usage",
+    "Type": "Static",
+    "ValidationCriteria": {
+      "MustSatisfy": "Contract application property keys MUST begin with the string \"x_\" unless it is a FOCUS-defined allocation property.",
```
shawnalpay, line 589: This properly reflects the spec (note the errant use of the word `allocation`, which was a copy paste error)... 😬 Should we fix it here?  It won't be a problem going forward because we're moving this requirement out in 1.4 via PR #2007, so this rule won't even exist anymore...

specification/requirements_model/releases/1.3/model_rules/datasets/cost_and_usage/objects/contractappliedobject.json
```diff
@@ -0,0 +1,2149 @@
...
+    "ModelVersionIntroduced": "1.3",
+    "Status": "Active",
+    "ApplicabilityCriteria": [],
+    "Order": 190,
+    "DatasetType": "CAU",
+    "DatasetId": "CostAndUsage",
+    "DatasetName": "Cost And Usage",
+    "Type": "Static",
+    "ValidationCriteria": {
+      "MustSatisfy": "\"ContractId\" MUST be present in a Cost and Usage *FOCUS Dataset* when the service provider supports *contract commitments*.",
```
shawnalpay, line 676: This properly reflects the wording of the spec, but it feels wrong now.  Should we change it here to say "...MUST be present in the Elements objects within the ContractApplied column..."?  Like my previous comment, this will be eliminated via the work of PR #2007, but it'll live on in perpetuity for our support of 1.3 datasets...

If we make a change, we'd need to do it for all the top-level properties.

specification/requirements_model/releases/1.3/model_rules/datasets/cost_and_usage/objects/contractappliedobject.json
```diff
@@ -0,0 +1,2149 @@
...
+            "ColumnName": "ContractApplied",
+            "Path": "$.Elements[*].ContractCommitmentAppliedQuantity"
+          },
+          {
+            "CheckFunction": "JSONCheckPathKeyExists",
+            "ColumnName": "ContractApplied",
+            "Path": "$.Elements[*].ContractCommitmentAppliedUnit"
+          }
+        ]
+      },
```
shawnalpay, line 310: Sooo... this is wrong in the spec.  (This was my PR... I'm ashamed to find this.)  It should include ContractId, and it doesn't.  

Shall we leave the RM reflecting the incorrect spec declaration to retain parity, or shall we repair it here and accept the gap between RM and spec?

specification/requirements_model/model_rules/attributes/datageneratorcalculatedsplitcostallocationhandling.json
```diff
@@ -0,0 +1,302 @@
...
+  "DataGeneratorCalculatedSplitCostAllocationHandling-A-000-M": {
+    "Function": "Composite",
+    "Reference": "DataGeneratorCalculatedSplitCostAllocationHandling",
+    "EntityType": "Attribute",
+    "EntityName": "Data Generator Calculated Split Cost Allocation Handling",
+    "EntityId": "DataGeneratorCalculatedSplitCostAllocationHandling",
+    "Notes": "",
+    "ModelVersionIntroduced": "1.3",
+    "Status": "Active",
+    "ApplicabilityCriteria": [],
```
shawnalpay: Is this the right place for this?

```suggestion
    "ApplicabilityCriteria": [DATA_GENERATOR_SPLIT_COST_ALLOCATION_SUPPORTED],
```

specification/requirements_model/releases/1.3/model_rules/attributes/invoicehandling.json
```diff
@@ -0,0 +1,127 @@
...
+    "Order": 20,
+    "ValidationCriteria": {
+      "MustSatisfy": "If an invoice-level charge appears on a customer invoice but cannot be expressed using existing FOCUS columns, data generators MUST include provider-defined columns (e.g., x_ChargeSubType) to capture the non-FOCUS-defined details needed to support invoice charges reconciliation using the FOCUS dataset.",
+      "Keyword": "MUST",
+      "Requirement": {},
+      "Condition": {},
+      "Dependencies": []
+    }
+  },
+  "InvoiceHandling-A-003-C": {
```
shawnalpay: Given the MAY, should this be an optional rule?  `InvoiceHandling-A-003-O` instead of `InvoiceHandling-A-003-C`?

specification/requirements_model/releases/1.3/model_rules/attributes/datageneratorcalculatedsplitcostallocationhandling.json
```diff
@@ -0,0 +1,302 @@
...
+    "EntityName": "Data Generator Calculated Split Cost Allocation Handling",
+    "EntityId": "DataGeneratorCalculatedSplitCostAllocationHandling",
+    "Notes": "",
+    "ModelVersionIntroduced": "1.3",
+    "Status": "Active",
+    "ApplicabilityCriteria": [],
+    "Type": "Dynamic",
+    "Order": 10,
+    "ValidationCriteria": {
+      "MustSatisfy": "A FOCUS dataset MUST include the following columns when the data generator supports data generator-calculated split cost allocation: AllocatedMethodId, AllocatedResourceId, AllocatedResourceName, AllocatedResourceTags.",
```
shawnalpay, line 99: There is no such column AllocatedResourceTags... bug in the 1.3 spec.  😬   Should we change it here?  And should we have conditions here?

```suggestion
      "MustSatisfy": "A FOCUS dataset MUST include the following columns when the data generator supports data generator-calculated split cost allocation: AllocatedMethodId, AllocatedResourceId, AllocatedResourceName, AllocatedTags.",
```

specification/requirements_model/releases/1.3/model_rules/datasets/cost_and_usage/objects/contractappliedobject.json
mike-finopsorg: We have the same issue on Columns. The normative text has both

MUST be of type Decimal.
and
MUST be a valid decimal value.

Once you have implemented a type Decimal check then a decimal value check is duplicative. I can make a DecimalValue check but in practice it will be doing exactly the same thing as TypeDecimal check, one fails both fail. Using the NotValue null at least offers something over the top of TypeDecimal. 

However, I will add we also have lines like

MUST be a valid decimal value when not null.

Which is backs us into a Dynamic rule. 

I guess the last option is to give them their own DecimalValue checkFunction and we just implement it in the validator as a SkipCheck?

specification/requirements_model/releases/1.3/model_rules/datasets/cost_and_usage/objects/contractappliedobject.json
mike-finopsorg: Same reply as [here](https://github.com/FinOps-Open-Cost-and-Usage-Spec/FOCUS_Spec/pull/1668#discussion_r2885751186) :)

specification/requirements_model/model_rules/datasets/cost_and_usage/objects/contractappliedobject.json
mike-finopsorg: Ahh good catch, the normative text here is written as
```
"ContractCommitmentID" MUST be equal to ResourceID...
```

specification/requirements_model/releases/1.3/model_rules/datasets/cost_and_usage/columns/listunitprice.json
```diff
@@ -359,7 +436,118 @@
...
+    "ModelVersionIntroduced": "1.3",
+    "Status": "Active",
+    "ApplicabilityCriteria": [],
+    "Order": 130,
+    "DatasetType": "CAU",
+    "DatasetId": "CostAndUsage",
+    "DatasetName": "Cost and Usage",
+    "Type": "Static",
+    "ValidationCriteria": {
+      "MustSatisfy": "ListCost MUST equal the product of ListUnitPrice and PricingQuantity when ListUnitPrice is not null and PricingQuantity is not null.",
```
shawnalpay, line 525: This is exactly the same check as `CAU-ListCost-C-011-C`.  Could we either 1) strike one rule or the other, or 2) have the model understand that only one of these needs to be checked?  It seems like a relatively "expensive" check for a dataset at scale, and it would be nice to avoid performing it twice.

specification/requirements_model/check_functions.json
mike-finopsorg: So all the Format* functions start with the prefix "Format" so FormatDecimal, FormatString... 
I think in hindsight we should go with FormatJSON, thoughts?

specification/requirements_model/model_rules/datasets/cost_and_usage/columns/allocatedmethoddetails.json
mike-finopsorg: The way I implemented it was as follows:

Validation:
    - Checks that extracted values are valid JSON
    - Supports both array paths (with [*]) and single value paths
    - If Path is not provided, validates the entire column value as valid JSON
    - Only validates non-null JSON values
    - Validates that the extracted element is itself valid JSON
    
    I guess I could handle the '$' path the same as not provided.

specification/requirements_model/releases/1.3/model_rules/datasets/cost_and_usage/columns/tags.json
mike-finopsorg: I think it makes more sense to be nested (well done Gemini), it doesn't hurt the model today as the rules are dynamic. I think fix upstream and adjust in the next model release.

specification/requirements_model/model_rules/datasets/cost_and_usage/columns/allocatedmethoddetails.json
mike-finopsorg: *sigh* ill fix as a larger pass vs doing the fix here.

specification/requirements_model/model_rules/attributes/datageneratorcalculatedsplitcostallocationhandling.json
mike-finopsorg: 👍

specification/requirements_model/releases/1.3/model_rules/attributes/datageneratorcalculatedsplitcostallocationhandling.json
mike-finopsorg: Added the applicability criteria, we need to decide a plan on wording matching intent or matching the spec.

specification/requirements_model/releases/1.3/model_rules/datasets/cost_and_usage/columns/listunitprice.json
mike-finopsorg: Im in favour of striking one rule or the other. I believe this its an over constraint that adds little value.

specification/requirements_model/model_rules/datasets/cost_and_usage/columns/allocatedmethoddetails.json
shawnalpay: Would adding an `OptionalArguments` be overkill?  Or would we maybe add this nuance of "make sure to handle for optional arguments in your code, otherwise things will bomb" to a future RM dev documentation?

specification/requirements_model/check_functions.json
shawnalpay: `FormatJSON` sounds good to me!

specification/requirements_model/releases/1.3/model_rules/datasets/cost_and_usage/columns/listunitprice.json
shawnalpay: Okay.  But is that a 1.3 thing or a 1.4 thing?

specification/requirements_model/releases/1.3/model_rules/datasets/cost_and_usage/objects/contractappliedobject.json
shawnalpay: Are you willing to go so far as to argue we should strike all instances of the second such requirement from the spec?  Obviously the decision was made early on in FOCUS development that both had value.

specification/requirements_model/model_rules/datasets/cost_and_usage/columns/allocatedmethoddetails.json
shawnalpay: Agreed to go with inclusion of $ in Path.

specification/requirements_model/releases/1.3/model_rules/datasets/cost_and_usage/columns/listunitprice.json
shawnalpay: Make it a 1.4 thing.  Discuss in TF-RM.  Resolving.

specification/requirements_model/model_rules/datasets/cost_and_usage/columns/contractapplied.json
shawnalpay: Agreed to $ Path.

specification/requirements_model/releases/1.3/model_rules/datasets/cost_and_usage/columns/tags.json
shawnalpay: Agree to change in 1.4.  No action here.  Resolving.

specification/requirements_model/tests/test_reference_prefix_for_columns.py
shawnalpay: Agreed we will retain, but set back to original version to have it still work for 1.2.  Will need to make sure all tests can work against all versions unless otherwise specified.

specification/requirements_model/releases/1.3/model_rules/datasets/cost_and_usage/objects/contractappliedobject.json
shawnalpay: Will bring possibility of striking this line from the spec to the Maintainers next week.  No action.

specification/requirements_model/releases/1.3/model_rules/datasets/cost_and_usage/objects/contractappliedobject.json
shawnalpay: Will be resolved in 1.4.  Leave as-is.  Resolving.

specification/requirements_model/releases/1.3/model_rules/datasets/cost_and_usage/objects/contractappliedobject.json
shawnalpay: Will be resolved in 1.4. Leave as-is. Resolving.

specification/requirements_model/releases/1.3/model_rules/datasets/cost_and_usage/objects/contractappliedobject.json
shawnalpay: Unresolve.  Capture in 1.3 errata doc.

specification/requirements_model/releases/1.3/model_rules/datasets/cost_and_usage/objects/contractappliedobject.json
shawnalpay: Unresolve.  Capture in 1.3 errata doc.

specification/requirements_model/releases/1.3/model_rules/datasets/cost_and_usage/objects/contractappliedobject.json
shawnalpay: Capture in 1.3 errata doc.

specification/requirements_model/releases/1.3/model_rules/attributes/invoicehandling.json
shawnalpay: Mike will research whether this impacts other examples.

specification/requirements_model/releases/1.3/model_rules/attributes/datageneratorcalculatedsplitcostallocationhandling.json
shawnalpay: Address in 1.3 errata doc.

specification/requirements_model/releases/1.3/model_rules/datasets/cost_and_usage/columns/tags.json
ijurica: That wasn’t an oversight 🙂 This was intentional, but please let me know if you think the decision was wrong 😄 

I chose not to nest the second and third requirements under the first because we were trying to **keep the nesting level under control**. In an earlier version I proposed a deeper nesting structure and received feedback that it **started to resemble pseudocode**, so I tried to keep the hierarchy flatter.

I also considered adding “When Tags is not null” to the other two requirements, but that felt slightly awkward. The first composite requirement applies specifically to the Tags column, (Capitalized) while the other two refer to provider-defined and user-defined tags themselves (lowercase). Conceptually, those rules still apply regardless of whether the Tags column is null; in practice, they simply don’t need to be evaluated when those tags are not present. Since these are Dynamic requirements, they cannot be evaluated in that case anyway, so the current structure seemed acceptable to me.

ljadvey (approved):

specification/requirements_model/check_functions.json
mike-finopsorg: Updated

specification/requirements_model/model_rules/datasets/cost_and_usage/columns/allocatedmethoddetails.json
mike-finopsorg: Updated and opened https://github.com/finopsfoundation/focus_validator/issues/138 for the validator to follow

specification/requirements_model/model_rules/datasets/cost_and_usage/columns/contractapplied.json
mike-finopsorg: Updated and opened https://github.com/finopsfoundation/focus_validator/issues/138 for the validator to follow

specification/requirements_model/releases/1.3/model_rules/datasets/cost_and_usage/columns/tags.json
mike-finopsorg: Closing this out the conversation can continue in 1.4 dev

specification/requirements_model/tests/test_reference_prefix_for_columns.py
mike-finopsorg: Updated

specification/requirements_model/releases/1.3/model_rules/attributes/invoicehandling.json
mike-finopsorg: Updated rule, added skips in the tests, and fix a small number of other rules impacted by this same type of issue.

specification/requirements_model/releases/1.3/model_rules/datasets/cost_and_usage/objects/contractappliedobject.json
mike-finopsorg: @shawnalpay im sorry I didn't take a note what we decided to do here.

Options
a) Do nothing
b) Remove requirement and make Dynamic
c) Implement new checkFunction and make validator map that to a skip action

TechinMama (approved):

specification/requirements_model/releases/1.3/model_rules/datasets/cost_and_usage/objects/contractappliedobject.json
shawnalpay: No action.  As with the related thread I linked above, I will raise with the Maintainers the possibility of striking the "valid decimal value".  Resolving.

shawnalpay (approved):
Having worked closely with @mike-finopsorg to review this over the last couple of days, I feel confident that it this PR is in good shape.  That said, the review revealed a handful of instances where the Requirements Model is correctly mirrors an incorrect spec requirement.  We will need to address this via a separate process/PR.

Approved!  Thank you @mike-finopsorg!

beau-nelford-anglepoint (approved):

jpradocueva:
Approved by the group without objections or abstentions during the March 5th Members’ call.

PR #1834, opened 2026-01-14
AI #1833: Added folder for 1.2 release of Requirements Model

### LINKED ISSUE #1833: [AI]  Create 1.2 RM folder as per TF-RM discussion about tracking releases
### Detailed Description

The TF-RM has decided that the best way to track the files related to different RM releases is to maintain a history of files in versioned folders

### Definition of Done

Folder tracking the 1.2 files added to the `requirements_model` folder

mike-finopsorg:
Clean copy of all the requirements model files as released into the releases/1.2 folder to track historic releases.

TF-RM has decided that keeping historical releases in folders enables bug fixes without causing new FOCUS releases which are done on a fixed schedule.


ijurica (approved):

shawnalpay (approved):
@mike-finopsorg 
This simple duplication works for me.  BUT!  Do we have any interest in trying to retain the commit history of these copies, using a solution like [this](https://stackoverflow.com/questions/16937359/git-copy-file-preserving-history) (or similar)?  Might cause more problems than it alleviates, but I'd like to explicitly decide against such a path, and I don't know if this was discussed in TF-RM.

Matt-Cowsert (approved):
I've given this a look. No concerns with what's in the PR, though I do expect we'll want to document the rationale and workflow for how resolving a bug in a prior RM model would look like (either in this PR or another one).

gparker-at-sf (approved):
👍

ljadvey (approved):

jpradocueva (approved):
Good approach going forward.

jpradocueva:
Approved by the group without objections or abstentions during the January 22nd Members’ call.

Issue #1091, opened 2025-05-23
[FR] Add column selection as a non-functional requirement 


Matt-Cowsert:
### 🧠 Problem Statement

Currently, not all FOCUS providers give Practitioners the ability to select which columns are provided in the FOCUS dataset _prior_ to exporting the report. Without clear mandates, generators implement divergent approaches, forcing practitioners to perform bespoke data cleaning and slowing adoption.

---

### 🧪 Use Case

**As a FinOps Practitioner performing cost allocation**
I need to select only Tags, BilledCost, EffectiveCost, and account structure columns for my allocation workflow
So that I can reduce dataset size, improve query performance, and simplify ETL processes without irrelevant columns

**As a FinOps Practitioner with limited data engineering resources**
I need to export only FOCUS standard columns (excluding x_ custom columns)
So that I can adopt FOCUS quickly without needing to accommodate provider-specific custom columns in my data pipelines

**As a FinOps Practitioner analyzing commitment optimization**
I need to select only commitment-related columns plus core cost/usage metrics
So that I can focus on commitment analysis without processing dozens of unused columns

**As a Data Engineer building FOCUS data pipelines**
I need to select different column subsets for different use cases (allocation vs. optimization vs. governance)
So that I can optimize storage costs, query performance, and data transfer volumes for each workflow

**As a FinOps Practitioner with bandwidth/cost constraints**
I need to minimize FOCUS dataset size by excluding unused columns
So that I can reduce egress costs, data transfer times, and storage requirements for large billing datasets

**As a FinOps Tooling Vendor ingesting FOCUS data**
I need to request only columns relevant to my product's functionality
So that I can optimize ingestion pipelines and reduce infrastructure costs for customers

**As a FOCUS Data Generator**
I need clear requirements for column selection capabilities
So that I can build consistent export experiences without guessing at practitioner needs or implementing bespoke solutions

#### Acceptance Criteria
**Column Selection Capabilities:**
- [ ] FOCUS datasets provide column selection mechanism before/during export
- [ ] Practitioners can select specific columns to include in exported datasets
- [ ] Practitioners can select predefined column groups (e.g., "core cost metrics," "commitment columns," "account structure")
- [ ] Column selection supports both FOCUS standard columns and custom (x_) columns
- [ ] Selection mechanism distinguishes between mandatory columns (always included) and optional columns (practitioner choice)

**Selection Granularity:**
[...]

PR #1904, opened 2026-01-28
AI #1838: Add TagDetails column to CostAndUsage

### LINKED ISSUE #1124: [FR] Add tag details for source attribution and eligibility
### Problem Statement

Practitioners face two interrelated challenges when working with tags in FOCUS datasets:

**Challenge 1: Tagging Eligibility is Opaque**
It's not obvious which charges are eligible to be tagged. As practitioners develop tag coverage targets and governance programs, they need a reference to understand what can be tagged for cost allocation purposes. This includes resources currently deployed and future offerings. Without this information, tag coverage reports penalize teams for resources that cannot be tagged, governance targets are unrealistic, and cost allocation efforts are undermined.

**Challenge 2: Tag Attribution is Missing**
When a tag value appears in the Tags column, practitioners cannot determine where that tag originated. Did it come from the resource directly? Was it inherited from a sub account, billing account, or other container? Which tag schema was used (user-defined vs. provider-defined, tags vs. labels)? When multiple levels have the same tag key with different values, which one "won" during finalization? Without this attribution, practitioners cannot:
- Troubleshoot unexpected tag values
- Understand inheritance behavior across providers
- Correlate FOCUS tag data with proprietary provider datasets
- Distinguish between directly-applied and inherited tags for governance purposes

**Current Gaps in FOCUS**
The existing Tags column provides finalized tag key-value pairs but lacks:
- Information about where each tag value originated (tag source)
- Identification of which tag schema produced the value
- Visibility into inheritance priority when multiple sources contribute tags
- Indication of whether a charge is eligible for tagging (and if so, at which levels)
- Distinction between "not tagged" (actionable) and "not taggable" (not actionable)

For observability use cases, labels function as tags and are used to segment bills by team. Understanding what's eligible for metadata-based allocation is critical.

---

#### Persona
**Primary:** FinOps Practitioner (cost allocation, chargeback, tag governance)
**Secondary:** Data Engineer (FOCUS ingestion, correlation with provider datasets)

#### User Story
**As a** FinOps practitioner,
**I need to** understand (1) which charges are eligible for tagging at what levels, and (2) where each finalized tag value originated,
**So that** I can build accurate tag coverage reports, establish realistic governance targets, perform cost allocation without penalizing teams for untaggable resources, and troubleshoot tag inheritance behavior.

#### Acceptance Criteria
- [ ] FOCUS provides a standardized mechanism for practitioners to determine which resources, services, or SKUs are eligible for tagging/labeling that will flow to cost and usage data
- [ ] The mechanism uses consistent terminology across providers regardless of whether they call them "tags," "labels," "metadata," or other terms
- [ ] The mechanism clearly distinguishes between:
  - Resources that are directly taggable with tags propagating to billing
  - Resources that can only inherit tags from parent resources or containers
  - Resources that are taggable but tags don't propagate to billing data
  - Resources that are not taggable at any level
- [ ] For each finalized tag value in a charge, practitioners can determine:
  - The source from which the tag originated (e.g., resource, sub account, billing account)
  - The tag schema that produced the value (e.g., user-defined tags, provider-defined labels)
- [ ] The mechanism defines or references inheritance priority so practitioners understand which source "wins" when multiple levels have the same tag key
- [ ] Practitioners can use the information to:
  - Generate tag coverage reports that exclude ineligible resources from coverage calculations
  - Identify gaps in tagging strategy (eligible but not tagged)
  - Correlate FOCUS tag data with proprietary provider datasets
  - Perform cost allocation without penalizing teams for untaggable charges
- [ ] The specification includes clear guidance for data generators on representing tag attribution and eligibility
- [ ] Temporal information indicates when a resource type became eligible (or ineligible) for tagging to support historical analysis

---

### Success Criteria

- [ ] Practitioners can distinguish between charges that are not tagged (actionable, eligible for tagging) and charges that are not taggable (inherent limitation of the resource or service), enabling tag coverage reports that exclude ineligible resources from coverage calculations
- [ ] For each finalized tag value in a charge, practitioners can determine the source from which the tag originated (e.g., resource, sub account, billing account) and the schema that produced it (e.g., user-defined vs. provider-defined), supporting troubleshooting and governance
- [ ] Tag eligibility and attribution information is consistently represented across data generators, so practitioners can apply uniform tag governance policies and coverage targets across providers
- [ ] Data generators have clear guidance on how to represent tag source attribution and tagging eligibility, including how to handle provider-specific inheritance hierarchies and tag schema differences

---

### Type of Request

- [ ] **Enhancement** – Refinement or extension of an existing FOCUS column, metric, or capability
- [x] **Net New** – A concept not yet represented in FOCUS or widely available from data generators
- [ ] **Enablement/Supporting Content** – Non-normative items like appendix explanations, examples, or supporting documentation

---

- ExxonMobil
- Domo
- Microsoft
- Walmart
- Twilio
- Northwestern Mutual

---

_No response_

---

https://learn.microsoft.com/en-us/azure/azure-resource-manager/management/tag-support

https://docs.aws.amazon.com/whitepapers/latest/tagging-best-practices/measuring-tagging-effectiveness-and-driving-improvements.html

https://cloud.google.com/resource-manager/docs/tags/tags-overview

https://focus.finops.org/focus-specification/v1-2/#designprinciples

Gated to FOCUS Working Group
- [1124 Provider Tagging Eligibility Ideation Doc](https://docs.google.com/document/d/1fNaS7p3GmeJvsPOP69sDlKiB09bnb1abdMFHufctZpc/edit?usp=drive_link)
- [1124 Provider Resource Tagging Eligibility](https://drive.google.com/drive/folders/1s5XgG72waBMvS7jwfMFdjbV_ApxvbJp5?usp=drive_link)
- [TaggingEligibility-SampleData](https://docs.google.com/spreadsheets/d/1oiBivrRPseiaVoMpXgG0rFwSA79zwm1UI9CyAOievSc/edit?gid=0#gid=0)

---

This could be achieved through a resource registry of sorts that captures resources across providers today. The expectation is that we can leverage existing content from providers (most have the largest percentage of spend resources covered) as well as inputs from practitioners and FinOps Vendors. 

Could also be a column, either in the Cost and Usage dataset or in a Resource dataset, or provided via metadata (wasn't taggable yesterday, is taggable today).

---

**Migrated from "Supported Feature Alignment":**

#### Current State Assessment

**Enhances existing feature:** Cost and Usage Attribution
**Current Gap:** The feature supports cost allocation using tags but doesn't:
- Identify which resources are taggable (leading to inaccurate coverage reports)
- Indicate where finalized tag values originated (preventing correlation with provider data)
- Define inheritance priority (leaving finalization behavior ambiguous)
- Distinguish between "not tagged" and "not taggable" (penalizing teams unfairly)

**Needed Capability:** Practitioners need tag attribution and eligibility information to calculate accurate coverage metrics, build realistic governance targets, and perform cost allocation without penalizing teams for untaggable resources.

#### Approach A: Enhance Existing "Cost and Usage Attribution" Feature

**Enhancement Description:**
Extend the Cost and Usage Attribution supported feature by adding one or more columns that provide tag attribution and eligibility details while preserving the existing Tags column behavior.

**Option A1: Single TagDetails Column (JSON)**
A new TagDetails column containing:
- For each tag: source (where it originated), schema (user-defined vs. provider-defined)
- Eligibility information for the charge
- Inheritance priority context

**Option A2: Separate Columns**
- TagDetails column for attribution (source and schema per tag)
- TaggingEligibility column for eligibility (directly taggable, inheritable only, not taggable)

**Pros:**
- Preserves existing Tags column behavior (no breaking changes)
- Allows practitioners to ignore details if they only need simple tag values
- Column selection can exclude details columns for those who don't need them

**Cons:**
- Significant data duplication across rows
- Increased dataset size, especially for providers with complex tag hierarchies

**Draft Enhancement - Directly Dependent Columns:**
- Tags (existing)
- TagDetails (proposed - attribution and/or eligibility)
- TaggingEligibility (proposed - if separated from TagDetails)

**Draft Enhancement - Supporting Columns:**
- SkuId
- ResourceType
- ServiceName
- BilledCost
- EffectiveCost

#### Approach B: Metadata-Based Eligibility with Column-Based Attribution

**Enhancement Description:**
Split the solution: eligibility goes into schema metadata (since it's consistent per resource type/SKU), while tag attribution remains in the dataset (since it varies per row).

**Eligibility in Metadata:**
- Define which tag sources are available (resource, sub account, billing account, etc.)
- Specify inheritance priority order
- Indicate which tag schemas are supported
- Note temporal changes in eligibility

**Attribution in Dataset:**
- TagDetails column showing source and schema for each finalized tag value

**Pros:**
- Avoids massive data duplication for eligibility (same eligibility repeated on every row)
- Eligibility information defined once, referenced many times
- Attribution remains row-specific where it belongs

**Cons:**
- Requires practitioners to join metadata with cost data for eligibility analysis
- More complex implementation for data generators
- Metadata access patterns vary across platforms

#### Approach C: New Standalone Supported Feature

**Feature Name:** Tag Coverage and Governance

**Description:**
A new supported feature enabling practitioners to understand tag attribution and eligibility across all providers. Supports accurate tag coverage reporting, realistic governance programs, and fair cost allocation.

**Directly Dependent Columns:**
- TagDetails (proposed - attribution information)
- TaggingEligibility (proposed - eligibility information)
- Tags (existing - for correlation)

**Supporting Columns:**
- SkuId
- ResourceType
- ServiceName
- ResourceId
- BilledCost
- EffectiveCost

**Enabled Capabilities:**
- **Tag Coverage Analysis:** Accurate coverage percentages excluding ineligible resources
- **Tag Governance:** Realistic targets based on what's actually taggable
- **Cost Allocation:** Allocation without penalizing teams for untaggable resources
- **Attribution Troubleshooting:** Identify where unexpected tag values originated
- **Provider Correlation:** Map FOCUS tags to provider-native constructs

**Pros:**
- Clear visibility for a major capability
- Dedicated feature documentation
- Explicit recognition of tag governance as a FinOps capability

**Cons:**
- Increases feature count
- May fragment related functionality (attribution, eligibility, tags)

---

**Migrated from "Introduced (Version)":**

1.4 (proposed)

#### Comparison of Approaches

| Aspect | Approach A (Column) | Approach B (Metadata + Column) | Approach C (New Feature) |
|--------|---------------------|-------------------------------|-------------------------|
| Data duplication | High (eligibility repeated) | Low (eligibility in metadata) | High (eligibility repeated) |
| Implementation complexity | Medium | Higher | Medium |
| Practitioner discoverability | Within existing feature | Split across metadata and data | Dedicated feature visibility |
| Flexibility for simple use cases | Can ignore columns | Must join metadata | Can ignore columns |
| Alignment with existing patterns | Follows column-based approach | Mixes patterns | Follows column-based approach |

#### Working Group Decision Required
The TF-2 discussion did not reach consensus on which approach is preferred. The working group should evaluate:
1. Data volume impact (especially for providers where most resources are taggable)
2. Practitioner workflow patterns (inline data vs. metadata lookups)
3. Alignment with existing FOCUS patterns
4. Phased delivery potential (eligibility first, then attribution, or vice versa)

---

---

## Maintainer Assessment

### Adoption Impact

- [ ] **Adoption Accelerator** – This feature would help organizations start using FOCUS
- [x] **Practitioner Enhancement** – Improves experience for existing FOCUS users
- [ ] **Technical Improvement** – Internal specification refinement

**Justification:**

Six organizations requested this, including enterprise practitioners (ExxonMobil, Walmart, Northwestern Mutual) and providers (Microsoft, Domo, Twilio). The problem is significant: tag governance programs penalize teams for untaggable resources and cannot troubleshoot inheritance behavior. However, organizations can adopt FOCUS without this capability. Tag coverage and governance are mature-usage workflows, not initial adoption gates.

- **Gap:** Practitioners cannot distinguish "not tagged" (actionable) from "not taggable" (inherent limitation). Tag coverage reports produce inflated uncovered percentages.
- **Org signals:** 6 orgs, including 3 Fortune 500 enterprises. Strong demand signal but consistently framed as a governance enhancement, not an adoption prerequisite.
- **Adoption position:** Enhances mature usage. Organizations adopt FOCUS for cost normalization first; tag governance maturity follows.

---

**Feature Description:**

This FR enhances the existing "Cost and Usage Attribution" supported feature by introducing tag eligibility and source attribution. The Tags column currently provides finalized key-value pairs but hides how values were determined. Practitioners building tag governance programs need two capabilities the feature cannot deliver today: whether a charge is eligible for tagging, and where each tag value originated (resource, sub account, billing account, inherited vs. direct).

- **Current limitation:** Cost and Usage Attribution supports tag-based allocation through the Tags column but treats all untagged charges identically. No mechanism distinguishes directly taggable resources from inherently untaggable ones.
- **Use case impact:** Tag coverage reporting, tag governance target-setting, cost allocation fairness, and tag inheritance troubleshooting are all impaired.
- **Structural gap:** No column or metadata captures tagging eligibility per charge or tag source attribution per tag value. The spec lacks the concept of "tag provenance."
- **Design note:** The FR proposes three implementation approaches (column-based, metadata-based hybrid, or new standalone feature). The working group did not reach consensus. Data volume impact is a key tradeoff: eligibility is consistent per resource type (favoring metadata), while attribution varies per row (favoring columns).

---

### Implementation Scope

**MVP Definition:**

Add a TaggingEligibility column (String, Conditional) to the Cost and Usage dataset indicating whether a charge is eligible for direct tagging, inheritable tagging only, or not taggable. This enables practitioners to exclude non-taggable charges from coverage calculations and set realistic governance targets. Defer tag source attribution (which source "won" during finalization) to a subsequent phase.

Satisfies Success Criteria 1 (distinguish not-tagged from not-taggable) and 3 (consistent representation across generators). Partially satisfies 4 (generators have guidance for eligibility; attribution guidance deferred).

**North Star Vision:**

Practitioners have full visibility into both tag eligibility and tag provenance. A TaggingEligibility column identifies which charges can be tagged and at what levels. A TagDetails column (or metadata reference) provides per-tag source attribution, schema identification, and inheritance priority. Together, these enable precise tag coverage metrics, inheritance troubleshooting, and fair cost allocation across all providers.

**Phasing Strategy:**

Two phases are warranted. Phase 1 (1.4): TaggingEligibility column establishing the not-tagged vs. not-taggable distinction. This is the highest-value, lowest-ambiguity component. Phase 2 (1.5): TagDetails for source attribution and schema identification. The working group should resolve the column vs. metadata design question during Phase 1 work, informed by data volume analysis from generator implementations.

---

### Specification Impact

**Impacted Parties:**
- [ ] Other: [specify]

Practitioners gain governance capabilities. Generators must expose tag eligibility and source attribution data they currently keep internal. Vendors must support new columns or metadata in their analytics platforms.

**Level of Ambiguity:**

4

The use cases are well-articulated and supported by detailed analysis, including sample data and three proposed approaches. However, fundamental design questions remain unresolved:

- Column vs. metadata vs. hybrid approach: the working group did not reach consensus
- Level of attribution detail: boolean eligibility vs. rich source hierarchy
- Data volume impact: eligibility repeated per row vs. per resource type in metadata
- Provider-specific tag hierarchies: each provider's inheritance model differs significantly
- How to handle temporal eligibility changes (resource type became taggable after a certain date)

Requester rated ambiguity at 3. This assessment raises to 4 because the lack of working group consensus on the implementation approach and the cross-provider variability in tag inheritance models represent significant unresolved design space.

---

### FinOps Scope Alignment

Requester checked Public Cloud, SaaS, and Data Center. This assessment agrees. Tag governance applies across any environment where metadata-based cost allocation is practiced.

## Community Support

  - Your **organization**

### LINKED ISSUE #1838: [AI] Draft skeleton PR for tag details (Column J key-driven approach)
### Detailed Description

Karl accepted ownership of drafting an initial PR for the tag details specification using the key-driven approach (Column J). This follows consensus that Column J is more efficient for frequent tag value extraction at scale, based on Karl's practical BigQuery testing. The skeleton should include structure for normative requirements with to-dos that can be assigned to other contributors (Michael, Irena, Matt) in the next meeting.

### Definition of Done

- [ ] Skeleton PR created with structural outline
- [ ] To-do placeholders included for normative requirements sections
- [ ] References Scenario 8 use cases for null handling discussion

kk09v:
PR for #1124 introducing the TagDetails column

To Do:
- Update /specification/requirements_model/model_rules/datasets/costandusage
- Add Use Case library examples
- Review PR #2078 restructuring this to align to approach in #2007 

BOLOs when reviewing (may no longer be an issue)
- Use of Scheme vs Schema
- Incorrectly using Tag Source vs Tag Scheme
- Tags in place of Tag
  - Column was originally TagsDetails
- References to Tags column should explicitly say Tags column to differentiate from Tags which is the name of a defined object


ijurica:
As this PR introduces a JSON format column, I’d like to mention the PR #1986

shawnalpay:
> As this PR introduces a JSON format column, I’d like to mention the PR #1986

It's a fair callout, @ijurica, but @mike-finopsorg and I are just now hammering out the details there.  Until we have alignment there, it should be business as usual with JSON columns, following the conventions we've previously laid out.

specification/datasets/cost_and_usage/columns/tagdetails.md
```diff
@@ -0,0 +1,227 @@
...
+* TagDetails root object MUST contain an object for each tag scheme present in Tags.
+  * The key for each tag scheme object MUST align to the tag prefix (up to but not including the forward slash) used to define the tag scheme in Tags.
+  * The key for the unprefixed user-defined tag scheme in Tags MUST be "Default".
+* Each tag scheme object MUST contain an object for each tag key used in that tag scheme.
+  * Each tag key object MUST contain FOCUS-defined tag properties.
+  * FOCUS-defined tag properties are subject to the additional requirements:
+    * Tag property key MUST match the spelling and casing specified for the FOCUS-defined property.
+    * Tag property value MUST be of the type specified for that property.
+    * Tag properties MUST adhere to additional normative requirements specific to that property.
+  * Each tag key object MUST contain an object with they key "AncestorTaggedSources".
```
aws-jrmarks:
```suggestion
  * Each tag key object MUST contain an object with the key "AncestorTaggedSources".
```

specification/datasets/cost_and_usage/columns/tagdetails.md
```diff
@@ -0,0 +1,227 @@
...
+
+The Tag Details column is a superset of [Tags](#tags) which includes additional properties describing tag eligibility and tag provenance from all [*tag sources*](#glossary:tag-source). Tag Details can be used to determine whether a [*charge*](#glossary:charge) was eligible to be tagged, improving the accuracy of tag coverage calculations. Tag Details can also be used to determine the origin of the [*finalized tag*](#glossary:finalized-tag) as well as find tag values present on ancestor sources.
+
+## Requirements
+
+### Column Requirements
+
+The TagDetails column adheres to the following requirements:
+
+* TagDetails SHOULD be present in a Cost and Usage [*FOCUS dataset*](#glossary:FOCUS-dataset) when the data generator supports setting user or provider-defined tags.
```
aws-jrmarks: I agree with SHOULD here, we've heard lots of feedback from customers that they'd like to be in control of when columns are populated that would drastically increase the size of the dataset (see AWS Split Cost Allocation in Legacy CUR as an example). If this was a MUST, then practitioners would be getting 3-7x larger reports by default, which seems unnecessary to force.

aws-jrmarks (commented):
Agree with SHOULD here, see comment in-line.

specification/datasets/cost_and_usage/columns/tagdetails.md
```diff
@@ -0,0 +1,227 @@
...
+
+The "TagSourceId" property adheres to the following requirements:
+
+* TagSourceId MUST be present in each tag key object in the Tags object.
+* TagSourceId MUST be present in each tag source object in the AncestorTaggedSources object.
+* TagSourceId MUST contain the identifier of the TagSource.
+
+## Overview
+
+### Object Scheme
```
qquigley:
```suggestion
### Object Schema
```
Checking the use of schema vs scheme. Ditto for line 84.

specification/datasets/cost_and_usage/columns/tagdetails.md
```diff
@@ -0,0 +1,227 @@
...
+### Column Requirements
+
+The TagDetails column adheres to the following requirements:
+
+* TagDetails SHOULD be present in a Cost and Usage [*FOCUS dataset*](#glossary:FOCUS-dataset) when the data generator supports setting user or provider-defined tags.
+* TagDetails MUST be of type String.
+* TagDetails MUST conform to [StringHandling](#attributes.stringhandling) requirements.
+* TagDetails MUST conform to [JsonObjectFormat](#attributes.jsonobjectformat) requirements.
+* TagDetails nullability is defined as follows:
+  * TagDetails MUST be null when Tags is null.
```
qquigley: Is this correct? If Tags is null but the resource was eligible for tagging, then we would want to know how with `UntaggedSources` in `TagDetails`, no?

specification/datasets/cost_and_usage/columns/tagdetails.md
mike-finopsorg: `TagDetails SHOULD be present in a Cost and Usage`

should be at the dataset Level

kk09v: Yeah, I would not have supported this column if #1091 wasn't included in this release. I personally view this column as very useful in a limited context, but certainly not necessary as part your daily ingestion jobs.

kk09v: Moved requirement to dataset level

specification/datasets/cost_and_usage/columns/tagdetails.md
```diff
@@ -0,0 +1,227 @@
...
+| TagSourceId | String | TRUE | The identifier of the *tag source* where the tag key is present. |
+
+### Example
+
+```json
+{
+  "Default": {
+    "Tags": {
+      "foo": {
+        "FocusColumn": "ResourceId",
```
shawnalpay: What is `FocusColumn`?  It's here several times in the example, but it's not anywhere in the requirements.

specification/datasets/cost_and_usage/columns/tagdetails.md
```diff
@@ -0,0 +1,227 @@
...
+  * The key for the unprefixed user-defined tag scheme in Tags MUST be "Default".
+* Each tag scheme object MUST contain an object for each tag key used in that tag scheme.
+  * Each tag key object MUST contain FOCUS-defined tag properties.
+  * FOCUS-defined tag properties are subject to the additional requirements:
+    * Tag property key MUST match the spelling and casing specified for the FOCUS-defined property.
+    * Tag property value MUST be of the type specified for that property.
+    * Tag properties MUST adhere to additional normative requirements specific to that property.
+  * Each tag key object MUST contain an object with the key "AncestorTaggedSources".
+    * The value of AncestorTaggedSources MUST be null when the tag key from the corresponding tag key object within the tag scheme is not present in any *tag source* other than the *tag source* which results in the *finalized tag*.
+    * When the value of AncestorTaggedSources is not null, each object in AncestorTaggedSources MUST have a key denoting the *tag source*.
```
shawnalpay: This requirement implies an Array structure instead of the Map structure shown in your example.

Your current example farther down follows the Map pattern:

```json
"AncestorTaggedSources": {
  "Subscription": {              // The label IS the Source Type
    "TagValue": "foo"
  },
  "Resource Group": {
    "TagValue": "bar"
  }
}
```

Whereas this normative requirement implies an Array structure like:

```json
"AncestorTaggedSources": [
  {
    "SourceType": "Subscription",   // You have to open the object to find out what it is
    "TagValue": "foo"
  },
  {
    "SourceType": "Resource Group",
    "TagValue": "bar"
  }
]
```

To reconcile this, I think we change the normative requirement to match your example.  Something like:

```suggestion
    * AncestorTaggedSources MUST be a JSON object where each key represents the tag source type (e.g., 'Subscription') and the value is an object containing the tag properties.
```

specification/datasets/cost_and_usage/columns/tagdetails.md
```diff
@@ -0,0 +1,227 @@
...
+| AncestorTaggedSources | _TagKey_ | Object | TRUE | An object containing all *tag sources* where the corresponding tag key was present which did not result in the *finalized tag*. |
+| _AncestorTaggedSource_ | AncestorTaggedSources | Object | Conditional | An object containing the the properties of the tag present in the *tag source*. |
+
+### Object Entries
+
+The tag key object and tag source objects contain the following properties:
+
+| Key | ValueType | Required | Description |
+| ----- | ---- | ---------- | ----------- |
+| TagValue | String | TRUE |  The value of the tag key. |
```
shawnalpay: You've got this listed here as a String -- but line 56 is currently:

> * TagValue MUST have the value of true (boolean) when the TagScheme does not support values.

Because Tags also allows String and Boolean, you likely need to allow similar here.

specification/datasets/cost_and_usage/columns/tagdetails.md
kk09v: I agree. Irena and I spoke about this during last week's meeting and I convinced her of the need to change some prior guidance/decision. Will be correcting today.

specification/datasets/cost_and_usage/columns/tagdetails.md
kk09v: Thanks, I called this out as one of the things to watch for as I was writing (and re-re-writing) it many times.

This is schema (the shape of the data). 
Line 84 should be tag scheme (the provider defined method of tagging)

specification/datasets/cost_and_usage/columns/tagdetails.md
kk09v: I see the confusion. You are correct that these should be key-value pairs with the key being the ancestor tag source and the value being the object. The object will contain at least the property (key-value pair) TagValue.

I think this will be much easier to write clearly and concisely after Mike's revision to how we're structuring these. Like I said in one of the meetings, I was having difficulties finding the right words to clearly describe (and differentiate) the situations, including {tags (the concept) and "Tags": the object} and {key of the tag (concept) and key inside of the json object}.

If you look at how I've written the rest of this, the way I've been referring to these is 
"ObjectKey":{
  "PropertyKey":"PropertyValue"
  } 
with the entire thing referred to as the _ObjectKey_ Object

So I agree that a revision is necessary, but right now the way this is written appears to be consistent with the way the rest of the PR is written.

specification/datasets/cost_and_usage/dataset.md
```diff
@@ -72,6 +72,7 @@ The specification for the Cost and Usage dataset defines a group of columns that
 | [Sub Account ID](#datasets.costandusage.subaccountid)                                               | Dimension          | Conditional   | True         | String    |
 | [Sub Account Name](#datasets.costandusage.subaccountname)                                           | Dimension          | Conditional   | True         | String    |
 | [Sub Account Type](#datasets.costandusage.subaccounttype)                                           | Dimension          | Conditional   | True         | String    |
+| [TagDetails](#datasets.costandusage.tagdetails)                                                     | Dimension          | Recommended   | True         | JSON      |
```
shawnalpay, line 74: @kk09v 
The reason the column presence requirements aren't showing up in this markdown, I believe, is that you need to merge in `working_draft`.  I'm going to leave that to you, though, given that it may introduce a pile of merge conflicts... 😅

kk09v: Thanks for the guidance. I think I'm half-way in-between two different directions on this.

specification/datasets/cost_and_usage/columns/tagdetails.md
kk09v: Yeah, this is an open item for discussion this week. It was included in the original proposal but as I was writing all of this out, I was conflicted.

specification/datasets/cost_and_usage/columns/tagdetails.md
```diff
@@ -0,0 +1,267 @@
...
+The TagDetails column adheres to the following requirements:
+
+* TagDetails MUST be of type String.
+* TagDetails MUST conform to [StringHandling](#attributes.stringhandling) requirements.
+* TagDetails MUST conform to [JsonObjectFormat](#attributes.jsonobjectformat) requirements.
+* TagDetails nullability is defined as follows:
+  * TagDetails MUST NOT be null unless all of the following are true:
+    * Tags column is null.
+    * Tags are not present in any other *tag sources*.
+    * No *tag sources* are supported for any user-defined or provider-defined tag scheme.
```
kk09v:
```suggestion
    * No *tag sources* are supported for any user-defined tag scheme.
```

Matt-Cowsert:
## TF2 Action Item – 2026-03-04

Karl walked through recent changes to PR #1904 including adjusted nullability requirements, tag source conditionality, and a new example scenario. The group identified several remaining changes needed before the PR can be pulled out of draft. Key items include tying the column to the column selection NFR, adding a glossary entry for "tag scheme," renaming properties for requirements model clarity, and researching null vs empty array handling.

**Scope:**
- [ ] Add conditionality on the column selection non-functional requirement (MUST if supported, SHOULD NOT if not)
- [ ] Add glossary entry for "tag scheme"
- [ ] Rename top-level tag properties to "finalized" variants for requirements model clarity
- [ ] Review examples for FOCUS column usage (incorporate Irena's findings)
- [ ] Research prior art on empty array vs null handling for JSON properties
- [ ] Remove provider-defined tag scheme from the nullability requirement
- [ ] Add additional example scenarios beyond the current GCP/untagged sources example

**Assignee:** @kk09v

specification/datasets/cost_and_usage/columns/tagdetails.md
```diff
@@ -0,0 +1,476 @@
...
+| AncestorTaggedSources | _TagKey_ | Object | TRUE | An object containing all *tag sources* where the corresponding tag key was present which did not result in the *finalized tag*. |
+| _AncestorTaggedSource_ | AncestorTaggedSources | Object | Conditional | An object containing the the properties of the tag present in the *tag source*. |
+
+### Object Entries
+
+The tag key object and tag source objects contain the following properties:
+
+| Key | ValueType | Required | Description |
+| ----- | ---- | ---------- | ----------- |
+| TagSource | String | Conditional | The type of *tag source* where the tag key is present. |
```
kk09v: This should not be conditional. It MUST be present in tag key object. 
However, it acts as the ObjectKey for the AncestorTaggedSource object so it does not need to be included in the properties of that object.

This will be easier to rewrite once we change the structure

specification/datasets/cost_and_usage/columns/tagdetails.md
```diff
@@ -0,0 +1,476 @@
...
+"TagValue" represents the tag value associated with the tag key for the corresponding tag scheme and *tag source*.
+
+The "TagValue" property adheres to the following requirements:
+
+* TagValue MUST be present in each tag key object in the Tags object.
+* When TagValue is directly contained in the tag key object, TagValue MUST represent the *finalized tag*.
+* TagValue MUST be present in each tag source object in the AncestorTaggedSources object.
+* TagValue MUST have the value of true (boolean) when the TagScheme does not support values.
+* Data generator MUST NOT alter tag values unless applying true (boolean) to valueless tags.
+* TagValue MAY be null when the the data generator supports setting a null value for a key-value pair type tag.
```
Matt-Cowsert:
```suggestion
* TagValue MAY be null when the data generator supports setting a null value for a key-value pair type tag.
```

Typo: "the the data generator."

specification/datasets/cost_and_usage/columns/tagdetails.md
```diff
@@ -0,0 +1,476 @@
...
+
+| Key | Parent | ValueType | Required | Description |
+| ----- | ---- | ---- | ---------- | ----------- |
+| Default | _root object_ | Object | TRUE | The object containing information about the unprefixed tag scheme in the Tags column. |
+| _TagScheme_ | _root object_ | Object | TRUE | One or more objects containing information about a prefixed tag scheme in the Tags column. |
+| Tags | _tag scheme object_ | Object | TRUE | An object containing all tag keys present in the Tags column. |
+| UntaggedSources | _tag scheme object_ | Array | TRUE | A list of sources which support this tag scheme (i.e. are taggable) that had no tags applied. |
+| _TagKey_ | Tags | Object | TRUE | An object containing the properties of the *finalized tag* as well as other *tag sources* where this key was present. |
+| AncestorTaggedSources | _TagKey_ | Object | TRUE | An object containing all *tag sources* where the corresponding tag key was present which did not result in the *finalized tag*. |
+| _AncestorTaggedSource_ | AncestorTaggedSources | Object | Conditional | An object containing the the properties of the tag present in the *tag source*. |
```
Matt-Cowsert:
```suggestion
| _AncestorTaggedSource_ | AncestorTaggedSources | Object | Conditional | An object containing the properties of the tag present in the *tag source*. |
```

Typo: "the the properties."

specification/datasets/cost_and_usage/columns/tagdetails.md
```diff
@@ -0,0 +1,476 @@
...
+
+### Object Entries
+
+The tag key object and tag source objects contain the following properties:
+
+| Key | ValueType | Required | Description |
+| ----- | ---- | ---------- | ----------- |
+| TagSource | String | Conditional | The type of *tag source* where the tag key is present. |
+| TagSourceId | String | TRUE | The identifier of the *tag source* where the tag key is present. |
+| TagValue | String or Boolean | TRUE |  The value associated with the tag key. |
```
Matt-Cowsert:
```suggestion
| TagValue | String, Boolean, or Number | TRUE |  The value associated with the tag key. |
```

The JSON schema (lines 252-265) allows `number` type and the second example scenario uses `12.2` (line 431). The table should match.

Worth noting: Number-typed tag values have schema and example support but no normative requirement text in the Content Requirements section (lines 54-59). The TF may want explicit guidance on when Number values appear.

specification/datasets/cost_and_usage/columns/tagdetails.md
```diff
@@ -0,0 +1,476 @@
...
+    * Tag property key MUST match the spelling and casing specified for the FOCUS-defined property.
+    * Tag property value MUST be of the type specified for that property.
+    * Tag properties MUST adhere to additional normative requirements specific to that property.
+  * Each tag key object MUST contain an object with the key "AncestorTaggedSources".
+    * The value of AncestorTaggedSources MUST be null when the tag key from the corresponding tag key object within the tag scheme is not present in any *tag source* other than the *tag source* which results in the *finalized tag*.
+    * When the value of AncestorTaggedSources is not null, each object in AncestorTaggedSources MUST have a key denoting the *tag source*.
+      * Each tag source object MUST contain FOCUS-defined tag properties.
+    * AncestorTaggedSources SHOULD contain objects for *tag sources* which did not result in the *finalized tag*.
+* Each tag scheme object MUST contain an array with the key "UntaggedSources".
+  * Untagged Sources array MUST contain all *tag sources* for the corresponding tag scheme which are eligible to be tagged for the *charge*.
```
Matt-Cowsert:
```suggestion
  * UntaggedSources array MUST contain all *tag sources* for the corresponding tag scheme which are eligible to be tagged for the *charge* but have no tags applied.
```

Current text reads as "all sources eligible to be tagged" which would include tagged sources. Adding "but have no tags applied" matches the description in the Object Schema table (line 92: "sources which support this tag scheme...that had no tags applied"). Also normalizes the key name from "Untagged Sources" (with space) to the exact key "UntaggedSources."

specification/datasets/cost_and_usage/columns/tagdetails.md
```diff
@@ -0,0 +1,476 @@
...
+  * The value of Untagged Sources MUST be null when there are no eligible *tag sources* which contain no tags.
```
Matt-Cowsert:
```suggestion
  * The value of UntaggedSources MUST be null when there are no eligible *tag sources* which contain no tags.
```

Same key-name normalization: "Untagged Sources" to "UntaggedSources."

specification/datasets/cost_and_usage/columns/tagdetails.md
```diff
@@ -0,0 +1,476 @@
...
+}
+```
+
+The corresponding Tags column would either be null or contain an empty object:
+
+```json
+{}
+```
+
+### Tags from multiple sources and schemas with different value types
```
Matt-Cowsert:
```suggestion
### Tags from multiple sources and schemes with different value types
```

Per the new glossary term [*tag scheme*](#glossary:tag-scheme). "Schemas" refers to data shape; "schemes" refers to tagging concepts.

specification/glossary.md
```diff
@@ -224,6 +224,10 @@ A sub account is an optional service-provider-supported construct for organizing
 
 A metadata label assigned to a resource to provide information about it or to categorize it for organizational and management purposes.
 
+<a name="glossary:tag-scheme"><b>Tag Scheme</b></a>
+
+A user-defined or service-provider-defined tagging concept which exists independently from other tagging concepts supported by the service provider (e.g. if both tags and labels are supported, each would be considered a tag scheme).
```
Matt-Cowsert:
```suggestion
A user-defined or service-provider-defined tagging concept which exists independently from other tagging concepts supported by the service provider (e.g., if both tags and labels are supported, each would be considered a tag scheme).
```

Missing comma after "e.g." per the convention used elsewhere in the glossary (e.g., Service definition).

specification/datasets/cost_and_usage/columns/tagdetails.md
```diff
@@ -0,0 +1,476 @@
...
+* TagDetails MUST conform to [JsonObjectFormat](#attributes.jsonobjectformat) requirements.
+* TagDetails nullability is defined as follows:
+  * TagDetails MUST NOT be null unless all of the following are true:
+    * Tags column is null.
+    * Tags are not present in any other *tag sources*.
+    * No *tag sources* are supported for any user-defined tag scheme.
+
+### Object Schema Requirements
+
+Tag Details consists of a valid JSON object which contains objects for one or more Tags schemes which each contain an object describing the tags as well as an array of eligible *tag sources* which do not contain tags.
```
Matt-Cowsert: Non-blocking: "one or more Tags schemes" could be read as referencing the Tags column name. Consider "one or more tag schemes" (lowercase) or linking the new glossary term: "[*tag schemes*](#glossary:tag-scheme)."

specification/datasets/cost_and_usage/columns/tagdetails.md
```diff
@@ -0,0 +1,476 @@
...
+
+When TagDetails is not null, the JsonObjectFormat for TagDetails adheres to the following requirements:
+
+* TagDetails root object MUST contain an object for each tag scheme present in Tags.
+  * The key for each tag scheme object MUST align to the tag prefix (up to but not including the forward slash) used to define the tag scheme in Tags.
+  * The key for the unprefixed user-defined tag scheme in Tags MUST be "Default".
+* Each tag scheme object MUST contain an object for each tag key used in that tag scheme.
```
Matt-Cowsert: Non-blocking: "each tag key used in that tag scheme" is ambiguous. In the ancestor-only example (lines 330-348), tag key "foo" appears in the TagDetails Tags object despite not being in the Tags column. "Used" could mean "present in any tag source for this charge" or "present in the Tags column." The ancestor-only pattern implies the former, but implementors could reasonably read it either way.

Classified as non-blocking because the fix requires a design decision on what "used" means, not a text replacement.

specification/datasets/cost_and_usage/columns/tagdetails.md
```diff
@@ -0,0 +1,476 @@
...
+
+<b>TagSource</b>
+
+"TagSource" denotes the type of *tag source* where the tag key is present.
+
+The "TagSource" property adheres to the following requirements:
+
+* TagSource MUST be present in each tag key object in the Tags object.
+* TagSource MUST contain the type of *tag source* where the tag key is present.
+* TagSource SHOULD NOT be present in each tag source object in the AncestorTaggedSources object.
```
Matt-Cowsert: Non-blocking: TagSource is redundant in AncestorTaggedSources because the key itself denotes the source type. If there's no valid reason a generator would include it, MUST NOT might be cleaner than SHOULD NOT.

specification/datasets/cost_and_usage/dataset.md
```diff
@@ -151,6 +152,8 @@ CostAndUsage MUST adhere to the following requirements:
   * CostAndUsage MUST include [SubAccountId](#datasets.costandusage.subaccountid) when the service provider supports a *sub account* construct.
   * CostAndUsage MUST include [SubAccountName](#datasets.costandusage.subaccountname) when the service provider supports a *sub account* construct.
   * CostAndUsage MUST include [SubAccountType](#datasets.costandusage.subaccounttype) when the service provider supports more than one possible SubAccountType value.
+  * CostAndUsage SHOULD include [TagDetails](#datasets.costandusage.tagdetails) when the data generator supports setting user or provider-defined tags and supports [Dataset Configuration](#attributes.datasetconfiguration).
```
Matt-Cowsert: Non-blocking: The SHOULD NOT sub-bullet (line 156) is nested under the SHOULD, which is unusual compared to other column requirements in this file. Most conditional columns use a single top-level bullet. Consider whether these should be sibling bullets or consolidated.

Matt-Cowsert (changes requested):
Solid foundation for a column that will significantly improve tag coverage analysis and provenance tracking. The main themes across this review:

1. **Normative ambiguities**: A few requirements could be read multiple ways by implementors (UntaggedSources scope, "each tag key used in that tag scheme").
2. **Type inconsistency**: The Object Entries table disagrees with the JSON schema and examples on TagValue types.
3. **Editorial**: Two typos, one naming inconsistency, one glossary punctuation fix.

Several TF2 action items from 3/4 appear still open (property renaming, additional examples). Not blocking this review round but flagging for tracking.

Matt-Cowsert:
## TF2 Action Item - March 11, 2026

Karl has made significant progress: removed the FOCUS column property, modified nullability requirements, added a second example with multiple tag schemes, added a glossary definition for "tag scheme," and added JSON schema replacing JTD. The remaining blocker is tag source conditionality. The current format makes it difficult to express that the same property exists in multiple objects with different requirements per parent object.

**Scope:**
- [ ] Restructure tag source normative requirements (should not be conditional in current form)
- [ ] Check for and resolve unresolved conversations
- [ ] Pull PR out of draft when ready

**Assignee:** @kk09v

specification/datasets/cost_and_usage/columns/tagdetails.md
kk09v: Justin, I added to that the inclusion requirements that this column SHOULD NOT be included if dataset configuration (column selection in export) is not supported. I believe this covers the concerns you and I had around this.

specification/datasets/cost_and_usage/columns/tagdetails.md
davidearney: This is still open and out there, so I'm going to add my two cents: 

I've specifically encountered scenarios with containers and VMs where tags weren't available, but metadata associated with untagged sources was still critical for cost identification. In those cases, 
having access to that metadata — even in the absence of finalized tags — was the only way to attribute costs meaningfully. Requiring TagDetails to be null whenever Tags is null would silently discard exactly that kind of useful context. 

I'd suggest changing to: 
`* TagDetails MAY be null when Tags is null.`

specification/datasets/cost_and_usage/dataset.md
kk09v: Merged from working draft and added the normative requirements.

shawnalpay:
Discussed how this PR intersects with concurrent PR #2007 in Mar 16 Maintainers.

@kk09v Agreement there was that you _will_ refactor this PR to leverage that format before pulling out of draft.  Given that I worked hard with @mike-finopsorg to get #2007 into fighting shape, I am more than happy to help you with that refactor if you get stuck, just let me know!

thecloudman (commented):
Is there a way to introduce tag details without actually adding in the tag key-vlaue pairs. I feel this would bloat datasets as we would essentially have three columns with some level of duplication, you would end up wit TagDetails, Tags, and Allocationtags all with fairly similar data which sort of goes against guidance specifcally around custom columns where providers should not introduce duplicate data if adding custom columns.

shawnalpay:
> Is there a way to introduce tag details without actually adding in the tag key-vlaue pairs. I feel this would bloat datasets as we would essentially have three columns with some level of duplication, you would end up wit TagDetails, Tags, and Allocationtags all with fairly similar data which sort of goes against guidance specifcally around custom columns where providers should not introduce duplicate data if adding custom columns.

It's a good point, @thecloudman, and one that I debated with myself (and with AI) for awhile before coming around on this.  Here was Gemini's reasoning when I asked it a few weeks ago:

----

This is a classic "normalization vs. usability" trade-off. While it seems redundant to include the tag value in both columns, there is a very strong architectural reason to keep it in **TagDetails**.

### The Rationale: "Contextual Binding"

If you remove the Tags content (the keys and values) from `TagDetails`, you break the link between the metadata and the tag itself.

#### 1. The "Join" Problem
If `TagDetails` only contained the extra info (Ancestors, Sources) but relied on `Tags` for the key/value pairs, a consumer would have to perform a complex, in-memory "join" between two JSON objects every time they wanted to answer a question like: 
*"Show me the provenance of the 'Environment' tag where the value is 'Production'."*

* **With Duplication:** You query `TagDetails` directly: "Find object where `TagKey` = 'Environment' AND `TagValue` = 'Production', then return `AncestorTaggedSources`."
* **Without Duplication:** You have to:
    1. Parse `Tags` to find that 'Environment' = 'Production'.
    2. Parse `TagDetails` to find the 'Environment' metadata object.
    3. Match them up in your code.
    4. Hope that the array indexes or keys align perfectly.

#### 2. The "Point-in-Time" Integrity
Tags are volatile. If a tag value changes from "Prod" to "Production," and you have two separate columns updating at potentially different parts of the ETL pipeline (or if one is nullable and the other isn't), you risk a mismatch. 

By storing the Tag Value inside the Tag Detail object, you are asserting: *"This specific metadata (Source=Resource) applies to THIS specific value (Value=baz)."*

#### 3. Handling "Multi-Value" Conflicts (The Advanced Case)
Some advanced tagging scenarios (like Azure Policy or GCP hierarchy) can result in conflicting values at different levels:
* **Subscription:** `CostCenter: 100`
* **Resource:** `CostCenter: 200`

The `Tags` column (the summary) creates a "Final Effective Value" (e.g., 200). The `TagDetails` column needs to show both the winner (200) and the loser (100 in `AncestorTaggedSources`). If you strip the "winner" value out of `TagDetails`, the structure becomes ambiguous. You would have a list of ancestors, but no clear indication of which one "won" without referencing the other column.

### Recommendation
**Keep the duplication.** Storage is cheap; compute complexity (especially for JSON parsing) is expensive. Making `TagDetails` a self-contained object ensures that any analysis of tag provenance or governance can be done using only that column, without needing to cross-reference the `Tags` column.

specification/datasets/cost_and_usage/columns/tagdetails.md
kk09v: I added a definition for tag schemes to the glossary whereas previously, it had only really been used in the Tags column definition.

kk09v:
```suggestion
Tag Details consists of a valid JSON object which contains objects for one or more *tag schemes* which each contain an object describing the tags as well as an array of eligible *tag sources* which do not contain tags.
```

specification/datasets/cost_and_usage/columns/tagdetails.md
```diff
@@ -0,0 +1,476 @@
...
+The TagDetails column adheres to the following requirements:
+
+* TagDetails MUST be of type String.
+* TagDetails MUST conform to [StringHandling](#attributes.stringhandling) requirements.
+* TagDetails MUST conform to [JsonObjectFormat](#attributes.jsonobjectformat) requirements.
+* TagDetails nullability is defined as follows:
+  * TagDetails MUST NOT be null unless all of the following are true:
+    * Tags column is null.
+    * Tags are not present in any other *tag sources*.
+    * No *tag sources* are supported for any user-defined tag scheme.
```
kk09v:
```suggestion
    * No *tag sources* are supported for any user-defined [*tag scheme*](glossary:tag-scheme).
```

specification/datasets/cost_and_usage/columns/tagdetails.md
kk09v: Yep, I had removed the option for Number because I could not think of a case in which number could be used. If I set a tag as product = 8675309, it would still show in tags as {"product":"8675309"} because I've put a number into a string field since the tag scheme supports strings. The definition of Tags states that the provider can't change the value (except TRUE for key-only tags), so it could never be a number. However, it later occurred to me that ProviderDefined tags could be numeric because they can define them on a per-key basis however they want.

specification/datasets/cost_and_usage/columns/tagdetails.md
kk09v: Yes. This was on my BOLO list

specification/datasets/cost_and_usage/dataset.md
kk09v: Agreed, this was done between Maintainers and TF2 last week. I had intended to ask in Maintainers, but I put my note on the wrong week. When I looked through, I didn't see any other column normative requirements that had 2 bullets. I believe this should have two bullets due to the two separate normative requirements, but I wasn't sure how to structure that best.

specification/glossary.md
kk09v: _Cringe_ this is a matter of stylistic debate, much like the Oxford comma.

- New Hart's Rules (the Oxford University Press style guide, which puts the Oxford in Oxford comma) discourages using a comma with this abbreviation. 
- Chicago Manual of Style encourages commas both before and after, though the before isn't applicable here because it's inside of parentheses.

Basically, we're inconsistent in our English language style, following one style guide some places and the other elsewhere. This truly is a "consistency review" topic, but I don't want to spend time debating rules of English any more than I want to start enforcing Robert's Rules in our meetings. 

I will begrudgingly accept your change for the sake of not having this debate.

specification/datasets/cost_and_usage/columns/tagdetails.md
kk09v: Agreed. Cleaned up and moved.

specification/datasets/cost_and_usage/columns/tagdetails.md
kk09v: Updated to "each tag key applied to any tag source in that tag scheme"

specification/datasets/cost_and_usage/columns/tagdetails.md
kk09v: I believe this is more clear in #2078

specification/datasets/cost_and_usage/columns/tagdetails.md
kk09v: This is resolved in #2078

specification/datasets/cost_and_usage/columns/tagdetails.md
kk09v: Resolved in the current version 
| TagValue | String, Boolean, or Number | TRUE |  The value associated with the tag key. |

specification/datasets/cost_and_usage/columns/tagdetails.md
kk09v: Yeah, the comment was still open but it was hidden because it was outdated. Here's what's currently reflected for nullability. 
  * TagDetails MUST NOT be null unless all of the following are true:
    * Tags column is null.
    * Tags are not present in any other *tag sources*.
    * No *tag sources* are supported for any user-defined tag scheme.

I even included an example that shows the Tags column is empty, but the TagDetails has ancestor tagged values present.

specification/datasets/cost_and_usage/columns/tagdetails.md
kk09v: This has been removed

kk09v:
@shawnalpay @thecloudman 

Continuing on the conversation here:
> Is there a way to introduce tag details without actually adding in the tag key-vlaue pairs. I feel this would bloat datasets as we would essentially have three columns with some level of duplication, you would end up wit TagDetails, Tags, and Allocationtags all with fairly similar data which sort of goes against guidance specifcally around custom columns where providers should not introduce duplicate data if adding custom columns.

I see the column as providing 3 key pieces of information:
- Tag Provenance
  - Where did the values in the Tags column come from?
- Ancestor Tags
  - If the charge record didn't have values in the tags column, can I look at its parent? I do this today, but I get the data through multiple chained API calls.
- Tag Eligibility
  - Was there a way that this _could have been_ tagged? The yes/no answer to this helps with tag coverage calculations. The values in UntaggedSources are breadcrumbs to find how to tag some of the non-obvious ones (e.g. providing tags in the API call itself for charged related to API usage).

I view the inclusion of the tags in the Tags column is necessary for the Tag Provenance piece. Take a look at my second example where there was tag inheritance applied and some values in the Tags column came from the resource while others came from an ancestor.

I also view the inclusion as necessary for Tag Eligibility. If those tags are not included, then the list of tag sources is not collectively exhaustive.

To be fair, both of these require the TagSource and TagSourceId, plus Provenance requires TagKey, but the the TagValue is truly redundant for the finalized tags. I don't think we're reducing the size much by cutting out that one property and if I was looking through this data, I'd view it as a miss to not have that included.

specification/datasets/cost_and_usage/dataset.md
```diff
@@ -153,6 +154,9 @@ CostAndUsage MUST adhere to the following requirements:
   * CostAndUsage MUST include [SubAccountId](#datasets.costandusage.subaccountid) when the service provider supports a *sub account* construct.
   * CostAndUsage MUST include [SubAccountName](#datasets.costandusage.subaccountname) when the service provider supports a *sub account* construct.
   * CostAndUsage MUST include [SubAccountType](#datasets.costandusage.subaccounttype) when the service provider supports more than one possible SubAccountType value.
+  * CostAndUsage SHOULD conform to the following requirements regarding the inclusion of [TagDetails](#datasets.costandusage.tagdetails):
+    * CostAndUsage SHOULD include [TagDetails](#datasets.costandusage.tagdetails) when the data generator supports setting user or provider-defined tags and supports [Dataset Configuration](#attributes.datasetconfiguration).
+    * CostAndUsage SHOULD NOT include [TagDetails](#datasets.costandusage.tagdetails) when the data generator does not support [Dataset Configuration](#attributes.datasetconfiguration).
   * CostAndUsage MUST include [Tags](#datasets.costandusage.tags) when the data generator supports setting user or provider-defined tags.
```
shawnalpay: This composite still needs finesse.

Here's one option:

```suggestion
  * When the data generator supports setting user or provider-defined tags, CostAndUsage MUST adhere to the following column presence requirements:
    * CostAndUsage SHOULD include [TagDetails](#datasets.costandusage.tagdetails) when the data generator supports [Dataset Configuration](#attributes.datasetconfiguration).
    * CostAndUsage SHOULD NOT include [TagDetails](#datasets.costandusage.tagdetails) when the data generator does not support [Dataset Configuration](#attributes.datasetconfiguration).
    * CostAndUsage MUST include [Tags](#datasets.costandusage.tags).
```

I think it's either that, or we just eliminate the composite completely:

```suggestion
  * CostAndUsage SHOULD include [TagDetails](#datasets.costandusage.tagdetails) when the data generator supports setting user or provider-defined tags and supports [Dataset Configuration](#attributes.datasetconfiguration).
  * CostAndUsage SHOULD NOT include [TagDetails](#datasets.costandusage.tagdetails) when the data generator does not support [Dataset Configuration](#attributes.datasetconfiguration).
  * CostAndUsage MUST include [Tags](#datasets.costandusage.tags) when the data generator supports setting user or provider-defined tags.
```

specification/datasets/cost_and_usage/dataset.md
```diff
@@ -153,6 +154,9 @@ CostAndUsage MUST adhere to the following requirements:
   * CostAndUsage MUST include [SubAccountId](#datasets.costandusage.subaccountid) when the service provider supports a *sub account* construct.
   * CostAndUsage MUST include [SubAccountName](#datasets.costandusage.subaccountname) when the service provider supports a *sub account* construct.
   * CostAndUsage MUST include [SubAccountType](#datasets.costandusage.subaccounttype) when the service provider supports more than one possible SubAccountType value.
+  * CostAndUsage SHOULD conform to the following requirements regarding the inclusion of [TagDetails](#datasets.costandusage.tagdetails):
+    * CostAndUsage SHOULD include [TagDetails](#datasets.costandusage.tagdetails) when the data generator supports setting user or provider-defined tags and supports [Dataset Configuration](#attributes.datasetconfiguration).
```
shawnalpay: As discussed in Maintainers a few weeks ago (and presumably not discussed in TF-2 since): 

**I object to our delivering a Recommended column without a very solid reason.**

I think we all agree/understand that generating this column will be hard to do.  (This is putting aside aside the argument that the dataset is much larger with this column, given that column selection would be facilitated via `Dataset Configuration`.) Presuming that few, if any data generators go to market with this column, then **we will have frustrated our practitioner base by defining a column that they can't presume will be in the dataset**, thus failing to have facilitated the basic use case of "what % of my charges _could_ have been tagged with Tag X?".  This FR was added to scope, and this PR exists as currently authored, because we believe in the priority of this practitioner need.

Is practitioner confusion/disappointment worth the benefit of having defined a north-star column that we may, presumably in some future version, require?  And if that is the plan, then why not simply require it now?

We have a handful of Recommended columns in FOCUS today, and in my opinion, they all suffer from this same problem: **practitioners expect them in FOCUS, and then they're not, and people are frustrated and disappointed**.  Consider `ServiceSubcategory`, which is Recommended.  The Feature Level is fine print that practitioners don't necessarily read, and it seems like the community expects this column in FOCUS dataset artifacts because it's in the spec.  I predict the same being true with `TagDetails`.

Consider also `AvailabilityZone`.  As I understand it (I was not directly involved at the time): the decision to make that column Recommended was the result of a long series of conversations where some felt strongly it should be included, and others felt strongly otherwise, and thus the only reasonable compromise was to make it Recommended.  I have not seen any such argument yet take place around `TagDetails`, and so I disagree with following that precedent.

Going forward, I believe strongly that we, as FOCUS authors, should either 1) believe enough in the utility of a new column that we require a `MUST` (and let data generators take conformance gaps if they don't agree with its utility), 2) slim down the scope of the column in some way to develop consensus around a `MUST`, or 3) postpone the column until we can get alignment on 1 or 2.

Thus, and to force conversation around this: I propose Option 1, making this column a `MUST`.  In my opinion, the confusion of an aspirational "ghost column" is worse than either the friction of forcing data generators to take a conformance gap, or the absence of the column at all.

```suggestion
    * CostAndUsage MUST include [TagDetails](#datasets.costandusage.tagdetails) when the data generator supports setting user or provider-defined tags and supports [Dataset Configuration](#attributes.datasetconfiguration).
```

shawnalpay (changes requested):

specification/datasets/cost_and_usage/dataset.md
kk09v: I would dislike the first suggestion if TagDetails and Tags weren't alphabetically next to each other (i.e. if we include AllocatedTags in this composite), but given that these are already grouped together, I find this option easier to read.

specification/datasets/cost_and_usage/dataset.md
kk09v: I'll check with everyone on this during the TF2 meeting this week before making or rejecting changes.

shawnalpay: Sounds good.  I'll also be bringing this to the Maintainers tomorrow.  We are swiftly running out of time in this release, so we have to align quickly.

shawnalpay: Just to summarize my wall of text above, we have four options:

1) Make it Conditional and keep as-is
2) Make it Conditional by slimming down scope to get consensus
3) Descope the column from 1.4 and consider in a future release
4) Leave it Recommended

In my opinion: for the reasons stated above, Option 4 is by far the worst option for the practitioner.

specification/datasets/cost_and_usage/dataset.md
shawnalpay: This was resolved in a separate thread.

shawnalpay:
Please see [this thread](https://github.com/FinOps-Open-Cost-and-Usage-Spec/FOCUS_Spec/issues/1124#issuecomment-4114258425) on parent Feature Request #1124.  We shall descope this work from FOCUS 1.4, given that the Maintainers today agreed that the PR in its current form would not survive the Member review and approval process.

Earlier today, @kk09v and I merged #2091 into this PR, so it does adhere to the refactored JSON format of other similarly-formatted columns.  This will serve as a useful reference when we resume this work in a future release.

We will discuss this PR and the parent FR in Mar 25 TF-2 this week.  In the meantime, please leave your feedback either here, or on #1124.

Closing this PR, but not deleting the branch.

specification/datasets/cost_and_usage/dataset.md
shawnalpay: Discussed at length in Mar 23 Maintainers.  See [this thread](https://github.com/FinOps-Open-Cost-and-Usage-Spec/FOCUS_Spec/issues/1124#issuecomment-4114258425) on FR #1124 for more, but: the short of it  is that we did not have consensus on this point, and thus we did not believe that this PR could survive the Member review and approval process.  Thus, we have descoped this work from 1.4.

Will leave this thread open, despite having closed the PR.  If and when we pick this PR back up, we will still need to address this concern.

specification/glossary.md
```diff
@@ -238,6 +238,10 @@ A sub account is an optional service-provider-supported construct for organizing
 
 A metadata label assigned to a resource to provide information about it or to categorize it for organizational and management purposes.
 
+<a name="glossary:tag-scheme"><b>Tag Scheme</b></a>
+
+A user-defined or service-provider-defined tagging concept which exists independently from other tagging concepts supported by the service provider (e.g., if both tags and labels are supported, each would be considered a tag scheme).
```
shawnalpay, line 243: Gemini's suggestion.

An independent metadata system defined by a user or service provider. If a platform supports multiple classification methods—such as both tags and labels—each constitutes a separate tag scheme.

PR #1816, opened 2026-01-11
FR #1091: Add Dataset Configuration attribute

flanakin:
## Summary

Introduces the **Dataset Configuration** attribute for FOCUS 1.4, which defines how practitioners can configure the structure and content of FOCUS datasets.

This attribute addresses issue #1091 (Column Selection) with a focused scope on column selection. Additional configuration options (row aggregation, time granularity, version selection, row filtering) have been documented in supporting content for potential integration in a separate change.

## Requirements

| Requirement Level | Description |
|------------------|-------------|
| MUST | Be configurable to include only a user-defined selection of columns |
| MUST | Adhere to all column-level specifications regardless of configuration |

## Changes

- **New**: `specification/attributes/dataset_configuration.md` - Attribute specification
- **New**: `supporting_content/attributes/dataset_configuration.md` - Design rationale, deferred requirements, and examples
- **Updated**: `specification/attributes/attributes.mdpp` - Added include
- **Updated**: `specification/datasets/cost_and_usage/dataset.md` - Added conformance reference
- **Updated**: `specification/datasets/contract_commitment/dataset.md` - Added conformance reference

## Deferred to Separate Changes

The following requirements were developed but deferred to keep this PR focused on the original #1091 scope:

- Row aggregation (summing metrics when dimensions are identical)
- Time granularity selection (daily, hourly, monthly)
- FOCUS version selection
- Row filtering by column values
- Configuration metadata structure (extending Dataset Instance metadata)

These are documented in `supporting_content/attributes/dataset_configuration.md`.

## Design Decisions

1. **Renamed from "Dataset Delivery"** to "Dataset Configuration" since "delivery" implies push semantics
2. **Column selection is MUST** (not SHOULD) - essential for practitioners to control data volume
3. **Expanded requirements deferred** - keeps this PR focused on the original column selection ask
4. **Requirements model rules deferred** - will be added after 1.3 requirements modeling is complete

## Test Plan

- [x] Build succeeds
- [x] Attribute follows existing patterns (ColumnHandling, DiscountHandling)

Closes #1091

🤖 Generated with [Claude Code](https://claude.ai/claude-code)

PR #1908, opened 2026-01-28
AI #1914: Reclassify unit price columns from Metric to Dimension

### LINKED ISSUE #1914: [AI] Create draft PR for unit price columns metric-to-dimension reclassification
### Detailed Description

During PR #1816  discussion, four unit price columns were identified as incorrectly tagged as metrics rather than dimensions. Michael originally included this fix due to aggregation requirements that were subsequently removed from the PR scope. Per TF consensus, this correction should be carved out into a separate draft PR and held until maintainers discuss the broader implications of metric vs. dimension classification.

### Definition of Done

- [ ] Create new draft PR with unit price column type changes
- [ ] Leave PR in draft status pending maintainer discussion
- [ ] Remove metric-to-dimension changes from PR #1816

flanakin:
## Summary

Reclassifies 4 unit price columns from Metric to Dimension. Unit prices are not summable and conflict with aggregation semantics — summing unit prices when rows are aggregated produces meaningless values.

## Changes

| Column | Change |
|--------|--------|
| ListUnitPrice | Metric → Dimension |
| ContractedUnitPrice | Metric → Dimension |
| PricingCurrencyListUnitPrice | Metric → Dimension |
| PricingCurrencyContractedUnitPrice | Metric → Dimension |

## Rationale

- Unit prices represent per-unit rates, not additive quantities
- Summing unit prices across rows (e.g., during row aggregation) produces meaningless results
- Classifying them as Dimension ensures they are treated as grouping attributes rather than summable values

Split from PR #1816 (Dataset Configuration) per reviewer feedback.

## Test plan

- [ ] Build succeeds
- [ ] No other columns reference unit price column types

🤖 Generated with [Claude Code](https://claude.ai/claude-code)

### DIFF
--- a/specification/datasets/cost_and_usage/columns/contractedunitprice.md
+++ b/specification/datasets/cost_and_usage/columns/contractedunitprice.md
@@ -41,7 +41,7 @@ The agreed-upon unit price for a single Pricing Unit of the associated SKU, incl
 
 | Constraint      | Value                                |
 |:----------------|:-------------------------------------|
-| Column type     | Metric                               |
+| Column type     | Dimension                            |
 | Feature level   | Conditional                          |
 | Allows nulls    | True                                 |
 | Data type       | Decimal                              |
--- a/specification/datasets/cost_and_usage/columns/listunitprice.md
+++ b/specification/datasets/cost_and_usage/columns/listunitprice.md
@@ -40,7 +40,7 @@ The suggested service-provider-published unit price for a single Pricing Unit of
 
 | Constraint      | Value                                |
 |:----------------|:-------------------------------------|
-| Column type     | Metric                               |
+| Column type     | Dimension                            |
 | Feature level   | Conditional                          |
 | Allows nulls    | True                                 |
 | Data type       | Decimal                              |
--- a/specification/datasets/cost_and_usage/columns/pricingcurrencycontractedunitprice.md
+++ b/specification/datasets/cost_and_usage/columns/pricingcurrencycontractedunitprice.md
@@ -42,7 +42,7 @@ The agreed-upon unit price for a single Pricing Unit of the associated SKU, incl
 
 | Constraint      | Value                                |
 |:----------------|:-------------------------------------|
-| Column type     | Metric                               |
+| Column type     | Dimension                            |
 | Feature level   | Conditional                          |
 | Allows nulls    | True                                 |
 | Data type       | Decimal                              |
--- a/specification/datasets/cost_and_usage/columns/pricingcurrencylistunitprice.md
+++ b/specification/datasets/cost_and_usage/columns/pricingcurrencylistunitprice.md
@@ -42,7 +42,7 @@ The suggested service-provider-published unit price for a single Pricing Unit of
 
 | Constraint      | Value                                |
 |:----------------|:-------------------------------------|
-| Column type     | Metric                               |
+| Column type     | Dimension                            |
 | Feature level   | Conditional                          |
 | Allows nulls    | True                                 |
 | Data type       | Decimal                              |


shawnalpay:
@flanakin 
Given that this is outside of the scope of 1.4: if you would like to bring this up for discussion, please gather at least one approval (ideally more) from amongst the Maintainers or Contributors.  We can then discuss this in a future Members meeting.

I have opinions on this, but I'll stay neutral for now.  That said: the concept of aggregation does not equal sum and only sum.  Aggregations include average, min, max, and median, all of which are applicable to unit prices.

specification/datasets/cost_and_usage/columns/contractedunitprice.md
shawnalpay, line 1: These changes would also need to be reflected in the dataset overview page, namely:

https://github.com/FinOps-Open-Cost-and-Usage-Spec/FOCUS_Spec/blob/unit-price-column-type/specification/datasets/cost_and_usage/dataset.md

udam-stitcher:
I don't think this is a good idea. While it's not summable, it can be used in aggregations and doing math (e.g., average unit prices, P*Q, etc.)

Not that ChatGPT is the ultimate decider here but I asked the simple question:

> Would a numeric column like a ListUnitPrice be considered a dimension or a metric?

Short answer: it’s a metric 👍

Longer, more useful answer (because this stuff always gets fuzzy in FinOps / analytics land):

## Dimension vs. Metric — the mental model

### Dimensions

Describe what something is
Usually categorical or descriptive
Used to group, filter, slice
Examples: Service, SKU, Region, Currency, UsageType

### Metrics

Describe how much
Numeric values you aggregate, compare, or calculate with
Examples: Cost, UsageQuantity, BilledAmount, EffectiveRate

## Where does ListUnitPrice land?

Even though it’s numeric, ListUnitPrice is a metric, because:
It represents a measurable value
You use it in calculations (e.g., effective rate, discounts, variance)
It participates in formulas like
`ExtendedListCost = UsageQuantity × ListUnitPrice`
`Discount = 1 - (EffectiveRate / ListUnitPrice)`

## Why people get tripped up here

Two common sources of confusion:

1. “But it doesn’t aggregate cleanly”

True — you usually don’t sum unit prices.
But metrics aren’t required to be summable.

There are:

- Additive metrics (Cost, Usage)
- Semi-additive metrics (Balances over time)
- Non-additive metrics (Rates, Percentages, Unit Prices)

ListUnitPrice is a non-additive metric.

2. “But it feels like an attribute of a SKU”

Also true — conceptually it describes the SKU.
But the rule of thumb is:
If it’s numeric and primarily used in calculations rather than grouping, it’s a metric.
In practice, you might store it alongside dimensional data, but you still treat it as a metric analytically.

shawnalpay:
@flanakin 
We discussed this PR in Feb 2 Maintainers per AI #1915.  Consensus opinion amongst the Maintainers was that Unit Prices are to remain Metrics and not Dimensions, largely per the argument laid out by @udam-stitcher [above](https://github.com/FinOps-Open-Cost-and-Usage-Spec/FOCUS_Spec/pull/1908#issuecomment-3824576986) (which aligns with the comment I left earlier [above](https://github.com/FinOps-Open-Cost-and-Usage-Spec/FOCUS_Spec/pull/1908#issuecomment-3814178555)).

If you would like time on the Maintainers agenda next week to make your case, let me know, and I will add it to the agenda.  Otherwise, I will close this PR in a week's time.

aws-jrmarks (commented):
Agree with Udam and Shawn here, unit price columns make more sense as Metrics given they are non-additive by themselves but can be combined with UsageQuantity or EffectiveRate in calculations, or in non-sum aggregations like average. I'd leave them as classified as Metric.

shawnalpay:
Closing this PR, per my previous message.

PR #2043, opened 2026-03-05
AI #1965: Add code wrapping CSS rules to fix PDF rendering issues

### LINKED ISSUE #1965: [AI] Address wrapping in SQL queries
### Detailed Description

We have examples of SQL queries going "off the page" in the PDF.  We need to address this.

Image

### Definition of Done

Update the spec content to avoid this wrapping.

mike-finopsorg:
## Problem
Code blocks in the generated PDF have horizontal scrollbars and truncate content that extends beyond the page width, making long lines unreadable in the printed specification.

## Solution
Added CSS rules to force pre-wrap behavior for `<pre>` and `<code>` elements in PDF rendering:
- `white-space: pre-wrap` - wraps long lines instead of overflowing
- `word-wrap: break-word` - breaks words when necessary to prevent overflow
- `overflow-wrap: break-word` - ensures wrapping at word boundaries

## Impact
- Code blocks now wrap to fit page width in PDFs
- No content truncation or horizontal scrolling
- Improves readability of specification examples and code samples in printed/PDF format


shawnalpay:
@mike-finopsorg Looking at the example we were discussing yesterday: should I expect the word `Usage` to appear in code block in the PDF generated by this branch?  It does not yet appear to.

Screenshot 2026-03-05 at 8 08 25 AM

ljadvey (approved):

shawnalpay (changes requested):
Needs revision, per my note above.  Conferred with @mike-finopsorg yesterday; apparently it works locally, but not remotely.  He will need to look into it.

mike-finopsorg:
I have forced font selection and updated a pile of packages to the latest, on my local Mac, local docker, and GitHub actions it now renders like attached:

Screenshot 2026-03-25 at 10 17 22 am

custom_linter_rules/rule_md_990.py
```diff
@@ -60,16 +59,3 @@ def next_token(self, context: PluginScanContext, token: MarkdownToken) -> None:
                     found_index = text_token.token_text.find(smart_char_item, start_index)
 
 
-        elif token.is_code_block:
```
shawnalpay, line 63: I see that the state tracking for code blocks, HTML blocks, and links (e.g., `self.__in_code_block`) was removed from the `next_token` method.

Does the `pymarkdownlnt` version handle this automatically under the hood?  Otherwise, removing these boundary checks will cause the linter to throw false positives on characters that legitimately appear inside code snippets, HTML tags, or URLs. Do we need to restore some sort of state-tracking logic?

mike-finopsorg: Added back in these cases. Thanks for raising

shawnalpay (approved):

shawnalpay:
Approved!  We don't need Member approval, but let's hold to merge until we have our batch of approved PRs tomorrow.  Thanks @mike-finopsorg!

.github/workflows/main.yml
```diff
@@ -17,9 +17,9 @@ jobs:
         shell: bash
         run: |
           /usr/bin/apt-get -y update
-          wget https://github.com/wkhtmltopdf/packaging/releases/download/0.12.6.1-2/wkhtmltox_0.12.6.1-2.jammy_amd64.deb
+          wget https://github.com/wkhtmltopdf/packaging/releases/download/0.12.6.1-3/wkhtmltox_0.12.6.1-3.jammy_amd64.deb
```
narlaadbe, line 20: Minor: Should we make the same change on the the other workflows as well?
https://github.com/FinOps-Open-Cost-and-Usage-Spec/FOCUS_Spec/blob/36220e116456c8e54704bbef41297ab8c5a9a7b3/.github/workflows/working_draft.yml#L21
https://github.com/FinOps-Open-Cost-and-Usage-Spec/FOCUS_Spec/blob/36220e116456c8e54704bbef41297ab8c5a9a7b3/.github/workflows/candidate_release.yml#L20

shawnalpay: Good point -- I agree. What do you think @mike-finopsorg?

mike-finopsorg: Good catch. updated

PR #2167, opened 2026-04-06
AI #2064: Replace &dollar; entities with literal $ and harden makescripts

### LINKED ISSUE #2064: [AI] Resolve $ vs &dollar; inconsistency in spec markdown files
### Detailed Description

Historically, dollar signs in certain contexts (when passed through PDF/HTML pipelines) caused rendering or build issues, so people had to encode them as entities.

Flexibility example files (e.g., [one_hundred_percent_utilization_with_commitment_discount_flexibility_with_1_resource.md](https://github.com/FinOps-Open-Cost-and-Usage-Spec/FOCUS_Spec/blob/working_draft/specification/appendix/examples/commitment_discount_flexibility_examples/one_hundred_percent_utilization_with_commitment_discount_flexibility_with_1_resource.md)) use literal `$` for dollar signs and render correctly in both HTML and PDF.  New commitment discount example files use `&dollar;` HTML entities.

This inconsistency was identified during PR #1877 review by @shawnalpay ([comment](https://github.com/FinOps-Open-Cost-and-Usage-Spec/FOCUS_Spec/pull/1877#discussion_r2937384576)) and directed at @mike-finopsorg for clarification. The Mar 16 Maintainers Meeting decided this should be resolved outside PR #1877.

**Options:**
1. Standardize on `&dollar;` everywhere — safest for future pipeline changes (e.g., LaTeX), but harder to read in raw markdown
2. Standardize on `$` everywhere — simpler, currently works in HTML and PDF
3. Document when each is appropriate (if there's a technical reason for the difference)

### Definition of Done

- @mike-finopsorg has provided a ruling on the preferred convention ($ vs &dollar;)
- Chosen convention is applied consistently across all FOCUS spec markdown files
- Contributor style guide / CONTRIBUTING.md is updated with the decision (Writing Style and Format) or other document capturing this decision

mike-finopsorg:
## Summary

- replace HTML &dollar; with literal $ across commitment discount, SaaS, and rounding appendix markdown examples
- update vendored MarkdownPP LaTeXRender heuristics to avoid false-positive $...$ rendering for currency/prose/table content
- preserve intentional inline LaTeX rendering behavior while suppressing noisy "Rendering:" output during spec build
-  disable pandoc dollar-math parsing and harden LaTeXRender for currency text

### Type of Change
- [X] Bug fix
- [ ] New feature / Specification update
- [ ] Editorial / Formatting

### Author Checklist
- [x] **AI Usage Guidelines:** I have read the [FOCUS AI Usage Guidelines](/guidelines/contributors/ai-usage-guidelines.md).
- [x] **Self Review Attestation:** I have performed a self-review of my own contribution.
- [x] **AI-Generated Examples:** If this PR includes examples generated by AI, I have manually verified that the data is mathematically accurate, logically consistent, and compliant with the FOCUS schema.

### DIFF
--- a/vendored/MarkdownPP/Modules/LaTeXRender.py
+++ b/vendored/MarkdownPP/Modules/LaTeXRender.py
@@ -22,7 +22,7 @@ from MarkdownPP.Transform import Transform
 #
 # Keep detection broad, then apply a content heuristic before rendering to
 # avoid treating currency/prose as LaTeX. Escaped dollars ("\$") are ignored.
-singlelinere = re.compile(r"(?<!\\)\$(\$?).+?(?<!\\)\$(\$?)")
+singlelinere = re.compile(r"(?<!\\)(?<!\$)(\$\$(?!\$)|\$(?!\$))(.+?)(?<!\\)\1(?!\$)")
 # $... or ...$ (or $$... or ...$$)
 startorendre = re.compile(r"^(?<!\\)\$(\$?)(?!\d)|^\S.*(?<!\\)\$(\$?)$")
 
--- a/vendored/MarkdownPP/Modules/LaTeXRender.py
+++ b/vendored/MarkdownPP/Modules/LaTeXRender.py
@@ -22,7 +22,8 @@ from MarkdownPP.Transform import Transform
 #
 # Keep detection broad, then apply a content heuristic before rendering to
 # avoid treating currency/prose as LaTeX. Escaped dollars ("\$") are ignored.
-singlelinere = re.compile(r"(?<!\\)(?<!\$)(\$\$(?!\$)|\$(?!\$))(.+?)(?<!\\)\1(?!\$)")
+# Also enforces no space immediately after the opening $ or before the closing $
+singlelinere = re.compile(r"(?<!\\)(?<!\$)(\$\$(?!\$)|\$(?!\$))(?!\s)(.+?)(?<!\s)(?<!\\)\1(?!\$)")
 # $... or ...$ (or $$... or ...$$)
 startorendre = re.compile(r"^(?<!\\)\$(\$?)(?!\d)|^\S.*(?<!\\)\$(\$?)$")
 
--- a/vendored/MarkdownPP/Modules/LaTeXRender.py
+++ b/vendored/MarkdownPP/Modules/LaTeXRender.py
@@ -19,12 +19,17 @@ from MarkdownPP.Module import Module
 from MarkdownPP.Transform import Transform
 
 # $...$ (or $$...$$)
-singlelinere = re.compile(r"\$(\$?)..*\$(\$?)")
+#
+# Keep detection broad, then apply a content heuristic before rendering to
+# avoid treating currency/prose as LaTeX. Escaped dollars ("\$") are ignored.
+singlelinere = re.compile(r"(?<!\\)\$(\$?).+?(?<!\\)\$(\$?)")
 # $... or ...$ (or $$... or ...$$)
-startorendre = re.compile(r"^\$(\$?)|^\S.*\$(\$?)$")
+startorendre = re.compile(r"^(?<!\\)\$(\$?)(?!\d)|^\S.*(?<!\\)\$(\$?)$")
 
 codere = re.compile(r"^(    |\t)")
 spancodere = re.compile(r'(`[^`]+\`)')  # code between backticks
+mathcommandre = re.compile(r"\\[A-Za-z]+")
+wordre = re.compile(r"[A-Za-z]{2,}")
 
 # Support for Pandoc style code blocks with attributes
 fencedcodere = re.compile(r"^((> *)?```\w*|(> *)?~~~~*(\s*{.*})?)$")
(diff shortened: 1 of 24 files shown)


vendored/MarkdownPP/Modules/LaTeXRender.py
```diff
@@ -19,12 +19,17 @@
 from MarkdownPP.Transform import Transform
 
 # $...$ (or $$...$$)
-singlelinere = re.compile(r"\$(\$?)..*\$(\$?)")
+#
+# Keep detection broad, then apply a content heuristic before rendering to
+# avoid treating currency/prose as LaTeX. Escaped dollars ("\$") are ignored.
+singlelinere = re.compile(r"(?<!\\)\$(\$?).+?(?<!\\)\$(\$?)")
```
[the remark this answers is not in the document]
mike-finopsorg: deeper fix applied.

vendored/MarkdownPP/Modules/LaTeXRender.py
```diff
@@ -19,12 +19,17 @@
 from MarkdownPP.Transform import Transform
 
 # $...$ (or $$...$$)
-singlelinere = re.compile(r"\$(\$?)..*\$(\$?)")
+#
+# Keep detection broad, then apply a content heuristic before rendering to
+# avoid treating currency/prose as LaTeX. Escaped dollars ("\$") are ignored.
+singlelinere = re.compile(r"(?<!\\)(?<!\$)(\$\$(?!\$)|\$(?!\$))(.+?)(?<!\\)\1(?!\$)")
```
shawnalpay: @mike-finopsorg Take this concern from Gemini for what it's worth.  I'm way out of my depth here, but it seems like it's making a reasonable argument.

----

## ⚠️ Potential Issue: Regex "Swallowing" Prose/Currency

While the new currency handling is a great improvement, the current regex for `singlelinere` in `vendored/MarkdownPP/Modules/LaTeXRender.py` is a bit too broad. Because it uses a lazy match (`.+?`) between any two dollar signs without whitespace constraints, it can accidentally "swallow" large chunks of a sentence if a line contains multiple currency values.

### The Problem
If a contributor writes a line like:
> `The cost is $1.00 and the difference is $0.05.`

1. **The Regex Match:** The current pattern will match from the first `$` to the second `$`, capturing `1.00 and the difference is ` as the content.
2. **The Heuristic:** Your `should_render` function will correctly identify this string as prose (no LaTeX commands, contains words) and return `False`.
3. **The Result:** Because the regex has already "consumed" that part of the string, the rest of the line is ignored. If there were actual LaTeX math later on that same line, it would never be seen by the preprocessor.

### The Recommended Fix
Update the regex to enforce **no whitespace** immediately after the opening `$` and no whitespace immediately before the closing `$`. This aligns with how Pandoc and GitHub Flavored Markdown distinguish math from currency.

```python
# Updated regex with whitespace constraints: (?!\s) and (?<!\s)
singlelinere = re.compile(r"(?<!\\)(?<!\$)(\$\$(?!\$)|\$(?!\$))(?!\s)(.+?)(?<!\s)(?<!\\)\1(?!\$)")
```

### Why this is safer:
* **Currency:** `$ 1.00 $` or `$1.00 and $2.00` will no longer match the regex at all, leaving the prose untouched.
* **Math:** `$\alpha + \beta$` will still match perfectly.
* **Compatibility:** This makes the pre-processor behavior consistent with the Pandoc `gfm-tex_math_dollars` logic you've enabled in the Makefile.


```suggestion
# Keep detection broad, then apply a content heuristic before rendering to
# avoid treating currency/prose as LaTeX. Escaped dollars ("\$") are ignored.
# Also enforces no space immediately after the opening $ or before the closing $
singlelinere = re.compile(r"(?<!\\)(?<!\$)(\$\$(?!\$)|\$(?!\$))(?!\s)(.+?)(?<!\s)(?<!\\)\1(?!\$)")
```

shawnalpay (approved):

shawnalpay:
Bringing this forward to Member review!

vendored/MarkdownPP/Modules/LaTeXRender.py
```diff
@@ -36,6 +42,49 @@ class LaTeXRender(Module):
...
+        # Explicit LaTeX command is a strong signal.
+        if mathcommandre.search(content):
+            return True
+
+        has_digit = re.search(r"\d", content) is not None
+        word_count = len(wordre.findall(content))
+
+        # Numeric prose patterns such as "$4,380 / 12 months / 50 users" or
+        # "$0.02 < Tolerance $1.00" are common false positives.
+        if has_digit and word_count > 0 and not mathcommandre.search(content):
```
Matt-Cowsert:
```suggestion
        if has_digit and word_count > 0:
```

The `not mathcommandre.search(content)` guard is redundant here. If `mathcommandre.search(content)` were True, the function would have already returned True at line 71. Removing it makes the control flow easier to follow.

vendored/MarkdownPP/Modules/LaTeXRender.py
```diff
@@ -19,12 +19,18 @@
...
 # $...$ (or $$...$$)
-singlelinere = re.compile(r"\$(\$?)..*\$(\$?)")
+#
+# Keep detection broad, then apply a content heuristic before rendering to
+# avoid treating currency/prose as LaTeX. Escaped dollars ("\$") are ignored.
+# Also enforces no space immediately after the opening $ or before the closing $
+singlelinere = re.compile(r"(?<!\\)(?<!\$)(\$\$(?!\$)|\$(?!\$))(?!\s)(.+?)(?<!\s)(?<!\\)\1(?!\$)")
 # $... or ...$ (or $$... or ...$$)
-startorendre = re.compile(r"^\$(\$?)|^\S.*\$(\$?)$")
+startorendre = re.compile(r"^(?<!\\)\$(\$?)(?!\d)|^\S.*(?<!\\)\$(\$?)$")
```
Matt-Cowsert, line 28:
```suggestion
startorendre = re.compile(r"^\$(?!\d)(\$?)|^\S.*(?<!\\)\$(\$?)$")
```

The `(?<!\\)` lookbehind after `^` in the first alternative is a no-op since there is no preceding character at position 0. Removing it reduces visual noise without changing behavior.

Matt-Cowsert (approved):
Three-layer fix: MarkdownPP heuristics filter currency false positives, Pandoc `-tex_math_dollars` disables dollar-math parsing, and `&dollar;` entities are replaced with literal `$` across appendix files. Documentation added to editorial and MarkdownPP guidelines.

Two non-blocking suggestions below.

vendored/MarkdownPP/Modules/LaTeXRender.py
shawnalpay: Agree.

vendored/MarkdownPP/Modules/LaTeXRender.py
shawnalpay: Here is what Gemini says.  That said: @mike-finopsorg went back and forth with his LLM endlessly, and Regex like this is always a blackhole.  We're not even going to be using this for the time being.  Despite my suggestion from Gemini below, **I vote for no action**.

----

I agree that `(?<!\\)` immediately following the `^` anchor is mathematically useless and removing it cleans up the visual noise.

However, we need to be careful with where the `(?!\d)` lookahead ends up in the suggested replacement. By moving it *before* the optional second dollar sign `(\$?)`, it introduces a subtle bug with double dollar signs. 

If we use `^\$(?!\d)(\$?)` and evaluate a string like `$$5`:
1. The regex matches the first `$`.
2. It checks the negative lookahead `(?!\d)`. The next character is the *second* `$`, which is not a digit, so the lookahead **passes**.
3. It then matches the second `$` and incorrectly accepts the string.

To get the best of both worlds (removing the useless lookbehind while keeping the lookahead protecting both single and double dollar signs from digits), we should leave `(?!\d)` at the end of the group:

```suggestion
startorendre = re.compile(r"^\$(\$?)(?!\d)|^\S.*(?<!\\)\$(\$?)$")
```

ijurica:
All instances of `&dollar` in the specification content have been replaced. The `$` symbol now renders correctly across tables and prose text in both HTML and PDF outputs.

ijurica (approved):

vendored/MarkdownPP/Modules/LaTeXRender.py
shawnalpay: Discussed with @Matt-Cowsert.  Resolving with no action.  We can iterate in the future if necessary.

shawnalpay:
Three approvals and no open threads.  Marking as Ready for Approval!

jpradocueva (approved):
Looks good to me.

ljadvey (approved):

jpradocueva:
Approved by the group without objections or abstentions during the April 9 Members’ call.

PR #2374, opened 2026-05-13
AI #2373: Downgrade Feature Levels of eight `Cost and Usage` columns

### LINKED ISSUE #2335: [FR] Revise Feature Levels to make FOCUS more globally applicable
### Problem Statement
Currently, the FOCUS specification requires several columns to be strictly `Mandatory` (e.g., `HostProviderName`, `ListCost`, `ContractedCost`) that are inherently designed around complex cloud infrastructure billing models. When attempting to apply FOCUS to the broader technology ecosystem—such as flat-fee SaaS applications, managed services, or single-product on-premise solutions—data generators are unable to accurately fulfill these `Mandatory` fields without fabricating redundant or artificial data. This creates a structural barrier preventing non-cloud technology vendors from achieving strict FOCUS conformance. 

---

### Use Case / User Story
**As a** SaaS Vendor or Non-Cloud Data Generator
**I need to** be able to map my flat-fee or non-unit-based billing data to the FOCUS specification without failing conformance checks on inapplicable columns
**So that** I can provide standardized FinOps data to my customers without inventing artificial metrics.

**As a** FinOps Practitioner
**I need** FOCUS data to be universally applicable across all my technology vendors (SaaS, On-Prem, Cloud)
**So that** I can ingest and query all technology costs in a single normalized dataset, even if the vendor does not use unit pricing or contractual commitments.

---

### Success Criteria
* The baseline set of `Mandatory` columns represents the absolute minimum viable data points applicable to 100% of B2B technology transactions, regardless of the underlying billing model.
* A single-product, flat-fee SaaS vendor can successfully pass a FOCUS conformance audit without creating dummy data.
* Complex Cloud Service Providers are still required to provide detailed metrics (like `ListCost` and `ContractedCost`) via `Conditional` nullability rules.
* Existing data schemas do not break for practitioners (the columns remain present in the schema but become conditionally nullable for specific vendor profiles).

* [X] Criterion 1
* [X] Criterion 2
* [X] Criterion 3
* [X] Criterion 4

---

### Type of Request
* [ ] **Standardization** - Data already exists across data generators but is not yet normalized in FOCUS
* [X] **Enhancement** - Refinement or extension of an existing FOCUS column, metric, or capability

---

* **FinOps Foundation (Focus Working Group)** - Adoption Blocker - Leaving complex cloud-specific columns as strictly `Mandatory` blocks adoption for SaaS, hardware, and niche software vendors who cannot naturally produce that data.

---

* [X] **Partially** - Some data generators support this, others don't

**If yes or partially, please provide links to documentation:**

This request accommodates vendors whose documentation explicitly shows flat-fee, non-unit, or non-committed billing models (e.g., standard SaaS seat licenses without public unit price lists).

---

(See attached or linked proposal spreadsheet detailing the specific column downgrades/upgrades and nullability constraints.)

---

Re-evaluate the Feature Level of the following columns to make the strict `Mandatory` core universally applicable:

1. **Downgrade to Conditional:** `HostProviderName`, `ContractedCost`, `ListCost`.
   * *Approach:* Apply conditions such as "Required when the data generator supports unit pricing concepts" or "Required when data generator supports contract commitments." This holds CSPs accountable while freeing SaaS/On-Prem vendors.
2. **Upgrade to Conditional:** `ChargeFrequency`. 
   * *Approach:* "Required when the data generator supports recurring subscriptions or multiple billing frequencies."

---

**Current Workaround:**
Vendors without this data either fail conformance or hardcode artificial values (e.g., `PricingQuantity = 1`, `PricingUnit = License`, `ServiceCategory = Software`) to pass checks, which creates noisy, redundant data for practitioners.

---

> **STOP:** The sections below are for FOCUS Maintainers only. Requesters do not need to fill these out. **Scroll to the bottom and click "Create" to submit your request.**

---

## Maintainer Assessment

### Adoption Impact
* [ ] **Adoption Blocker** * [ ] **Adoption Accelerator** * [ ] **Practitioner Enhancement** * [ ] **Technical Improvement** **Justification:**

```
FOCUS Maintainers complete during triage.
```

---

**Feature Description:**

```
FOCUS Maintainers describe how this request advances FinOps capabilities or defines new supported features.
```

---

### Implementation Scope

**MVP Definition:**

```
FOCUS Maintainers define minimum implementation that provides value.
```

**North Star Vision:**

```
FOCUS Maintainers describe ideal complete solution.
```

**Phasing Strategy:**

```
FOCUS Maintainers determine if/how this should be implemented in phases.
```

---

### Specification Impact

**Impacted Parties:**
* [ ] FinOps Practitioner
* [ ] FOCUS Data Generator
* [ ] Vendor Supporting FOCUS
* [ ] Other: [specify]

**Level of Ambiguity:**

```
Rating and reasoning.
```

---

### FinOps Scope Alignment
* [ ] Public Cloud 
* [ ] Software-as-a-Service (SaaS)
* [ ] Data Center 
* [ ] Licensing
* [ ] AI 
* [ ] Custom 

## Community Support

* Add a **comment** below including:
  * Your **organization**
  * A **brief explanation** of why this is important to you (e.g., use case, urgency)
* FOCUS Staff & Maintainers will aggregate supporting orgs over time.

### LINKED ISSUE #2336: [AI] Present strawman for Feature Level revisions
### Detailed Description

_No response_

### Definition of Done

Generate and present spreadsheet to Maintainers of proposed changes to Feature Levels.

### LINKED ISSUE #2373: [AI] Draft PRs of Feature Level revisions and new column
### Detailed Description

_No response_

### Definition of Done

Draft PR of Feature Level downgrades.
Draft PR of Feature Level upgrades.
Draft PR introducing Data Generator Name.

shawnalpay:
## Summary

This pull request introduces the first phase of the **Universal Technology Cost Schema** proposal. It downgrades 8 structurally biased columns in the Cost and Usage dataset from `Mandatory` to `Conditional`. 

Historically, these columns were made Mandatory based on their prevalence in Public Cloud billing. However, as FinOps expands into SaaS, AI, and Internal IT, enforcing these columns forces data generators in those categories to emit "fake data" or empty strings. By tying these columns to specific **Presence Conditions**, we preserve 100% of the fidelity for Public Cloud datasets while allowing alternative business architectures (like token-based AI or flat-fee SaaS) to natively adopt the FOCUS standard.

For more information on the full proposal, see these documents:

- [Proposal: Revising FOCUS Feature Levels](https://docs.google.com/document/d/1nMV3z0Nf08pVPPwbamhmbElCg_Rq5nfZZBS1Wq7JkQQ/edit?tab=t.0#heading=h.iqphd0hlmtx7)
- [Cost and Usage Feature Levels and Conditions](https://docs.google.com/spreadsheets/d/1D6DaaxOVUGBVuAPMu1T2g-69ABkN6KuMvmaKL7Mgjkw/edit?gid=300003991#gid=300003991)
- [FOCUS Implementation Guide: Public Cloud](https://docs.google.com/document/d/14-pKbbaIL-YllfSBzbyrctdemDK52KdhuhhwDUJnIB4/edit?tab=t.0#heading=h.fnpwzk52x702)
- [FOCUS Feature Conditions Reference](https://docs.google.com/document/d/1yVsinNlL1MthJNHlEo1xGDUA2TfVF06OVLc5mi9BS8o/edit?tab=t.0)

## Changes Included
**1. Column Definition Updates (`Feature level` changed from `Mandatory` to `Conditional`):**
* `billingaccountname.md`
* `contractedcost.md`
* `hostprovidername.md`
* `invoiceissuername.md`
* `listcost.md`
* `pricingquantity.md`
* `pricingunit.md`
* `servicecategory.md`

**2. Dataset Table Updates (`dataset.md`):**
* Updated the Cost and Usage summary table to reflect the new `Conditional` feature levels for the 8 columns above.

**3. Requirement Rule Updates (`dataset.md`):**
* Appended the appropriate `when the *operating model*...` condition clauses to the requirement list, directly wiring the downgraded columns to their governing Presence Conditions:
  * `BillingAccountName` mapped to `includes billing account naming`
  * `HostProviderName` mapped to `includes host and service provider differences`
  * `InvoiceIssuerName` mapped to `includes payable invoices`
  * `ListCost` mapped to `publishes list unit prices`
  * `PricingQuantity` & `PricingUnit` mapped to `includes unit pricing`
  * `ServiceCategory` mapped to `includes a service hierarchy`

## Note to Reviewers
This PR isolates the **downgrades** only. The remaining proposals from the working group document (upgrading 3 columns from Recommended to Conditional, and introducing `Data Generator Name` as a new Mandatory column) will be submitted as separate PRs for isolated review.

### Type of Change
* [ ] Bug fix
* [ ] Editorial / Formatting

### Author Checklist


specification/datasets/cost_and_usage/dataset.md
```diff
@@ -146,14 +146,14 @@ CostAndUsage MUST adhere to the following requirements:
...
   * CostAndUsage MUST include [RegionName](#datasets.costandusage.regionname) when the *operating model* [includes regions](#conditions.includesregions).
   * CostAndUsage MUST include [ResourceId](#datasets.costandusage.resourceid) when the *operating model* [includes provisioned resources](#conditions.includesprovisionedresources).
   * CostAndUsage MUST include [ResourceName](#datasets.costandusage.resourcename) when the *operating model* [includes provisioned resources](#conditions.includesprovisionedresources).
   * CostAndUsage MUST include [ResourceType](#datasets.costandusage.resourcetype) when the *operating model* [includes provisioned resources](#conditions.includesprovisionedresources) and [includes resource type assignment](#conditions.includesresourcetypeassignment).
-  * CostAndUsage MUST include [ServiceCategory](#datasets.costandusage.servicecategory).
+  * CostAndUsage MUST include [ServiceCategory](#datasets.costandusage.servicecategory) when the *operating model* [includes a service hierarchy](#conditions.includesservicehierarchy).
```
Matt-Cowsert: Adding context from the 2026-05-11 Maintainers call where this downgrade was discussed:

* Multiple maintainers preferred `ServiceCategory` to stay Mandatory and revisit the allowed-values list instead of conditioning the column. The concern was that cross-provider comparability (a core practitioner value) degrades when the spec's primary categorization column becomes optional.
* An alternative gating proposal was raised: mandatory when the operating model supports multiple services, rather than mandatory when a categorization hierarchy exists.
* All participants agreed the allowed-values list needs to be revisited regardless of the Mandatory/Conditional decision.
* No consensus was reached; the action item was for maintainers to review the proposal and provide column-specific feedback.

Separately, the condition `IncludesServiceHierarchy` is defined as evaluating true "when the operating model supports a hierarchy of service categories (e.g., Service, Sub-service)." That gates on whether *multiple levels* exist. But `ServiceCategory` is itself a single level; the second level (`ServiceSubcategory`) is the one whose presence would track a hierarchy.

Two paths worth weighing:

1. Keep `ServiceCategory` Mandatory (consistent with the call discussion) and apply this condition to `ServiceSubcategory` instead. Pair with the allowed-values revision the maintainers agreed to. ServiceCategory has 19 spec-defined allowed values including "Other" as a catch-all, so every service maps to exactly one. Cross-provider analysis depends on ServiceCategory existing uniformly.
2. Or change the condition to gate on "operating model supports multiple services" (the alternative raised on the call) or "operating model includes service categorization at all," rather than "includes a service hierarchy." The allowed-values expansion still needs to happen either way.

As drafted, `IncludesServiceHierarchy` gates a column it doesn't logically govern, and the open allowed-values question from the call hasn't been resolved.

Matt-Cowsert (changes requested):
The downgrade is internally consistent and the condition wiring is sound for seven of the eight columns. `BillingAccountName`, `ContractedCost`, `HostProviderName`, `InvoiceIssuerName`, `ListCost`, `PricingQuantity`, and `PricingUnit` each map cleanly to a condition that tracks an operating-model dimension. The "HostProviderName MUST match ServiceProviderName in all other cases" pattern is the fake-data anti-pattern this proposal is designed to fix.

One concern on `ServiceCategory` (inline on the requirements line): the 2026-05-11 Maintainers call ended without consensus on this specific downgrade. Multiple maintainers preferred to keep `ServiceCategory` Mandatory and revisit the allowed-values list instead, and an alternative gating condition ("mandatory when the operating model supports multiple services") was raised but not resolved. The condition currently wired (`IncludesServiceHierarchy`) also has a mapping issue: it gates on whether multiple levels of categorization exist, but `ServiceCategory` is the single top level. Worth surfacing back to that open thread.

One downstream reference goes stale with this change: `guidelines/data-generators/dataset-mapping-documentation.md` uses `ListCost` as the "Formula Mapping" example with Feature Level = `Mandatory`. After this PR lands, that cell should read `Conditional`. Your call whether to fix inline here or sweep in a follow-up docs PR.

udam-stitcher:
I'm surprised to see the group go down this route, so I may be missing something.

What this list is based on. What datasets were used to determine these as needing to be conditional? I missed last week's maintainers meeting due to travel. On the surface, this change seems misguided IMO.

(Details below) 6 of the 8 here seem like very necessary things, regardless of the dataset or provider is, and are very standard pieces of data that any provider should be able to provide, in order to enable FinOps use cases. 

IMO, the goal should not be to force some data, regardless of how little value they provide downstream, into FOCUS, but rather to guide data generators on which columns are most important for downstream FinOps use cases. If not, it would seem that someone could argue that the date columns and BilledCost are the only necessary data.

## Cost metrics:
listcost
contractedcost

Cost metric must be aggregateable across datasets. Unless we somehow believe that these concepts don't apply to non-cloud datasets (which is absolutely not the case), we should leave these in. Every offering out there has a List price, no matter what it is. Even if there were no concept of a negotiated discount, that would simply set the dataset's contracted cost equal to the list cost, which is easy for the data generator to do. Similarly, if there are no prepayments, the BilledCost will equal the List and Contracted. Otherwise, with that same argument, EffectiveCost should also be made conditional? since many providers or some customers don't have/use prepayments.


## Basic fields available in any Invoice:
invoiceissuername.md
pricingquantity.md
pricingunit.md

These are simply the most commonly available pieces of information across all cloud, SaaS, and AI I've looked at, and even in the most aggregated billing scenarios (invoice-based billing), they still show up. What benefit would it provide for practitioners of FOCUS data to not know what pricing quantity and unit were (which also correspond to unit price)? Even the most basic reconciliation use cases would depend on some level of validation that the cost of a charge is based on the quantity you're being charged for, at some rate. 

## Highest level of categorization for reporting
servicecategory.md

Without this, organizations would simply have to do manual mapping to categorize their costs. This column gives providers a simple way to guide downstream users in categorizing aggregated cost data (e.g., how much of my spending is for AI vs. databases vs. something else). Seems like low effort to categorize for SaaS/AI or anyone else for that matter, and provides huge downstream value. 

## This is a concept that's prevalent in most vendors, regardless of Cloud/SaaS/AI
billingaccountname.md

This column allows nulls, but making it mandatory was intended to ensure vendors understand the value of a human-readable/recognizable identifier for the highest-level accounts that incur costs. Low effort (especially since nullable), high value downstream.

## Fine to downgrade (probably should have been Conditional to begin with)
hostprovidername.md

Agree, this is mostly a Cloud/SaaS construct, and conditional seems fine here. 

Let me know if I'm off base here.

shawnalpay:
@udam-stitcher 
Thanks for taking the time to write up this feedback. Before diving in, I want to reference the proposal documents I linked in the parent post, which we discussed at length in last week's Maintainers meeting, as they provide the full context and matrices for how this logic is applied:

- [Proposal: Revising FOCUS Feature Levels](https://docs.google.com/document/d/1nMV3z0Nf08pVPPwbamhmbElCg_Rq5nfZZBS1Wq7JkQQ/edit?tab=t.0#heading=h.iqphd0hlmtx7)
- [Cost and Usage Feature Levels and Conditions](https://docs.google.com/spreadsheets/d/1D6DaaxOVUGBVuAPMu1T2g-69ABkN6KuMvmaKL7Mgjkw/edit?gid=300003991#gid=300003991)
- [FOCUS Implementation Guide: Public Cloud](https://docs.google.com/document/d/14-pKbbaIL-YllfSBzbyrctdemDK52KdhuhhwDUJnIB4/edit?tab=t.0#heading=h.fnpwzk52x702)
- [FOCUS Feature Conditions Reference](https://docs.google.com/document/d/1yVsinNlL1MthJNHlEo1xGDUA2TfVF06OVLc5mi9BS8o/edit?tab=t.0)

I completely agree with your underlying goal here: we absolutely cannot let data generators off the hook for critical FinOps data. We both want the highest-fidelity downstream data possible. 

Let's name the rub up front: **my proposal here is to move away from our existing pattern where we ask data generators to fill in columns that are not applicable to their operating model**. I believe that FOCUS' future lies in its ability to have each Technology Category natively supported by FOCUS, not asking them to fit into a pattern that biases towards Public Cloud. That's an uncomfortable shift to propose, I know, but I believe it to be the most extensible one.  

The premise of your feedback relies on the idea that forcing a column to be present forces the data to be valuable. When we force a Mandatory column onto a business model where that concept doesn’t natively exist, we force the generator to emit artificial data, and I believe that this actively pollutes our FinOps reporting.

I should also say that `Conditional` != `Recommended`.  Downgrading to "Conditional" does *not* mean these fields become optional for relevant Technology Categories. Because of the Presence Conditions, if a vendor has a rate-card or an invoice, these columns act as strict Mandatory fields for them. This proposal simply ensures that an internal IT ledger or an open-source scraper isn't forced to invent an "Invoice Issuer" or a "Contracted Cost" for a transaction that never had an invoice or a contract to begin with. ALL of these would still be applicable to Public Cloud providers. In the links above, you can see my first stab at a proposed "Implementation Guide" for Public Cloud, which would make clear that, for the typical Public Cloud provider, all of these continue to be effectively Mandatory.

**1. Cost Metrics (`ListCost`, `ContractedCost`)**
* **Inapplicable Sectors:** Internal IT / Showback (e.g., Apptio, ServiceNow), Outcome-based FinOps tooling (e.g., ProsperOps), Open-Source Infrastructure.
* Regarding the premise that "Every offering out there has a List price..." — this is not a universally applicable truth across all Technology Categories. Consider internal cost allocation as a percentage, outcome-based pricing for tooling vendors, or bespoke professional services. There is no concept of a public MSRP in these cases, and thus a `ListCost` is not applicable.
* Regarding `ContractedCost`: if this is to remain a Mandatory column, then we need to consider harmonizing the Feature Level of all contract-related columns.  Below is the current state of affairs.  Why would `ContractedCost` be Mandatory, but not `ContractedUnitPrice`?
  - CostAndUsage MUST include ContractApplied when the service provider supports contract commitments.
  - CostAndUsage MUST include ContractedCost.
  - CostAndUsage MUST include ContractedUnitPrice when the service provider supports negotiated pricing concepts.

**2. Invoicing and Units (`InvoiceIssuerName`, `PricingQuantity`, `PricingUnit`)**
* **Inapplicable Sectors:** Open-source telemetry scrapers (e.g., Kubecost), Internal ITSM Platforms, Milestone or Flat-Tier Enterprise Software.
* You noted these show up in all Cloud/SaaS/AI invoices. But FOCUS is a *Universal Technology* schema, not just a *commercial vendor* schema. Scrapers pulling from a local cluster do not issue invoices. Internal IT chargeback engines do not issue invoices.
* I should note that I proposed downgrading `InvoiceIssuerName` in part to provoke discussion of the current schema asymmetry: if `Invoice Issuer Name` is universally Mandatory, why is `Invoice ID` currently Conditional?  Whatever the outcome of this proposal, do you agree that those columns should have the same Feature Level?  And if not, why not?
* Regarding units (and this is related to the `ListCost` discussion above): Forcing a $100,000 flat-fee enterprise site license to hardcode "Quantity: 1" and "Unit: Each" does not provide unit-economic value; it's arbitrary filler. In FOCUS 1.4, we analyzed a series of real-world invoices, and the vast majority of them did NOT include Pricing Quantity or Pricing Unit.  

**3. Service (`ServiceCategory`)**
* I will admit that I proposed this downgrade based on the fact that the current structure is heavily biased towards Public Cloud.  Where is hardware?  Where is telco?  Facilities? Putting all of these in `Other` is an inaccurate representation of relevant providers.  I have proposed PR #2379 to refresh `ServiceCategory`, and I'd be happy to get onboard with retaining this as `Mandatory` if we can agree on a list that feels applicable across the entire technology landscape.

**4. Account (`BillingAccountName`)**
* **Inapplicable Sectors:** Open-source cluster scrapers, Single-Tenant SaaS, Self-serve developer tiers (where friendly account alias hierarchies don't natively exist).
* It sounds like you're saying that this column was introduced with an explicit understanding that some providers would not have such a structure, but FOCUS would encourage them to do so -- is that right?  And if so, are we sure we wish to retain that gap, especially given that FOCUS has now moved beyond Public Cloud?  I think providing a BillingAccountId is still critical, but that it would be reasonable to make this conditional on the provider's support of account naming.

This topic is on the docket for May 18 Maintainers today, and the plan is to go through each of these in detail.  Look forward to discussing further.

specification/datasets/cost_and_usage/dataset.md
shawnalpay: @Matt-Cowsert 
Based on the May 11 Maintainers conversation, I have issued PR #2379 for augmenting the allowed values in ServiceCategory.  If we can align on a set of allowed values that feels applicable across all technology providers, I'd be happy to retain it as Mandatory in this PR.

Re: ServiceSubcategory: I have indeed applied this condition to that column via PR #2375 -- have a look there.

ijurica:
### Single-value conditions 

I’m struggling with the idea that, in cases where only a single value exists, providers may not need to explicitly expose it.

#### North Star

I understand this would represent **a material change**, but my preference would be that **providers always expose the value explicitly, even when only a single value exists**, to avoid practitioners needing to infer it from documentation. This applies both at the feature design level and in terms of nullability handling.

#### Design Principle

This pattern is applied to a subset of dimensions, but we don’t currently have a design principles that defines the criteria for which dimensions this should apply to. 

#### Service Category

Regardless of the direction we take, I would not expect Service Category to be included in this set.

> Includes Service Hierarchy: ... the operating model includes more than one level of service categorization

I believe Service Category should remain mandatory. While I agree we should revisit the allowed values, I would prefer that we continue requiring providers to map to one of the existing categories (with the option for consumers to override if needed), rather than making this dimension conditional.

### Default values in mixed-provider datasets

At a higher level, I’m not sure whether the specification should primarily:

a) optimize for providers (e.g., introduce additional conditions that simplify individual provider datasets), or  
b) optimize for practitioners who frequently merge datasets across multiple providers into a unified model.

In practice, both concerns are valid, but they lead to different design trade-offs. Personally, I’m generally leaning more toward option b), since mixed-provider datasets seem likely to be a very common downstream use case.

Regardless of the direction, I think we should explicitly provide guidance for mixed-provider datasets, since this is a common downstream use case and introduces non-trivial handling of conditional columns and defaults.

When data is combined across multiple providers into a unified dataset, some conditional columns may be present for only a subset of providers depending on whether the underlying concept is supported.

For a set of key dimensions and metrics that are modeled as conditional, we may need to explicitly define default values that apply regardless of whether those values are provided by data generators or derived by consumers.

**Example guidance for default values**

- If unit pricing concept-related columns are not included:
  - Pricing Unit: based on `UnitFormat`, default to Units or Count (at least that'd my interpretation)
  - List/Contracted Unit Price: default to 1
  - Pricing Quantity: derived such that Cost = Price × Quantity

- If pricing is percentage-based:
  - PricingUnit: Percentage (decimal representation)
  - PricingQuantity: Base amount on which the percentage is applied
  - UnitPrice: Percentage expressed as a decimal (e.g., 0.30 = 30%)
  - Cost = UnitPrice × Quantity

- If list/public pricing or contracted pricing-related columns are not included (e.g., for savings reporting based on negotiated discounts):
   ***Note:** We need to revisit Contracted Cost and Contracted Unit Price definitions!!!*
  - List Price defaults to Contracted Price (or vice versa, depending on availability)

jpradocueva (approved):
From the perspective of the requirements model, these changes are valid and well-formed.

jpradocueva (commented):

Snapshot totals: 10 conversation comments, 3 reviews, and 2 inline review comments.

| --- | --- | --- |
| F-001 | Submit inline comment | The first Billing Account glossary occurrence is unlinked; correcting the link does not change schema or normative behavior. |
| F-002 | Submit inline comment | The Operating Model glossary link omits required emphasis; correcting it does not change schema or normative behavior. |
| F-003 | Submit inline comment | The first Billing Account glossary occurrence is unlinked; correcting it does not change schema or normative behavior. |
| F-004 | Submit inline comment | The Operating Model glossary link omits required emphasis; correcting it does not change schema or normative behavior. |
| F-005 | Submit inline comment | The Operating Model glossary link omits required emphasis; correcting it does not change schema or normative behavior. |
| F-006 | Submit inline comment | The first Service glossary occurrence is unlinked; correcting it does not change schema or normative behavior. |
| F-007 | Submit inline comment | BillingAccountName has contradictory Feature Levels across schema representations; synchronizing the catalog does not change normative behavior. |

specification/conditions/conditions_overview.md
```diff
@@ -10,14 +10,17 @@ Conditions are sorted alphabetically by name.
 | ------------------ | ------------- | ----------- |
 | [Includes Aggregation Level Currency Differences](#conditions.includesaggregationlevelcurrencydifferences) | Billing | Operating model includes the representation of billing currency and payment currency at different aggregation levels on payable invoices. |
 | [Includes Availability Zones](#conditions.includesavailabilityzones) | Resource | Operating model includes deploying resources or services within an availability zone. |
-| [Includes Billing and Payment Currency Differences](#conditions.includesbillingpaymentcurrencydifferences) | Billing | Operating model includes billing and payment in different currencies. |
+| [Includes Billing Account Naming](#conditions.includesbillingaccountnaming) | Account | Operating model includes assigning a display name for the billing account. |
```
jpradocueva, line 13: ### F-001 — error — [Line 13](https://github.com/FinOps-Open-Cost-and-Usage-Spec/FOCUS_Spec/blob/0ee3c50cc3ec88b6d24d854291a93bec4983b5f1/specification/conditions/conditions_overview.md#L13)

**Changed content:** `Operating model includes assigning a display name for the billing account.`

**Rule:** `editorial-conventions-linking-rule-06` — Writing Specification Content & Review Guidelines > Editorial Conventions > Linking Rule; position 6. “Link the first remaining occurrence in reading order.”

**Engineering context:** Billing account is a defined FOCUS glossary concept, and this is its first non-heading occurrence in the file.

**Problem:** The phrase is not linked to `#glossary:billing-account`.

**Impact:** Readers lose the required first-occurrence path to the concept’s precise definition.

**Suggested correction:** Link the glossary term.

```suggestion
| [Includes Billing Account Naming](#conditions.includesbillingaccountnaming) | Account | Operating model includes assigning a display name for the [*billing account*](#glossary:billing-account). |
```

specification/conditions/includesbillingaccountnaming.md
```diff
@@ -0,0 +1,26 @@
+# Includes Billing Account Naming
+
+The Includes Billing Account Naming condition represents a verifiable state where the [operating model](#glossary:operating-model) includes assigning a display name for the billing account.
```
jpradocueva, line 3: ### F-002 — error — [Line 3](https://github.com/FinOps-Open-Cost-and-Usage-Spec/FOCUS_Spec/blob/0ee3c50cc3ec88b6d24d854291a93bec4983b5f1/specification/conditions/includesbillingaccountnaming.md#L3)

**Changed content:** `[operating model](#glossary:operating-model)`

**Rule:** `editorial-conventions-glossary-terms-05` — Writing Specification Content & Review Guidelines > Editorial Conventions > Glossary terms; position 5. “Link with `[*term*](#glossary:term)` format.”

**Engineering context:** Glossary links italicize their anchor text to identify defined FOCUS concepts.

**Problem:** The destination is correct, but the anchor text is not italicized.

**Impact:** The new definition violates the required glossary-link syntax.

**Suggested correction:** Use `[*operating model*](#glossary:operating-model)`.

jpradocueva, line 3: ### F-003 — error — [Line 3](https://github.com/FinOps-Open-Cost-and-Usage-Spec/FOCUS_Spec/blob/0ee3c50cc3ec88b6d24d854291a93bec4983b5f1/specification/conditions/includesbillingaccountnaming.md#L3)

**Changed content:** `assigning a display name for the billing account`

**Rule:** `editorial-conventions-linking-rule-06` — Writing Specification Content & Review Guidelines > Editorial Conventions > Linking Rule; position 6. “Link the first remaining occurrence in reading order.”

**Engineering context:** Billing account is a defined FOCUS glossary concept.

**Problem:** Its first non-heading occurrence is unlinked.

**Impact:** Readers are not directed to the specification’s definition.

**Suggested correction:** Use `[*billing account*](#glossary:billing-account)`.

specification/conditions/includeshostserviceproviderdifferences.md
```diff
@@ -0,0 +1,26 @@
+# Includes Host-Service Provider Differences
+
+The Includes Host-Service Provider Differences condition represents a verifiable state where the [operating model](#glossary:operating-model) includes public delineation between host and service provider.
```
jpradocueva, line 3: ### F-004 — error — [Line 3](https://github.com/FinOps-Open-Cost-and-Usage-Spec/FOCUS_Spec/blob/0ee3c50cc3ec88b6d24d854291a93bec4983b5f1/specification/conditions/includeshostserviceproviderdifferences.md#L3)

**Changed content:** `[operating model](#glossary:operating-model)`

**Rule:** `editorial-conventions-glossary-terms-05` — Writing Specification Content & Review Guidelines > Editorial Conventions > Glossary terms; position 5. “Link with `[*term*](#glossary:term)` format.”

**Engineering context:** Glossary links italicize the defined FOCUS term.

**Problem:** The link omits the required emphasis.

**Impact:** The new definition violates the prescribed glossary syntax.

**Suggested correction:** Use the required glossary markup.

```suggestion
The Includes Host-Service Provider Differences condition represents a verifiable state where the [*operating model*](#glossary:operating-model) includes public delineation between host and service provider.
```

specification/conditions/includesmultipleservices.md
```diff
@@ -0,0 +1,26 @@
+# Includes Multiple Services
+
+The Includes Multiple Services condition represents a verifiable state where the [operating model](#glossary:operating-model) includes multiple services.
```
jpradocueva, line 3: ### F-005 — error — [Line 3](https://github.com/FinOps-Open-Cost-and-Usage-Spec/FOCUS_Spec/blob/0ee3c50cc3ec88b6d24d854291a93bec4983b5f1/specification/conditions/includesmultipleservices.md#L3)

**Changed content:** `[operating model](#glossary:operating-model)`

**Rule:** `editorial-conventions-glossary-terms-05` — Writing Specification Content & Review Guidelines > Editorial Conventions > Glossary terms; position 5. “Link with `[*term*](#glossary:term)` format.”

**Engineering context:** Glossary links must italicize their anchor text.

**Problem:** The `operating model` link omits the required emphasis.

**Impact:** The new condition definition violates the glossary-link convention.

**Suggested correction:** Use `[*operating model*](#glossary:operating-model)`.

jpradocueva, line 3: ### F-006 — error — [Line 3](https://github.com/FinOps-Open-Cost-and-Usage-Spec/FOCUS_Spec/blob/0ee3c50cc3ec88b6d24d854291a93bec4983b5f1/specification/conditions/includesmultipleservices.md#L3)

**Changed content:** `includes multiple services`

**Rule:** `editorial-conventions-linking-rule-06` — Writing Specification Content & Review Guidelines > Editorial Conventions > Linking Rule; position 6. “Link the first remaining occurrence in reading order.”

**Engineering context:** Service is a defined FOCUS glossary concept.

**Problem:** Its first non-heading occurrence is unlinked.

**Impact:** Readers are not directed to the precise definition used by the condition.

**Suggested correction:** Use `[*services*](#glossary:service)`.

specification/datasets/cost_and_usage/dataset.md
```diff
@@ -98,7 +98,7 @@ CostAndUsage MUST adhere to the following requirements:
   * CostAndUsage SHOULD include [AvailabilityZone](#datasets.costandusage.availabilityzone) when the *operating model* [includes availability zones](#conditions.includesavailabilityzones).
   * CostAndUsage MUST include [BilledCost](#datasets.costandusage.billedcost).
   * CostAndUsage MUST include [BillingAccountId](#datasets.costandusage.billingaccountid).
-  * CostAndUsage MUST include [BillingAccountName](#datasets.costandusage.billingaccountname).
+  * CostAndUsage MUST include [BillingAccountName](#datasets.costandusage.billingaccountname) when the *operating model* [includes billing account naming](#conditions.includesbillingaccountnaming).
```
jpradocueva, line 101: ### F-007 — error — [Line 101](https://github.com/FinOps-Open-Cost-and-Usage-Spec/FOCUS_Spec/blob/0ee3c50cc3ec88b6d24d854291a93bec4983b5f1/specification/datasets/cost_and_usage/dataset.md#L101)

**Changed content:** `CostAndUsage MUST include BillingAccountName ... when the operating model includes billing account naming.`

**Rule:** `validation-schema-accuracy-mathematical-schema-accuracy-01` — Writing Specification Content & Review Guidelines > Validation & Schema Accuracy > Mathematical & Schema Accuracy; position 1. “AI reviewers MUST rigorously calculate, parse, and verify all data within examples (especially JSON snippets and tables). Flag any mathematical inconsistencies or hallucinated data.”

**Engineering context:** The dataset catalog is a schema summary whose Feature Level should agree with the detailed column constraint and normative presence rule.

**Related locations:** Dataset catalog line 19; `billingaccountname.md` line 33.

**Problem:** The requirement and column definition make BillingAccountName Conditional, but the dataset catalog still reports Mandatory.

**Impact:** Implementers receive contradictory presence obligations depending on which schema representation they read.

**Suggested correction:** Change the Billing Account Name catalog row to `[Conditional](#conditions.includesbillingaccountnaming)`.

specification/conditions/includesbillingaccountnaming.md
jpradocueva, line 3:
```suggestion
The Includes Billing Account Naming condition represents a verifiable state where the [*operating model*](#glossary:operating-model) includes assigning a display name for the billing account.
```

specification/conditions/includesmultipleservices.md
jpradocueva, line 3:
```suggestion
The Includes Multiple Services condition represents a verifiable state where the [*operating model*](#glossary:operating-model) includes multiple services.
```

specification/conditions/includeshostserviceproviderdifferences.md
```diff
@@ -0,0 +1,26 @@
+# Includes Host-Service Provider Differences
```
flanakin, line 1: nit: I'm assuming this was discussed, but this hyphen feels incorrect. Also, our column is Host Provider Name, so should this be "Host Provider Differences"?

flanakin (approved):

PR #2360, opened 2026-05-08
AI #2359: Add `Principal ID` column to `Cost and Usage` dataset

shawnalpay:
## Summary

This PR introduces the `PrincipalId` column to the Cost and Usage dataset, identifying the `principal`: an entity defined in an identity and access management model (e.g., user, service account) to which access to resources or services is granted. `PrincipalId` is commonly used to report and audit cost by the `principal` that incurred a charge, preserving a principal-level audit trail across PaaS, SaaS, and GenAI billing.

Per the TF2 recommendation, the actor columns are delivered incrementally, starting with `PrincipalId`. Two PRs are stacked on this branch. #2553 delivers `CredentialId` and `RequesterDetails` under the same presence condition and retargets `working_draft` once this PR merges; sequencing was raised at the Maintainers call on 2026-08-10 and settled on 2026-09-08. Draft #2495 holds the `ConsumerId` work carved out on 2026-06-30, so it is not lost and can be picked up after `PrincipalId` lands.

### Key Changes
* **New Column:** `PrincipalId`, an identifier of the `principal` to which access to a resource or service is granted.
* **Definition Anchor:** `Principal` is defined by reference to the standard identity and access management principal rather than a FOCUS-specific abstraction, and is distinguished from the credential presented with an individual request and from the organizational container within which the request runs. The definition names no particular provider's model, so it holds for a marketplace purchase and for single sign-on, where the identity provider and the charged service differ.
* **New Condition:** `IncludesRequesterAttribution`, keyed on whether the operating model attributes usage or purchase of resources or services to requesters, with the dataset-level presence requirement for `PrincipalId` wired to it. Keying it on the requester rather than the principal means a data generator that publishes credential identifiers but cannot resolve a principal still evaluates it to true. `CredentialId` and `RequesterDetails` share the condition on #2553. It sits under `Account` in the operating model conditions overview, the closest of the existing categories.
* **Nullability:** `PrincipalId` is null when a charge is not associated with a principal, and not null when it is. The requirement turns on the association rather than on whether the service provider can determine the principal.
* **Uniqueness:** `PrincipalId` is a unique identifier within the service provider, the form `ResourceId`, `CapacityReservationId`, and `CommitmentDiscountId` use.
[...]

Issue #2359, opened 2026-05-08
[AI] Draft PR proposing the addition of Consumer ID and Principal ID


shawnalpay:
### Definition of Done

Draft PR proposing the addition of Consumer ID to Cost and Usage.

shawnalpay:
I completed #2360.  There's still much to discuss, but it's now out of draft.  Closing this PR.

PR #2401, opened 2026-05-20
AI #2359 - Ext. 1: Align Actor Attribution examples with adopted conventions

ijurica:
## Summary

This PR is an extension of PR #2360, focusing on refinements and additions to the illustrative scenarios in the appendix section.

### Overall Rationale/Guideline

The proposed changes aim to align the appendix examples with a consistent interpretation of what Principal and Consumer represent, and under which circumstances either identifier can or cannot be resolved.

Recognizing that presence and nullability requirements remain under discussion, the examples focus on whether actor identities can be determined from the transaction context itself, while aiming to make the rationale for null values explicit wherever possible, rather than reflecting scenarios where the information is withheld or omitted for implementation-specific reasons or provider/customer choice (cardinality, supported scopes, dual opt-in, etc.).

#### 1. Editorial updates in Scenarios 2 and 5

- Added explicit `(PrincipalId)` and/or `(ConsumerId)` references for consistency across scenarios in this appendix.

##### Direct IaaS

> 3. **Direct IaaS:** A data engineer at **AeroScale** directly provisions an Aura Web compute cluster. Because the infrastructure is consumed directly by the principal, no downstream application actor exists.

- The original Direct IaaS scenario described a data engineer directly provisioning and consuming a compute cluster, with `ConsumerId` set to null. According to the glossary definition, a consumer is a downstream actor that consumes resources or services. If the principal directly consumes the resource, the principal is also the consumer, and `ConsumerId` should therefore be populated (as in Scenario 5), rather than null.

##### Network Edge Processing

> 3. **Network Edge Processing:** At **Acme Corp**, traffic routed through an Aura Web edge network is authenticated at the gateway (`PrincipalId`), but downstream consumers (`ConsumerId`) are unknown due to aggregated network traffic.

- This scenario was added to illustrate a case where the request initiator (`PrincipalId`) is known, while downstream consumers (`ConsumerId`) are null not due to a design choice to suppress or limit data for cardinality reasons, but because the consumer identity cannot be resolved from the available data.

##### Original

> 4. **Seat-Based SaaS:** **Acme Corp** pays for SprintCanvas project management licenses. The provider does not attribute an infrastructure actor, but the opaque `ConsumerId` allows safe user-level chargeback.

##### Replacement

> 4. **Seat-Based SaaS:** **Acme Corp** pre-purchases a number of SprintCanvas seats assigned to individual users. Since the entitlement is established through a contractual agreement, there is no originating requestor (`PrincipalId` is `null`), while the opaque identifier allows safe user-level chargeback (`ConsumerId`).

##### Rationale

- The original wording referenced an infrastructure/application actor distinction that no longer aligns with the updated glossary terminology.
- The scenario illustrates a case where `ConsumerId` can be resolved for user-level attribution and chargeback, while the record is derived from a contractual agreement rather than an explicit request, resulting in no originating requestor (`PrincipalId` is `null`).
- This aligns with the broader goal of making the rationale for populated versus null actor identifiers more explicit across all appendix examples.

#### 4. Added Billing System Charge Example(s)

> 6. **Billing System Charge:** **Acme Corp** receives a promotional credit or is charged applicable tax from LatticeScale. Since the charge is generated by the provider's billing system rather than initiated by a request for a resource or service, the principal cannot be resolved (`PrincipalId` is `null`). As the charge applies at the account level, it cannot be associated with a specific end-user (`ConsumerId` is `null`).

- Tax and Credit scenarios illustrate charges generated by the provider's billing system rather than by a request for a resource or service, resulting in `PrincipalId` being `null`.
- As the charge is account-level and not attributable to a specific end-user, `ConsumerId` is also `null`.

### Type of Change
* [ ] Bug fix
* [ ] Editorial / Formatting

### Author Checklist


specification/appendix/actor_attribution.md
shawnalpay, line 1: @ijurica 
Great suggestion to explicitly map `(PrincipalId)` and `(ConsumerId)` directly in the text... it makes the examples read so much clearer. Introducing the out-of-band billing system charges (Credits/Taxes/Corrections) is also a great way to explain *why* both columns might legitimately be `null`.

I suggest a couple of minor structural tweaks to keep the matrix punchy:

* **Combined Tax & Credit:** Since both scenarios result in the exact same table output (`null` / `null` based on the same billing system logic), let's merge them into a single "Billing System Charge" row. 
* **Restored the SaaS Scenario:** I moved the SaaS example back to a single row. Splitting it into "Purchase" vs. "Usage" starts bleeding a bit too much into transaction-lifecycle logic, which distracts from the core goal of the appendix. More importantly, we need to preserve that classic "pure SaaS" use case where the provider gives us the downstream user (`ConsumerId`) but leaves the requesting actor empty (`PrincipalId` = `null`). 

See below.

1. **Generative AI API:** **Acme Corp** uses an internal bot to summarize notes via a LatticeScale API. Security audits the service account (`PrincipalId`), while FinOps allocates token costs to the employee (`ConsumerId`).
2. **Multi-Tenant PaaS:** A shared BI engine (`PrincipalId`) runs a query on OmniQuery for a specific client of **GearPeak Outdoors**. Capturing the downstream client (`ConsumerId`) prevents costs from pooling in overhead.
3. **Direct IaaS:** A data engineer at **AeroScale** directly provisions an Aura Web compute cluster (`PrincipalId`). Because the infrastructure is consumed directly by the principal, no downstream application actor exists, leaving the consumer omitted (`ConsumerId` is null).
4. **Seat-Based SaaS:** **Acme Corp** pays for SprintCanvas project management licenses. The provider does not attribute an infrastructure actor (`PrincipalId` is null), but the opaque identifier allows safe user-level chargeback (`ConsumerId`).
5. **Direct PaaS Usage:** A data scientist at **Acme Corp** logs into a LatticeScale managed notebook using their individual SSO credential (`PrincipalId`). Because the user authenticated directly to the platform and is the sole consumer of the compute (`ConsumerId`), the same opaque identifier is populated in both columns.
6. **Billing System Charge:** **Acme Corp** receives a promotional credit or is charged applicable tax from LatticeScale. Since the charge is generated by the provider's billing system rather than initiated by a request for a resource or service, the principal cannot be resolved (`PrincipalId` is null). As the charge applies at the account level, it cannot be associated with a specific end-user (`ConsumerId` is null).

| Scenario | Technology Category | Data Generator | PrincipalId | ConsumerId |
| :--- | :--- | :--- | :--- | :--- |
| 1 | **Generative AI API** | LatticeScale | `svc-acme-docbot-prod` | `emp_hash_84729x` |
| 2 | **Multi-Tenant PaaS** | OmniQuery | `svc-bi-reporting-engine` | `client_gearpeak_001` |
| 3 | **Direct IaaS** | Aura Web | `user_d_engineer_993` | `null` |
| 4 | **Seat-Based SaaS** | SprintCanvas | `null` | `user_uuid_554321` |
| 5 | **Direct PaaS Usage** | LatticeScale | `dev_uuid_112233` | `dev_uuid_112233` |
| 6 | **Billing System Charge** | LatticeScale | `null` | `null` |

Matt-Cowsert (commented):
Thanks Irena, the explicit `(PrincipalId)` / `(ConsumerId)` annotations make the scenarios much easier to follow, and the billing-system charges (Correction, Credit, Tax) are a useful way to show when both columns are null.

shawnalpay has already proposed a structural revision (restoring Direct IaaS and the single-row Seat-Based SaaS, merging Credit and Tax), so I won't re-open those threads. One cross-cutting point sits underneath that discussion:

The narratives explain nullability in terms of whether the actor is *resolvable* from the transaction ("cannot be resolved," "unknown due to aggregated network traffic," and in the rationale, "an authenticated actor exists, making a null `PrincipalId` unlikely"). The adopted `PrincipalId` and `ConsumerId` nullability requirements key off a different test: a value MUST be null when a charge "is not attributed to a [principal/consumer] by the service provider." In every row here the two framings produce the same null / not-null values, so there is no data inconsistency. The gap is that the prose explains the *why* differently from the normative column text it illustrates.

This is also the crux of the Seat-Based SaaS disagreement: under the "attributed by the service provider" rule, `PrincipalId` can be null even when an initiating actor logically exists, because nullability follows whether the provider attributes the charge, not whether an actor is knowable. Worth aligning the scenario narratives to the provider-attribution framing, or flagging explicitly that the resolvability framing is the open question for the nullability discussion.

specification/appendix/actor_attribution.md
shawnalpay: Copy/pasting @ijurica's response from Slack first:

> Scenario 6 (Billing System Charge) works for me, and I’m happy to incorporate it.
> 
> On Scenario 3 and 4, I have a slightly different interpretation.
> 
> For Scenario 3 (Direct IaaS), the current narrative describes a case where a data engineer directly provisions and consumes a compute cluster, with `ConsumerId` set to null. Based on the glossary definition, a consumer is an actor that consumes the resource or service. In this case, if the principal is also the one consuming the resource, then the principal and consumer are the same actor, and `ConsumerId` should be populated (as in Scenario 5), rather than null.
> 
> If we want to demonstrate cases where `PrincipalId` is known but `ConsumerId` is unknown (or vice versa), I would prefer scenarios that illustrate non-resolvability of Principal or Consumer, rather than framing nullability primarily as a provider decision not to include it.
> 
> So, I' like to retain Network Edge Processing and/or the Seat-Based SaaS (Purchase ), since they reflect situations where the consumer is not observable or not yet assigned.

Good point about Scenario 3. If the data engineer provisions *and* consumes the compute cluster, they are inherently the consumer. Having the `ConsumerId` be `null` in that case just points to a lazy billing export from the provider, not true non-resolvability.

Good idea to focus on *true non-resolvability* instead. Let's swap out the Direct IaaS example for your **Network Edge Processing** scenario. It illustrates a situation where the requester is known, but the downstream consumers are mathematically impossible to resolve because of aggregated traffic (`PrincipalId` = Populated, `ConsumerId` = `null`).

Since the Edge Processing scenario covers that `[Populated, null]` quadrant, we actually don't need the SaaS Purchase scenario to do that job anymore. Because of that, I’d love to keep the "Seat-Based SaaS" example as a single, simple row. It cleanly demonstrates the opposite `[null, Populated]` state without introducing the extra complexity of procurement lifecycles (Purchase vs. Usage). 

Here's a revised version. Lmk what you think.

1. **Generative AI API:** **Acme Corp** uses an internal bot to summarize notes via a LatticeScale API. Security audits the service account (`PrincipalId`), while FinOps allocates token costs to the employee (`ConsumerId`).
2. **Multi-Tenant PaaS:** A shared BI engine (`PrincipalId`) runs a query on OmniQuery for a specific client of **GearPeak Outdoors**. Capturing the downstream client (`ConsumerId`) prevents costs from pooling in overhead.
3. **Network Edge Processing:** At **Acme Corp**, traffic routed through an Aura Web edge network is authenticated at the gateway (`PrincipalId`), but downstream consumers (`ConsumerId` is null) are unresolvable due to aggregated network traffic.
4. **Seat-Based SaaS:** **Acme Corp** pays for SprintCanvas project management licenses and assigns them to individual users.  In this case, there was authorized request coming from the customer (`PrincipalId` is null), but the opaque identifier allows safe user-level chargeback (`ConsumerId`).
5. **Direct PaaS Usage:** A data scientist at **Acme Corp** logs into a LatticeScale managed notebook using their individual SSO credential (`PrincipalId`). Because the user authenticated directly to the platform and is the sole consumer of the compute (`ConsumerId`), the same opaque identifier is populated in both columns.
6. **Billing System Charge:** **Acme Corp** receives a promotional credit or is charged applicable tax from LatticeScale. Since the charge is generated by the provider's billing system rather than initiated by a request for a resource or service, the principal cannot be resolved (`PrincipalId` is null). As the charge applies at the account level, it cannot be associated with a specific end-user (`ConsumerId` is null).

| Scenario | Technology Category | Data Generator | PrincipalId | ConsumerId |
| :--- | :--- | :--- | :--- | :--- |
| 1 | **Generative AI API** | LatticeScale | `svc-acme-docbot-prod` | `emp_hash_84729x` |
| 2 | **Multi-Tenant PaaS** | OmniQuery | `svc-bi-reporting-engine` | `client_gearpeak_001` |
| 3 | **Network Edge Processing** | Aura Web | `svc-edge-gateway-prod` | `null` |
| 4 | **Seat-Based SaaS** | SprintCanvas | `null` | `user_uuid_554321` |
| 5 | **Direct PaaS Usage** | LatticeScale | `dev_uuid_112233` | `dev_uuid_112233` |
| 6 | **Billing System Charge** | LatticeScale | `null` | `null` |

ijurica: I wanted to make the rationale for why PrincipalId or ConsumerId is null more consistent and explicit across all scenarios. Based on that, I updated the **Seat-Based SaaS** scenario to the following:

> 4. **Seat-Based SaaS:** **Acme Corp** pre-purchases a number of SprintCanvas seats assigned to individual users. Since the entitlement is established through a contractual agreement, there is no originating requestor (`PrincipalId` is `null`), while the opaque identifier allows safe user-level chargeback (`ConsumerId`).

@shawnalpay @Matt-Cowsert, thoughts?

shawnalpay: @ijurica Looks good to me!

shawnalpay (approved):
LGTM!

Issue #2322, opened 2026-04-29
[General Feedback] Inconsistent casing of tag keys in the Tags column observed across real-world FOCUS exports (e.g., "Application" vs "APPLICATION")


SantiagoBarreroEmaster:
### Submitting Organization (Optional)

TI724

### Feedback Summary

The FOCUS Specification (v1.x) does not provide explicit guidance on tag-key casing within the `Tags` column. In real-world FOCUS exports from Microsoft Azure, the same logical tag key appears with inconsistent casing across rows of the same dataset (e.g., `Application`, `APPLICATION`, `application`), forcing every consumer to implement defensive case-insensitive matching. We request clarification from the working group on whether tag keys should be normalized at export time or whether case-insensitive consumption is the expected pattern, so that this expectation can be documented in the spec and applied consistently across providers and implementers.

### Type of Feedback

- [ ] Bug — incorrect or contradictory behavior in the current spec
- [ ] Typo or grammar
- [x] Clarity improvement
- [ ] Field naming or label inconsistency
- [x] Definition or behavior inconsistency
- [ ] Minor correction (non-breaking)
- [ ] Other (please explain)

### Affected Section or Field (Optional)

Tags column / spec.md#tags

### Real-world context

This feedback emerged from real implementation work conducted by **TI724**, a FinOps services organization currently progressing through the FinOps Certified Service Provider (FCSP) certification. As part of our adoption of the FOCUS standard, we built an internal solution called **FOCUS Query** — a lightweight standalone tool (FastAPI backend with a DuckDB analytical engine, plus a React frontend) that ingests FOCUS-formatted CSV files and executes three use cases drawn directly from the official FOCUS Use Case Library:

- *Compare billed cost per subaccount to budget*
- *Update budgets for each application*
- *Update budgets with billed costs*

The implementation primarily processes FOCUS exports originating from **Microsoft Azure** environments, which is where the inconsistency described below was observed.

### Observed casing variations

When parsing the `Tags` column across multiple FOCUS-formatted Azure billing exports, the same logical tag key was represented with different casing depending on how the tag was originally written by the resource owner. The most common variations we observed include:

- `Application` (PascalCase)
- `APPLICATION` (uppercase)
- `application` (lowercase, less common)

These variations appear within the **same `Tags` column**, often within the **same dataset**, since Azure preserves the casing originally entered by the user when the tag was created.

### Impact on implementers

[...]

PR #691, opened 2025-01-28
WI #540: Ensure all user-defined Tags are encapsulated within the Tags column

beau-nelford-anglepoint:
This change address [Issue #540](https://github.com/FinOps-Open-Cost-and-Usage-Spec/FOCUS_Spec/issues/540) and encapsulates the updates to multiple tag schemes in the documentation, decided by a previous survey among the FinOps community.

image

PR #2434, opened 2026-06-01
FR #2325: Restore user-defined tag key preservation requirement to Tags column

### LINKED ISSUE #2325: [FR] Standardize tag key handling in the Tags column
### Problem Statement

Practitioners consuming FOCUS data cannot rely on tag keys in the Tags column being a faithful representation of what was entered at the source. The current spec constrains data generators from altering tag values, but places no equivalent constraint on tag keys, so generators may transform user-defined keys (or leave them as-is) without any normative expectation either way.

In real-world FOCUS exports, the absence of this constraint surfaces as inconsistent representation of the same logical tag key within a single dataset. The originating feedback ([#2322](https://github.com/FinOps-Open-Cost-and-Usage-Spec/FOCUS_Spec/issues/2322)) reports observing `Application`, `APPLICATION`, and `application` for the same conceptual key in Azure-sourced exports, forcing every consumer to implement defensive matching logic and risking silent data loss for implementers unaware of the variation. Without spec-level expectations on key handling, consumers cannot tell whether observed variation reflects source-data fidelity, generator transformation, or generator inconsistency, and cannot share normalization logic across implementations.

### Use Case / User Story

**As a** FinOps practitioner consuming FOCUS data, **I need** predictable tag key handling in the Tags column across data generators, **so that** I can perform cost allocation by tag without writing per-generator normalization logic to compensate for unspecified key behavior.

### Success Criteria

- Tag keys in the Tags column are a faithful representation of what users entered at the source, consistently across all data generators that support tagging.
- Practitioners can build cost-allocation reports that aggregate by tag key without applying provider-specific normalization to compensate for generator-side key handling.
- When tag key variation appears within a dataset (e.g., differences in casing), practitioners can interpret it as source-data fidelity rather than generator inconsistency.
- Data generators have clear guidance on how to handle tag keys when producing FOCUS-compliant exports.

### Type of Request

Enhancement - Refinement or extension of an existing FOCUS column, metric, or capability

(Refining the Tags column by restoring a previously-defined normative requirement and adding clarifying guidance. Also defensible as Tech Debt given the requirement existed prior to PR #691.)

- **Walmart** (Karl Kraft) - Nice to Have - Sponsor for v1.5; raised the gap during maintainers discussion of #2322 and identified the original normative requirement that was removed.
- **TI724** (Santiago Barrero, original feedback author) - Nice to Have - Maintains custom case-insensitive matching logic in their FOCUS Query tool to handle Azure exports; standardized expectations would reduce a brittle workaround.

Yes - Widely available across data generators

The behavior the proposed constraint covers (data generators not altering user-defined tag keys) is consistent with how major cloud generators handle keys today, though their approaches differ:

- **AWS** treats tag keys as case-sensitive natively, so user-entered casing is preserved in exports. (Note: AWS IAM enforces an additional validation that prevents tags differing only by casing on the same IAM resource, so platform-wide case-sensitivity has a service-specific exception.)
- **Microsoft Azure** treats tag keys as case-insensitive for operations but preserves the user-entered casing when surfacing tags in cost reports, which is the source of the variation reported in #2322.
- **Google Cloud** restricts label keys to lowercase characters at input, so generators are not altering user input downstream; users cannot enter mixed-case keys to begin with.

In all three cases, what reaches the FOCUS export matches what the user provided. The proposed requirement is therefore compatible with current generator behavior rather than introducing new constraints. SaaS and non-cloud data generator behavior has not been verified for this FR.

### Data Generator Documentation Links (Optional)

- Microsoft Azure: [Use tags to organize your Azure resources and management hierarchy](https://learn.microsoft.com/en-us/azure/azure-resource-manager/management/tag-resources). Documents case-insensitive operations with case-preservation surfacing in cost reports.
- AWS: [Tagging your AWS resources](https://docs.aws.amazon.com/tag-editor/latest/userguide/tagging.html). "Tag keys are case sensitive." Page also notes IAM-specific validation against case-only differences.
- Google Cloud: [Labels overview](https://docs.cloud.google.com/resource-manager/docs/labels-overview). Label keys must use lowercase characters only.

- Originating feedback: [#2322](https://github.com/FinOps-Open-Cost-and-Usage-Spec/FOCUS_Spec/issues/2322)
- PR that removed the original requirement: [#691](https://github.com/FinOps-Open-Cost-and-Usage-Spec/FOCUS_Spec/pull/691) (WI #540, merged 2025-02-27). Diff replaced "Providers MUST NOT alter user-defined Tag keys or values." with "Providers MUST NOT alter tag values unless applying true (boolean) to valueless tags."
- Maintainers discussion (Slack `#maintainers`, 2026-04-29): consensus from Mike Fuller, Karl Kraft, Shawn Alpay that observed key variation is provider-preserved behavior, not generator transformation, and is the correct thing to constrain via "MUST NOT alter."

1. Restore a normative requirement to `specification/columns/tags.md` constraining data generators from altering user-defined tag keys. Determine whether to:
   - Combine with the existing values requirement (e.g., "Providers MUST NOT alter user-defined tag keys or values unless applying true (boolean) to valueless tags."), or
   - Keep as a separate bullet for clarity.
2. Add non-normative content (spec body or appendix) explaining:
   - Tag key representation is preserved by data generators as originally entered by the user.
   - Consumers should expect to perform case-insensitive matching when their use case requires aggregating logically equivalent keys, while remaining aware that some providers (e.g., AWS) may treat differently-cased keys as semantically distinct.
3. Confirm phrasing makes the constraint scope clear. Per Karl's note in the Slack thread, the requirement is intended to constrain generator-side transformation of what the user provided, not user-side tagging behavior (e.g., a user setting `Application` does not constitute "alteration" even if `APPLICATION` previously existed).

**Current Workaround:** Consumers implement case-insensitive matching in their own pipelines. TI724's FOCUS Query tool is one example.

**Urgency/Timeline:** Sponsored by Karl Kraft (Walmart) for v1.5 under the new sponsorship process.

**Historical Context:** The protection on tag keys existed in FOCUS prior to [PR #691](https://github.com/FinOps-Open-Cost-and-Usage-Spec/FOCUS_Spec/pull/691), which split the original "Providers MUST NOT alter user-defined Tag keys or values." bullet into a values-only requirement and dropped the keys half. PR #691 was approved by a number of maintainers and merged through normal review, so this FR does not assert the removal was inadvertent or in error. It only proposes restoring the keys protection now that practitioner feedback (#2322) has surfaced a concrete gap created by its absence.

**Related Requests:**
- Originating feedback: [#2322](https://github.com/FinOps-Open-Cost-and-Usage-Spec/FOCUS_Spec/issues/2322)
- Historical: [PR #691](https://github.com/FinOps-Open-Cost-and-Usage-Spec/FOCUS_Spec/pull/691)

### LINKED ISSUE #2433: [AI] Draft PR restoring user-defined tag key preservation requirement to Tags column
**Assignee(s):** @thecloudman, @kk09v
**Parent Issue:** #2325
**Target Date:** 2026-10-01

### Detailed Description

Draft PR for FR #2325. Restores a normative requirement constraining data generators from altering user-defined tag keys in the `Tags` column, and adds non-normative guidance on key preservation and consumer-side case-sensitivity expectations.

#### Background

The protection on tag keys existed prior to [PR #691](https://github.com/FinOps-Open-Cost-and-Usage-Spec/FOCUS_Spec/pull/691) (merged 2025-02-27), which split the original "Providers MUST NOT alter user-defined Tag keys or values" bullet into a values-only requirement and dropped the keys half. Originating feedback #2322 (TI724) surfaced the resulting gap: Azure exports represent the same logical key as `Application`, `APPLICATION`, and `application` within a single dataset, forcing every consumer to implement defensive case-insensitive matching. Sponsored by Karl Kraft (Walmart) for v1.5.

#### Phrasing scope

Per Karl's note in the maintainers thread (2026-04-29), the requirement constrains generator-side transformation of what the user provided — not user-side tagging behavior. A user setting `Application` does not constitute "alteration" even when `APPLICATION` previously existed. The restored bullet is a parity-with-input constraint, not a cross-row uniqueness constraint. Reviewers should not read it as requiring keys to be unique or normalized within a dataset.

### Definition of Done

- [ ] Draft PR opened against `working_draft` covering:
  - [ ] `tags.md` — new normative bullet under the User-defined tags section: user-defined tag keys match the provided keys (kept separate from the existing value-match bullet per FR §1 option B).
  - [ ] `tags.md` — non-normative section on tag key preservation and consumer-side case-sensitivity expectations (AWS case-sensitive native, Azure case-preserving, GCP lowercase-only).
  - [ ] `allocatedtags.md` — parallel bullet so the constraint extends to allocated-charge tag handling.
- [ ] Out of scope, deferred to separate follow-up: provider-defined tag key handling; `x_` extension column conformance; export-time key normalization.
- [ ] Present in TF-2 for review.
- [ ] Maintainer approval received on the PR.
- [ ] PR merged to `working_draft` by 2026-10-01 (v1.5 milestone).

thecloudman:
## Summary

Restores a normative requirement — removed in [PR #691](https://github.com/FinOps-Open-Cost-and-Usage-Spec/FOCUS_Spec/pull/691) — that user-defined tag keys in the `Tags` column faithfully represent the keys entered at the source. Originating feedback #2322 (TI724) surfaced the gap: Azure exports represent the same logical key as `Application`, `APPLICATION`, and `application` within a single dataset, forcing every consumer to implement defensive case-insensitive matching. Sponsored by Karl Kraft (Walmart) for v1.5.

Per the maintainers discussion (2026-04-29), the constraint governs generator-side transformation of what the user provided — it is a parity-with-input requirement, not a cross-row uniqueness or normalization requirement. A user setting `Application` does not constitute "alteration" even when `APPLICATION` exists elsewhere.

## Design Context

PR #691 split the original "Providers MUST NOT alter user-defined Tag keys or values" bullet into a values-only requirement (now `Tag values MUST match the provided values…`) and dropped the keys half. This PR restores the keys protection as a separate bullet (per FR #2325 §1 option B), keeping the boolean-valueless exception scoped to values where it belongs.

The new bullet carries an `except for any tag scheme-specific prefix` clause so it does not contradict the adjacent bullets that require data generators to prepend scheme-specific prefixes to user-defined keys.

## Type of Change

- [x] New feature / Specification update

## Changes

* `tags.md` — new normative bullet under the User-defined tags section: `User-defined tag keys MUST match the keys provided at the source, except for any tag scheme-specific prefix.`
* `tags.md` — new non-normative "Tag Key Representation" section explaining source-fidelity of key casing and consumer-side case-sensitivity expectations (case-sensitive, case-preserving, and fixed-charset platforms).
* `allocatedtags.md` — parallel key-preservation bullet so the constraint extends to allocated-charge tag handling.
* ~~Requirements-model rules `CAU-Tags-C-024-M` and `CAU-AllocatedTags-C-018-M` (ModelVersionIntroduced 1.5), wired into the respective user-defined-tags composites.~~

## Out of Scope (deferred)

* Provider-defined tag key handling (different semantics — provider-controlled, prefixed).
* `x_` extension column conformance (raised as discussion point 3 in #2322).
* Export-time key normalization (FR explicitly defers to consumer-side matching).

## Test plan

- [ ] CI build passes (`make` in `specification/`).
- [ ] Requirements-model tests pass (`pytest requirements_model/tests/`).
- [ ] Spec renders correctly; new non-normative section appears under Tags.
- [ ] Markdown lint (`pymarkdownlnt`) passes.

## Author Checklist

- [x] I have read and agree to the [AI Usage Guidelines](guidelines/contributors/ai-usage-guidelines.md).
- [x] My contribution complies with the FOCUS Specification's contributing guidelines.
- [x] I have run the spec build locally where possible (`build_json.py --build-only` passes and same-file rule references resolve; full `make` and `pytest` blocked by missing `pymarkdownlnt` / `markdown-pp` / `pytest` in the local environment — CI will run the full gate).
- [x] I have updated the documentation accordingly (column `.md` normative + non-normative content, requirements-model rules).

Closes #2433.
Relates to #2325, #2322.


specification/datasets/cost_and_usage/columns/tags.md
```diff
@@ -68,6 +69,18 @@ The table below represents a finalized dataset with these *resources*. It also s
 
 Because the Virtual Machine Resource did not have an `env` tag, it inherited tag, `env:prod` (italicized), from its parent *sub account*. Conversely, because the Virtual Machine Resource already has a `team` tag (`team:web`), it did not inherit `team:ops` from its parent *sub account*.
 
+## Tag Key Representation
+
+User-defined tag keys in Tags are a faithful representation of the keys entered at the source, aside from any *tag scheme*-specific prefix the data generator prepends. A data generator does not normalize the casing or otherwise transform the characters of a user-defined tag key.
```
kk09v:
```suggestion
User-defined tag keys in Tags are a faithful representation of the keys entered at the source, aside from any *tag scheme*-specific prefix the data generator prepends. A data generator does not normalize the casing or otherwise transform the characters of a user-defined tag key but may define constraints on the tag keys set. E.g., data generator may limit the number of characters or set of characters supported, as long as these limitations are applied prior to setting the tags.
```

kk09v:
We need to move the Requirements Model changes to 1.5 folder.

specification/requirements_model/releases/1.4/model_rules/datasets/cost_and_usage/columns/allocatedtags.json
Matt-Cowsert, line 1: Discussed on 2026-06-15, this topic will be discussed in the TF-RM meeting on 2026-06-16 and then guidance given offline to the authors. @jpradocueva

specification/datasets/cost_and_usage/columns/allocatedtags.md
```diff
@@ -20,6 +20,7 @@ AllocatedTags MUST adhere to the following requirements:
 * Data generator-defined tags MUST adhere to the following requirements:
   * Data generator-defined tag keys MUST be prefixed with a predetermined, data generator-specified tag key prefix that is unique to each corresponding provider-specified [*tag scheme*](#glossary:tag-scheme).
   * Data generator-specified tag key prefixes SHOULD be publicly documented.
+* User-defined tag keys MUST match the keys provided at the source, except for any *tag scheme*-specific prefix.
```
flanakin: nit: Do we need to clarify what "at the source" means?

flanakin (dismissed):

specification/datasets/cost_and_usage/columns/tags.md
```diff
@@ -68,6 +69,18 @@ The table below represents a finalized dataset with these *resources*. It also s
 
 Because the Virtual Machine Resource did not have an `env` tag, it inherited tag, `env:prod` (italicized), from its parent *sub account*. Conversely, because the Virtual Machine Resource already has a `team` tag (`team:web`), it did not inherit `team:ops` from its parent *sub account*.
 
+## Tag Key Representation
+
+User-defined tag keys in Tags are a faithful representation of the keys entered at the source, aside from any *tag scheme*-specific prefix the data generator prepends. A data generator does not normalize the casing or otherwise transform the characters of a user-defined tag key but may define constraints on the tag keys set. E.g., data generator may limit the number of characters or set of characters supported, as long as these limitations are applied prior to setting the tags.
```
flanakin: 🤖 [AI][Claude] ⚠️ **Should fix**

Three small editorial items on this line:

1. The sentence-leading `E.g.,` deviates from the spec's established convention — inline examples use the parenthetical `(e.g., …)` form everywhere else (389 occurrences; this is the only sentence-leading `E.g.,` in the spec).
2. Missing article: `data generator may limit` → `a data generator may limit`.
3. `constraints on the tag keys set` reads awkwardly; `the set of supported tag keys` is clearer.

```suggestion
User-defined tag keys in Tags are a faithful representation of the keys entered at the source, aside from any *tag scheme*-specific prefix the data generator prepends. A data generator does not normalize the casing or otherwise transform the characters of a user-defined tag key but may define constraints on the set of supported tag keys (e.g., a data generator may limit the number of characters or set of characters supported), as long as these limitations are applied prior to setting the tags.
```

flanakin (commented):
🤖 [AI][Claude] **PR Review**

**Summary:** Clean restoration of the user-defined tag key preservation requirement. The normative bullet is well-formed (schema-level subject, state-based, correct glossary linking and `except for` exception with precedent in `numeric_format.md`), and the requirements-model wiring is correct — `CAU-Tags-C-024-M` and `CAU-AllocatedTags-C-018-M` are placed in the correct composites with non-conflicting `Order` values, the build succeeds, and all 128 requirements-model tests pass. One editorial item in the new non-normative prose.

### ⚠️ Should fix (1)

1. `tags.md` line 74 — sentence-leading `E.g.,` deviates from the spec's parenthetical-example convention (0 other instances in the spec vs. 389 uses of `(e.g., …)`), plus a missing article and awkward `the tag keys set` phrasing.

### ✅ What's done well

* Correct scoping: keeps the boolean-valueless exception on the values bullet and restores keys as a separate bullet per FR §1 option B.
* `except for any *tag scheme*-specific prefix` clause cleanly avoids contradicting the adjacent prefix-prepending bullets.
* Requirements-model rules carry `ModelVersionIntroduced: 1.5` and are wired into both the `Items` and `Dependencies` of the matching user-defined-tags composite.
* Non-normative section accurately captures the case-sensitive / case-preserving / fixed-charset platform distinction from the originating feedback.

specification/datasets/cost_and_usage/columns/allocatedtags.md
Matt-Cowsert:
```suggestion
* User-defined tag keys MUST match the keys specified by the end-user, except for any *tag scheme*-specific prefix.
```

specification/datasets/cost_and_usage/columns/tags.md
```diff
@@ -23,6 +23,7 @@ Tags MUST adhere to the following requirements:
   * Provider-defined tag keys MUST be prefixed with a predetermined, provider-specified tag key prefix that is unique to each corresponding provider-specified [*tag scheme*](#glossary:tag-scheme).
   * Provider-specified tag key prefixes SHOULD be publicly documented.
 * User-defined tags MUST adhere to the following requirements:
+  * User-defined tag keys MUST match the keys provided at the source, except for any *tag scheme*-specific prefix.
```
thecloudman: Same issue as flanakin's comment on the AllocatedTags equivalent of this bullet: "at the source" isn't grounded, and it collides with the existing **Tag Source** glossary term (`#glossary:tag-source`), which defines a grouping construct, not "point of origin." The sibling values-bullet at line 21 already establishes the pattern for this exact concept without "source" language (`Tag values MUST match the provided values...`). Suggest mirroring it here for consistency:

```suggestion
  * User-defined tag keys MUST match the provided keys, except for any *tag scheme*-specific prefix.
```

specification/datasets/cost_and_usage/columns/tags.md
thecloudman: Building on flanakin's editorial suggestion above — this line also inherits the "source" ambiguity flagged on the AllocatedTags bullet (collision with the `#glossary:tag-source` glossary term). Folding that fix into the same suggestion:

```suggestion
User-defined tag keys in Tags are a faithful representation of the keys as entered by the user, aside from any *tag scheme*-specific prefix the data generator prepends. A data generator does not normalize the casing or otherwise transform the characters of a user-defined tag key but may define constraints on the set of supported tag keys (e.g., a data generator may limit the number of characters or set of characters supported), as long as these limitations are applied prior to setting the tags.
```

specification/datasets/cost_and_usage/columns/tags.md
```diff
@@ -68,6 +69,18 @@ The table below represents a finalized dataset with these *resources*. It also s
 
 Because the Virtual Machine Resource did not have an `env` tag, it inherited tag, `env:prod` (italicized), from its parent *sub account*. Conversely, because the Virtual Machine Resource already has a `team` tag (`team:web`), it did not inherit `team:ops` from its parent *sub account*.
 
+## Tag Key Representation
+
+User-defined tag keys in Tags are a faithful representation of the keys entered at the source, aside from any *tag scheme*-specific prefix the data generator prepends. A data generator does not normalize the casing or otherwise transform the characters of a user-defined tag key but may define constraints on the set of supported tag keys (e.g., a data generator may limit the number of characters or set of characters supported), as long as these limitations are applied prior to setting the tags.
+
+Tag key casing reflects the source rather than a data generator convention, and the representation a data generator surfaces depends on the source platform. For example:
```
thecloudman: Same "source" ambiguity as the normative bullet above (collides with the `#glossary:tag-source` glossary term, which means something unrelated — a grouping construct). This line uses "the source" and "source platform" back to back; suggest grounding both in the actor/concept actually being described:

```suggestion
Tag key casing reflects how the user entered the key rather than a data generator convention, and the representation a data generator surfaces depends on the platform. For example:
```

specification/datasets/cost_and_usage/columns/tags.md
```diff
@@ -68,6 +69,18 @@ The table below represents a finalized dataset with these *resources*. It also s
...
+
+* Some platforms treat tag keys as case-sensitive, so keys entered with different casing (e.g., `Application` and `application`) are distinct keys and are surfaced as entered.
+* Some platforms treat tag keys as case-insensitive for operations while preserving the casing entered when the key was created, so the same logical key can appear with different casing across rows (e.g., `Application`, `APPLICATION`, and `application`).
+* Some platforms restrict tag keys to a fixed character set (e.g., lowercase only) at the source, so variation does not arise.
```
thecloudman: "at the source" here is also redundant — the sentence is already about what a platform does, so "the source" adds nothing but the ambiguity flagged elsewhere in this file.

```suggestion
* Some platforms restrict tag keys to a fixed character set (e.g., lowercase only), so variation does not arise.
```

specification/datasets/cost_and_usage/columns/tags.md
```diff
@@ -68,6 +69,18 @@ The table below represents a finalized dataset with these *resources*. It also s
...
+
+Because casing reflects source fidelity, a consumer aggregating logically equivalent keys can apply case-insensitive matching, while remaining aware that some platforms treat keys differing only by casing as semantically distinct.
```
thecloudman: Closing out the same "source" cleanup as the rest of this section:

```suggestion
Because casing reflects how the user entered the key, a consumer aggregating logically equivalent keys can apply case-insensitive matching, while remaining aware that some platforms treat keys differing only by casing as semantically distinct.
```

specification/datasets/cost_and_usage/columns/allocatedtags.md
thecloudman: Both resolve flanakin's ambiguity concern, but suggest going with the pattern already established in the spec rather than introducing "end-user" as new terminology: the sibling values-bullet for this same column (`tags.md:21`) already reads `Tag values MUST match the provided values unless true (boolean) is applied to valueless tags.` — no actor language needed. Mirroring that for keys keeps the two parallel requirements consistent with each other:

```suggestion
* User-defined tag keys MUST match the provided keys, except for any *tag scheme*-specific prefix.
```

specification/datasets/cost_and_usage/columns/tags.md
kk09v:
```suggestion
Tag key casing reflects how the tag key was provided rather than a data generator convention, and the representation a data generator surfaces depends on the platform. For example:
```
More similar to the above phrasing. There may not be a user at all; this could be programmatic.

specification/datasets/cost_and_usage/columns/tags.md
kk09v:
```suggestion
Because casing reflects how the key was provided, a consumer aggregating logically equivalent keys can apply case-insensitive matching, while remaining aware that some platforms treat keys differing only by casing as semantically distinct.
```

kk09v (dismissed):

specification/requirements_model/releases/1.4/model_rules/datasets/cost_and_usage/columns/allocatedtags.json
kk09v: Removed

specification/datasets/cost_and_usage/columns/allocatedtags.md
Matt-Cowsert: @thecloudman @kk09v can one of you resolve this comment as addressed? With Michael being out, I don't want to rely on him resolving on his own before members review.

mike-finopsorg (dismissed):

specification/datasets/cost_and_usage/columns/allocatedtags.md
Matt-Cowsert: As @flanakin is OOO this week and I've verified this change myself, I'm going to resolve this comment. If we need to re-open, we can.

specification/datasets/cost_and_usage/columns/allocatedtags.md
```diff
@@ -20,6 +20,7 @@ AllocatedTags MUST adhere to the following requirements:
 * Data generator-defined tags MUST adhere to the following requirements:
   * Data generator-defined tag keys MUST be prefixed with a predetermined, data generator-specified tag key prefix that is unique to each corresponding provider-specified [*tag scheme*](#glossary:tag-scheme).
   * Data generator-specified tag key prefixes SHOULD be publicly documented.
+* User-defined tag keys MUST match the provided keys, except for any *tag scheme*-specific prefix.
```
jpradocueva: @thecloudman I do realize that we have similar statements from previous releases, but now we are emphasizing our Editorial and Normative guidelines. Will it be acceptable if we frame the requirement as:

_"The user-defined tag keys in AllocatedTags MUST match the provided keys, except for any tag scheme-specific prefix"_

This makes it clear that the entity receiving the constraint is the AllocatedTags

jpradocueva: The same applies to the "Tags" requirement listed below.

jpradocueva (commented):
@thecloudman I left a comment about the entity that receives the constraint. I do realize that we already have this problem, as we have a similar requirement from a previous release. I will bring this topic to the TF-RM for discussion.

ljadvey (dismissed):

specification/requirements_model/releases/1.4/model_rules/datasets/cost_and_usage/columns/allocatedtags.json
ijurica, line 1: Could we revert the changes to tags.json and allocatedtags.json and keep this PR focused on the spec content?

Per our agreement, requirements model files are out of scope for this PR, and keeping the diff limited to the spec content will make review cleaner. 

The only change there is a trailing newline addition, and I believe all updates to RM files should be handled separately.

specification/requirements_model/releases/1.4/model_rules/datasets/cost_and_usage/columns/tags.json
ijurica, line 1: Similar to [this](https://github.com/FinOps-Open-Cost-and-Usage-Spec/FOCUS_Spec/pull/2434/files#r3554154055), I suggest we revert the changes to tags.json.

specification/datasets/cost_and_usage/columns/allocatedtags.md
ijurica: ### Current wording aligned with guidelines

While I understand the intent, I would like to mention that the current wording (`User-defined tag keys MUST match...`) actually aligns with the currently-established pattern used by all other sub-element bullets in this column (`Tag values MUST match the provided values...`, `Provider-defined tag keys MUST be prefixed...`, `User-defined tag keys MUST NOT include...`, etc.) and is consistent with the allowed subject form `<term> key` / `<term> value` listed in the guidelines (e.g., `Tag keys`).

### Impact if adopted

Adding `in AllocatedTags` as a subject qualifier here would create an asymmetry with the other seven sub-element bullets in the same column, and would also apply to `Tags` and `SkuPriceDetails`.

### Suggestion

I'd **suggest** we first **evaluate** whether to introduce a `<term> key in <ColumnId>` subject pattern in the Editorial and Normative guidelines, and if adopted, **adjust the guidelines** accordingly and **apply it** consistently across all three key-value columns in a **separate PR(s)**. That way this PR stays focused on restoring the content requirement.

---

***Note:** If you guys agree with the suggestion, we should add this to the backlog.*

ijurica: While I believe the normative subject is aligned with guidelines, I did notice two other points worth addressing.😃 

### Approved conditional keyword

Per the Normative Requirements Guidelines, conditional logic in requirements MUST be expressed using one of the approved conditional keywords: `when`, `unless`, `only when`, `except when`. The current wording uses `except for`, which is not on the list.

I'd suggest we align the new bullet with the approved keywords by replacing `except for` with `except when`, in both `Tags` and `AllocatedTags`:

### Contextual grouping parent bullet in AllocatedTags

With the introduction of an additional requirement whose subject is `User-defined tag keys`, `AllocatedTags` now has two sibling top-level bullets sharing the same subject. This is the pattern that already justifies the `User-defined tags MUST adhere to the following requirements:` parent grouping used in `Tags`, and `AllocatedTags` already applies the analogous grouping to `Data generator-defined tags`.

I'd suggest we introduce the same contextual grouping parent bullet in `AllocatedTags` to keep the two columns structurally aligned:

---

### Suggestion

Since there's a comment between the two bullets, I can't attach an applicable final suggestion inline, so I'm including it here as a quote:

> * User-defined tags MUST adhere to the following requirements:
>   * User-defined tag keys MUST match the provided keys, except when a *tag scheme*-specific prefix is prepended.
>   * User-defined tag keys in all user-defined *tag schemes* MUST include a predetermined, data generator-specified tag key prefix that is unique to each corresponding user-defined *tag scheme* when the data generator has more than one user-defined *tag scheme*.

specification/datasets/cost_and_usage/columns/tags.md
```diff
@@ -23,6 +23,7 @@ Tags MUST adhere to the following requirements:
   * Provider-defined tag keys MUST be prefixed with a predetermined, provider-specified tag key prefix that is unique to each corresponding provider-specified [*tag scheme*](#glossary:tag-scheme).
   * Provider-specified tag key prefixes SHOULD be publicly documented.
 * User-defined tags MUST adhere to the following requirements:
+  * User-defined tag keys MUST match the provided keys, except for any *tag scheme*-specific prefix.
```
ijurica: See [this](https://github.com/FinOps-Open-Cost-and-Usage-Spec/FOCUS_Spec/pull/2434/files#r3554585043) comment for reasoning

```suggestion
  * User-defined tag keys MUST match the provided keys, except when a *tag scheme*-specific prefix is prepended.
```

specification/datasets/cost_and_usage/columns/tags.md
```diff
@@ -68,6 +69,18 @@ The table below represents a finalized dataset with these *resources*. It also s
...
+
+Tag key casing reflects how the tag key was provided rather than a data generator convention, and the representation a data generator surfaces depends on the platform. For example:
+
+* Some platforms treat tag keys as case-sensitive, so keys entered with different casing (e.g., `Application` and `application`) are distinct keys and are surfaced as entered.
+* Some platforms treat tag keys as case-insensitive for operations while preserving the casing entered when the key was created, so the same logical key can appear with different casing across rows (e.g., `Application`, `APPLICATION`, and `application`).
+* Some platforms restrict tag keys to a fixed character set (e.g., lowercase only), so variation does not arise.
+
+Because casing reflects how the key was provided, a consumer aggregating logically equivalent keys can apply case-insensitive matching, while remaining aware that some platforms treat keys differing only by casing as semantically distinct.
+
```
ijurica: Going through the section, I found myself wondering whether it belongs in the **column definition**, or whether it might fit better in **supporting content**. I used the following two criteria:

### Criterion 1 — Input-time vs. post-input scope

Initial `StringHandling` draft touched on how providers could or should prevent things like empty strings, whitespace-only strings, or leading/trailing spaces at input, and the feedback at the time was that this **is out of FOCUS scope**. `StringHandling` and FOCUS in general focus on post-input — how a string value, once stored in the provider's system, is passed through in its original form — not on what source platforms allow, prevent, or normalize at input time.

Analogous to `StringHandling`, the tag key requirement specifies that whatever the source stored, the data generator passes through into Cost and Usage. Following the same precedent, content that describes source-platform input-time behavior for tag keys (case-sensitive vs. case-insensitive storage, restricted character sets) sits outside that scope.

Some might argue the same could apply to `Finalized Tags`. But the difference is that in the case of finalization, FOCUS expects to receive the already-processed result from the data generator, and the consumer has no ability to apply their own finalization convention — so the column definition must describe what to expect. **Casing is the opposite situation**: the original is preserved into Cost and Usage, and the consumer retains discretion, so pre-FOCUS platform behavior does not need to be described.

### Criterion 2 — Consumer usability guidance

So far, FOCUS has included consumer-oriented guidance only where a usage risk is a direct consequence of a FOCUS design decision. For example, the Usability Constraints sections in ListCost, ContractedCost, and quantity-related metrics address a double-counting risk that arises directly from how FOCUS defines those column values, so guidance belongs in the column definition.

Tag key casing is a different situation. The Requirements bullet takes the deliberate position that whatever the source stored appears in Cost and Usage. Whether Application and application were entered as two distinct keys by mistake or intentionally reflects a source-side convention that predates FOCUS. FOCUS does not create that ambiguity — it preserves it as-is — so advising consumers on how to compensate for it is not analogous to the double-counting case.

Even though I think guidance on handling tag keys that differ only by casing is useful, I believe it sits **outside the scope of the FOCUS column definition**. It may fit better in supporting content, rather than in the Tags column definition, since it **addresses source-side conventions that FOCUS deliberately preserves rather than an artifact FOCUS itself introduces**.

## Applying the criteria to the section

**Block 1**

> "User-defined tag keys in Tags are a faithful representation of the keys entered at the source, aside from any *tag scheme*-specific prefix the data generator prepends."

Restates the Requirements bullet.

**Block 2**

> "A data generator does not normalize the casing or otherwise transform the characters of a user-defined tag key but may define constraints on the set of supported tag keys (e.g., a data generator may limit the number of characters or set of characters supported), as long as these limitations are applied prior to setting the tags."

Describes what a data generator may do at input time. **Potentially out of scope under Criterion 1.**

**Block 3**

> "Tag key casing reflects how the tag key was provided rather than a data generator convention, and the representation a data generator surfaces depends on the platform. For example:"
>
> * "Some platforms treat tag keys as case-sensitive, so keys entered with different casing (e.g., `Application` and `application`) are distinct keys and are surfaced as entered."
> * "Some platforms treat tag keys as case-insensitive for operations while preserving the casing entered when the key was created, so the same logical key can appear with different casing across rows (e.g., `Application`, `APPLICATION`, and `application`)."
> * "Some platforms restrict tag keys to a fixed character set (e.g., lowercase only), so variation does not arise."

Describes how source platforms handle tag keys at input. **Potentially out of scope under Criterion 1.**

**Block 4**

> "Because casing reflects how the key was provided, a consumer aggregating logically equivalent keys can apply case-insensitive matching, while remaining aware that some platforms treat keys differing only by casing as semantically distinct."

Consumer guidance on handling casing when aggregating tag keys. **Potentially out of scope under Criterion 2.**

specification/datasets/cost_and_usage/columns/tags.md
```diff
@@ -68,6 +69,18 @@ The table below represents a finalized dataset with these *resources*. It also s
 
 Because the Virtual Machine Resource did not have an `env` tag, it inherited tag, `env:prod` (italicized), from its parent *sub account*. Conversely, because the Virtual Machine Resource already has a `team` tag (`team:web`), it did not inherit `team:ops` from its parent *sub account*.
 
+## Tag Key Representation
+
+User-defined tag keys in Tags are a faithful representation of the keys entered at the source, aside from any *tag scheme*-specific prefix the data generator prepends. A data generator does not normalize the casing or otherwise transform the characters of a user-defined tag key but may define constraints on the set of supported tag keys (e.g., a data generator may limit the number of characters or set of characters supported), as long as these limitations are applied prior to setting the tags.
```
Matt-Cowsert:
```suggestion
User-defined tag keys in Tags are a faithful representation of the provided keys, aside from any *tag scheme*-specific prefix the data generator prepends. A data generator does not normalize the casing or otherwise transform the characters of a user-defined tag key but may define constraints on the set of supported tag keys (e.g., a data generator may limit the number of characters or set of characters supported), as long as these limitations are applied prior to setting the tags.
```

The latest commit replaced "source" phrasing with "provided" phrasing throughout this file: the Requirements bullet now reads `MUST match the provided keys`, and the later paragraphs in this section read "how the tag key was provided" and "how the key was provided". That change avoids a collision with the **Tag Source** glossary term (`#glossary:tag-source`), which names a grouping construct rather than a point of origin. This opening sentence is the one remaining "entered at the source" instance. Aligning it to "the provided keys" completes that cleanup and keeps the sentence consistent with the normative bullet it restates.

specification/requirements_model/releases/1.4/model_rules/datasets/cost_and_usage/columns/allocatedtags.json
Matt-Cowsert: Confirmed: the 1.5 requirements-model rules (CAU-Tags-C-024-M and CAU-AllocatedTags-C-018-M) are removed from the 1.4 model rules in the current revision, matching the guidance from the TF-RM discussion. This thread can be resolved.

specification/datasets/cost_and_usage/columns/allocatedtags.md
Matt-Cowsert: The applied wording (`User-defined tag keys MUST match the provided keys, except for any *tag scheme*-specific prefix.`) addresses the ambiguity my suggestion was aimed at, and it keeps the keys bullet parallel with the existing values bullet. My suggestion is superseded, and this thread can be resolved.

specification/datasets/cost_and_usage/columns/allocatedtags.md
jpradocueva: This topic was discussed in the TF-RM on Tuesday, July 14. The issue is now tracked in AI #2535

specification/datasets/cost_and_usage/columns/tags.md
davidearney: Thank you. I was reading and reading your longer reasoning statement above and the line was not very readable. This change makes a big difference.

davidearney (dismissed):

specification/datasets/cost_and_usage/columns/tags.md
Matt-Cowsert: Discussed in TF-2 on 2026-07-22 and we agreed to move this statement into supporting content (vs. the appendix or within the column definitions).

specification/requirements_model/releases/1.4/model_rules/datasets/cost_and_usage/columns/allocatedtags.json
thecloudman: Done — reverted in commit 005e8b61. Both 1.4 RM files (tags.json and allocatedtags.json) are back to working_draft state; only the trailing-newline change remained and that is now removed. RM rule updates for this PR's content will be handled in a separate PR.

thecloudman: Done — reverted in commit 005e8b61. Both 1.4 RM files (tags.json and allocatedtags.json) are back to working_draft state; only the trailing-newline change remained and that is now removed. RM rule updates for this PR's content will be handled in a separate PR.

specification/requirements_model/releases/1.4/model_rules/datasets/cost_and_usage/columns/tags.json
thecloudman: Done — same commit (005e8b61) reverts both files. Closing.

specification/datasets/cost_and_usage/columns/tags.md
thecloudman: Applied in commit 005e8b61 — both tags.md and allocatedtags.md now use 'except when a tag scheme-specific prefix is prepended'. Also applied the same fix to the parallel bullet in allocatedtags.md for consistency.

specification/datasets/cost_and_usage/columns/tags.md
thecloudman: Applied in commit 005e8b61 — 'keys entered at the source' replaced with 'provided keys', consistent with the normative bullet above and avoiding the tag-source glossary collision.

specification/datasets/cost_and_usage/columns/tags.md
thecloudman: Thanks for the suggestion. Going with Matt's variant (commit 005e8b61) which makes the narrower change of replacing 'keys entered at the source' with 'provided keys' — this avoids the tag-source glossary collision while keeping the rest of the sentence intact. Closing this thread.

specification/datasets/cost_and_usage/columns/tags.md
thecloudman: Superseded by Matt's suggestion applied in commit 005e8b61 ('provided keys' replacing 'the source'). Closing.

specification/datasets/cost_and_usage/columns/tags.md
thecloudman: Superseded by Matt's suggestion applied in commit 005e8b61 ('provided keys' replacing 'the source'). Closing.

specification/datasets/cost_and_usage/columns/tags.md
thecloudman: Superseded by Matt's suggestion applied in commit 005e8b61 ('provided keys' replacing 'the source'). Closing.

specification/datasets/cost_and_usage/columns/tags.md
thecloudman: Superseded by Matt's suggestion applied in commit 005e8b61 ('provided keys' replacing 'the source'). Closing.

specification/datasets/cost_and_usage/columns/allocatedtags.md
thecloudman: Both points applied in commit 73f3603c. The conditional keyword fix was already in the prior commit (005e8b61); this commit adds the 'User-defined tags MUST adhere to the following requirements:' parent grouping bullet in allocatedtags.md, matching the structure in tags.md.

specification/datasets/cost_and_usage/columns/tags.md
thecloudman: Agreed, and acted on the TF-2 decision. The ## Tag Key Representation section has been removed from the Tags column definition in commit 73f3603c and moved to supporting_content/datasets/cost_and_usage/columns/tags.md for reference.

specification/datasets/cost_and_usage/columns/allocatedtags.md
thecloudman: Addressed — 'at the source' has been replaced with 'provided' throughout (commit 005e8b61 for the normative bullet, commit 73f3603c removed the Tag Key Representation section entirely). Closing.

specification/datasets/cost_and_usage/columns/allocatedtags.md
thecloudman: Noted — no change needed in this PR since AI #2535 is tracking the normative subject question separately. Closing this thread.

thecloudman (commented):
ready for member review

Matt-Cowsert (approved):
Approving on the current head (`73f3603c`). The restored user-defined tag key preservation bullet is well-formed and consistent with the sibling requirements, the conditional wording matches the guidelines, and the Tag Key Representation content moved to supporting content per our TF-2 decision on 2026-07-22. The normative-subject question is tracked separately in AI #2535, so it's not a blocker here.

Note for other reviewers: the earlier approvals were dismissed because they predated the 2026-07-23 edits, so we'll need fresh approvals on this head before official sign-off.

ijurica (approved):

davidearney (approved):
Looks like all previous concerns have been addressed. LGTM.

specification/datasets/cost_and_usage/columns/allocatedtags.md
```diff
@@ -20,7 +20,9 @@ AllocatedTags MUST adhere to the following requirements:
 * Data generator-defined tags MUST adhere to the following requirements:
   * Data generator-defined tag keys MUST be prefixed with a predetermined, data generator-specified tag key prefix that is unique to each corresponding provider-specified [*tag scheme*](#glossary:tag-scheme).
   * Data generator-specified tag key prefixes SHOULD be publicly documented.
-* User-defined tag keys in all user-defined *tag schemes* MUST include a predetermined, data generator-specified tag key prefix that is unique to each corresponding user-defined *tag scheme* when the data generator has more than one user-defined *tag scheme*.
+* User-defined tags MUST adhere to the following requirements:
+  * User-defined tag keys MUST match the provided keys, except when a *tag scheme*-specific prefix is prepended.
```
flanakin: Should we clarify what "provided" refers to? Where it's provided and by who?

flanakin: The "except" could be interpreted as "we don't care what you do when a tag scheme is prefixed". Might be worth rewording or maybe breaking out into a separate req. When no schem, MUST match + when schema, MUST match "{scheme}{delim}{key}"?

specification/datasets/cost_and_usage/columns/tags.md
```diff
@@ -23,6 +23,7 @@ Tags MUST adhere to the following requirements:
   * Provider-defined tag keys MUST be prefixed with a predetermined, provider-specified tag key prefix that is unique to each corresponding provider-specified [*tag scheme*](#glossary:tag-scheme).
   * Provider-specified tag key prefixes SHOULD be publicly documented.
 * User-defined tags MUST adhere to the following requirements:
+  * User-defined tag keys MUST match the provided keys, except when a *tag scheme*-specific prefix is prepended.
```
flanakin: Same comment about "except" and "provided" as allocatedtags.md

flanakin (approved):

specification/datasets/cost_and_usage/columns/allocatedtags.md
kk09v: Although this is "more important," I think we move these bullets below the description of the tag scheme-specific prefix so that we've established the concept before relying on it.

```suggestion
  * User-defined tag keys MUST match the provided keys, when the user defined tag scheme does not include a tag scheme-specific prefix. 
  * User-defined tag keys MUST match the provided keys before the tag scheme-specific prefix is prepended, when the user defined tag scheme includes a tag scheme-specific prefix.
```

kk09v: `Provide: make available for use; supply`
There are myriad different methods which a tag can be "provided." It could be user interaction in a the portal, it could be inherited from another resource/billing account, it could be in the header of an API call, etc. It doesn't seem like you take issue with the use of this word, only the fact that it hasn't been defined. I would argue that our usage of this consistent with the dictionary definition and it's unnecessary to enumerate the methods which a provider (cringe at the collision here) may allow tag keys and values to be supplied.

If there are concerns about "provided" then "supplied" is a synonym that seems to fit just as well here.

thecloudman: 🤖 [AI][Claude Code] ✅ **Implemented**

Reviewed by thecloudman

Agreed — the single-bullet "except when...prepended" phrasing leaves the prefixed case unconstrained, which contradicts the PR intent (prefix ≠ license to alter the key). Seconding this split; keeping "provided" as-is per the sibling values-bullet precedent and kk09vs rationale above.

specification/datasets/cost_and_usage/columns/tags.md
thecloudman: 🤖 [AI][Claude Code] ✅ **Implemented**

Reviewed by thecloudman

Same fix as proposed on the AllocatedTags equivalent of this bullet — the "except" branch currently leaves the prefixed case unconstrained. Splitting into two explicit bullets so the provided-key match still applies once a prefix exists, keeping "provided" as-is (matches the sibling values-bullet, `Tag values MUST match the provided values...`):

```suggestion
  * User-defined tag keys MUST match the provided keys, when the user defined tag scheme does not include a *tag scheme*-specific prefix.
  * User-defined tag keys MUST match the provided keys before the *tag scheme*-specific prefix is prepended, when the user defined tag scheme includes a *tag scheme*-specific prefix.
```

specification/datasets/cost_and_usage/columns/allocatedtags.md
Matt-Cowsert: "Provided" leaves both the source and the point in the pipeline open, and the fix folds into the same line. The wording that closes it:

* User-defined tag keys MUST match the tag keys assigned to the *tag source* when no *tag scheme*-specific prefix is prepended.
* User-defined tag keys MUST match the *tag scheme*-specific prefix followed by the tag key assigned to the *tag source* when a *tag scheme*-specific prefix is prepended.

"Assigned to the *tag source*" names where the keys come from using a term this column's introduction already links, which keeps the subject schema-level rather than naming an actor.

Worth noting for scope: "Tag values MUST match the provided values" two bullets up carries the same ambiguity, but this PR does not touch that line, so changing it would be its own edit.

specification/datasets/cost_and_usage/columns/tags.md
Matt-Cowsert: Both points land in one change here. Splitting on whether a prefix is present gives each case its own MUST, so the prefixed case is stated rather than left as an exception, and "provided" gets a source.

```suggestion
  * User-defined tag keys MUST match the tag keys assigned to the *tag source* when no *tag scheme*-specific prefix is prepended.
  * User-defined tag keys MUST match the *tag scheme*-specific prefix followed by the tag key assigned to the *tag source* when a *tag scheme*-specific prefix is prepended.
```

"Assigned to the *tag source*" answers where the keys come from using a term this column's introduction already links, which keeps the subject schema-level instead of naming an actor.

This is the same wording proposed on `allocatedtags.md`, so the two columns stay parallel. It stays in prose rather than `{scheme}{delimiter}{key}`: FOCUS does not define a delimiter separately from the prefix, and the example further down this file (`userDefinedTagScheme2/foo`) shows the delimiter living inside the prefix, so naming it separately would introduce a concept the spec does not carry today.

specification/datasets/cost_and_usage/columns/allocatedtags.md
kk09v:
```suggestion
  * User-defined tag keys MUST match the tag keys assigned to the *tag source* when no *tag scheme*-specific prefix is prepended.
  * User-defined tag keys MUST match the *tag scheme*-specific prefix followed by the tag key assigned to the *tag source* when a *tag scheme*-specific prefix is prepended.
```
To match verbiage from Tags

ijurica (approved):

jpradocueva (approved):
From the perspective of the Requirements Model, these changes are valid and well-formed.

jpradocueva:
Approved by the group without objections or abstentions during the August 6 Members’ call.

PR #2472, opened 2026-06-17
AI #2467: add condition entity to schema for rm 1.5

### LINKED ISSUE #2467: [AI] Add Condition entity to schema and folder structure for RM 1.5
### Detailed Description

Specification 1.5 has added Conditions entity, we need to setup the schema and folder structure to allow these new entities

### Definition of Done

Schema and folders updated for Condition entities

mike-finopsorg:
## Summary

Add Condition entity type and rename ApplicabilityCriteria to Conditions (RM 1.5)

Introduce the Condition entity to the 1.5 requirements model and rename the
applicability concept to "Conditions", keeping versions 1.2 to 1.4 unchanged.

Schema (1.5):
* Add "Condition" to the EntityType enum and "N" to the rule-id type class.
* Exempt Attribute and Condition rules from the non-null DatasetId/DatasetName
  requirement; non-exempt entities must carry non-null string values.
* Make DatasetId/DatasetName nullable; add DatasetType "ATT" to all attribute
  rules and the three dataset keys to every attribute entry.
* Rename the ApplicabilityCriteria key to Conditions (top-level and per-rule),
  and reshape the top-level Conditions into a registry keyed by condition
  EntityId, each linking to its CON- rule.

Conditions content:
* Add 25 condition rules sourced from specification/conditions, each a
  Composite -000-C root with -001-C (true) and -002-C (false) children. Rules
  whose MustSatisfy contains a conditional clause are named -C.
* Register all 25 conditions in conditions.json with Description and ModelRuleId.

Reference migration:
* Migrate the 163 per-rule flag references to condition EntityIds, driven by
  the authoritative column-to-condition pairs in specification/datasets.

Tooling and tests:
* Rename applicability_criteria.json to conditions.json; make build_json.py and
  build_tables.py version-aware for the renamed key.
* Add version-aware accessors in conftest and route the affected tests through
  them; split the reference-prefix test into pre-1.5 and 1.5+ variants.
* Add a registry-to-rule consistency test (1.5+) and update contributor docs.

### Type of Change
* [ ] Bug fix
* [X] New feature / Specification update
* [ ] Editorial / Formatting

### Author Checklist


jpradocueva (approved):

specification/requirements_model/releases/1.5/model_rules/datasets/cost_and_usage/columns/resourcetype.json
```diff
@@ -8,9 +8,8 @@
     "Notes": "",
     "ModelVersionIntroduced": "1.2",
     "Status": "Active",
-    "ApplicabilityCriteria": [
-      "BILLING_BASED_ON_PROVISIONED_RESOURCES_SUPPORTED",
-      "RESOURCE_TYPE_ASSIGNMENT_SUPPORTED"
+    "Conditions": [
+      "IncludesProvisionedResources"
```
ijurica: `IncludesResourceTypeAssignment` condition is missing

Quote from CostAndUsage dataset.md

>   * CostAndUsage MUST include [ResourceType](#datasets.costandusage.resourcetype) when the *operating model* [includes provisioned resources](#conditions.includesprovisionedresources) and [includes resource type assignment](#conditions.includesresourcetypeassignment).

specification/requirements_model/releases/1.5/model_rules/datasets/cost_and_usage/costandusage.json
```diff
@@ -1807,9 +1807,8 @@
     "Notes": "",
     "ModelVersionIntroduced": "1.2",
     "Status": "Active",
-    "ApplicabilityCriteria": [
-      "BILLING_BASED_ON_PROVISIONED_RESOURCES_SUPPORTED",
-      "RESOURCE_TYPE_ASSIGNMENT_SUPPORTED"
+    "Conditions": [
+      "IncludesProvisionedResources"
```
ijurica: same as [here](https://github.com/FinOps-Open-Cost-and-Usage-Spec/FOCUS_Spec/pull/2472/files#r3476454385), `IncludesResourceTypeAssignment` condition is missing

Quote from CostAndUsage dataset.md

>   * CostAndUsage MUST include [ResourceType](#datasets.costandusage.resourcetype) when the *operating model* [includes provisioned resources](#conditions.includesprovisionedresources) and [includes resource type assignment](#conditions.includesresourcetypeassignment).

specification/requirements_model/releases/1.5/model_rules/conditions/includesaggregationlevelcurrencydifferences.json
```diff
@@ -0,0 +1,86 @@
+{
```
davidearney, line 1: One finding: includesaggregationlevelcurrencydifferences.json is the only condition file with a lowercase filename (all 24 others are PascalCase). On Linux CI this will cause a mismatch if any tooling constructs the filename from the EntityId. Worth a comment before merge.

specification/requirements_model/releases/1.5/model_rules/datasets/cost_and_usage/columns/pricingcurrencycontractedunitprice.json
```diff
@@ -8,9 +8,8 @@
     "Notes": "",
     "ModelVersionIntroduced": "1.2",
     "Status": "Active",
-    "ApplicabilityCriteria": [
-      "PUBLIC_PRICE_LIST_SUPPORTED",
-      "VIRTUAL_CURRENCY_SUPPORTED"
+    "Conditions": [
+      "IncludesPricingBillingCurrencyDifferences"
```
Matt-Cowsert: This column-level `Conditions` lists only `IncludesPricingBillingCurrencyDifferences`, but PricingCurrencyContractedUnitPrice has compound, per-tier presence requirements: it is required when the operating model includes virtual currency and includes list unit prices, and recommended when it includes pricing and billing currency differences and includes list unit prices. So `IncludesVirtualCurrency` and `IncludesListUnitPrices` are both absent here. The matching `-D-` presence rules in `costandusage.json` carry the same single condition, and the MUST-tier rule's MustSatisfy text names virtual currency and unit prices that the `Conditions` array does not. The pre-1.5 `ApplicabilityCriteria` on these rules listed both conditions per tier. The single column-level array cannot capture the per-tier split on its own, so this is worth a look alongside the `-D-` rules rather than a one-line fix.

specification/requirements_model/releases/1.5/model_rules/datasets/cost_and_usage/columns/pricingcurrencyeffectivecost.json
```diff
@@ -8,9 +8,8 @@
     "Notes": "",
     "ModelVersionIntroduced": "1.2",
     "Status": "Active",
-    "ApplicabilityCriteria": [
-      "PUBLIC_PRICE_LIST_SUPPORTED",
-      "VIRTUAL_CURRENCY_SUPPORTED"
+    "Conditions": [
+      "IncludesPricingBillingCurrencyDifferences"
```
Matt-Cowsert: This column-level `Conditions` lists only `IncludesPricingBillingCurrencyDifferences`, but PricingCurrencyEffectiveCost has compound, per-tier presence requirements: it is required when the operating model includes virtual currency and includes list unit prices, and recommended when it includes pricing and billing currency differences and includes list unit prices. So `IncludesVirtualCurrency` and `IncludesListUnitPrices` are both absent here. The matching `-D-` presence rules in `costandusage.json` carry the same single condition, and the MUST-tier rule's MustSatisfy text names virtual currency and unit prices that the `Conditions` array does not. The pre-1.5 `ApplicabilityCriteria` on these rules listed both conditions per tier. The single column-level array cannot capture the per-tier split on its own, so this is worth a look alongside the `-D-` rules rather than a one-line fix.

specification/requirements_model/releases/1.5/model_rules/datasets/cost_and_usage/columns/pricingcurrencylistunitprice.json
```diff
@@ -8,9 +8,8 @@
     "Notes": "",
     "ModelVersionIntroduced": "1.2",
     "Status": "Active",
-    "ApplicabilityCriteria": [
-      "PUBLIC_PRICE_LIST_SUPPORTED",
-      "VIRTUAL_CURRENCY_SUPPORTED"
+    "Conditions": [
+      "IncludesPricingBillingCurrencyDifferences"
```
Matt-Cowsert: This column-level `Conditions` lists only `IncludesPricingBillingCurrencyDifferences`, but PricingCurrencyListUnitPrice has compound, per-tier presence requirements: it is required when the operating model includes virtual currency and includes list unit prices, and recommended when it includes pricing and billing currency differences and includes list unit prices. So `IncludesVirtualCurrency` and `IncludesListUnitPrices` are both absent here. The matching `-D-` presence rules in `costandusage.json` carry the same single condition, and the MUST-tier rule's MustSatisfy text names virtual currency and unit prices that the `Conditions` array does not. The pre-1.5 `ApplicabilityCriteria` on these rules listed both conditions per tier. The single column-level array cannot capture the per-tier split on its own, so this is worth a look alongside the `-D-` rules rather than a one-line fix.

Matt-Cowsert (commented):
Thanks @mike-finopsorg. This is a large, careful migration and the Condition entity work holds up well. The schema changes (the EntityType enum, the new rule-id class, and the DatasetId and DatasetName nullability with the Attribute and Condition exemption), the 25 condition rule files and the `conditions.json` registry (all 25 entries resolve to a real `CON-<EntityId>-N-000-C` root, with distinct descriptions), the version-aware tooling and test accessors, and the guidelines rename are all consistent.

One theme to raise, plus two confirmations of existing threads.

**Several columns carry fewer conditions than their presence requirements specify.** @ijurica already flagged ResourceType (missing `IncludesResourceTypeAssignment`). The same pattern affects the three `PricingCurrency` price columns. In each case the structured `Conditions` matches the column's overview-table cell, which holds one condition per column, rather than the column-presence requirement bullets, which is where the compound and per-tier triggers live:

| Column | `Conditions` in this PR | Presence requirements specify |
|---|---|---|
| ResourceType | `IncludesProvisionedResources` | provisioned resources and resource type assignment |
| PricingCurrencyContractedUnitPrice | `IncludesPricingBillingCurrencyDifferences` | MUST: virtual currency and list unit prices. SHOULD: pricing and billing currency differences and list unit prices |
| PricingCurrencyEffectiveCost | `IncludesPricingBillingCurrencyDifferences` | same as above |
| PricingCurrencyListUnitPrice | `IncludesPricingBillingCurrencyDifferences` | same as above |

For the three PricingCurrency columns, the MUST, SHOULD, and "all other cases" presence rules in `costandusage.json` all carry the same single `IncludesPricingBillingCurrencyDifferences`. The MUST-tier rules read "supports prices in virtual currency and publishes unit prices exclusive of discounts" but name neither `IncludesVirtualCurrency` nor `IncludesListUnitPrices`, and the SHOULD tier drops `IncludesListUnitPrices`. The pre-1.5 `ApplicabilityCriteria` on these rules listed both conditions per tier, so the second one looks lost in translation. Inline notes are on each of the three columns.

One guardrail so this does not over-correct: the four SKU columns (`SkuId`, `SkuMeter`, `SkuPriceDetails`, `SkuPriceId`) carried both `PUBLIC_PRICE_LIST_SUPPORTED` and `UNIT_PRICING_SUPPORTED` in the old model, but their presence bullets and table cells both specify only `IncludesUnitPricing`, so collapsing them to `IncludesUnitPricing` is correct and they should not gain `IncludesListUnitPrices`.

Each column's conditions live in two places that need to match: the per-column file `-000-C` root and the dataset rollup `-D-` presence rules. `test_column_000_dataset_dep_applicability` checks that those two agree, so updating only one side will surface as a test failure.

**Confirming @davidearney's filename note:** `includesaggregationlevelcurrencydifferences.json` is the only lowercase condition filename, and its in-file EntityId and rule IDs are already PascalCase, so a rename to `IncludesAggregationLevelCurrencyDifferences.json` resolves it with no content change.

**Test coverage:** nothing in the suite reconciles a column's `Conditions` against the presence requirements. `test_conditions_registry` covers registry-to-rule integrity and `test_column_000_dataset_dep_applicability` covers the two model locations agreeing, but a column whose conditions are incomplete in both places passes cleanly. A follow-up check that compares each column's conditions to its presence bullets would keep this from recurring as more columns gain conditions.

specification/requirements_model/releases/1.5/model_rules/conditions/includeslistunitprices.json
```diff
@@ -0,0 +1,86 @@
...
+    "Status": "Active",
+    "Conditions": [],
+    "Type": "Dynamic",
+    "Order": 10,
+    "ValidationCriteria": {
+      "MustSatisfy": "IncludesListUnitPrices MUST evaluate to true when IncludesUnitPricing is true and the operating model includes standard, non-discounted unit prices.",
+      "Keyword": "MUST",
+      "Requirement": {},
+      "Condition": {},
+      "Dependencies": []
```
ijurica: `IncludesListUnitPrices` normative requirements reference another Condition:

> IncludesListUnitPrices MUST evaluate to true when [IncludesUnitPricing](#conditions.includesunitpricing) is true and the operating model includes standard, non-discounted unit prices.
> IncludesListUnitPrices MUST evaluate to false when [IncludesUnitPricing](#conditions.includesunitpricing) is false or the operating model does not include standard, non-discounted unit prices.

Given the explicit reference to `IncludesUnitPricing` in the `MustSatisfy` text, should `Dependencies` point to `CON-IncludesUnitPricing-N-001-C` / `-N-002-C`? Or is the textual reference intentionally left unmodeled here?

specification/requirements_model/releases/1.5/model_rules/conditions/includeslistunitprices.json
```diff
@@ -0,0 +1,86 @@
...
+    "Status": "Active",
+    "Conditions": [],
+    "Type": "Dynamic",
+    "Order": 20,
+    "ValidationCriteria": {
+      "MustSatisfy": "IncludesListUnitPrices MUST evaluate to false when IncludesUnitPricing is false or the operating model does not include standard, non-discounted unit prices.",
+      "Keyword": "MUST",
+      "Requirement": {},
+      "Condition": {},
+      "Dependencies": []
```
ijurica: see [this comment](https://github.com/FinOps-Open-Cost-and-Usage-Spec/FOCUS_Spec/pull/2472/files#r3500796860)

specification/requirements_model/releases/1.5/model_rules/conditions/includesaggregationlevelcurrencydifferences.json
mike-finopsorg: Good find. I have corrected the file names to all lower like all the other entities.

specification/requirements_model/releases/1.5/model_rules/conditions/includeslistunitprices.json
mike-finopsorg: Resolved in commit

specification/requirements_model/releases/1.5/model_rules/conditions/includeslistunitprices.json
mike-finopsorg: Resolved in commit

specification/requirements_model/releases/1.5/model_rules/datasets/cost_and_usage/columns/pricingcurrencycontractedunitprice.json
mike-finopsorg: Fixing in commit

specification/requirements_model/releases/1.5/model_rules/datasets/cost_and_usage/columns/pricingcurrencylistunitprice.json
mike-finopsorg: Fixing in commit

specification/requirements_model/releases/1.5/model_rules/datasets/cost_and_usage/columns/pricingcurrencyeffectivecost.json
mike-finopsorg: Fixing in commit

specification/requirements_model/releases/1.5/model_rules/datasets/cost_and_usage/columns/resourcetype.json
mike-finopsorg: Fixed in commit

specification/requirements_model/releases/1.5/model_rules/datasets/cost_and_usage/costandusage.json
mike-finopsorg: fixed in commit

caligo-erik (approved):

ljadvey (approved):

thecloudman:
**`ACCOUNT_NAMING_SUPPORTED` was dropped without a replacement condition**

Old `applicability_criteria.json` had 27 flags; the new `conditions.json` has 25. One of the two removed, `ACCOUNT_NAMING_SUPPORTED`, was still in active use:

`cost_and_usage/columns/billingaccountname.json`, rule `CAU-BillingAccountName-C-003-C`:
- Before: `"ApplicabilityCriteria": ["ACCOUNT_NAMING_SUPPORTED"]`
- After: `"Conditions": []`

The `MustSatisfy` text is unchanged ("...MUST NOT be null when the invoice issuer supports assigning a display name for the billing account"), so the rule is still conditional in prose but no longer linked to a condition. The guidelines doc (line 390) also still references this flag.

Can we add a matching condition (e.g. `IncludesAccountNaming`) and re-link the rule, or confirm the drop is intentional?

*Reviewed by Graham Murphy.*

Matt-Cowsert (approved):
Thanks @mike-finopsorg. Approving. Coming back to this after the latest round of fixes, the migration holds up well.

I traced every rule that carried an `ApplicabilityCriteria` array in the 1.5 model through to its new `Conditions` array. Of 1514 such rules, all but 13 map faithfully onto the new condition vocabulary, and referential integrity is clean in both directions: every condition a rule references is present in `conditions.json`, and every registered condition is referenced by at least one rule. The model builds and the suite passes at 180 passed, 44 skipped.

Eleven of the 13 deviations are correct and deliberate:

* The four `Sku*` composite roots and `CAU-CostAndUsage-D-058-C` through `D-061-C` drop `PUBLIC_PRICE_LIST_SUPPORTED` and keep only `IncludesUnitPricing`. That matches the current dataset text: "CostAndUsage MUST include SkuId when the *operating model* includes unit pricing."
* `CAU-CostAndUsage-D-049`, `D-053`, and `D-057` move from the `-C` to the `-O` suffix and empty their condition arrays, matching their "in all other cases" MAY text. The rename leaves no dangling `ModelRuleId` references anywhere in the model.

The three `PricingCurrency` price columns, `ResourceType`, the condition file names, and the `IncludesListUnitPrices` dependency on `IncludesUnitPricing` all look right now, so the threads from the earlier rounds are resolved.

Two follow-on action items carry the rest, rather than adding work to this PR.

**#2541** picks up the two remaining deviations, both on `SkuPriceId`. `CAU-SkuPriceId-C-014-C` and `CAU-SkuPriceId-C-015-C` are gated on `IncludesUnitPricing`, where their text points at `IncludesListUnitPrices` and `IncludesNegotiatedPricing` respectively, and sibling rules carrying the same trigger clauses map to those narrower conditions. Both already carried `IncludesUnitPricing` in the commit I looked at last round, so they are not fallout from this round's fixes; I missed them the first time. That item also covers the flag vocabulary in `writing-requirements-model-guidelines.md`, where the key rename landed but the example values and the reference table are still 1.2-to-1.4 flags.

**#2540** picks up the dropped `ACCOUNT_NAMING_SUPPORTED` flag that @thecloudman raised. Confirmed: `CAU-BillingAccountName-C-003-C` went from `["ACCOUNT_NAMING_SUPPORTED"]` to `[]` while its text still reads "BillingAccountName MUST NOT be null when the invoice issuer supports assigning a display name for the billing account." The `-C` suffix stays legitimate, since the "when" clause independently justifies it under the scope-precedence rule, so there is no rule-id inconsistency. The root cause sits upstream of this PR rather than in the model: `specification/conditions/` defines 25 conditions and none of them covers account naming, so there is nothing in the current condition set for the rule to point at. Adding an `IncludesAccountNaming` condition is a specification-side change.

One more note for @thecloudman on the 27-to-25 reduction: the other flag that went away is `PRICING_BILLING_AGGREGATION_DIFFERENCES_SUPPORTED`, and zero rules referenced it, so dropping it loses nothing. `BILLING_PAYMENT_AGGREGATION_DIFFERENCES_SUPPORTED` maps correctly to `IncludesAggregationLevelCurrencyDifferences` on both of the rules that used it.

jpradocueva:
Approved by the group without objections or abstentions during the July 16 Members’ call.

PR #2473, opened 2026-06-17
AI #2415: Add scoped detail configuration

### LINKED ISSUE #2358: [FR] Add Actor dimensions to Cost and Usage
### Problem Statement

Practitioners managing SaaS and GenAI costs (e.g., Anthropic, OpenAI, GitHub Copilot) report that attribution is the primary driver for chargeback and cost allocation. However, modern cloud and AI architectures often suffer from a "multiplexer" problem: the entity authorized to call the service (the infrastructure Principal) is frequently different from the entity benefiting from the service (the downstream Consumer).

Currently, FOCUS lacks a standardized mechanism to distinguish between these two layers of attribution. If practitioners attempt to use a single column or custom extensions, they face two significant risks:
1. **The Infrastructure Blind Spot:** Overwriting the system actor (IAM Role/Service Account) with a human user ID destroys the audit trail for DevOps and Security.
2. **The PII Compliance Trap:** Ingesting raw user data (names, emails) into billing datasets introduces significant compliance risks (GDPR, CCPA).

Practitioner quotes (hypothetical):
- "Our security team blocked the ingestion of AI usage reports because the API payload contains employee email addresses; we need a standard for opaque actor strings."
- "My AWS bill shows a Service Role spent $50k on Bedrock, but I have no way to attach the 1,000 individual tenants who actually sent the prompts without breaking the schema."

The gap is structural: FOCUS needs a standardized, multi-layered, and privacy-conscious way to represent the **Actors** (system and end-actor) responsible for a charge.

---

### Use Case / User Story

**Story 1: Resolution of Asymmetric Attribution (The Multiplexer)**
**As a** FinOps practitioner managing a shared AI platform
**I need** separate columns for the System Actor (`PrincipalId`) and the End-Actor (`ConsumerId`)
**So that** I can simultaneously audit which system credentials authorized the spend and which internal business unit consumed the output.

**Story 2: Secure Business-Level Chargeback**
**As a** FinOps practitioner managing SaaS and AI platforms
**I need** a standardized column to capture non-PII, opaque consumer identifiers
**So that** I can securely map costs back to internal cost centers via our IdP without exposing personal data in the raw billing dataset.

---

### Success Criteria

- [x] Two net-new columns (`PrincipalId` and `ConsumerId`) are added to the Cost and Usage dataset.
- [x] Formal glossary definitions for **Principal** and **Consumer** are added to establish the boundary between infrastructure-layer and application-layer actors.
- [x] Normative BCP-14 requirements explicitly mandate the use of opaque identifiers and prohibit plain-text PII when an alternative identifier is available.
- [x] An informative "Note" blockquote is included regarding pseudonymized data and privacy framework compliance (GDPR Recital 26).

---

### Type of Request

- [ ] **Standardization** 
- [ ] **Enhancement**
- [x] **Net New**
- [ ] A concept not yet represented in FOCUS
- [ ] **Enablement/Supporting Content**
- [ ] **Tech Debt** 

---

- FinOps Foundation 

---

SaaS and AI vendors increasingly support opaque identifiers. The Anthropic Enterprise Analytics API provides a `user_id` string alongside PII. AWS identifies assumed roles and session names via ARNs. By providing two columns, FOCUS aligns with the native data structures of both traditional CSPs and modern AI APIs.

---

**Target Scope for Actor Attribution:**

| Actor Type | Column | Example Opaque Value (Safe) | Example PII (Unsafe) |
|---|---|---|---|
| **System Actor** | `PrincipalId` | `arn:aws:iam::123:role/ChatBot-Prod` | N/A |
| **End-Actor** | `ConsumerId` | `user_01AbCdEfGh` | `jane.smith@company.com` |
| **Service Account**| `PrincipalId` | `svc-finops-automation@project.iam` | N/A |
| **SSO Identity** | `ConsumerId` | `9a8b7c6d-1234-abcd-9876543210ef` | `John Doe` |

---

1. **Define Glossary Terms**: Establish "Principal" and "Consumer" as the two layers of actor-based attribution.
2. **Add new column definitions**: `principalid.md` and `consumerid.md`.
3. **Draft Normative Rules**: 
   - Both columns MUST be String type and nullable.
   - `ConsumerId` MUST NOT contain plain-text PII if an alternative opaque identifier is available.
4. **Draft Informative Guidance**: Cross-reference the columns to explain how they resolve the "multiplexer" attribution problem in a privacy-compliant manner.

---

**Why "Actor" and "Attribution" instead of "Identity":** The term "Identity" is a high-gravity word that often triggers unnecessary legal/compliance friction. "Actor" is a technically accurate, clinical term that describes any entity (machine or human) interacting with the system, and "Attribution" focuses on the FinOps utility rather than the personal nature of the data.

---

---

### Adoption Impact

- [x] **Adoption Accelerator** – This feature would help organizations start using FOCUS for AI and SaaS.

**Justification:**
As FinOps expands beyond IaaS, user-level and system-level attribution are the primary vectors for cost analysis. Without a safe, standard way to ingest the "who" and "what," enterprise security teams will block the ingestion of these critical datasets. This framework provides the necessary guardrails to accelerate FOCUS adoption in the AI/SaaS space.

### Specification Impact

**Impacted Parties:**
- [x] FinOps Practitioner
- [x] FOCUS Data Generator
- [x] Vendor Supporting FOCUS

**Level of Ambiguity: 3**
While the schema change is straightforward, introducing the concept of pseudonymized data and multi-layer attribution sets a new precedent for FOCUS that requires careful alignment between technical and legal stakeholders.

---

### FinOps Scope Alignment

- [x] Public Cloud
- [x] Software-as-a-Service (SaaS)
- [x] AI 
- [x] Licensing

### LINKED ISSUE #2415: [AI] Draft Delivery Grain attribute PR for FOCUS
**Assignee(s):** @finops-DaveMoreau
**Parent Issue:** #2358

**Detailed Description**
TF2 discussed how high-cardinality actor columns (Principal ID, Consumer ID) under FR #2358 can dramatically increase FOCUS dataset size, and that adding them SHOULD be an intentional, opt-in act rather than a default. The group reached directional consensus to document a "delivery grain" / opt-in concept (modeled on the existing split cost allocation opt-in mechanism), but rejected the current framing that places it inside a column's "content constraints" section. It is unresolved whether the concept belongs as a new standalone attribute or as a subset of the existing Dataset Configuration attribute. Dave Moreau volunteered to draft a rough PR proposing how to model it. Matt committed to synthesizing the discussion into a scoping write-up and sharing it with Dave on 2026-05-27 (the basis for this AI). This PR is expected to be a prerequisite for PR #2360 (Principal ID / Consumer ID columns) to gain approval and merge.

**Definition of Done**
- [ ] Draft PR opened (kept in draft) tied to FR #2358 proposing how to model the delivery grain / opt-in concept
- [ ] Placement options explored (new standalone attribute vs. subset of the Dataset Configuration attribute) rather than committing prematurely
- [ ] Prerequisite relationship to PR #2360 captured in the PR description

finops-DaveMoreau:
## Summary

This PR extends the Dataset Configuration attribute with scoped detail configuration for FOCUS 1.5.

Data generators can make optional, higher-cardinality detail available for documented areas of a FOCUS dataset. Practitioners can select the detail needed for those areas without FOCUS defining a request payload, property name, or transport mechanism. A provider can expose the selection through an API parameter, export setting, query interface, or another access mechanism.

The glossary gains four entries: Companion Artifact, Detail Representation, Detail Scope, and Detail Variant. A detail scope is a subset of records, identified by values of FOCUS columns representing dimensions (e.g., ServiceName, ResourceType), for which more than one detail variant is offered. The records delivered when no additional detail is selected count as a detail variant.

The normative requirements establish:

* When a detail scope is offered, a FOCUS dataset is configurable to select each offered detail variant and includes only one detail variant for each detail scope.
* The user-defined selection of columns includes the columns documented for a selected detail variant, so those columns are delivered without an exception to the column-selection requirement.
* A FOCUS dataset is configurable to select one detail representation when more than one is offered for a selected detail variant. Whether that selection applies to the complete dataset or to each detail scope is left to the data generator.
* Detail scope documentation includes the column values that identify each detail scope, the offered detail variants and the columns each populates (FOCUS or custom), and whether each variant uses Data Generator-Calculated Split Cost Allocation Handling.
* Detail scope documentation includes the offered detail representations, their relationship to other records representing the same data, and the columns that relate companion artifact records to dataset artifact records. The documentation is accessible to practitioners.
* The sum of each summable metric column for the records in a detail scope is preserved across all offered detail variants, so variants of the same scope reconcile to the same totals.
* Records with identical values in all delivered columns other than summable metrics, including custom columns, should be represented as a single record while preserving the aggregate value of each summable metric column.

The supporting content describes illustrative detail representations:

* Inline: additional populated columns on the same records
* Expanded: more detailed records replace the less detailed records in the same dataset artifact
* Companion artifact: detail delivered in a separate file or table whose structure FOCUS does not define

It also explains that Data Generator-Calculated Split Cost Allocation Handling is a defined subset of scoped detail configuration, and that record-minimization guidance does not replace column-specific aggregation guidance, including the guidance for `PricingQuantity`, `ListCost`, and `ContractedCost`.

### Open Questions for TF-2

* **Term names.** This PR uses detail variant and detail representation, as @ijurica proposed, in place of detail level and delivery method. "Level" implies an ordinal scale and collides with "Feature level" in Content Constraints tables. "Delivery method" sits next to DeliveryHandling's delivery mechanism. The Replacement representation is renamed Expanded because Replacement is a CorrectionHandling correction style. Data coverage is folded into the Detail Scope definition rather than defined separately. These names are the author's proposal; TF-2 can confirm them or choose others.
* **Default output as a detail variant.** The Detail Variant entry counts the default output as a variant. Without that, a detail scope with only one optional variant would not trigger the configurability requirement, and the opt-in would not be guaranteed. TF-2 can confirm or revise this.
* **Column selection and detail variant columns.** The column-selection requirement, introduced in 1.4, says a FOCUS dataset is configurable to include only a user-defined selection of columns, while selecting a detail variant brings in the columns documented for that variant. This PR treats those columns as part of the user-defined selection: a column list and a detail variant choice are both user configuration, so the delivered column set is the union of the two and "only" keeps its force. The 1.4 requirement is therefore unchanged, and `ATT-DatasetConfiguration-A-001-M` stays accurate. Alternatives: add `except when a column is documented for a selected detail variant` to the column-selection requirement, which states the carve-out explicitly but amends text introduced in 1.4; or leave both requirements unqualified and rely on the detail variant requirement taking precedence implicitly, which a validator reading A-001 alone would treat as a conflict. Which resolution does TF-2 prefer?
* **Detail scopes and delivery scopes.** The Delivery Scope entry allows non-temporal, logical groupings, so a detail scope could be read as one kind of delivery scope. This PR keeps them separate because the delivery scope requirements would misapply. DeliveryHandling requires a complete snapshot for each delivery scope under Overwrite, and CorrectionHandling places Delta and Ledger corrections within a delivery scope. Treating each detail scope as a delivery scope would make it a separate unit of supersession and correction. Records in a companion artifact are also outside any dataset artifact delivery. A detail scope instead determines which records have selectable detail variants. Should the two stay separate, and should Detail Scope be renamed to avoid confusion with Delivery Scope?
* **Changing a detail variant within a delivery scope.** Under Overwrite, the next dataset artifact supersedes earlier ones, so a new variant selection takes effect cleanly. Under Append (Delta or Ledger), dataset artifacts delivered at the previous variant remain alongside records at the new variant within the same delivery scope, so the combined artifacts include more than one detail variant for a detail scope. Restating earlier records at the new variant would also count summable metrics twice unless the earlier records are reversed. Should a variant change take effect only at a delivery scope boundary (e.g., the next billing period), require reversal records, or be left to documentation?
* **Companion artifacts and native datasets.** A companion artifact's structure is not defined by FOCUS, which makes it close to a native dataset artifact. DatasetCompleteness requires custom columns for all native dataset columns unless an exclusion is documented. If companion artifact columns count as native dataset columns, the detail they hold (e.g., `x_UserId`) would be required in the FOCUS dataset, which undoes the opt-in. Should the glossary state that companion artifact columns are not native dataset columns for this purpose, or define a companion artifact as native data? This decision also governs the Schema metadata column list: ColumnDefinition covers every column provided in dataset instance artifacts, so companion artifact columns sit outside it today only because a companion artifact is not a dataset artifact.
* **Detail variants and dataset instances.** The Dataset Instance entry already covers instances that differ by custom column inclusions. Does selecting a detail variant produce a distinct dataset instance, or is it configuration of one instance? The answer determines when a new Schema metadata entry is required and whether the dataset-instance merge rules in BilledCost and EffectiveCost apply.
* **PrincipalId and the default column set.** PrincipalId (#2360) is required when IncludesRequesterAttribution applies, and a default column set must include all applicable FOCUS columns. Together, high-cardinality actor detail is in every offered default column set. Options: exempt columns populated only by a non-default detail variant, tie PrincipalId to a selected detail variant, or accept PrincipalId in the default column set.
* **Record minimization and Ledger corrections.** In the Ledger correction style, a reversal and its corrected record can share every non-summable value. The record-minimization SHOULD would merge them into a net record and lose the audit trail. Should the requirement carry an exception, or is a supporting content note enough?

### Type of Change

* [ ] Bug fix
* [ ] Editorial / Formatting

### Author Checklist


finops-DaveMoreau:
Pushed for visibility and discussion. There are still parts that I am cleaning up, but I won't have time before TF-2 meeting.

specification/attributes/dataset_configuration.md
```diff
@@ -30,6 +45,8 @@ A practitioner configures their FOCUS Cost and Usage dataset to include only the
 
 Even though columns like `CommitmentDiscountId` and `ResourceId` are excluded, the included cost columns (`BilledCost`, `EffectiveCost`) still reflect commitment discounts correctly. The dataset remains conformant to the FOCUS specification because each included column follows all requirements for that column, including requirements that reference columns not in the dataset.
 
+Another practitioner configures their FOCUS Cost and Usage dataset to receive a dataset instance with different granularity configurations for different sets of records. The dataset instance includes records for AI services configured with `PrincipalId` and `ConsumerId`, records for another service configured with `PrincipalId`, and records for other services configured without actor-level columns. The data generator documents the applicability filter for each granularity configuration and describes how cost metrics are represented when a separate delivered dataset instance represents the same underlying usage or charges.
```
Matt-Cowsert: `PrincipalId` and `ConsumerId` are used here as concrete columns, but neither exists in `working_draft` yet. Per issue #2415, they are introduced in PR #2360, which is expected to merge after this change, so this example would reference columns the spec does not define at the time this lands. Consider phrasing the actor columns as illustrative (e.g., "future actor-level identifiers such as PrincipalId and ConsumerId"), or sequencing this change with #2360 so the columns exist when the example ships. The same applies to the `PrincipalId`, `ConsumerId`, and `SessionId` references in the supporting content.

specification/attributes/dataset_configuration.md
```diff
@@ -1,20 +1,35 @@
...
 ## Requirements
 
 Dataset conforming to DatasetConfiguration attribute MUST adhere to the following requirements:
 
 * *FOCUS dataset* MUST be configurable to include only a user-defined selection of columns.
-* *FOCUS dataset* MUST adhere to all column-level specifications defined in the FOCUS schema, regardless of the user's chosen configuration (e.g., column selection).
+* *FOCUS dataset* MUST be configurable to select a granularity configuration for each applicability filter when more than one granularity configuration is offered for that applicability filter.
+* *FOCUS dataset* MUST adhere to all column-level specifications defined in the FOCUS schema, regardless of the selected or default configuration (e.g., column selection, granularity configuration).
+* *FOCUS dataset* MUST have mutually exclusive applicability filters for granularity configurations within a delivered dataset instance.
+* *FOCUS dataset* granularity configuration documentation MUST adhere to the following requirements when a delivered dataset instance uses one or more granularity configurations:
```
Matt-Cowsert: These nested bullets make "granularity configuration documentation" the subject and use identify and describe as the verbs. AGENTS.md asks normative subjects to be schema-level entities (e.g., FOCUS dataset) and to describe a verifiable state rather than a behavior. Data Generator-Calculated Split Cost Allocation Handling is a useful model here: it states the documentation obligation in its non-normative Description and keeps the Requirements bullets on column states (e.g., "the sum of FOCUS dataset column across allocated charges MUST match..."). Consider moving these documentation obligations into supporting prose, or restating them as states, so the normative section stays state-based.

specification/attributes/dataset_configuration.md
```diff
@@ -1,20 +1,35 @@
...
 
 Common scenarios where dataset configuration is valuable include:
 
 * **Managing Scale**: Trim large exports to reduce time and cost of data preparation
 * **Reducing Noise**: Tailor datasets for specific workflows (e.g., cost allocation, commitment analysis)
+* **Managing Granularity**: Select dataset instances that expose the level of detail needed for a workflow (e.g., service-level, resource-level, actor-level)
 * **Lowering Barriers**: Strip away technical complexity for spreadsheet users
 * **Enabling Comparison**: Remove custom (`x_`) columns for standardized cross-provider reporting
 
+Dataset instance granularities identify the levels of detail represented by records in a [*dataset instance*](#glossary:dataset-instance). Each granularity configuration defines one level of detail for records that match a documented applicability filter (e.g., ServiceName).
```
Matt-Cowsert:
```suggestion
Dataset instance granularities identify the levels of detail represented by records in a dataset instance. Each granularity configuration defines one level of detail for records that match a documented applicability filter (e.g., ServiceName).
```

AGENTS.md links a glossary term only on its first occurrence per source file. `dataset instance` is already linked in the opening paragraph ("...selecting [*dataset instances*](#glossary:dataset-instance)..."), so this later mention can stay plain text.

Matt-Cowsert (commented):
Thanks for getting this up before TF-2, David. The draft frames the granularity problem well, and modeling high-cardinality detail as a Dataset Configuration choice rather than a per-column flag fits cleanly with the existing column-selection mechanism. The six open questions are well chosen for task force discussion. A few observations on scope and on the normative requirements, plus one small linking fix.

**Scope against issue #2415.** The issue's Definition of Done asks for placement options to be explored, a new standalone attribute versus a subset of the Dataset Configuration attribute, rather than settling on one early. This draft extends Dataset Configuration with normative requirements, which reads as having chosen the subset approach. It would help task force readers to see the standalone-attribute alternative represented alongside it, or a short note on why the subset placement won out, so the placement decision stays visible going into TF-2.

**Forward references to columns not yet in the spec.** `PrincipalId`, `ConsumerId`, and `SessionId` appear as concrete columns in the Example and in the supporting content, but none of these exist in `working_draft` today. Per issue #2415 they arrive in PR #2360, which is expected to merge after this change. As written, those references would point at columns the spec does not yet define once this lands. Two options keep the examples self-consistent: phrase the actor columns as clearly illustrative (e.g., "a future actor-level identifier such as PrincipalId"), or sequence this change together with #2360. Noted inline on the Example too.

**Documentation obligations written as normative requirements.** The nested "granularity configuration documentation MUST/SHOULD identify/describe..." bullets make documentation the subject and use identify and describe as the verbs. AGENTS.md asks normative subjects to be schema-level entities and to describe a state rather than a behavior. The closest precedent, Data Generator-Calculated Split Cost Allocation Handling, keeps the documentation obligation in its non-normative Description and reserves the Requirements bullets for column states. Worth settling whether these obligations live as normative requirements or as supporting prose while the draft is still forming. Detail inline.

specification/attributes/dataset_configuration.md
finops-DaveMoreau: I used those column names because Shawn use them in his slide at the day 2 keynote at FinOps X. I can make up different names, but these were meant as just examples that would make sense for the scope that the cloud provider defined for this granularity config. All we can infer from this is that the columns specified aren't mandatory FOCUS columns because they are necessarily not mandatory if they are opt-in.

specification/attributes/dataset_configuration.md
```diff
@@ -14,7 +15,17 @@ Common scenarios where dataset configuration is valuable include:
...
+* *FOCUS dataset* MAY offer configurable detail levels for provider-defined detail scopes.
+* *FOCUS dataset* detail-level configuration MUST be a mapping of provider-defined scope keys to detail level values when configurable detail levels are offered.
+* *FOCUS dataset* detail-level value MUST be either a string identifying the requested level or an object containing a `Level` property (string) and an optional `DeliveryType` property (string).
+* *FOCUS dataset* MUST include dimension columns required by a selected detail level even when those columns are not present in the selected column list.
+* *FOCUS dataset* detail-level documentation MUST adhere to the following requirements when configurable detail levels are offered:
+  * *FOCUS dataset* detail-level documentation MUST include all offered scope keys and their allowed level values.
+  * *FOCUS dataset* detail-level documentation MUST include the effect of each detail level on dataset records, including columns added, effect on row counts, and relationship to other delivered dataset instances that represent the same underlying usage or charges.
```
Matt-Cowsert: This bullet covers three separate things: the columns added, the effect on record counts, and the relationship to other delivered dataset instances. AGENTS.md asks each requirements bullet to state one verifiable thing and to split bullets that combine multiple obligations. Splitting also makes each part testable on its own:

```suggestion
  * *FOCUS dataset* detail-level documentation MUST include the columns added by each detail level.
  * *FOCUS dataset* detail-level documentation MUST include the effect of each detail level on record counts.
  * *FOCUS dataset* detail-level documentation MUST include the relationship of each detail level to other delivered dataset instances that represent the same underlying usage or charges.
```

supporting_content/attributes/dataset_configuration.md
```diff
@@ -259,7 +341,7 @@ Major cloud providers support various configuration options:
 
 ## Configuration Metadata
 
-The Dataset Configuration attribute requires that FOCUS datasets include metadata describing the selected configuration options (`DatasetConfiguration-A-003-M`). This section evaluates what changes would be needed to support this requirement within the existing metadata structure.
+The Dataset Configuration attribute requires documentation for offered granularity configurations, but it does not yet define structured metadata for all selected configuration options. This section evaluates what changes would be needed to support structured configuration metadata within the existing metadata structure.
```
Matt-Cowsert:
```suggestion
The Dataset Configuration attribute requires documentation for offered detail level configurations, but it does not yet define structured metadata for all selected configuration options. This section evaluates what changes would be needed to support structured configuration metadata within the existing metadata structure.
```

This line uses "granularity configurations," the old name. The normative file and the rest of this section now call the feature "detail level configuration."

supporting_content/attributes/dataset_configuration.md
```diff
@@ -274,36 +356,59 @@ The FOCUS metadata system has four sections:
...
 
-### What Needs to be Tracked
+### What Needs to Be Tracked
 
 Configuration metadata should describe the options applied when generating a dataset artifact:
 
 | Configuration Option | Metadata Needed                                                |
 |----------------------|----------------------------------------------------------------|
 | Column selection | List of included columns (or excluded columns) |
+| Granularity configuration | Selected level of detail, selection method, applicability filter, representation mode, privacy-sensitive identifier columns, and relationship to dataset instances representing the same underlying usage or charges |
```
Matt-Cowsert:
```suggestion
| Detail level configuration | Selected detail levels per scope key (level and delivery-type), columns added by each level, privacy-sensitive identifier columns, and relationship to dataset instances representing the same underlying usage or charges |
```

This row still uses the old vocabulary (granularity configuration, applicability filter, representation mode). The rest of this PR replaces those with detail level configuration, scope keys, levels, and delivery types, so the row names a feature the normative file no longer defines.

Matt-Cowsert (commented):
David, this is a real simplification over the original applicability-filter design, and it reads more cleanly. The two prior threads are resolved by the redesign: the actor columns in the example are illustrative rather than concrete forward references, and the documentation obligations now read as states on FOCUS dataset detail-level documentation rather than identify and describe behaviors. Closing both, plus the glossary-link thread whose target text was removed. Two structural questions on the normative requirements, since that is the part to get right.

**Whether the opt-in is required.** Every detail-level requirement says the provider MAY offer detail levels, so a provider can choose not to offer the opt-in at all. The reason to model this as configuration is to let the practitioner decide when to take on the extra rows. When offering it is optional, that control is not guaranteed. Split cost allocation, which the group pointed to as the model, uses MUST requirements rather than MAY. A conditional requirement would cover this: when a provider can deliver a finer detail level for a scope, the FOCUS dataset MUST be configurable to request it. The least-granular-default rule you already have would pair with it. Did you intend a conditional MUST here, or is keeping it fully optional deliberate for a first pass?

**Whether to define the request format.** The requirements also spell out the format of the configuration request: that the DetailLevel field is a mapping of scope keys to levels, and that each value is either a string or an object with Level and DeliveryType properties. Every other Dataset Configuration requirement describes what the dataset or its columns look like, not the format of the request the practitioner sends. Worth deciding whether the spec should define the request format at all, or instead require only that the capability and its documentation exist, and leave the request shape to the supporting content and any future configuration metadata. Split cost allocation is a good comparison again: it sets rules for the resulting data, not for how a practitioner asks for it.

Two smaller cleanups noted inline: stale terminology in the supporting content's Configuration Metadata section that did not carry over to the new vocabulary, and one requirements bullet that bundles several obligations.

specification/attributes/dataset_configuration.md
Matt-Cowsert: Resolved by the redesign. The example now uses a generic user identifier column and provider-defined scope keys instead of naming PrincipalId, ConsumerId, or SessionId, so it no longer points at columns the spec has not defined yet. Closing this.

specification/attributes/dataset_configuration.md
Matt-Cowsert: The redesign addresses this. The documentation requirements now read as states on FOCUS dataset detail-level documentation (MUST include ...) rather than making documentation the actor of identify and describe. Closing this; one residual on a bullet that bundles several obligations is noted in the review summary.

specification/attributes/dataset_configuration.md
Matt-Cowsert: The sentence this suggestion targeted was removed in the redesign, so the duplicate dataset instance link is gone. Closing this.

ijurica:
### Determinism of the detail level mechanism

Introducing a parallel mechanism such as a "detail level" as a convenience layer on top of explicit column selection is a good idea. However, I believe that the mapping between a requested detail level and the resulting dataset artifact must remain deterministic, at least for FOCUS columns.

To achieve this, FOCUS would need to:

- explicitly define the allowed detail level values within the FOCUS specification itself;
- define which FOCUS columns are associated with each detail level (potentially through a many-to-many relationship between detail levels and columns);
- allow providers to include custom columns, as long as those additions are documented in provider-specific documentation.

Otherwise, "detail level" becomes a provider-specific alias layer rather than a standardized FOCUS capability.

***Note:** I am not sure whether detail level can be adequately addressed within the existing Dataset Configuration attribute. My current view is that an additional FOCUS entity may be required to properly define and govern this concept.*

### Record aggregation in dataset artifacts

AI surfaced this topic because some of the columns we are considering have exceptionally high cardinality, but the underlying requirement is broader. It applies equally to all FOCUS and custom columns, regardless of what they represent. 

My expectation was that the Dataset Configuration attribute would be extended to enable optimization of the resulting dataset artifact size after the user-defined column selection has been applied.

A draft version of the additional requirement:

> * FOCUS dataset SHOULD represent records sharing identical values across all dimensions and non-summable metrics (e.g., unit prices) in the user-defined selection of columns as a single record with summed values for all summable metrics (e.g., costs and quantities).

***Note on double-counting:** There is one concern I have not yet fully thought through — how the proposed aggregation would interact with columns where double-counting is a known issue, specifically PricingQuantity, ListCost, and ContractedCost. These are all classified as summable metrics, so under the current draft they would be summed across aggregated records — but that would compound the double-counting problem rather than reduce it. This needs further consideration*

In [this spreadsheet](https://docs.google.com/spreadsheets/d/133QAFqgS8xMIn2Fd4ykOM0zmdfsX9XVdzl_iFKGYsOM/edit?gid=0#gid=0), I have described several scenarios and the expected resulting datasets using sample data.

Matt-Cowsert:
## TF2 Action Item - 2026-07-15

PR #2473 (draft, under the now-closed AI #2415) adds a detail-level configuration to the Dataset Configuration attribute so practitioners can opt into finer-grained records for specific cost areas. In review, Irena raised that detail levels should deterministically define which FOCUS columns are included at each level rather than leaving it to provider discretion, which may call for a new FOCUS entity to model the mapping. She also asked for a requirement that, when high-cardinality columns are excluded, data is aggregated to a level that minimizes dataset size. Dave agreed to add a draft requirement covering the aggregation point so the group can debate concrete text. The determinism question is scoped for the next TF2 session.

**Scope:**
- [ ] Add a draft requirement that, when high-cardinality columns are excluded, data is aggregated to a level that minimizes dataset size

**Assignee:** @finops-DaveMoreau

Matt-Cowsert:
## TF2 Action Item - 2026-07-22

The group agreed to move away from the granularity framing on this PR, which reads as time granularity and caused confusion. The mechanism the group asked for is a service-specific opt-in: a data generator documents, using FOCUS dimension criteria, which columns are populated for which services. The criteria referenced in that documentation should be FOCUS dimensions (e.g., Service Name, Resource Type) so a practitioner can verify what was promised against the delivered data, while the extra detail columns the generator adds are not required to be FOCUS columns. This pairs with the dual opt-in discussed on the Principal ID work: the generator opts in and documents the coverage, then the consumer opts in to include those columns. Dave has already stepped back from the granularity wording and will rework the PR next week, after going through Irena's notes and reviewing similar patterns already in the spec. Continues AI #2415.

**Scope:**
- [ ] Replace the granularity terminology with a service-specific opt-in framing for which columns are populated per service
- [ ] State that the criteria used to document coverage reference FOCUS dimensions, while the added detail columns need not be FOCUS columns
- [ ] Align the wording with the dual opt-in model (generator documents coverage, consumer opts in to include)
- [ ] Resolve whether Service Name is the right basis or only an example, given its indirect tie to the SKU
- [ ] Incorporate Irena's existing review notes on the PR

**Assignee:** @finops-DaveMoreau

Matt-Cowsert:
## TF2 Action Item - 2026-07-29

@finops-DaveMoreau pushed the detail configuration requirements ahead of the TF2 call, covering detail scope, detail level, aggregation of identical values, the columns documented per level, delivery method selection, and join keys when detail arrives in separate files.

The stated reason for staying in draft was not understanding why split cost allocation and detail configuration are separate concepts, when detail configuration always splits costs, and when a practitioner should reach for one over the other. @ijurica answered it: split cost allocation predates detail configuration and is a narrower, better defined subset of the same pattern. The same sequence played out with commitment discount IDs for reserved instances and savings plans before the broader case was covered. @ijurica then said the PR reads as ready to come out of draft.

Two wording cautions came out of the discussion. FOCUS defines only the Cost and Usage dataset today, so detail delivered in a separate dataset would be an `x_` dataset, and split cost allocation could not be delivered separately under the current dataset set. @finops-DaveMoreau noted that as FOCUS defines more of these columns, `PrincipalId` and `ConsumerId` among them, the assumption that split targets are non FOCUS fields stops holding. @Matt-Cowsert said some providers already deliver split cost allocation outside Cost and Usage, that the guidance should state the expectation without discarding usefulness where providers differ, and that parts of this may be pulled back for a later release.

@finops-DaveMoreau also raised whether a data generator using split cost allocation should be required to declare it in the detail configuration. @Matt-Cowsert called that a reasonable request. No decision was taken.

**Scope:**
- [ ] Reference split cost allocation as a defined subset of detail configuration, so a reader can tell which one applies
- [ ] Check the separate dataset wording against FOCUS defining only Cost and Usage today, so the `x_` dataset constraint reads correctly
- [ ] Decide whether split cost allocation has to be declared in the detail configuration, and record the outcome on the PR
- [ ] Move PR #2473 out of draft

**Assignee:** @finops-DaveMoreau

finops-DaveMoreau:
Implemented the TF2 follow-ups from 2026-07-29:

* Clarified that service-specific detail configuration is the broader opt-in/documentation mechanism and Data Generator-Calculated Split Cost Allocation Handling is a defined subset for detail levels that split an origin charge into allocated charges.
* Added a DatasetConfiguration documentation requirement to disclose whether each offered detail level uses DataGeneratorCalculatedSplitCostAllocationHandling.
* Reworded separately delivered detail as a provider-defined companion artifact unless FOCUS defines a standard dataset for that detail.
* Updated the PR summary and working notes to match the current service-specific detail framing.

Validation run:

* Targeted markdown lint passed for the changed spec/supporting/working files.
* Requirements model tests passed: 180 passed, 44 skipped.
* Full `make` did not complete locally because `pandoc` and `wkhtmltopdf` are not installed in this environment. (I will try to install tonight)

finops-DaveMoreau:
Renamed the feature terminology from service-specific detail configuration to scoped detail configuration so the mechanism is not tied to `ServiceName` or a single service boundary.

The supporting content now states that a detail scope can apply to one service, multiple services, or records defined by non-service dimension criteria. `ServiceName` remains only an example of a FOCUS dimension criterion.

specification/attributes/dataset_configuration.md
```diff
@@ -40,8 +57,8 @@ Dataset Configuration
 
 ## Description
 
-Defines configuration options for controlling the structure and content of a FOCUS dataset.
+Defines configuration options for controlling the structure, content, and scoped detail of a FOCUS dataset.
 
 ## Version Introduced
 
-1.4
+1.4 (column selection); 1.5 (scoped detail configuration)
```
qquigley, line 64: I think editorial guidelines and precedent support this line staying as `1.4` with additional changes marked in the 1.5 changelog. Not a hill I'll die on or even scrape my knee on, but something that stuck out to me.

specification/attributes/dataset_configuration.md
```diff
@@ -1,20 +1,33 @@
...
 * **Managing Scale**: Trim large exports to reduce time and cost of data preparation
 * **Reducing Noise**: Tailor datasets for specific workflows (e.g., cost allocation, commitment analysis)
+* **Managing Detail**: Include optional detail for cost areas where detailed attribution is needed (e.g., per-user or per-feature costs for a shared service)
 * **Lowering Barriers**: Strip away technical complexity for spreadsheet users
 * **Enabling Comparison**: Remove custom (`x_`) columns for standardized cross-provider reporting
 
 ## Requirements
 
-Dataset conforming to DatasetConfiguration attribute MUST adhere to the following requirements:
+FOCUS dataset conforming to DatasetConfiguration attribute MUST adhere to the following requirements:
```
qquigley: What's the rationale for adding "FOCUS" here? If I'm understanding correctly, the [normative requirements](https://github.com/FinOps-Open-Cost-and-Usage-Spec/FOCUS_Spec/pull/2473/changes#r3800388224) say that the original use of just "Dataset" is correct here.

finops-DaveMoreau: Agreed. That seems dumb. I'll fix.

finops-DaveMoreau:
Reverted the DatasetConfiguration requirements anchor back to `Dataset conforming to DatasetConfiguration attribute...`. I agree there was no ambiguity worth resolving there; the attribute name and repository context already make the scope clear, and this restores the existing requirement-anchor pattern.

specification/attributes/dataset_configuration.md
Matt-Cowsert, line 64: ## TF2 Action Item - August 19, 2026

@qquigley commented that editorial guidelines and precedent keep the version introduced at 1.4, with the additions recorded in the 1.5 changelog. @finops-DaveMoreau raised the underlying question in the meeting: Dataset Configuration was introduced in 1.4, and the individual properties defined inside it do not carry their own introduction version.

@matt-cowsert pointed at the precedent for columns introduced in one release and updated in another. @finops-DaveMoreau noted this is an attribute rather than a column, and @ijurica confirmed attributes follow the same practice, holding the version introduced at first introduction even where the attribute later picked up material changes. @finops-DaveMoreau agreed to follow @qquigley's comment.

**Scope:**
- [ ] Keep the version introduced at 1.4 on the Dataset Configuration attribute

**Assignee:** @finops-DaveMoreau

specification/attributes/dataset_configuration.md
```diff
@@ -14,7 +15,19 @@ Common scenarios where dataset configuration is valuable include:
 Dataset conforming to DatasetConfiguration attribute MUST adhere to the following requirements:
 
 * *FOCUS dataset* MUST be configurable to include only a user-defined selection of columns.
-* *FOCUS dataset* MUST adhere to all column-level specifications defined in the FOCUS schema, regardless of the user's chosen configuration (e.g., column selection).
+* *FOCUS dataset* MUST adhere to all column-level specifications defined in the FOCUS schema, regardless of the selected configuration (e.g., column selection or detail level).
+* *FOCUS dataset* MUST be configurable to select one detail level for a detail scope when the same data coverage can be delivered at more than one detail level.
```
Matt-Cowsert, line 19: `detail scope`, `detail level`, `delivery method`, and `data coverage` carry the ten MUST bullets in this list, and none of the four is defined in anything that ships. They are defined in `supporting_content/attributes/dataset_configuration.md`, which `specification/spec.mdpp` does not include, and the glossary has no entry for any of them. The non-normative intro at the top of this file does not define them either. `delivery method` has a second problem: it sits a page from DeliveryHandling's `delivery mechanism`, which means Overwrite and Append. Glossary entries would be one way to close this, definitions in this file's intro another. Happy to open a follow-on action item for whichever the group prefers.

specification/attributes/dataset_configuration.md
```diff
@@ -14,7 +15,19 @@ Common scenarios where dataset configuration is valuable include:
...
+* *FOCUS dataset* MUST be configurable to select one delivery method for all detail scopes or for each detail scope when a selected detail level can be delivered through more than one delivery method.
```
Matt-Cowsert, line 20: Do you mean the generator must offer both selection granularities, or that it may offer whichever one it chooses? `for all detail scopes or for each detail scope` reads both ways. If either granularity satisfies the requirement, saying so explicitly in one bullet would close it. If both are required, this is two bullets, which is also what the Normative Requirements Guidelines ask for when a bullet joins independent conditions with `or` that produce distinct constraints. Either way you can resolve this comment once the wording says which.

specification/attributes/dataset_configuration.md
```diff
@@ -14,7 +15,19 @@ Common scenarios where dataset configuration is valuable include:
...
+* *FOCUS dataset* MUST include the columns documented for a selected detail level.
```
Matt-Cowsert, line 21:
```suggestion
* *FOCUS dataset* MUST include the columns documented for a selected detail level when a detail level is selected, regardless of a user-defined selection of columns.
```

A practitioner who narrows columns and also selects a detail level hands the generator two instructions it cannot both satisfy: the first bullet here reads `*FOCUS dataset* MUST be configurable to include only a user-defined selection of columns`. An earlier revision of this list settled that with `even when those columns are not present in the selected column list`. The `when` clause matches the two bullets above it, which are the only ones here gated on a detail level existing.

specification/attributes/dataset_configuration.md
```diff
@@ -14,7 +15,19 @@ Common scenarios where dataset configuration is valuable include:
...
+* *FOCUS dataset* detail-scope documentation MUST include FOCUS dimension criteria that identify the data coverage of each offered detail scope.
```
Matt-Cowsert, line 22: An earlier revision of this list opened the detail block with `*FOCUS dataset* MAY offer configurable detail levels for provider-defined detail scopes.` That bullet came out and nothing replaced it, so this documentation block now reads as an obligation on every FOCUS dataset: a generator that offers no detail scopes has no detail-scope documentation and misses seven MUST bullets. Lines 19 and 20 kept their `when` clauses. Restoring the MAY bullet, or adding `when the FOCUS dataset offers a detail scope` to each of these seven, would close it.

specification/attributes/dataset_configuration.md
```diff
@@ -14,7 +15,19 @@ Common scenarios where dataset configuration is valuable include:
...
+* *FOCUS dataset* detail-scope documentation MUST include all offered detail levels for each offered detail scope.
+* *FOCUS dataset* detail-scope documentation MUST include the columns populated for each offered detail level.
+* *FOCUS dataset* detail-scope documentation MUST include whether each offered detail level uses [DataGeneratorCalculatedSplitCostAllocationHandling](#attributes.datagenerator-calculatedsplitcostallocationhandling).
+* *FOCUS dataset* detail-scope documentation MUST include the available delivery methods for each offered detail level.
+* *FOCUS dataset* detail-scope documentation MUST include the relationship of each delivery method to other delivered [*dataset artifacts*](#glossary:dataset-artifact) or provider-defined companion artifacts that represent the same underlying usage or charges.
+* *FOCUS dataset* detail-scope documentation MUST include the columns used to relate a provider-defined companion artifact to the corresponding less-detailed dataset artifact when the detail is delivered outside the corresponding dataset artifact.
```
Matt-Cowsert, line 28:
```suggestion
* *FOCUS dataset* detail-scope documentation MUST include the columns used to relate a provider-defined companion artifact to the corresponding less-detailed dataset artifact when the detail is delivered outside the corresponding dataset artifact.
* *FOCUS dataset* detail-scope documentation MUST be accessible to practitioners.
```

Seven bullets say what this documentation must contain and none say a practitioner can reach it, and the mechanism runs on the practitioner reading it before selecting a detail level. DeliveryHandling closes its documentation block with the identical bullet: `*FOCUS dataset* delivery mechanism documentation MUST be accessible to practitioners.` Tags and the InvoiceDetail dataset carry an accessibility bullet too, at SHOULD and with a different audience respectively, so MUST here matches the closest analog rather than the weakest.

specification/attributes/dataset_configuration.md
```diff
@@ -14,7 +15,19 @@ Common scenarios where dataset configuration is valuable include:
...
+* *FOCUS dataset* SHOULD represent records with identical values in all delivered dimension columns and non-summable metric columns as a single record.
```
Matt-Cowsert, line 29: Does `delivered dimension columns` reach the custom columns a detail level populates? The spec answers two ways. The glossary defines a dimension as a specification-defined categorical attribute, which would exclude `x_UserId` from this file's Example section. But `specification/appendix/json_object_examples/contract_commitment_applicability_examples.md` uses `"Dimension": "x_BillingModel"` under a Custom Scope heading, and DataGeneratorCalculatedSplitCostAllocationHandling says `FOCUS dataset column representing a dimension`, where FOCUS dataset column covers custom columns. On the narrow reading, two records differing only by `x_UserId` qualify for merging here and the practitioner loses the detail they selected. Naming custom columns explicitly in this bullet would settle it regardless of how the broader question lands.

specification/attributes/attributes_overview.md
```diff
@@ -11,7 +11,7 @@ Attributes serve as reusable containers for requirements that enforce consistenc
 | [Custom Column Handling](#attributes.customcolumnhandling) | Defines column ID naming, formatting, and value requirements for [*custom columns*](#glossary:custom-column) appearing in a [*FOCUS dataset*](#glossary:FOCUS-dataset). |
 | [Data Generator-Calculated Split Cost Allocation Handling](#attributes.datagenerator-calculatedsplitcostallocationhandling) | Allows data generators to provide granular cost information based on specific documented methods. |
 | [Dataset Completeness](#attributes.datasetcompleteness) | Defines requirements for a *FOCUS dataset* to include *custom columns* not represented in [*FOCUS columns*](#glossary:FOCUS-column). |
-| [Dataset Configuration](#attributes.datasetconfiguration) | Defines the rules for customizing a dataset's schema. |
+| [Dataset Configuration](#attributes.datasetconfiguration) | Defines the rules for customizing a dataset's schema and level of detail. |
```
Matt-Cowsert, line 14:
```suggestion
| [Dataset Configuration](#attributes.datasetconfiguration) | Defines the rules for customizing a dataset's schema and scoped detail. |
```

The attribute file now uses `scoped detail` in both its opening paragraph and its Description field. This row still carries the earlier phrasing.

Matt-Cowsert (commented):
The new requirements in Dataset Configuration are written over four terms the specification never defines: detail scope, detail level, delivery method, and data coverage. All four are defined in `supporting_content/`, which `spec.mdpp` does not include, and the glossary has no entry for any of them. A practitioner reading the published spec gets ten MUST bullets and no way to tell what a detail scope is.

The rest is inline. The documentation block starting at line 22 lost its applicability gate when the `MAY offer configurable detail levels` bullet came out, so it now reads as an obligation on every FOCUS dataset. Line 21 and the first bullet in the list pull against each other when a practitioner narrows columns and also selects a detail level. Line 20 leaves it open whether both selection granularities are required. And the detail-scope documentation has no accessibility requirement, which Delivery Handling carries on its equivalent block.

Line 29 turns on whether a custom column counts as a dimension column, and the spec answers that two ways: the glossary calls a dimension specification-defined, while the Contract Commitment Applicability examples use `x_BillingModel` as one. Worth settling, since it decides whether the selected detail survives.

The version-introduced item from the August 19 task force call is still open.

specification/attributes/dataset_configuration.md
Matt-Cowsert, line 19: ## TF2 Action Item - August 26, 2026

`detail scope`, `detail level`, `delivery method`, and `data coverage` carry the MUST bullets in the dataset configuration documentation block, and none of the four is defined in a file the specification build includes. Their definitions sit in supporting_content/attributes/dataset_configuration.md, which specification/spec.mdpp does not pull in.

Task Force 2 (TF-2) covered this on 2026-08-26. @matt-cowsert's guidance was that a term the specification does not define does not exist for a reader, and that most of the supporting content text can move straight into glossary entries. @finops-DaveMoreau accepted the direction on the call.

**Scope:**
- [ ] Add glossary entries for detail scope, detail level, delivery method, and data coverage
- [ ] Source the entry text from supporting_content/attributes/dataset_configuration.md

**Assignee:** @finops-DaveMoreau

specification/attributes/dataset_configuration.md
```diff
@@ -1,11 +1,12 @@
 # Dataset Configuration
```
ijurica, line 1: The scoped detail configuration introduced in this PR uses several new concepts (`detail scope`, `detail level`, `delivery method`, `provider-defined companion artifact`) that are not defined in the glossary. Since these are used as normative subjects in `DatasetConfiguration` requirements and are likely to be referenced from other attributes going forward, I suggest formalizing them as glossary entries.

### Renames to consider

Before formalizing glossary terms, I'd propose renaming two of the terms currently used in this PR. 

My reasoning:

| Current PR term | Proposed term | Why |
|---|---|---|
| detail level | **Detail Variant** | `Level` implies ordinal ranking (less → more detailed). Variants may also differ non-ordinally within the same scope (e.g., same cardinality but different added dimensions). `Variant` keeps that door open without committing to a scale. |
| delivery method | **Detail Representation** | `Delivery method` collides with the `DeliveryHandling` attribute, which governs how a FOCUS dataset itself is delivered (Overwrite / Append). The concept here is different — how detail is represented in or alongside a dataset artifact — and deserves a name that does not overload existing terminology. |

### Proposed glossary entries

**Detail Scope**
A defined subset of FOCUS dataset records for which one or more Detail Variants are offered. A Detail Scope is identified using criteria based on FOCUS dimensions.

**Detail Variant**
A provider-defined specification of the detail provided within a Detail Scope.

**Detail Representation**
The manner in which data at a selected Detail Variant is represented in relation to the corresponding less-detailed data within the same Detail Scope.

**Companion Artifact**
A provider-defined artifact delivered alongside a FOCUS dataset artifact. A Companion Artifact provides data at a selected Detail Variant for a Detail Scope of that FOCUS dataset.

---
### General Notes:

* **On provider-defined.** In these glossary definitions I deliberately used the term provider-defined to allow either the data generator or the service provider to be the responsible party — whichever fits the implementation in practice. 

* **On `usage or charges`.** The phrase `same underlying usage or charges` is used in several places in this PR. Since `DatasetConfiguration` is not Cost-and-Usage-specific (other datasets conform to it — e.g., `ContractCommitment`), I'd suggest replacing it with dataset-neutral phrasing (e.g., `same underlying data`) where applicable.

specification/attributes/dataset_configuration.md
finops-DaveMoreau: 🤖 [AI][Claude Code] ✅ **Implemented**

Version Introduced for the Dataset Configuration attribute is back to `1.4` in 87c245c5. Following attribute precedent, the scoped detail additions will be recorded in the 1.5 changelog rather than in the attribute's Version Introduced section.

finops-DaveMoreau: 🤖 [AI][Claude Code] ✅ **Implemented**

Completed this TF2 action item in 87c245c5: Version Introduced for the Dataset Configuration attribute is back to `1.4`, following @qquigley's comment and the attribute precedent @ijurica confirmed. The scoped detail additions will be recorded in the 1.5 changelog.

specification/attributes/dataset_configuration.md
finops-DaveMoreau: 🤖 [AI][Claude Code] ✅ **Implemented**

In fcda8332 this bullet now reads: `*FOCUS dataset* MUST include the columns documented for a selected detail level, regardless of the user-defined selection of columns.`

The `when a detail level is selected` clause was left out because `for a selected detail level` already limits the requirement to selected detail levels. Giving documented detail columns precedence over column selection also keeps the selected detail from being merged back together by the record-minimization SHOULD.

specification/attributes/dataset_configuration.md
finops-DaveMoreau: 🤖 [AI][Claude Code] ✅ **Implemented**

In fcda8332 the detail-scope documentation bullets are nested under `When *FOCUS dataset* offers a detail scope, *FOCUS dataset* detail-scope documentation MUST adhere to the following requirements:`, following the condition grouping bullet form in the Normative Requirements Guidelines. A FOCUS dataset that offers no detail scope is no longer held to these bullets. The wording of the nested bullets is unchanged.

specification/attributes/dataset_configuration.md
finops-DaveMoreau: 🤖 [AI][Claude Code] ✅ **Implemented**

In fcda8332, `*FOCUS dataset* detail-scope documentation MUST be accessible to practitioners.` is the last bullet in the detail-scope documentation block, matching the equivalent DeliveryHandling bullet.

specification/attributes/attributes_overview.md
finops-DaveMoreau: 🤖 [AI][Claude Code] ✅ **Implemented**

Applied in fcda8332. The Dataset Configuration row in the attributes overview now reads `Defines the rules for customizing a dataset's schema and scoped detail.`, matching the attribute's introduction and Description.

specification/attributes/dataset_configuration.md
finops-DaveMoreau: 🤖 [AI][Claude Code] ✏️ **Wording update**

In 1dd689e8 the condition bullet for the detail-scope documentation block now reads `When a detail scope is offered, *FOCUS dataset* detail-scope documentation MUST adhere to the following requirements:`. The earlier wording placed `FOCUS dataset` inside the condition, where the Normative Requirements Guidelines call for a precise abstraction level, and the abstract FOCUS dataset does not itself offer detail scopes. The passive form matches the existing `when a default column set is offered` condition in the same attribute. The nested bullets are unchanged.

specification/attributes/dataset_configuration.md
finops-DaveMoreau: 🤖 [AI][Claude Code] ✅ **Implemented**

Fixed in a978845c: the DatasetConfiguration requirements anchor is back to `Dataset conforming to DatasetConfiguration attribute MUST adhere to the following requirements:`. Per the Normative Requirements Guidelines, attribute anchors name the targeted entity type as plain `Dataset` or `Column`, while `*FOCUS dataset*` is used as the grammatical subject of the requirement bullets beneath it.

specification/attributes/dataset_configuration.md
finops-DaveMoreau: 🤖 [AI][Claude Code] ✅ **Implemented**

In ad243df2 the record-minimization bullets no longer depend on the Dimension and Metric glossary entries, which both define only FOCUS-defined columns. They now read:

* `*FOCUS dataset* SHOULD represent records with identical values in all delivered FOCUS dataset columns, other than FOCUS dataset columns representing summable metrics, as a single record.`
* `*FOCUS dataset* SHOULD preserve the aggregate value of each FOCUS dataset column representing a summable metric when records are represented as a single record.`

The glossary defines a FOCUS dataset column as either a FOCUS column or a custom column. Two records that differ only in `x_UserId` therefore stay separate, and custom summable columns keep their sums. The Dimension and Metric glossary entries are unchanged in this PR.

specification/attributes/dataset_configuration.md
finops-DaveMoreau: 🤖 [AI][Claude Code] ✅ **Implemented**

Either granularity satisfies the requirement. In ad243df2 the bullet reads:

`*FOCUS dataset* MUST be configurable to select one detail representation when more than one detail representation is offered for a selected detail variant.`

Whether the selection is offered for the complete dataset or separately for each detail scope is left to the data generator, and the Detail Representations section of the supporting content describes both. Detail representation replaces delivery method; that rename is listed under Open Questions for TF-2 in the PR description.

specification/attributes/dataset_configuration.md
finops-DaveMoreau: 🤖 [AI][Claude Code] 🔄 **Implemented, pending TF-2 confirmation**

In ad243df2 the glossary defines Detail Scope, Detail Variant, Detail Representation, and Companion Artifact, and the DatasetConfiguration requirements use those terms. Two terms from this comment changed: detail level is now detail variant, and delivery method is now detail representation, which also removes the overlap with DeliveryHandling's delivery mechanism. Data coverage is no longer a separate term; the Detail Scope entry defines the records a scope covers.

The names are listed under Open Questions for TF-2 in the PR description, so this thread stays open until TF-2 confirms them.

finops-DaveMoreau: 🤖 [AI][Claude Code] 🔄 **Implemented, pending TF-2 confirmation**

In ad243df2 glossary entries exist for Detail Scope, Detail Variant, Detail Representation, and Companion Artifact, with text drawn from the supporting content and @ijurica's proposed entries. They use proposed names in place of detail level and delivery method, and data coverage is folded into the Detail Scope definition. The Detail Variant entry also counts the default output as a variant, so every detail scope offers a real opt-in choice.

Both the names and the default-variant point are listed under Open Questions for TF-2 in the PR description. Leaving this action item open for TF-2 to confirm.

specification/attributes/dataset_configuration.md
finops-DaveMoreau: 🤖 [AI][Claude Code] 🔄 **Implemented, pending TF-2 confirmation**

In ad243df2 both renames are adopted (detail level → detail variant, delivery method → detail representation), along with glossary entries for Detail Scope, Detail Variant, Detail Representation, and Companion Artifact. Two changes from the proposed text:

* The Replacement representation is renamed Expanded, because Replacement is already a CorrectionHandling correction style.
* The Companion Artifact entry says its structure is not defined by FOCUS instead of calling it provider-defined, because the glossary uses both "service-provider-defined" and "data generator-defined". This keeps your intent that either party can define it.

`usage or charges` is replaced with dataset-neutral phrasing in the new requirements and the detail representation guidance. Leaving this open until TF-2 confirms the names.

specification/attributes/attributes_overview.md
flanakin: "Scope" is a loaded term. I would suggest we stay away from that given the FinOps Framework definition, which isn't quite how we're using it here. "Level of detail" is arguably safer.

specification/attributes/dataset_configuration.md
```diff
@@ -1,11 +1,12 @@
...
 
-Dataset Configuration allows FinOps practitioners to tailor the structure and content of a [*FOCUS dataset*](#glossary:FOCUS-dataset). Datasets provided by data generators are often massive, and their ingestion can lead to excessive storage costs and slow processing times. By removing large, static, or irrelevant columns, FinOps practitioners can optimize the dataset for better performance and lower storage costs.
+Dataset Configuration allows FinOps practitioners to tailor the structure, content, and scoped detail of a [*FOCUS dataset*](#glossary:FOCUS-dataset). Datasets provided by data generators are often massive, and their ingestion can lead to excessive storage costs and slow processing times. By selecting only the columns and detail needed for a given workflow, FinOps practitioners can optimize the dataset for better performance and lower storage costs.
```
flanakin: "Scoped detail" isn't clear and honestly isn't necessary. I would remove it. We're covered by "structure and content" already. 
"For a given workflow" also feels unnecessary. Workflow implies a specific implementation. Leaving it out keeps the meaning without implicitly constraining scenarios.
```suggestion
Dataset Configuration allows FinOps practitioners to tailor the structure and content of a [*FOCUS dataset*](#glossary:FOCUS-dataset). Datasets provided by data generators are often massive, and their ingestion can lead to excessive storage costs and slow processing times. By selecting only the columns and detail needed, FinOps practitioners can optimize the dataset for better performance and lower storage costs.
```

specification/attributes/dataset_configuration.md
```diff
@@ -1,11 +1,12 @@
...
 
 Common scenarios where dataset configuration is valuable include:
 
 * **Managing Scale**: Trim large exports to reduce time and cost of data preparation
 * **Reducing Noise**: Tailor datasets for specific workflows (e.g., cost allocation, commitment analysis)
+* **Managing Detail**: Include optional detail for cost areas where detailed attribution is needed (e.g., per-user or per-feature costs for a shared service)
```
flanakin: We're using "detail" 3x in this. Can we rephrase?

specification/attributes/dataset_configuration.md
```diff
@@ -1,11 +1,12 @@
...
+Dataset Configuration allows FinOps practitioners to tailor the structure, content, and scoped detail of a [*FOCUS dataset*](#glossary:FOCUS-dataset). Datasets provided by data generators are often massive, and their ingestion can lead to excessive storage costs and slow processing times. By selecting only the columns and detail needed for a given workflow, FinOps practitioners can optimize the dataset for better performance and lower storage costs.
 
 Common scenarios where dataset configuration is valuable include:
 
 * **Managing Scale**: Trim large exports to reduce time and cost of data preparation
 * **Reducing Noise**: Tailor datasets for specific workflows (e.g., cost allocation, commitment analysis)
+* **Managing Detail**: Include optional detail for areas of a dataset where detailed attribution is needed (e.g., per-user or per-feature costs for a shared service)
```
flanakin: nit: We're using "detail" 3x in this. Can we rephrase?

specification/attributes/dataset_configuration.md
```diff
@@ -14,9 +15,25 @@ Common scenarios where dataset configuration is valuable include:
 Dataset conforming to DatasetConfiguration attribute MUST adhere to the following requirements:
 
 * *FOCUS dataset* MUST be configurable to include only a user-defined selection of columns.
-* *FOCUS dataset* MUST adhere to all column-level specifications defined in the FOCUS schema, regardless of the user's chosen configuration (e.g., column selection).
+* *FOCUS dataset* MUST adhere to all column-level specifications defined in the FOCUS schema, regardless of the selected configuration (e.g., column selection or detail variant selection).
```
flanakin: This feels unclear to me. 
```suggestion
* *FOCUS dataset* MUST adhere to all column-level specifications defined in the FOCUS schema, regardless of the selected configuration (e.g., columns or amount of detail).
```

specification/attributes/dataset_configuration.md
flanakin: I disagree with this change. "Level of detail" is a very common term. "Detail scope" and "detail variant" are absolutely not common terms. I would argue they are more confusing. Not to mention the near-miss overlap with scopes as defined by the FinOps Framework. This feels eerily similar, while not being the same. But my bigger issue is the lack of clarity. "Level of detail" is clear as the "amount of detail". Unless we want to change "detail"?

Thinking out loud: Our goal is to define when we want a column populated. Is there a simpler way to convey that than "detail scope" and "detail variant"?

specification/attributes/dataset_configuration.md
```diff
@@ -14,9 +15,26 @@ Common scenarios where dataset configuration is valuable include:
...
+* When a [*detail scope*](#glossary:detail-scope) is offered, *FOCUS dataset* MUST adhere to the following requirements:
+  * *FOCUS dataset* MUST be configurable to select each [*detail variant*](#glossary:detail-variant) offered for a *detail scope*.
+  * *FOCUS dataset* MUST include only one *detail variant* for each *detail scope*.
+  * *FOCUS dataset* MUST preserve the sum of each [*FOCUS dataset column*](#glossary:FOCUS-dataset-column) representing a summable [*metric*](#glossary:metric) for the records in a *detail scope* across all offered *detail variants*.
+  * *FOCUS dataset* user-defined selection of columns MUST include the columns documented for a selected *detail variant*.
```
Matt-Cowsert:
```suggestion
  * *FOCUS dataset* user-defined selection of columns MUST include the columns documented for a selected *detail variant* that are not populated when no additional detail is selected.
```

Your union model holds for a variant a practitioner opts into, not for the default one. The Detail Variant entry counts the default output as a variant and the documentation bullet below covers the columns populated for every offered variant, so the union of a five-column list like the Example's and the default variant's documented columns is just those columns. The qualifier keeps your framing and limits the union to the columns a variant adds, so "only" keeps its force inside a detail scope. Your supporting content at line 263 already reads that way: the single column list covers the columns contributed by every selected variant. That rests on the default output staying a variant, which is open question 2.

specification/attributes/dataset_configuration.md
```diff
@@ -14,9 +15,26 @@ Common scenarios where dataset configuration is valuable include:
...
+    * *FOCUS dataset* detail scope documentation MUST include each *detail representation* offered for each *detail variant*.
+    * *FOCUS dataset* detail scope documentation MUST include the relationship between records delivered at each *detail representation* and other records representing the same underlying data in [*dataset artifacts*](#glossary:dataset-artifact) or [*companion artifacts*](#glossary:companion-artifact) (e.g., records that replace or supplement other records).
+    * *FOCUS dataset* detail scope documentation MUST include the columns that relate records in a *companion artifact* to the related records in the *dataset artifact* when a *companion artifact* is offered.
+    * *FOCUS dataset* detail scope documentation MUST be accessible to practitioners.
+* *FOCUS dataset* SHOULD represent records with identical values in all delivered *FOCUS dataset columns*, other than *FOCUS dataset columns* representing summable *metrics*, as a single record.
+* *FOCUS dataset* SHOULD preserve the aggregate value of each *FOCUS dataset column* representing a summable *metric* when records are represented as a single record.
```
Matt-Cowsert:
```suggestion
* *FOCUS dataset* SHOULD represent records with identical values in all delivered *FOCUS dataset columns*, other than *FOCUS dataset columns* containing summable numeric values (e.g., costs and quantities), as a single record.
* *FOCUS dataset* SHOULD preserve the aggregate value of each *FOCUS dataset column* containing summable numeric values when records are represented as a single record.
```

The Record Minimization guidance says the merge preserves aggregates for custom columns too, and these bullets do not reach them. The glossary defines a metric as a FOCUS-defined column, so a summable `x_TokenCount` falls in the identity set of the first bullet and the merged record keeps one record's value rather than the sum. The August 26 thread closed on the opposite expectation. The Normative Requirements Guidelines qualifier form is `FOCUS dataset column containing numeric values`, with summable added here. Line 22 keeps the metric wording, as DataGeneratorCalculatedSplitCostAllocationHandling does, because a custom column only one variant populates cannot have its sum preserved across variants.

specification/attributes/dataset_configuration.md
```diff
@@ -14,9 +15,26 @@ Common scenarios where dataset configuration is valuable include:
...
 * *FOCUS dataset* MAY offer a default column set.
-* *FOCUS dataset* default column set MUST include all applicable [*FOCUS columns*](#glossary:FOCUS-column) when a default column set is offered.
+* *FOCUS dataset* default column set MUST include all applicable FOCUS columns when a default column set is offered.
```
Matt-Cowsert:
```suggestion
* *FOCUS dataset* default column set MUST include all applicable *FOCUS columns* when a default column set is offered.
```

The glossary link moved up to the detail scope documentation bullet, which is right under link-once, but the italics came off with it. Repeat occurrences keep them in this file (`*detail representation*` and `*FOCUS dataset columns*`), in `custom_column_handling.md`, and in `dataset_completeness.md`.

Matt-Cowsert (commented):
A practitioner who narrows the column list still gets the columns documented for the default variant back inside a detail scope. Treating detail variant columns as part of the user-defined selection still swallows the column list: the Detail Variant entry counts the default output as a variant, and detail scope documentation has to list the columns populated for every offered variant, so the union of a short list and the columns documented for the default variant is just those documented columns. The suggestion on that bullet limits the union to the columns a variant adds. Whether it is needed depends on the default output staying a detail variant, which is open question 2, so I would rather settle both here.

Six smaller items sit on the lines they affect.

Every v1.5 pull request has to be in member review by September 24.

specification/attributes/dataset_configuration.md
```diff
@@ -30,6 +48,12 @@ A practitioner configures their FOCUS Cost and Usage dataset to include only the
 
 Even though columns like `CommitmentDiscountId` and `ResourceId` are excluded, the included cost columns (`BilledCost`, `EffectiveCost`) still reflect commitment discounts correctly. The dataset remains conformant to the FOCUS specification because each included column follows all requirements for that column, including requirements that reference columns not in the dataset.
 
+A data generator offers a detail scope for records where Service Name is "Example AI Service". The detail scope documentation identifies three detail variants for that scope: the default variant, which does not populate per-user detail; a "user" detail variant, which populates the custom column `x_UserId`; and a "feature" detail variant, which populates the custom column `x_FeatureName`.
+
+Neither the "user" nor the "feature" detail variant is more detailed than the other, since each populates a column the other does not. The records in the detail scope sum to the same totals for each summable metric under either selection.
```
Matt-Cowsert:
```suggestion
Neither the "user" nor the "feature" detail variant is more detailed than the other, since each populates a column the other does not. The records in the detail scope sum to the same totals for each summable metric under any of the three selections.
```

The sum-preservation requirement covers all offered detail variants, the default one included. Naming two selections invites a reader to treat the default output as outside the rule, which is the same reading that would let a generator skip the detail variant obligations for it.

specification/glossary.md
```diff
@@ -141,6 +145,18 @@ A physical representation of a specific [*dataset instance*](#glossary:dataset-i
...
+
+A subset of records in a [*FOCUS dataset*](#glossary:FOCUS-dataset), identified by values of [*FOCUS columns*](#glossary:FOCUS-column) representing [*dimensions*](#glossary:dimension) (e.g., [Service Name](#datamodel.costandusage.servicename), [Resource Type](#datamodel.costandusage.resourcetype)), for which more than one [*detail variant*](#glossary:detail-variant) is offered. A detail scope can correspond to one [*service*](#glossary:service), multiple services, or records identified without reference to a service.
+
+<a name="glossary:detail-variant"><b>Detail Variant</b></a>
+
+An offered option for a [*detail scope*](#glossary:detail-scope) that determines which columns are populated and the records that result. The records delivered for a detail scope when no additional detail is selected also constitute a detail variant. A detail scope's variants do not have to be a sequence from less to more detail. For any two, each can populate columns the other does not. Detail variants for the same detail scope represent the same underlying charges, so each summable [*metric*](#glossary:metric) sums to the same value for every variant.
```
Matt-Cowsert:
```suggestion
An offered option for a [*detail scope*](#glossary:detail-scope) that determines which columns are populated and the records that result. The records delivered for a detail scope when no additional detail is selected also constitute a detail variant. A detail scope's variants do not have to be a sequence from less to more detail. For any two, each can populate columns the other does not. Detail variants for the same detail scope represent the same underlying data, so each summable [*metric*](#glossary:metric) sums to the same value for every variant.
```

@ijurica asked for dataset-neutral phrasing because DatasetConfiguration is not Cost and Usage specific, naming ContractCommitment as her example and `same underlying data` as the substitute. A Contract Commitment detail scope does not carry charges, and a glossary entry is where a reader picks the term up.

supporting_content/attributes/dataset_configuration.md
```diff
@@ -225,6 +234,56 @@ When introduced, time granularity will allow practitioners to choose temporal re
...
+
+The columns populated for a selected detail variant are part of the delivered dataset artifact, so they appear in the Schema metadata like any other column. ColumnDefinition contains one object for every column provided in the dataset instance artifacts that reference the parent Schema object. A practitioner reading the Schema metadata therefore sees a single column list covering the user-defined column selection together with the columns contributed by every selected detail variant across all detail scopes, with one entry for a column that more than one detail scope contributes.
+
+Columns that appear only in a companion artifact are not part of a dataset artifact, so they are not in that column list. When the same column is also populated in the dataset artifact, such as under the inline detail representation, it is provided in the dataset artifact and is included.
+
+Changing the selected detail variant for a detail scope adds or removes columns in the delivered dataset artifact. Schema metadata requires a new Schema collection object whenever the structure of the dataset instance artifacts changes, so each variant change produces a new Schema collection object.
```
Matt-Cowsert:
```suggestion
Changing the selected detail variant for a detail scope adds or removes columns in the delivered dataset artifact under the inline and expanded detail representations. Schema metadata requires a new Schema collection object whenever the structure of the dataset instance artifacts changes, so those variant changes produce a new Schema collection object. A variant whose detail is delivered only in a companion artifact leaves the dataset artifact structure unchanged.
```

The paragraph above says columns appearing only in a companion artifact are not part of a dataset artifact, so a variant delivered that way changes no dataset artifact column. As written this paragraph promises a new Schema collection object for every variant change, which would have a generator publishing schema objects that describe the same structure.

supporting_content/attributes/dataset_configuration.md
```diff
@@ -225,6 +234,56 @@ When introduced, time granularity will allow practitioners to choose temporal re
...
+
+A provider that natively measures usage at the actor grain can offer actor attribution as scoped detail without requiring split cost allocation. When a provider starts from a coarser [*origin charge*](#glossary:origin-charge) and distributes it across actors or workloads using an allocation method, the actor-attribution detail uses [Data Generator-Calculated Split Cost Allocation Handling](#attributes.datagenerator-calculatedsplitcostallocationhandling).
+
+### Relationship to Split Cost Allocation
+
+Scoped detail configuration is the broader opt-in and documentation mechanism for selecting additional detail. Data Generator-Calculated Split Cost Allocation Handling is a defined subset of that pattern for detail variants that split an origin charge into [*allocated charges*](#glossary:allocated-charge). A detail variant can add scoped detail without split cost allocation when the provider already measures the underlying usage or charges at that detail.
```
Matt-Cowsert:
```suggestion
Scoped detail configuration is the broader opt-in and documentation mechanism for selecting additional detail. Data Generator-Calculated Split Cost Allocation Handling is a defined subset of that pattern for detail variants that split an origin charge into [*allocated charges*](#glossary:allocated-charge). A detail variant can add scoped detail without split cost allocation when the provider already measures the underlying data at that detail.
```

@ijurica asked for `same underlying data` in place of `usage or charges` because DatasetConfiguration is not Cost and Usage specific, and the new requirements already read that way. This clause covers any dataset's detail variant, so the charge wording narrows it, even though the split cost allocation sentence beside it is charge-specific by design.

specification/attributes/attributes_overview.md
Matt-Cowsert: Applied and confirmed in the current file: the Dataset Configuration row now reads "Defines the rules for customizing a dataset's schema and scoped detail." Closing this.

PR #2471, opened 2026-06-17
AI #2466: Add DatasetType and DatasetName to the schema for Attributes in prep for Conditions to be added to RM 1.5

mike-finopsorg:
## Summary

Add DatasetType, DatasetId, and DatasetName to attribute rules

Add the dataset identity keys to every rule under model_rules/attributes for 1.5, setting DatasetType to "ATT" and DatasetId/DatasetName to null (attributes are not scoped to a specific dataset instance).

Update model_schema.json so all ModelRules require the three keys, allow DatasetId and DatasetName to be string or null, and enforce non-null string values via the allOf conditional when EntityType is not Attribute.

### Type of Change
* [ ] Bug fix
* [X] New feature / Specification update
* [ ] Editorial / Formatting

### Author Checklist
* [X] **AI Usage Guidelines:** I have read the [FOCUS AI Usage Guidelines](/guidelines/contributors/ai-usage-guidelines.md).
* [X] **Self Review Attestation:** I have performed a self-review of my own contribution.
* [X] **AI-Generated Examples:** If this PR includes examples generated by AI, I have manually verified that the data is mathematically accurate, logically consistent, and compliant with the FOCUS schema.

PR #2375, opened 2026-05-13
AI #2373: Upgrade Feature Levels of three `Cost and Usage` columns

shawnalpay:
## Summary

This pull request introduces the "Upgrade" phase of the **Universal Technology Cost Schema** proposal. 

By shifting FOCUS to rely more heavily on the `Conditional` feature level to accommodate diverse business architectures (like SaaS and AI), we must simultaneously strengthen the schema for the providers that *do* support complex billing constructs. If a concept is a fundamental driver of mathematical cost or native taxonomy for a Technology Category, it cannot be left as a loose "Recommendation." 

This PR upgrades **Availability Zone**, **Charge Frequency**, and **Service Subcategory** from `Recommended` to `Conditional`. By tying these directly to their relevant Presence Conditions, we ensure that when a data generator supports these concepts (e.g., a Public Cloud provider charging for cross-zone data transfer or a SaaS vendor offering annual vs. monthly billing), their inclusion is strictly enforced for financial reconciliation.

For more information on the full proposal, see these documents:

- [Proposal: Revising FOCUS Feature Levels](https://docs.google.com/document/d/1nMV3z0Nf08pVPPwbamhmbElCg_Rq5nfZZBS1Wq7JkQQ/edit?tab=t.0#heading=h.iqphd0hlmtx7)
- [Cost and Usage Feature Levels and Conditions](https://docs.google.com/spreadsheets/d/1D6DaaxOVUGBVuAPMu1T2g-69ABkN6KuMvmaKL7Mgjkw/edit?gid=300003991#gid=300003991)
- [FOCUS Implementation Guide: Public Cloud](https://docs.google.com/document/d/14-pKbbaIL-YllfSBzbyrctdemDK52KdhuhhwDUJnIB4/edit?tab=t.0#heading=h.fnpwzk52x702)
- [FOCUS Feature Conditions Reference](https://docs.google.com/document/d/1yVsinNlL1MthJNHlEo1xGDUA2TfVF06OVLc5mi9BS8o/edit?tab=t.0)

## Changes Included
**1. Column Definition Updates (`Feature level` changed from `Recommended` to `Conditional`):**
* `availabilityzone.md`
* `chargefrequency.md`
* `servicesubcategory.md`

**2. Dataset Table Updates (`dataset.md`):**
* Updated the Cost and Usage summary table to reflect the new `Conditional` feature levels for the 3 columns above.

**3. Requirement Rule Updates (`dataset.md`):**
[...]

PR #2507, opened 2026-07-01
FR #2335: Add feature-level rubric principles guideline

### LINKED ISSUE #2335: [FR] Revise Feature Levels to make FOCUS more globally applicable

Matt-Cowsert:
## Summary

Adds `guidelines/contributors/feature-level-rubric-guidelines.md`: the principles for setting a FOCUS column's feature level and nullability when the column is proposed, rather than reopening it after the column merges.

The core idea is that **Mandatory is a bar a column earns, not the default it starts from.** A column is Mandatory only when its concept exists for every operating model and a value is naturally producible; anything a reasonable operating model would leave entirely null is Conditional, gated on an operating model Condition. This targets the adoption barrier of forcing SaaS, AI, and internal-IT generators to emit empty or padded columns to clear a Mandatory bar set on cloud-shaped assumptions.

The guideline keeps feature level and nullability as separate axes, and turns the principles into four answerable decision inputs (necessity, applicability variance, producibility, derivability) with a decision-routing diagram. Borderline columns default to Conditional under an interim boundary rule until a companion mechanics guideline calibrates the threshold from generator data.

This is the first of two parts: the load-bearing principles here, and the data-driven mechanics (the applicability matrix, the back-test against current columns, and a machine-readable check) in a companion guideline to follow. Two supporting-content documents accompany it: the blind-rater testing evidence and the design notes.

This PR touches `guidelines/` and `supporting_content/` only. It does not change `specification/`, and is outside the spec build and lint path. Relates to #2335. Opened as a draft for maintainer feedback on the principles before the mechanics companion follows.

### Changelog

* **2026-07-19 (late evening):** Readability split of the two densest passages, with no content change: principle 5 (honest nulls, never fabrication) broken into shorter standalone sentences, and decision input 4 (derivability) reorganized into six nested sub-bullets under its unchanged lead sentence; no clause, example, or qualifier dropped, no claims merged, and no list numbering shifted. Validated before landing by a round-9 blind-rater check run as an independent evaluation, with the rater packet rebuilt and the three raters dispatched by a session separate from the one that authored the edits: all determinate sentinels held their round-8 outcomes, and every quote a rater attributed to the edited sections verified verbatim against the split text. Round 8 recorded in the blind-rater evidence doc.
* **2026-07-19:** Readability pass, validated by a fresh three-model blind-rater sentinel check before landing (all anchors held; EffectiveCost's grounding strengthened, with all three raters quoting the rewritten sentence as dispositive): the Overview's four-outcome list trimmed to its core claims, principle 6's "manufactured" unified to "fabricated" to match principle 5, and the input 2 definitional-equality example corrected to read accurately against EffectiveCost's spec requirements (equal to BilledCost for ordinary charges, computed from it where commitments amortize, so the concept stays universal). A reviewer suggestion removing a stale "two-layer output" reference from the design notes applied as written.
* **2026-07-13 (evening):** Two-Layer Output replaced by Normative Content Defined by the Rubric, a reviewer suggestion applied as written: the rubric's outputs are operating model Conditions, Feature Levels, and nullability, with category-based expectations informative only and out of scope, to be addressed separately. Recommended and Optional now read as exceptions throughout, never expected leveling outcomes: the Overview carries a fourth outcome (a column enters or remains in the schema only by earning Mandatory or Conditional), the paragraph under the Key Concepts table states the rubric assigns only those two levels, and input 1 notes that a leveling discussion leaning toward Recommended signals insufficient conviction in the column's utility, so the column is held out of the schema rather than admitted at a weaker level. Mechanical: Contents anchor updated for the section rename, a dangling informative-layer reference reworded, and list markers restored to asterisks per the repo linter.
* **2026-07-13 (afternoon):** Per the maintainer 1:1 and follow-up review comments: useful-but-not-needed columns are now held out of the schema rather than routed to Recommended, Optional is never a feature-level outcome, and the specification's existing Recommended and Optional columns are named as exceptions to re-level or retire over time; derivability defaults to leaving a fully derivable column out, with keeping one framed as a deliberate promotion for Supported Feature correctness or usability; Terms renamed to Key Concepts, the intrinsic/operating model gate terms and the requirements-model digression dropped, and the "Recording Why a Column is Conditional" section removed (its output lives in presence and nullability requirements per the normative requirements guidelines); the closing section repurposed as Relationship to the Companion Guideline; a Conditional column may link to more than one operating model Condition; flowchart labels line-broken so GitHub's renderer keeps text inside the node shapes.
* **2026-07-13 (morning):** Per second-round review feedback, the retrospective "The Problem This Addresses" section is removed: its desired outcomes now appear in the Overview in prescriptive form, and the change rationale stays in the design notes and this PR. Contents list reformatted as bullets. Design notes aligned with the directional-derivation principle; round 7 blind-rater re-check of the simplified text recorded in the evidence doc (all sentinels held; the cost/price family surfaced one open scope question on the substitution test).
* **2026-07-12:** Guideline simplified for readability: heavy de-duplication, the summary section replaced by a contents list, an illustration on each principle, derivation symmetry replaced with the directional derivation principle (a derived column can be no more present than its sources; a derived column's absence never pulls a source down), "fill value" reframed as "manufactured value," and the in-flux cost/price examples removed pending the open level-change PRs.
* **2026-07-08:** The Rubric in Brief plain-language summary added after round 6 confirmed it changes no leveling outcome; principle 6 names its gating Condition; design-notes terminology aligned; round 6 recorded in the evidence doc.
* **2026-07-06:** Restructured for standalone readability (terms defined before use, self-contained principles, one home for the interim boundary rule); first-round and 1:1 review feedback from ijurica applied, including the qualified "operating model Condition" term; blind-rater evidence and design notes added to supporting content; round 5 re-check held.
* **2026-07-02:** Principles refined from the blind-rater and adversarial review: definitional-equality versus gap-filling distinction, derivability scoped to in-dataset consumer computation, the concept fixed by the column's Description, Condition admissibility (no circular Conditions), and the interim boundary rule; clarity punch lists applied.
* **2026-07-01:** Reframed around Mandatory as an earned high bar with Conditional as the boundary default; decision-routing diagram added; plain-language pass.
* **2026-06-30:** Initial draft.

### Type of Change
* [ ] Bug fix
* [ ] Editorial / Formatting

### Author Checklist
* [ ] **AI-Generated Examples:** If this PR includes examples generated by AI, I have manually verified that the data is mathematically accurate, logically consistent, and compliant with the FOCUS schema. _(N/A: prose worked examples only, no CSV/JSON data.)_


Matt-Cowsert:
I've revised the principles based on maintainer feedback and two rounds of blind-rater testing, and updated the PR description to match. The headline change: Mandatory is now framed as a high bar a column earns, not the default it starts from. A column only lands Mandatory when its concept exists for every operating model and a value is naturally producible, so no reasonable data generator has to carry it null down every row. Anything a reasonable operating model would leave entirely null is Conditional. This is the change that most directly serves data generator adoption: we stop forcing SaaS, AI, and internal-IT generators to emit empty or padded columns to clear a bar set on cloud-shaped assumptions.

The revision also drops the "fallback" framing, simplifies the necessity call (useful-but-not-needed is Recommended, Optional reserved for the rare discretionary column), scopes derivability to same-dataset values, and adds the smaller clarifications the rater tests surfaced. The full current shape is in the updated description.

A note on the level-change PRs, #2374 and #2375. The point of the rubric here is to check those level changes against explicit inputs rather than intuition. Run that way, it backs the applicability-gated downgrades: the cost columns that not every generator can produce (ListCost, ContractedCost) come out Conditional. It comes out differently on a couple of the necessity-gated columns: ServiceCategory holds at Mandatory and ServiceSubcategory at Recommended under these principles. That is less a disagreement than a reason to look at those columns with the inputs, and eventually the generator data, in front of us rather than settling them by gut feel.

The exact tipping point, how many operating models lacking a column moves it from Mandatory to Conditional, is left to the companion mechanics guideline, calibrated from real generator data. This stays part 1 of 2.

Happy to walk it section by section, tagging each as strong alignment, minor pushback, or hard disagreement, and merging the agreed sections first.

Matt-Cowsert:
This is a clarity and accuracy pass over the guideline, on top of the principle revisions from earlier. It does not change any leveling decision. The goal was to make the doc read cleanly for a maintainer seeing it cold and to line every example and reference up with the current spec.

It went in as two commits: a first pass of readability and fact-check fixes that leave the sentences earlier blind-rater testing depends on untouched, and a second, deliberately minimal pass for the few edits that do touch those sentences.

A few of the accuracy fixes worth calling out:

* The gate-recording example had SkuId's and ResourceId's null rules swapped. It now reads that SkuId is null on tax rows and ResourceId is null when the charge is not related to a resource, matching skuid.md and resourceid.md.
* The Conditions registry is now located as the existing `specification/conditions/` section rather than described as a future artifact, and the machine-readable check refers to the model's `ApplicabilityCriteria` field by name where it maps the older criteria keys.
* The ChargeClass example is scoped to a correction to a previously closed billing period, matching chargeclass.md.
* Input 2 (applicability variance), which was one long paragraph, is split into labeled sub-bullets with no wording changes, since it is the densest part of the rubric.

After the second commit I re-ran the blind-rater check on the columns whose level depends on the touched sentences. The outcomes held: BilledCost, ChargeClass, and EffectiveCost stay Mandatory, InvoiceId and ListCost stay Conditional, the derivability check still does not demote BillingAccountName on a provider-held name lookup, and the refinement carve-out still leaves AvailabilityZone in place since the region primary it refines is itself Conditional.

Still a draft, still looking for maintainer feedback on the principles before the mechanics companion follows.

guidelines/contributors/feature-level-rubric-guidelines.md
```diff
@@ -0,0 +1,158 @@
...
+
+These principles set the level on purpose, keep the two questions apart, and make Mandatory something a column earns rather than something it is assumed to be.
+
+## Core Principles
+
+These principles turn on the operating model. Operating model is a defined term in the specification glossary; for leveling purposes, read it as the set of characteristics that decide which FOCUS concepts apply to a data generator (whether it includes regions, commitment discounts, virtual currency, and so on), independent of its category.
+
+1. **Level by operating model, not by technology category.** A category (cloud, SaaS, PaaS, data center, AI) describes the data generator. It does not constrain the schema and has no say in the level. When applicability varies, name a Condition from the operating-model Conditions registry instead. Each Condition reads "the operating model includes X" (for example, includes regions, or includes commitment discounts). The registry is the specification's Conditions section (specification/conditions/): a single list of operating-model Conditions that the specification maintains and that each Conditional column links to.
```
ijurica: I wonder whether we should avoid mentioning Category in this document altogether.

The rubric consistently establishes that feature levels are determined by concepts and the Conditions that describe whether those concepts exist in a generator's operating model, not by technology categories. Introducing Category—even with repeated caveats that category defaults are not ceilings—potentially adds more cognitive overhead than value and, more importantly, risks encouraging implementers to reason from "I am a SaaS provider" rather than "does my operating model include this concept?"

In practice, category defaults can easily become implementation shortcuts. A generator may conclude that a column is "not for us" simply because its category is shown as typically not supporting it, even though its own operating model may satisfy the relevant Condition—or evolve to satisfy it in the future.

I think we would end up with a cleaner, more future-proof model by keeping the specification and its guidelines focused exclusively on Concepts and Conditions.

I suggest relocating any category-based guidance to separate educational materials.

guidelines/contributors/feature-level-rubric-guidelines.md
```diff
@@ -0,0 +1,158 @@
...
+    * **Read only presence gates for this test.** Read only the Conditions that gate the column's presence for this test. A Condition that instead gates nullability (whether the value may be null when the column is present) leaves the level Mandatory and feeds input 3, not this one; do not read a nullability gate as an applicability gate.
+3. **Producibility.** When the column applies, can the operating model produce a meaningful value, or is the honest answer null? Set Allows nulls = False only when a meaningful value is genuinely available on every row where the column is present, not manufactured to look present. Otherwise set Allows nulls = True, and use null wherever the value is not meaningful or not available. This sets nullability, never the level, and never allows fabrication.
```
ijurica: I think we're still missing the consumer perspective, particularly for cross-data-generator use cases.

The rubric focuses on how an individual generator should model its data, but consumers typically analyze datasets combined from multiple generators. In that context, we should pay more attention to the implications for Supported Features, especially their SQL queries and expected semantics.

Although we state that generators should never fabricate values, there may be cases where defining a meaningful default value (whether populated by the generator or by the consumer) produces more predictable and useful Supported Feature behavior than excluding records with NULLs and degrading the feature's results. I think this trade-off deserves explicit consideration.

---

### Example

Cost Comparison feature depends on `BilledCost`, `ContractedCost`, `EffectiveCost`, `ListCost`. 

If `ContractedCost` is null for generators that don't assert the contracted-pricing concept, this query silently miscounts once instances are merged. Three ways to handle it, with different consequences:

* **(a) Generator-side default** (current spec behavior for ContractedCost — defaults to ListCost)
  * Predictable, but obligates every generator to backfill — in tension with Principle 4 ("never fabricate").
  * Typically simpler consumer SQL, but increases data volume and pushes logic upstream

* **(b) Consumer-side default at query time**
  * Same result, but shifts the obligation to the consumer/query layer — needs to be documented per Supported Feature, not left to a column description.
  * Typically more complex SQL, but lower data redundancy

* **(c) Exclude nulls**
  * Understates coverage and biases results toward generators that populate the column
  * Not acceptable without explicit definition of resulting population.
  * Excluding NULLs changes the effective scope of the Supported Feature result and breaks cross-generator comparability unless explicitly defined in the feature specification.

---

### Conclusion

Supported Features, especially when they rely on columns that are Conditional in presence and non-nullable when present, should be treated as a critical design input.

In most such cases, option (c) (excluding nulls) is likely the least desirable, as it degrades the feature by silently reducing coverage and introducing bias in cross-generator results.

Wherever possible, we should therefore explicitly **define a meaningful defaulting strategy or resolution mechanism for defaults**, instead of leaving behavior implicit or relying on NULL handling in queries.

At this point, I'm not sure whether the defaulting strategy should consistently be:
* generator-side,
* consumer/query-side within Supported Features,
* or whether this decision should be made on a case-by-case basis per Supported Feature and supporting conditional column, based on a trade-off between implementation complexity and data footprint/size impact.

guidelines/contributors/feature-level-rubric-guidelines.md
```diff
@@ -0,0 +1,158 @@
...
+
+5. **Derived-pair symmetry.** When a column and the column it derives from are both carried, they take the same level. A cost and the unit price it derives from are leveled together. This settles the cost-versus-price split by principle, not column by column. Which level the pair shares comes from the four decision inputs, run per column as usual: when the inputs place both members at the same level, that is the pair's level, and when they would place the two members on opposite sides of the Mandatory-versus-Conditional line, the pair is a boundary case and takes the interim level from The Two Axes, Conditional, unless the working group records why a specific operating model is exceptional.
```
ijurica: Reminders:

- There is an issue with the current description of the ContractedCost and ListCost columns in the introduction section. The description assumes a strictly invertible relationship between cost and unit price (C = P × Q), which does not always hold in practice — this is already evident from the normative requirements of the cost columns. We plan to update these descriptions accordingly, after which it will be clear that **cost columns are not examples of derivable columns**.

- I agree that unit prices are derivable (P = C / Q). However, we intentionally included unit prices in the dataset to support unit price reconciliation scenarios, where we aim to avoid rounding errors and ensure analytical consistency. This reflects an implicit “cake vs ingredients” principle that has influenced design decisions, although it has not yet been formally documented. In essence, derivable columns may still be promoted to first-class dataset elements when they improve the correctness or usability of Supported Features.

---

A few concerns:

- The term "derived pair" is problematic:
  - The term "pair" is misleading. A derived column is not necessarily derived from a single other column. The example itself (ListCost = ListUnitPrice × PricingQuantity) involves three columns rather than two.
    *Note: As previously mentioned, in this specific case, the derivable column is the unit price, not the cost — we will address this in the specification, but I'm not yet sure whether I would recommend moving the normative requirements from the cost columns to the unit price columns.*
  - "derived" describes only the computed column, not the columns it is derived from. The relationship is between a derived column and one or more source columns, not between two equivalent members of a pair.
- Rather than requiring all columns to "share the same level," consider specifying that derivable columns inherit the Feature Level (and Condition, if applicable) from the columns they are derived from.
- The current principle is expressed in terms of a specific example (cost and unit price), but I think it should instead describe derivation relationships more generally. The same reasoning would apply to other derivable columns.

guidelines/contributors/feature-level-rubric-guidelines.md
```diff
@@ -0,0 +1,158 @@
...
+
+## The Problem This Addresses
+
+Today a column's feature level is an output, not a choice. The level falls out of how the presence requirement happens to be written: an unconditional `MUST` gives Mandatory, a conditional `MUST` gives Conditional, a `SHOULD` gives Recommended, and a `MAY` gives Optional. Nothing decides what the requirement should be in the first place, so each column is leveled on its own.
+
+Two problems followed. First, two separate questions got treated as one: whether a column is present (the level), and whether its value may be null (nullability). "Always present, value may be null" is not the same as "present only when it applies." Second, similar columns drifted to different levels with no rule to settle it, such as the split between cost and price columns (ListCost is Mandatory while ListUnitPrice is Conditional) and the different handling of Account and Provider columns (SubAccountId is Conditional while provider identity columns such as ServiceProviderName are uniformly Mandatory).
```
ijurica: * **two separate questions got treated as one: whether a column is present (the level), and whether its value may be null (nullability):** 
   * I'm not clear on what concrete case this refers to. The current specification already carries Feature level and Allows nulls as two separate fields in every column's Content Constraints table, so they are not structurally conflated today. Could you provide a specific example of a column where this conflation led to an incorrect leveling outcome?
   * On the contrary, if a column is not nullable across all use cases, that seems like evidence it should be Mandatory — the two axes aligning is not the same as them being confused.
* **similar columns**
   * The criterion for "similar columns" should be explicit, because column naming alone can be misleading (e.g., ContractApplied, ContractedUnitPrice, ContractedCost).
   * The second example pairs SubAccountId (Conditional) and ServiceProviderName (Mandatory) as an instance of "different handling of Account and Provider columns." It is not clear to me why these two columns are presented as a pair that should be compared — one is a sub-account identifier, the other is a provider name. What makes them "similar" in the sense that their different Feature Levels indicate a problem?

guidelines/contributors/feature-level-rubric-guidelines.md
```diff
@@ -0,0 +1,158 @@
...
+
+3. **Two axes, kept apart.** Level and nullability are decided separately. Level says whether the column is present, and when. Nullability says whether the value may be null when the column is present.
+
+4. **Honest nulls, never fabrication.** When a value is not meaningful or not available, it is null; FOCUS never asks a data generator to invent a placeholder. Two things follow, and they set the high bar for Mandatory. A value that would be null for a whole class of operating models makes the column Conditional (input 2 under The Four Decision Inputs, below), not Mandatory carried null throughout for those models. A level that would force a generator to fabricate or pad a value is never correct and must change. A column is never kept Mandatory by leaning on a manufactured or default value to fill it; whether a column can borrow a related value where its own is absent is a detail of that column's definition, not a reason to level it Mandatory.
```
ijurica: * I agree with the intent of this principle — we should not ask generators to invent meaningless placeholder values. 
  * However, the principle as stated needs more precision to prevent misuse.
    * There is a risk that a provider could interpret "never fabricate" too broadly — for example, claiming they should not "fabricate" a SkuId or SkuPriceId because they don't currently expose such identifiers in their native billing data. But if that provider's operating model includes unit pricing, then by definition they have quantifiable goods or services with specific prices. What the specification asks for is a unique identifier for something that already exists in their system — that is not fabrication, it is identification.
    * The safeguard against this misuse is clear definitions. If we clearly define what unit pricing means and what SkuId and SkuPriceId represent in that context, then a provider who asserts IncludesUnitPricing understands what is expected, and producing identifiers that unambiguously identify the concepts defined in the specification is an obligation, not fabrication.
  * Additionally, when a column's value has only one possible correct answer given the provider's operating model and the column's definition, populating that value is not fabrication — it is conformance. The "never fabricate" principle should target cases where no meaningful value exists, not cases where the provider simply hasn't surfaced an identifier for a concept they do support. (examples, BillingAccountType, SubAccountType)

* The statement "A value that would be null for a whole class of operating models makes the column Conditional" is missing a **key element**: it does not mention under what **condition** the value would be null. 
  * The test should not be whether a value happens to be null across some group of operating models, but whether there exists an **identifiable operating-model Condition under which the column would be null across all records and all use cases in a given dataset instance**.

guidelines/contributors/feature-level-rubric-guidelines.md
```diff
@@ -0,0 +1,158 @@
...
+
+> **Note:** "Condition" here means an operating-model gate: a named entry in the Conditions registry that marks the operating models where a concept exists. Asserting a Condition (principle 2) is not a separate registration step in the current specification: each Condition is a verifiable state that holds when the operating model includes the named characteristic. How a generator's asserted Conditions are recorded is mechanics for the companion guideline. This is a different thing from the per-rule `Condition` field that also appears in the requirements model (inside each rule's `ValidationCriteria`) and gates an individual validation rule; the spelling is the same, so context, not capitalization, tells the two apart. One Condition may gate a column's presence (feeding input 2, and making the column Conditional) or its nullability (feeding input 3, leaving the level unchanged). These are two roles for the same registry entry; only the presence role bears on the level.
```
ijurica: - I wonder whether we're **overloading the term Condition**.
  This note exists largely to explain that the operating-model "Conditions" discussed here are different from the ValidationCriteria.Condition field already used elsewhere in the specification. That makes me wonder whether we're using the same term for two different concepts that should instead have distinct names (e.g. operating model conditions instead of using Condition as the umbrella term)

- It also introduces the term **Conditions registry**, which I don't think has been defined previously.

guidelines/contributors/feature-level-rubric-guidelines.md
Matt-Cowsert: Fair on both counts. The conflation I meant is in how leveling arguments run, not in the schema fields; the concrete case is ContractedCost, which stays Mandatory with generators that have no negotiated pricing directed to default it to ListCost, so the presence question never gets asked. I reworded the paragraph to say that and cite that case (b68438d5). I also dropped the SubAccountId / ServiceProviderName example; you're right it doesn't hold up, since under this rubric those two should differ, and I stated the similarity criterion instead: columns bound to the same concept or linked by derivation. Agreed on your last point too: the two axes aligning is evidence for Mandatory, not a confusion. The problem is only when nullability is used to dodge the presence question.

guidelines/contributors/feature-level-rubric-guidelines.md
Matt-Cowsert: I want to keep the category mentions in the principles, and I think we agree on where the real risk lives. Every category mention in the doc is negative (it has no say, it never caps, never write a category rule). Those lines exist to catch the reader who shows up thinking "I'm SaaS, so this column isn't for me"; we can't retire that habit without naming it. The part I changed (b68438d5): the informative layer of category defaults now says it is expected to live outside the specification, in educational materials, which is where you suggested category guidance belongs. If that layer stays out of the spec and the normative layer stays category-free, does that cover the concern? Worth a few minutes when we talk.

guidelines/contributors/feature-level-rubric-guidelines.md
Matt-Cowsert: Agreed, and this is now in the doc (b68438d5). Principle 4 says fabrication means inventing a value for a concept the operating model lacks, and producing the representation of a concept the model does have is not fabrication: availability is judged against the operating model, not against what the current billing export surfaces. So a model that includes unit pricing has SKUs to identify, and producing SkuId is identification, not fabrication; and a value with only one correct answer given the model and the definition (BillingAccountType) is populated as a matter of conformance. Producibility got the same clause. Your safeguard is the same one the doc leans on elsewhere: clear definitions plus asserted Conditions make it an obligation.

guidelines/contributors/feature-level-rubric-guidelines.md
Matt-Cowsert: I took the terminology points in b68438d5: principle 5 no longer says pair, and it is phrased as a derived column and its source columns, direction included. The rule itself is the open question, and your reminders make it bigger than wording. I ran the three candidate rules against the list-price family and they pull in different directions: symmetry as written can drag ListCost down to the interim Conditional; inheritance from sources would promote ListUnitPrice to Mandatory, since ListCost and PricingQuantity are both Mandatory; and if C = P × Q is not faithfully invertible once the cost descriptions are fixed, then neither rule binds this family, each column levels on its own inputs, and today's split stops being drift and becomes defensible. I lean toward that third reading. It needs a losslessness qualifier on derivability (input 4) rather than a symmetry principle doing the work, and it also decides whether cost-versus-price stays in the problem statement as an example. Let's work through it when we talk.

guidelines/contributors/feature-level-rubric-guidelines.md
Matt-Cowsert: Both taken, with one constraint. "Condition" for the operating-model gates comes from the specification itself (the Conditions section), so this doc can't rename the concept, but the coinage was mine and it's gone: "Conditions registry" now reads "Conditions section" throughout, matching the spec's own name, and the operating-model qualifier is used where the two meanings could collide (b68438d5). Renaming the per-rule ValidationCriteria Condition field is requirements-model schema work; I'll carry it as an option for the companion rather than this doc.

guidelines/contributors/feature-level-rubric-guidelines.md
Matt-Cowsert: This is the real gap, and I read it as the counterpart the rubric needs rather than a competing principle. Where I land on the three options: (a) generator-side defaults are the pattern the rubric retires, and (c) silent exclusion is not acceptable, so (b) is the default, with the resolution defined per Supported Feature rather than left to each query author. One mechanical point makes (b) cleaner than it looks: under the rubric the no-contract generator's column is absent, not null, so the feature defines a merge-time rule over dataset instances (take ListCost where ContractedCost is absent), not per-row null handling. What I'd add to this doc is the write-back: a Conditional level is only complete when the Supported Features that list the column define their semantics under absence, recorded with the leveling decision. The per-feature strategy choice (generator-side, consumer-side, or case-by-case) belongs to the companion and the Supported Features work, and the back-test has to pair any re-level with that migration, or it breaks Cost Comparison the way you modeled. I held off adding the write-back text until we talk; want to settle the wording then?

guidelines/contributors/feature-level-rubric-guidelines.md
Matt-Cowsert: Reworked in the latest commits. The problem statement now frames the conflation as decision practice rather than schema structure, and cites the concrete case you asked for: ContractedCost is Mandatory with a default to list cost, so the presence question never gets asked. The account/provider example is gone, and the drift problem is scoped to columns bound to the same concept or linked by derivation, which is what the derivation symmetry principle actually addresses. The restructure also moved all definitions ahead of first use, which should help with the back-and-forth reading you described.

guidelines/contributors/feature-level-rubric-guidelines.md
Matt-Cowsert: Partly applied, and worth discussing live. The informative category-defaults layer is now explicitly expected to live outside the specification in educational materials, which I think addresses the main risk you raised (readers deciding "not for us" and never reaching the Conditions). I kept the category mentions inside the principles because every one of them is negative (categories have no say, never cap, never get written into a rule), and never naming the habit leaves the most common wrong mental model uncorrected. If it still reads as promoting categories after the restructure, let's work through it on our next call.

guidelines/contributors/feature-level-rubric-guidelines.md
Matt-Cowsert: Added. Principle 5 now draws the boundary directly: fabrication means inventing a value for a concept the operating model lacks; producing the representation of a concept the model does have is identification, not fabrication. Producibility is judged against the operating model rather than the generator's current export, with SkuId as the worked illustration.

guidelines/contributors/feature-level-rubric-guidelines.md
Matt-Cowsert: Applied in two steps. The principle is now derivation symmetry (principle 7 after the restructure): derivation runs from one or more source columns to a derived column, a directional relation, not a pair of equals. I also corrected the direction from our call: a unit price can derive from cost and pricing quantity, but a cost cannot derive from a unit price, since rows with no priced quantity (tax, say) have nothing to multiply. The rule's final shape is still open, and your cake-versus-ingredients point (why we carry a derivable unit price at all: price sheet reconciliation, rounding) is the piece I want to work into the rationale with you in TF1 alongside the pricing description work.

guidelines/contributors/feature-level-rubric-guidelines.md
Matt-Cowsert: Agreed on both points, and this went further than an editorial fix. The guideline now uses the qualified term "operating model Condition" everywhere, so bare "Condition" no longer appears as an umbrella term, and "Conditions registry" is replaced with the section's own name. Following our call, I want to pursue renaming the concept itself to Operating Model Conditions, in the spec's Conditions section and the RM 1.5 schema work (#2467, #2471, #2472) while those are still open. That is a separate change from this PR, and I'll raise it with Mike and the maintainers; the guideline here reads correctly under either outcome.

guidelines/contributors/feature-level-rubric-guidelines.md
Matt-Cowsert: This is the comment I want to spend real time on, so I'd rather work it through together than reply past it. Short version of where my head is: you've identified the missing counterpart to the generator-side rules. The rubric bans padding but says nothing about what replaces it for cross-generator Supported Features, and silence defaults consumers into silent exclusion, which is the degradation your Cost Comparison example shows. The likely shape is a write-back edge from the necessity input (feature definitions state how they behave across Conditional columns), with the mechanics in the companion. Let's shape it on a call before I draft text.

guidelines/contributors/feature-level-rubric-guidelines.md
Matt-Cowsert: Coming back to your second bullet, which my earlier replies skipped past. The key element you named (an identifiable operating model Condition under which the column would be null across all records and all use cases in a dataset instance) is carried by input 2 rather than principle 6: input 2 gates Conditional on the operating model Condition that marks the models where the concept exists, and its presence-gate discriminator draws the same scope you described, a gate counts only when an operating model could lack the characteristic outright, so that no row in any use case of that model ever has a value and the column is absent as a whole. I also tightened principle 6's closing sentence to name the gating Condition, so the sentence you quoted no longer reads as a bare null-prevalence test.

guidelines/contributors/feature-level-rubric-guidelines.md
```diff
@@ -0,0 +1,152 @@
...
+
+Today a column's feature level is an output, not a choice. It falls out of how the presence requirement is written: an unconditional `MUST` gives Mandatory, a conditional `MUST` gives Conditional, a `SHOULD` gives Recommended, a `MAY` gives Optional. Nothing decides what the requirement should be.
+
+So Mandatory became the default rather than a high bar, and columns only some operating models can produce were mandated anyway. Forcing those generators to emit an empty or padded column is the adoption barrier this guideline targets.
+
+Two habits sustain it:
+
+* **Level and nullability get argued as one question.** They are separate, and the specification already stores them as separate fields. Conflating them keeps a column Mandatory for models that lack its concept, on the reasoning that those generators can leave it null or fill it from another column, so the presence question is never asked.
+* **Derivation has no leveling rule.** Columns linked by derivation are leveled independently, so a derived column can be mandated redundantly, or a source column dragged to the level of something computed from it.
+
```
ijurica: The current "The Problem This Addresses" section reads more like change rationale than guideline content. In addition, a significant portion of this section is already covered by the normative principles later in the document (for example, Mandatory as an earned level, separation of level and nullability, and derivation rules).

I don't think the guideline needs to preserve the history of what was wrong before this change; that context belongs in the PR/Feature Request.

However, I think the intent behind the section is still valuable. Rather than keeping a retrospective problem statement, I would suggest incorporating the desired outcomes into the Overview in a prescriptive form:
* columns applicable only to some operating models should not be treated as universally required;
* nullability should not be used as a substitute for applicability decisions;
* derivation relationships should not determine feature level by themselves.

This keeps the guideline focused on the intended behavior while avoiding duplication and separating the "why are we introducing this change?" rationale from the long-lived guidance.

guidelines/contributors/feature-level-rubric-guidelines.md
```diff
@@ -0,0 +1,152 @@
...
+Level and nullability are two axes, decided separately. Level says whether the column is present, and when. Nullability says whether the value may be null when the column is present. Mandatory is a bar a column earns by applying to every operating model, not the default it starts from.
+
+This is the first of two parts: the principles here, the mechanics (procedure, data-driven tests, machine-readable check) in a companion guideline. The bar to meet: two people applying these principles to the same column reach the same level, without the author in the room.
+
+**Contents:** [The Problem This Addresses](#the-problem-this-addresses) · [Terms](#terms) · [Core Principles](#core-principles) · [The Interim Boundary Rule](#the-interim-boundary-rule) · [The Four Decision Inputs](#the-four-decision-inputs) · [Recording Why a Column is Conditional](#recording-why-a-column-is-conditional) · [Two-Layer Output](#two-layer-output) · [Applying the Principles: A Worked Example](#applying-the-principles-a-worked-example) · [Scope of This Guidance](#scope-of-this-guidance)
```
ijurica: Separately from the document structure discussion, I suggest changing the Contents section from an inline list to a bulleted list.

```suggestion
**Contents:** 

* [The Problem This Addresses](#the-problem-this-addresses)
* [Terms](#terms) 
* [Core Principles](#core-principles)
* [The Interim Boundary Rule](#the-interim-boundary-rule)
* [The Four Decision Inputs](#the-four-decision-inputs)
* [Recording Why a Column is Conditional](#recording-why-a-column-is-conditional)
* [Two-Layer Output](#two-layer-output)
* [Applying the Principles: A Worked Example](#applying-the-principles-a-worked-example)
* [Scope of This Guidance](#scope-of-this-guidance)
```

guidelines/contributors/feature-level-rubric-guidelines.md
```diff
@@ -0,0 +1,162 @@
...
+Today a column's feature level is an output, not a choice. It falls out of how the presence requirement is written: an unconditional `MUST` gives Mandatory, a conditional `MUST` gives Conditional, a `SHOULD` gives Recommended, a `MAY` gives Optional. Nothing decides what the requirement should be.
+
+So Mandatory became the default rather than a high bar, and columns only some operating models can produce were mandated anyway. Forcing those generators to emit an empty or padded column is the adoption barrier this guideline targets.
+
+Two habits sustain it:
+
+* **Level and nullability get argued as one question.** They are separate, and the specification already stores them as separate fields. Conflating them keeps a column Mandatory for models that lack its concept, on the reasoning that those generators can leave it null or fill it from another column, so the presence question is never asked.
+* **Derivation has no leveling rule.** Columns linked by derivation are leveled independently, so a derived column can be mandated redundantly, or a source column dragged to the level of something computed from it.
+
+## Terms
```
ijurica: I wonder whether "Terms" is the best name for this section. The specification already uses **Term** as a glossary entry:

> **Term:** An agreement specified on a [*contract*](#glossary:contract) or [*invoice*](#glossary:invoice).

Would **"Key Concepts"** (or a similar heading) be a better fit?

```suggestion
## Key Concepts
```

guidelines/contributors/feature-level-rubric-guidelines.md
```diff
@@ -0,0 +1,162 @@
...
+* **Operating model Condition.** A named entry in the specification's Conditions section (specification/conditions/), each reading "the operating model includes X" (includes regions, includes commitment discounts). The Conditions section is the single list, and every Conditional column links to one. The qualified term is always written, so it is never confused with the per-rule `Condition` field in the requirements model, which gates one validation rule, or with lowercase conditions in requirement text. One operating model Condition may gate a column's presence (making it Conditional) or its nullability (level unchanged); only the presence role bears on the level.
+* **Intrinsic gate and operating model gate.** The two gates a Conditional column records. An intrinsic gate is value variation that depends on the row (SkuId is null on tax rows), written as a row-level rule. An operating model gate is presence variation that depends on what the model includes (regions, virtual currency), written as an operating model Condition. Category variation is always an operating model gate.
```
ijurica: I wonder whether we need the terms **Intrinsic gate** and **Operating model gate** at all.

My understanding is that they largely restate concepts that already exist elsewhere in the rubric:

- **Operating model Condition** determines whether a Conditional dataset or column requirement applies (i.e., presence at the dataset-instance level).
- **Column Nullability **(or the corresponding row-level rule) determines when a present column may legitimately be NULL on individual records.

Introducing two additional terms for these concepts seems to increase the conceptual overhead without adding much explanatory value.

guidelines/contributors/feature-level-rubric-guidelines.md
Matt-Cowsert: Applied in d2a1e1a3. The retrospective section is gone and the Overview now states the three outcomes prescriptively. The change rationale stays in the design notes and the PR description.

guidelines/contributors/feature-level-rubric-guidelines.md
```diff
@@ -0,0 +1,156 @@
...
+* [The Interim Boundary Rule](#the-interim-boundary-rule)
+* [The Four Decision Inputs](#the-four-decision-inputs)
+* [Recording Why a Column is Conditional](#recording-why-a-column-is-conditional)
+* [Two-Layer Output](#two-layer-output)
+* [Applying the Principles: A Worked Example](#applying-the-principles-a-worked-example)
+* [Scope of This Guidance](#scope-of-this-guidance)
+
+## Key Concepts
+
+* **Operating model.** The collective set of business concepts underlying a FOCUS-compliant dataset. For leveling, read it as the characteristics that decide which FOCUS concepts apply to a generator (regions, commitment discounts, virtual currency, and so on), independent of category.
```
Matt-Cowsert, line 30: If we agree with this simplified definition, we'll need to update the glossary entry too

guidelines/contributors/feature-level-rubric-guidelines.md
```diff
@@ -0,0 +1,155 @@
...
+* **Operating model Condition.** A named entry in the specification's Conditions section (specification/conditions/), each reading "the operating model includes X" (includes regions, includes commitment discounts). The Conditions section is the single list, and every Conditional column links to one.
```
Matt-Cowsert: We should probably discuss how we're meant to distinguish between Capital "Condition" and lower case "conditions" in the spec. Here's some additional context to think about:

The qualified term is always written, so it is never confused with the per-rule `Condition` field in the requirements model, which gates one validation rule, or with lowercase conditions in requirement text. One operating model Condition may gate a column's presence (making it Conditional) or its nullability (level unchanged); only the presence role bears on the level.

guidelines/contributors/feature-level-rubric-guidelines.md
```diff
@@ -0,0 +1,142 @@
...
+* **Consume a machine-readable derivation source.** Input 4 needs a per-column record of what each column derives from and by what relation (arithmetic, lookup, or conversion).
+* **Point every Conditional column at the Conditions section.** Require each Conditional column to link to a Condition there, and map the older `ApplicabilityCriteria` keys (for example `REGION_SUPPORTED`) onto the matching Condition IDs (for example `IncludesRegions`).
+* **Plan the back-test's re-leveling.** It will flag existing Mandatory columns meeting neither necessity basis (administrative and audit columns are the common case); the guidance says whether to re-level or record the dataset-structural basis that keeps them Mandatory.
+
+The back-test needs the Supported Features to name the columns each use case depends on: a column needed on the use-case basis but absent from every Directly Dependent Columns list has nothing for input 1 to point at. The principles stand on their own, and a team can adopt them without waiting for the mechanics.
```
ijurica: I think this section currently mixes a summary of this document with a description of the companion guideline.

The first paragraph ("In scope: ...") repeats the structure the reader has just seen and does not add much value. I would remove it.

Consider repurposing this section to clarify the boundary between this guideline and the companion guideline. Since the split of responsibilities is still being defined, I would avoid wording that implies the companion guideline already has a finalized scope.


```suggestion
## Relationship to the Companion Guideline(s)

This guideline defines the principles for assigning feature levels and nullability. The companion guideline is intended to define the mechanics needed to apply these principles consistently and support conformance validation.

The following topics should be considered for the companion guideline:

- The step-by-step leveling procedure.
- How a generator's asserted operating model Conditions are recorded.
- The necessity-boundary test between refining and breaking.
- The applicability test, including the matrix and the threshold at which a column moves between Mandatory and Conditional.
- The disposition of variance that does not map to an operating model Condition.
- The back-test against current columns.
- Where the informative layer is recorded.
- The machine-readable check, including:
  - reading all relevant requirement surfaces,
  - consuming a machine-readable derivation source,
  - linking Conditional columns to the Conditions section,
  - planning the re-leveling of existing columns.
```

Matt-Cowsert: Applied in 657fe8a9 with your structure. Two small deviations: the heading is singular because the linter rejects "(s)" in title case, and I kept the closing sentence that the principles stand alone. The Recommended/Optional stance from our call is in 834dc209.

guidelines/contributors/feature-level-rubric-guidelines.md
Matt-Cowsert: agreed and removed, thanks for flagging

guidelines/contributors/feature-level-rubric-guidelines.md
Matt-Cowsert: agreed and restructured, resolving

guidelines/contributors/feature-level-rubric-guidelines.md
```diff
@@ -0,0 +1,150 @@
...
+Borderline applicability, the concept present for most models and absent for a few, is settled by the companion's test; until then such a column takes the Conditional default from the Interim Boundary Rule.
+
+## Two-Layer Output
+
+The rubric produces two layers; only one is normative.
+
+1. **Normative (no categories).** Per column: a level, a nullability, and zero or more operating model Conditions. Always *Conditional, gated on the operating model Condition that the model includes regions*, never *Mandatory for cloud, Optional for SaaS*.
+2. **Informative (categories, non-binding).** Per category: the operating model Conditions typically met in that category today. These are defaults, not ceilings; a generator may exceed them (principle 2). This layer belongs outside the specification, in educational materials; where it is recorded is settled by the companion.
+
+A category-free normative layer is what lets a level be conformance-tested and stay valid as operating models change. The informative layer records today's expectations without binding tomorrow's generators.
```
ijurica: I suggest keeping this Guideline and this section focused on the normative content produced by applying the rubric and moving the informative layer discussion to the appropriate guidance material (potentially the companion guideline), where its purpose, format, and maintenance can be defined separately.

```suggestion
## Normative Content Defined by the Rubric

The rubric provides guidance for defining the normative content related to feature leveling and column nullability.

The rubric helps identify:
- applicable operating model Conditions,
- Feature Levels for FOCUS datasets and FOCUS columns,
- Nullability for FOCUS columns.

Feature Levels are defined independently of category. Conditional columns are expressed through operating model Conditions rather than category-specific rules such as "Mandatory for cloud" or "Optional for SaaS".

> **Note:** Category-based expectations are informative only and are out of scope for this rubric. They should be addressed separately, potentially in the companion guideline or other guidance material.
```

guidelines/contributors/feature-level-rubric-guidelines.md
```diff
@@ -0,0 +1,150 @@
...
...ing a derivable column anyway is a deliberate promotion, made when the duplicate improves a Supported Feature's correctness or usability (a unit price carried for price-sheet reconciliation), and the kept column is then leveled by the other inputs with the derivation recorded. It does not clash with principle 7: derivability keeps a redundant column out of `MUST`, while principle 7 fixes the direction of any derivation among columns that are kept. Derivable means a consumer can compute the value from other Mandatory columns in the same dataset instance. It covers arithmetic, conversion (a cost in another currency), and a lookup whose mapping is present in the dataset as a distinct artifact. A value needing a provider-held mapping is not derivable, nor is one whose only in-dataset link is co-occurring on the same rows (a display name resolved from an identifier through a provider-only table); such a column is leveled by inputs 1 through 3. Direction matters: a unit price derives from a cost and its pricing quantity, but a cost does not derive from a unit price, since rows with no priced quantity have nothing to multiply. The companion supplies the machine-readable derivation source.
```
ijurica: I suggest mentioning `Recommended` (and especially `Optional`) in this guideline only as **Exceptions** rather than as expected leveling outcomes.

From a consumer perspective, it is difficult to justify investing in functionality that depends on a column that may or may not be present across conforming datasets.

Matt-Cowsert: Agreed, and 7cb4734a makes it run through the whole doc rather than living only in input 1. The Overview now carries it as a fourth outcome (a column enters or remains in the schema only by earning Mandatory or Conditional, and Recommended/Optional persist only as exceptions on existing columns pending re-leveling), the paragraph under the Key Concepts table states the rubric assigns only those two levels, and input 1 now says a leveling discussion that leans toward Recommended is itself the signal that conviction in the column's utility is missing, so the column is held out of the schema rather than admitted at a weaker level. Your consumer rationale is the one the doc leans on: a consumer cannot build on a column that may never arrive.

guidelines/contributors/feature-level-rubric-guidelines.md
Matt-Cowsert: Applied as you wrote it; the only edits after the fact are swapping the dash bullets for asterisks to satisfy the repo linter and updating the Contents entry to the new section name (7cb4734a).

supporting_content/guidelines/feature-level-rubric-design-notes.md
```diff
@@ -0,0 +1,62 @@
...
+* **Calibration data.** The prevalence threshold waits on cell-level column-population evidence from real non-cloud generator exports: conformance submissions are the natural source, targeted requests to non-cloud generators the alternative. Until then the interim boundary rule governs, and it is expected to govern the whole current release cycle.
+
+## Relationship to the Guideline and the Companion
+
+The principles guideline carries what is stable: the operating-model basis, the earned-Mandatory bar, the two axes, the directional derivation principle, the four decision inputs, the presence-versus-nullability discriminator, the two-layer output, and the interim boundary rule. The companion mechanics guideline carries the machinery described here: the matrix, the calibrated threshold, the records, the back-test, and a machine-readable check. The split is deliberate: the principles can be adopted and applied with the interim rule now, and nothing in them changes when calibration replaces the interim rule with a measured threshold.
```
qquigley: Is "two-layer output" still applicable here? It exists in [older commits](https://github.com/FinOps-Open-Cost-and-Usage-Spec/FOCUS_Spec/pull/2507/changes/ea5d49f0ccbff84d064faf73c0a94448d4d8f97e#diff-4729f5be0cd10c0ca29c83d5cb1832736a6ccbcebd2807bbb133c311fe47dcfaR75-R83) but not this current branch.

qquigley:
```suggestion
The principles guideline carries what is stable: the operating-model basis, the earned-Mandatory bar, the two axes, the directional derivation principle, the four decision inputs, the presence-versus-nullability discriminator, and the interim boundary rule. The companion mechanics guideline carries the machinery described here: the matrix, the calibrated threshold, the records, the back-test, and a machine-readable check. The split is deliberate: the principles can be adopted and applied with the interim rule now, and nothing in them changes when calibration replaces the interim rule with a measured threshold.
```

Matt-Cowsert: thanks for flagging. applied your suggestion

guidelines/contributors/feature-level-rubric-guidelines.md
```diff
@@ -0,0 +1,158 @@
...
+
+## The Four Decision Inputs
+
+Every leveling decision answers four questions.
+
+1. **Necessity.** Is the data needed, so that without it something breaks rather than just degrades? A column is needed on either of two bases:
```
ijurica: Personally, I would remove Recommended and Optional from the FOCUS spec all together.

For me, a column should be included (mandatory or conditional) if:

- it is required by at least one Supported Feature, or
- it is required to satisfy the normative requirements, i.e., conditions of another column.

Matt-Cowsert:
Discussed in Maintainers on 2026-08-17. Closing in favor of #2604

PR #2424, opened 2026-05-29
AI #2364: Add `SKU Price` dataset

shawnalpay:
## Summary

This PR introduces the **SKU Price** dataset, establishing a standardized schema for cloud provider catalog rates, internal multipliers, negotiated unit prices, and volume-based tiers. By normalizing rate card properties, this dataset enables practitioners to perform precise rate lookups, analyze commitment discounts, calculate unit economics, and understand the cost mechanics of payable vs. consumable prices.

Sample data using this data model is in the appendix section **Examples: SKU Price** (`specification/appendix/sku_price_examples/`), with the CSVs under `specification/data/sku_price_examples/`. The earlier working artifacts remain available: the data model spreadsheet [here](https://docs.google.com/spreadsheets/d/1vtkK5cNCwZIXM0LbZOs9WTw9LN1pZWUU4DZXe_FXPoc/edit?gid=2001743556#gid=2001743556) and the sample data folder [here](https://drive.google.com/drive/folders/1KVEMTelpIbf-8ci1jEtQhMl-NnsMI7BY).

> **Note for reviewers:** The definitions of `ListUnitPrice` and `ContractedUnitPrice` are being revised under [FR #2492](https://github.com/FinOps-Open-Cost-and-Usage-Spec/FOCUS_Spec/issues/2492). This PR uses the FOCUS 1.4 wording for both columns. It will be updated to match #2492 once that work is approved. Approving this PR to move to member review is not approval of those two definitions.

### 1. New Dataset: `SKU Price`
* Added `sku_price/dataset.md` and updated `data_model.md` to integrate the new dataset.
* Establishes the core relationships between Cost and Usage (via `SKU Price ID`) and Contract Commitments (via `Contract ID`).
* Defines a row uniqueness key (`ServiceProviderName`, `SkuPriceId`, `ContractId`, `SkuPriceEffectiveStart`, `PricingCurrency`) and requires validity periods not to overlap for a given `ServiceProviderName`, `SkuPriceId`, `ContractId`, and `PricingCurrency`. Whether `QuantityTierMinimum` belongs in the key is open under [AI #2679](https://github.com/FinOps-Open-Cost-and-Usage-Spec/FOCUS_Spec/issues/2679); the appendix still describes a six-column key.

### 2. New Columns (22)
Added definitions and constraints for 22 new columns tailored for rate card modeling, categorized as follows:
[...]

PR #2549, opened 2026-07-21
AI #2530: Split the compound duration format bullet in ContractCommitmentDurationType

### LINKED ISSUE #1057: [FR] Separate dataset for SKUs
### 🧠 Problem Statement

---
### 🧪 Use Case

---
Planning and estimation

---
### 📨 Type of Request

---
TBD

---

---
### 📊 Impacted Parties

---
### 🌫️ Level of Ambiguity
TBD

---

---

---

  - Your **organization**

### LINKED ISSUE #2530: [AI] Split the compound duration format bullet in ContractCommitmentDurationType and PurchaseDurationType
**Parent Issue:** #1057

**Detailed Description**

Review of the SKU Price dataset (PR #2424) flagged that the `ContractCommitmentDurationType` requirement bundles three constraints into a single SHOULD bullet and duplicates the column's Expected Format section:

> ContractCommitmentDurationType SHOULD be expressed with a quantity and time unit, where quantity is a positive integer, and time-unit is a standardized unit of time, either singular or plural (e.g., "1 Day", "1 Year", "3 Months", "3 Years").

This runs against the one-verifiable-state-per-bullet authoring rule, and the quantity and unit detail is already stated in the column's Expected Format section. The bullet is an inherited pattern from the published Contract Commitment dataset.

**Update (July 20, 2026):** PR #2424 renamed the SKU Price column to `PurchaseDurationType`, so the two datasets no longer share a Column ID. The compound bullet itself is unchanged and identical in both apart from the subject name, so both columns are still fixed together to keep the shared duration format consistent.

**Definition of Done**

- [ ] The compound SHOULD bullet is split into single-constraint bullets, or the format detail is consolidated into the Expected Format section, in `specification/datasets/contract_commitment/columns/contractcommitmentdurationtype.md`
- [ ] The same change is applied to `specification/datasets/sku_price/columns/purchasedurationtype.md` so both columns keep the same format structure
- [ ] No change to either column's allowed values, feature level, or nullability

Matt-Cowsert:
## Summary

Resolves #2530.

### Scope

AI #2530 covers two columns that carry the same compound bullet. The fix is split across two PRs because the columns sit in different datasets, and only one of those datasets exists on `working_draft` today.

| Column | Dataset | Where the fix lands |
| :--- | :--- | :--- |
| `ContractCommitmentDurationType` | Contract Commitment | This PR |
| `PurchaseDurationType` | SKU Price | #2424 |

`specification/datasets/sku_price/` is introduced by #2424 and is not on `working_draft` yet, so `PurchaseDurationType` cannot be edited from this branch. #2424 already carries the same replacement bullet in `specification/datasets/sku_price/columns/purchasedurationtype.md`, so both columns land on identical format wording apart from the subject name.

The rest of this description covers the `ContractCommitmentDurationType` change only.

`ContractCommitmentDurationType` carried a single SHOULD bullet that bundled three constraints and restated detail already given in the column's Expected Format section:

> ContractCommitmentDurationType SHOULD be expressed with a quantity and time unit, where quantity is a positive integer, and time-unit is a standardized unit of time, either singular or plural (e.g., "1 Day", "1 Year", "3 Months", "3 Years").

This PR replaces it with a single verifiable constraint that names the format directly:

> ContractCommitmentDurationType SHOULD use the "[Numeric Value] [Unit]" format (e.g., "1 Day", "1 Year", "3 Months", "3 Years").

The quantity and unit detail now lives only in the Expected Format section, which the Content Constraints table already references as the column's Value format. The adjacent bullet recommending the unit of time be one of the allowed values is unchanged and continues to carry that constraint.

Two authoring rules motivate the change:

* Splitting Requirements: a requirement must be split when it contains a hidden constraint expressed as a definition (`ColumnA MUST be Z, where Z is defined as Y`). The `where quantity is..., and time-unit is...` clause is exactly that shape.
* DRY: each normative requirement belongs in exactly one place. Positive-integer and singular-or-plural were stated both in the requirement and in Expected Format.

The replacement borrows the sentence shape the `UnitFormat` attribute uses across its six format requirements:

> *FOCUS dataset column* including a unit quantity SHOULD use the `<unit-quantity> <plural-units>` format (e.g., `1000 Tokens`, `1000 Characters`).

The placeholder notation differs. `UnitFormat` writes format templates as backticked angle-bracket strings, while this bullet keeps the double-quoted `"[Numeric Value] [Unit]"` notation the column's Expected Format section already uses, so the requirement and the section it names stay consistent with each other.

No requirement changes meaning. Allowed values, feature level, nullability, and data type are untouched.

### Changelog

* **2026-08-20:** Added a note under Allowed Values recording that this column's time units express a term rather than a measurement, and intentionally differ from the time-based units in Unit Format. Addresses review feedback from @kk09v and matches the equivalent note already on `PurchaseDurationType` in #2424, so both columns read the same way. The allowed values themselves are unchanged.

### Test Plan

* Ran `python3 enhanced_markdown_lint.py --config markdownlnt.cfg scan` on the changed file: clean.
* Ran `make` from `specification/`: includes validation, lint, and the spec.md / spec.html / spec.pdf build all completed.
* Confirmed with pandoc that `"[Numeric Value] [Unit]"` renders as literal text rather than a reference link, matching how the existing Expected Format section already renders it.

### Type of Change
* [ ] Bug fix
* [x] Editorial / Formatting

### Author Checklist


ijurica (approved):

Arjun-Trident (approved):

specification/datasets/contract_commitment/columns/contractcommitmentdurationtype.md
kk09v, line 19: With the "Don't repeat ourselves" principal, I believe we should link this to the allowed time based units instead of explicitly stating them here. This will ensure consistency; see my note below about a discrepancy between these two lists.

https://github.com/FinOps-Open-Cost-and-Usage-Spec/FOCUS_Spec/blob/working_draft/specification/attributes/unit_format.md#allowed-time-based-unit-names

kk09v, line 35: Quarter is not currently an allowed time based unit in the attribute. We should either add quarter to that list or eliminate it from this list.
https://github.com/FinOps-Open-Cost-and-Usage-Spec/FOCUS_Spec/blob/working_draft/specification/attributes/unit_format.md#allowed-time-based-unit-names

Matt-Cowsert: Karl, linking this table to the UnitFormat list came up on #2424 in July, and I landed on keeping the two separate.

UnitFormat governs how the unit of measure for a quantity is expressed. Every column that declares it is a measurement-unit column: PricingUnit, ConsumedUnit, CommitmentDiscountUnit, ContractCommitmentUnit. This one is a Dimension holding a categorical term classifier, and its Value format points at its own Expected Format section, so it sits outside that set by design. The lists also diverge in both directions: this column carries Week and Quarter, which appear nowhere else in the specification as measurement units, and omits Second through Microsecond, which mean nothing for a commitment term.

I've pushed a note under Allowed Values recording that the divergence is deliberate, matching the one already on PurchaseDurationType in #2424, so both columns read the same way. If you'd rather we actually reconcile the two lists, that is a bigger change than this PR and I'd take it to Task Force 1.

Matt-Cowsert: Karl, the Quarter question is already open on #2600, and I'd rather settle it there than here.

Michael's supporting content for that PR raises the same point and offers the same two options you did: add Quarter to the reduction ladder, or drop it from the allowed values. His change touches this same Requirements section and is written against the bullet this PR replaces, so it will pick up the split wording and land on this file either way.

Two things worth carrying into that discussion. Week has the same gap as Quarter, since neither appears in the UnitFormat list, so the fix should cover both or neither. And ContractCommitmentDurationType shipped in 1.4, so removing an allowed value breaks anyone already emitting "1 Quarter" and breaks parity with PurchaseDurationType in #2424, which carries the same table.

This PR is scoped to splitting one compound bullet, so I'm leaving the values alone. I'll carry your Quarter point over to #2600 so it doesn't get lost.

jpradocueva (approved):

Validation completed for PR head `3a2a2f6f93a4890f1600da5211edba3dd46a14f7`
using `updates-agents` at `f8770501040b62db0c7aebe1bbe4de122f829266`.

No errors or warnings were identified within the supported Markdown validation
scope.

Content-validated documents: 1
Unsupported files: 0
Incomplete documents: 0
Errors: 0
Warnings: 0
Suggestions: 0
Unresolved claims: 0
Evidence complete: true

kk09v (approved):

jpradocueva:
Approved by the group without objections or abstentions during the August 27 Members' call.

PR #2552, opened 2026-07-21
FR #2335: Scope the feature-level rubric to Mandatory and Conditional (Option B)

### LINKED ISSUE #2546: [AI] Draft an Option B feature-level guideline PR scoped to Mandatory and Conditional
**Parent Issue:** #2335

**Detailed Description**
Maintainers reviewed the revised feature-level guideline in PR #2507 and agreed the next revision should evaluate only the Mandatory and Conditional levels. Those two levels carry the asymmetry questions that matter most right now, while the criteria for Recommended remain hard to describe deterministically. Rather than continue iterating in PR #2507, the next step is a clean Option B draft with the Recommended and Optional evaluation put on hold for a later pass. PR #2507 stays in draft as the record of the fuller Option A approach. Committed for the next maintainers call, with an earlier share and a ping to the maintainers group for offline feedback once ready.

**Definition of Done**
- [ ] New draft PR under FR #2335 containing guideline content scoped to Mandatory and Conditional evaluation
- [ ] Recommended and Optional evaluation explicitly deferred to a later revision, not silently dropped
- [ ] PR #2507 left in draft and referenced as the Option A alternative
- [ ] Maintainers pinged for offline feedback ahead of the next maintainers call

Matt-Cowsert:
## Summary

A narrowed alternative to #2507, per the maintainers discussion on 2026-07-20: this revision of the feature-level guideline evaluates only Mandatory and Conditional. #2507 stays in draft as the record of the fuller approach (Option A). This is Option B, and it touches the same three files, so only one of the two merges.

The guideline starts from a column the working group has decided to add, and decides which of the two levels it takes and whether its value may be null.

### What changed from #2507

* **Four decision inputs become two.** Applicability sets the level; producibility sets nullability. The flowchart follows.
* **Necessity and derivability are removed.** Both answered a prior question, whether a proposed column belongs in the schema at all, rather than which of Mandatory or Conditional an admitted column takes. Deciding between the two levels does not depend on either.
* **A new Scope of This Revision section names both deferrals:** the criteria for Recommended and Optional, and the admission tests. It also names what the narrowing gives up.
* **A new principle states the Supported Features write-back.** A Conditional level is not complete until the Supported Features that list the column as directly dependent account for its absence. The principle states that obligation and leaves the per-feature answer to the Supported Features work. It comes from the consumer-perspective thread on #2507.
* **No position is taken on the future of Recommended or Optional.** #2507's stance, that useful-but-not-needed columns are held out of the schema and that the existing Recommended and Optional columns are exceptions to re-level or retire, is not carried forward. The columns holding those levels are unaffected, and their meaning is unchanged.

### What did not change

The interim boundary rule, the applicability and producibility tests, the presence-versus-nullability discriminator, and the directional derivation principle all carry over from #2507.

### What the narrowing costs

The guideline can no longer be applied end to end to a net-new column proposal, because it presumes the decision to add the column has already been made. Until the deferred revision lands, that admission decision stays where it has always been, with the working group's judgment on the proposal.

### Supporting content

* `feature-level-rubric-design-notes.md` records the rationale for the narrowing, the two design questions that moved into deferred territory (the refinement carve-out on display-name columns, and what becomes of Recommended and Optional), and a new question on variance gates.
* `feature-level-rubric-blind-rater-evidence.md` maps the earlier testing rounds onto the narrowed scope and records the two rounds run against this revision, the second of them against the text as it now stands.

### Variance gates

An earlier revision left this open. It is answered in the guideline now, and the reasoning is the part to push on.

Six of the twenty-five shipped Conditions gate on two values differing, or on more than one value occurring, rather than on whether a concept exists at all. Between them they gate twelve columns across all three datasets, so this is a pattern rather than a stray entry. Applicability alone returns Mandatory for every one of them, because the underlying concept is universal: every model prices in some currency, and every model has a billing account with some type.

Input 1 now says such a gate is admissible when the column's value is fully determined by another column in the same dataset wherever the Condition is false, in which case the column is Conditional on that Condition, and not admissible otherwise, in which case it stays Mandatory. PricingCurrency equals BillingCurrency when a model prices and bills in one currency, so that gate holds. A single billing account type is recoverable from no other column, so that one does not.

Two places to push. The guideline defers derivability as an admission test and then leans on derivability here. Those are different questions, admission into the schema versus presence in a dataset instance, but it is a fair challenge. And the specification does not apply this consistently today: ServiceProviderName and BillingCurrency are Mandatory and repeat on every row, while BillingAccountType is Conditional because it would repeat. No column levels change here, and when to apply these criteria to already-published columns stays a separate decision.

### Verification

Lint passes and the mermaid flowchart renders.

The blind-rater check has now been run against the text as it stands, covering principle 7 and input 1's variance sub-bullet, which were both untested when this PR was last updated. It found one gap, in the variance sub-bullet, and the fix is included here. The sub-bullet said what a column's level becomes when a variance gate is not admissible but not when it is, so a reader had to supply that step alone, and readers did not supply it the same way. The sentence now names both outcomes. Everything else held, including the two sentences the earlier check showed can move an outcome. The blind-rater evidence document records the results in full, along with two caveats that matter for reading them: the guideline states the answer outright for several of the columns tested, and the packet supplied a roster of sibling column levels that raters leaned on where the guideline itself is silent.

Scope note: this PR touches `guidelines/` and `supporting_content/` only and does not change `specification/`, so it is outside the spec build and lint path.

Relates to #2335. Drafted per AI #2546.

### Changelog

**2026-08-02, later.** Completed input 1's variance sub-bullet so it states a level for both outcomes. The sentence establishing that a gate is admissible now also says the column is Conditional on that Condition, matching the sentence beside it that already gave the Mandatory outcome for gates that are not admissible. Blind-rater testing against the current text showed readers did not supply that missing step consistently on their own, and recorded both the testing and this edit in the evidence document.

**2026-08-02.** Stated a rule in input 1 for Conditions that gate on variance rather than on absence, resolving what the previous revision carried as an open question: such a gate is admissible when the column's value is fully determined by another column in the same dataset wherever the Condition is false. Retitled the neighboring sub-bullet from "Only presence gates belong here", which excluded what the new one admits, leaving its body unchanged. Scoped principle 7 prospectively, so adopting the guideline does not reopen a level the specification already ships, with Scope of This Revision saying so and the design notes recording the size of the existing gap (24 of the 46 columns leveled Conditional are directly dependent columns of at least one Supported Feature, spanning 10 features). Corrected the design notes' count of variance-shaped Conditions from "at least one" to six of twenty-five, gating twelve columns. Reworded the Key Concepts pointer to the FOCUS Feature Level section so it claims no more than that section defines, matched "feature levels" and "nullability" casing to the rest of the guideline, retitled the worked-example section as plural, and recorded the fourth and fifth post-round-9 editing passes in the evidence document.

**2026-07-27.** Corrected wording that contradicted itself or the specification (the Overview's applicability claim, the Interim Boundary Rule's all-null case, input 1's presence gate, input 2's producibility scope). Named the deferred destinations and recorded the variance-gate question. Removed restatement by merging principles 5 and 6 and cutting the Overview's preview list. Added principle 7, the Supported Features write-back. Moved the post-round-9 edit record into the evidence document. Cut self-narration and filler. Replaced the overloaded verb "carry" with scope-specific ones, following @ijurica's review.

### Type of Change
* [ ] Bug fix
* [ ] Editorial / Formatting

### Author Checklist
* [ ] **AI-Generated Examples:** If this PR includes examples generated by AI, I have manually verified that the data is mathematically accurate, logically consistent, and compliant with the FOCUS schema. _(N/A: this PR adds prose guidance only and contains no FOCUS data.)_


guidelines/contributors/feature-level-rubric-guidelines.md
```diff
@@ -0,0 +1,165 @@
...
+* [The Two Decision Inputs](#the-two-decision-inputs)
+* [Normative Content Defined by the Rubric](#normative-content-defined-by-the-rubric)
+* [Applying the Principles: A Worked Example](#applying-the-principles-a-worked-example)
+* [Relationship to the Companion Guideline](#relationship-to-the-companion-guideline)
+
+## Scope of This Revision
+
+FOCUS has four feature levels: Mandatory, Conditional, Recommended, and Optional. This revision covers two of them.
+
+Start here once the working group has already decided a column is worth carrying in a dataset. This guideline then picks which of Mandatory or Conditional that column takes, and whether its value may be null.
```
ijurica: ## Terminology: "carry" vs. spec's "include"/"present"

The rubric uses "carry" as its operative verb for dataset/column membership (e.g. "columns... decided to carry", "worth carrying in a dataset", "carries the commitment columns", "no reasonable generator would carry the column null"). 

The spec consistently uses:
- **"include" / "included in"** — normative verb for dataset membership, e.g. glossary: *"A column included in a FOCUS dataset"*; dataset Requirements: *"ContractCommitment MUST include ContractCommitmentId"*. Also the exact verb the model generates from `Function: Presence` rules (`"CostAndUsage MUST include ServiceCategory."`).
- **"present"** — used in metadata-schema context, e.g. *"columns present in the dataset instance artifact"*.

Since this rubric feeds into how Presence rules get worded, an undefined third term risks producing language that doesn't map cleanly onto the model's actual phrasing.

**Suggestion:** replace "carry" with "include" (or "present") throughout, or define it explicitly if it's meant to mean something distinct from both.

Matt-Cowsert: Thanks for catching this Irena. I split the verb by scope rather than unifying on "include", because the eleven instances were doing three jobs: dataset membership ("belongs in a dataset", "each dataset that defines it"), the generator's publication obligation ("publishes the commitment columns", "would publish the column null"), and holding a level ("columns still hold them"). One verb across all three reads right for the first and wrong for the other two, since a generator does not include a column, it publishes one.

The first group is what maps onto Presence rules. If you'd rather have the model's own verb there, I can move "belongs in" and "defines" to "include". Say the word and I'll take it.

Matt-Cowsert:
closed in favor of #2604

PR #2558, opened 2026-07-24
FR #1974: add adjusted validation path for flexibility examples

### LINKED ISSUE #1974: [FR] Ensure specification examples pass FOCUS Validator
### Problem Statement

The FOCUS specification repository contains numerous worked examples across appendices and supporting content (commitment discounts, pricing scenarios, usage patterns, etc.). If an example contains all files then the full example should be ran through the validator (missing columns, validation issues, etc.), else if it is a small subset of columns for an example then it should validate only the columns in the example.

This undermines confidence in the examples as authoritative references and creates friction for practitioners attempting to use them as templates for their own FOCUS implementations.

Key issues observed when validating current examples against FOCUS 1.2:

- Missing MUST columns (e.g., BillingCurrency, InvoiceIssuerName, ChargeDescription, ChargeClass, ServiceSubcategory, InvoiceId)
- ID columns containing purely numeric values that trigger type-inference failures in CSV parsers (e.g., BillingAccountId parsed as integer instead of string)
- Null values in conditionally-required columns (e.g., ContractedCost, ListCost, ListUnitPrice)
- Non-compliant PricingUnit formats (e.g., "Instance Hours", "GB-Month" instead of "Hours", "GB")
- InvoiceIssuerName/ProviderName mismatches triggering third-party billing rules incorrectly

---

### Use Case / User Story

**As a** FOCUS data generator implementing the specification
**I need** the specification's own examples to pass the FOCUS Validator without errors
**So that** I can use them as authoritative reference implementations and validate my output against a known-good baseline

**As a** FOCUS specification contributor
**I need** a repeatable process to validate examples against each spec version
**So that** new examples and spec revisions maintain validator compliance automatically

---

### Success Criteria

- [ ] All CSV/data examples in the FOCUS specification repository pass the FOCUS Validator with zero failures (excluding inherent OR-composite child false-negatives)
- [ ] Examples include all MUST and RECOMMENDED columns defined in the target spec version
- [ ] Column values use correct types, formats, and unit specifications
- [ ] A validation step exists (CI or manual) to catch regressions when examples are modified
- [ ] Process is version-aware and can iterate when new spec versions (e.g., 1.3) add columns or change rules

---

### Type of Request

- [ ] **Enhancement** – Refinement or extension of an existing FOCUS column, metric, or capability
- [ ] **Net New** – A concept not yet represented in FOCUS or widely available from data generators
- [x] **Enablement/Supporting Content** – Non-normative items like appendix explanations, examples, or supporting documentation

---

- FinOps Foundation

---

_No response_

---

- FOCUS Validator project (validates CSV/Parquet against spec rules using DuckDB)
- FR #1488 commitment discount examples — recently upgraded to validator compliance as proof of concept
- FOCUS v1.2 specification — current target version
- FOCUS v1.3 Validator — in development by Mike and team, will introduce new columns and rules

Validation results from FR #1488 example upgrade (before/after):

| Metric            | Before | After                     |
| ----------------- | ------ | ------------------------- |
| Failures per file | 46-50  | 1 (inherent OR-composite) |
| Missing columns   | 6      | 0                         |
| Type mismatches   | 37     | 0                         |
| Null violations   | 13     | 0                         |
| Format violations | 3      | 0                         |

Common root causes discovered during FR #1488 remediation:

| Issue                          | Root Cause                                            | Fix Pattern                                             |
| ------------------------------ | ----------------------------------------------------- | ------------------------------------------------------- |
| BillingAccountId type mismatch | Purely numeric IDs parsed as integers by CSV loaders  | Prefix with alphabetic characters (e.g., "ACCT-")       |
| ContractedCost null violations | Purchase and on-demand rows missing contracted values | Set to appropriate amounts per charge type              |
| PricingUnit format failures    | Multi-word or hyphenated units                        | Use FOCUS Unit Format spec ("Hours", "GB", "GB/Second") |
| BilledCost third-party rule    | InvoiceIssuerName != ProviderName                     | Match exactly for first-party billing scenarios         |
| PricingQuantity type mismatch  | Integer decimals serialized without decimal places    | Format with explicit decimal places (e.g., "1.00")      |

---

26 CSV example files across 4 groups in `specification/data/`:

**Group 1: Commitment Discount Scenarios — Generic (7 files, 13 columns each)**

Location: `commitment_discount_scenarios/`

| File                                        | Rows | Status              |
| ------------------------------------------- | ---- | ------------------- |
| commitment_discount_purchase_scenario_1.csv | 1    | Missing 33+ columns |
| commitment_discount_purchase_scenario_2.csv | 3    | Missing 33+ columns |
| commitment_discount_purchase_scenario_3.csv | 4    | Missing 33+ columns |
| commitment_discount_usage_scenario_1.csv    | 1    | Missing 33+ columns |
| commitment_discount_usage_scenario_2.csv    | 1    | Missing 33+ columns |
| commitment_discount_usage_scenario_3.csv    | 2    | Missing 33+ columns |
| commitment_discount_usage_scenario_4.csv    | 2    | Missing 33+ columns |

Columns present: BillingPeriodStart, BillingPeriodEnd, ChargePeriodStart, ChargePeriodEnd, ChargeCategory, ChargeFrequency, PricingCategory, ResourceId, BilledCost, EffectiveCost, CommitmentDiscountId, CommitmentDiscountQuantity, CommitmentDiscountUnit

**Group 2: Commitment Discount Flexibility (4 files, 23 columns each)**

Location: `commitment_discount_flexibility/`

| File                                                                                      | Rows | Status              |
| ----------------------------------------------------------------------------------------- | ---- | ------------------- |
| one_hundred_percent_utilization_with_commitment_discount_flexibility_with_1_resource.csv  | 5    | Missing 23+ columns |
| one_hundred_percent_utilization_with_commitment_discount_flexibility_with_2_resources.csv | 6    | Missing 23+ columns |
| one_hundred_percent_utilization_without_commitment_discount_flexibility.csv               | 4    | Missing 23+ columns |
| zero_percent_utilization_without_commitment_discount_flexibility.csv                      | 5    | Missing 23+ columns |

Note: Contains non-standard column `x_CommitmentDiscountUnitPrice` and `SkuPriceid` (lowercase 'i' typo).

**Group 3: SaaS Examples — Simple & Spend Agreements (9 files, 50 columns each)**

Location: `saas_examples/simple_agreements/`, `saas_examples/spend_agreements/`

| File                          | Rows | Status           |
| ----------------------------- | ---- | ---------------- |
| simple_saas_agreements_a1.csv | 1    | Needs validation |
| simple_saas_agreements_a2.csv | 1    | Needs validation |
| simple_saas_agreements_a3.csv | 1    | Needs validation |
| simple_saas_agreements_b.csv  | 1    | Needs validation |
| simple_saas_agreements_c.csv  | 3    | Needs validation |
| saas_spend_agreements_a1.csv  | 4    | Needs validation |
| saas_spend_agreements_a2.csv  | 13   | Needs validation |
| saas_spend_agreements_b1.csv  | 5    | Needs validation |
| saas_spend_agreements_b2.csv  | 14   | Needs validation |

Note: These files include columns beyond the current spec version (CapacityReservationId, CapacityReservationStatus, HostProviderName, ServiceProviderName, SkuMeter, SkuPriceDetails, Tags).

**Group 4: SaaS Examples — Virtual Currency (6 files, 35 columns each)**

Location: `saas_examples/`

| File                                  | Rows | Status           |
| ------------------------------------- | ---- | ---------------- |
| virtual_currency_pricing_model_a1.csv | 1    | Needs validation |
| virtual_currency_pricing_model_a2.csv | 3    | Needs validation |
| virtual_currency_pricing_model_b1.csv | 1    | Needs validation |
| virtual_currency_pricing_model_b2.csv | 3    | Needs validation |
| virtual_currency_pricing_model_b3.csv | 3    | Needs validation |
| virtual_currency_pricing_model_c.csv  | 2    | Needs validation |

Note: Contains BOM character in header. Includes PricingCurrency, PricingCurrencyContractedUnitPrice, PricingCurrencyEffectiveCost, PricingCurrencyListUnitPrice columns. Missing ServiceCategory, ServiceSubcategory, SubAccountId, SubAccountName, and several other columns.

1. Run the Validator against all 26 files and catalogue failures per group
2. Remediate groups in priority order: Group 1 and 2 (most gaps), then Group 4 (BOM + missing columns), then Group 3 (validation only)
3. Re-validate after each group remediation
4. Re-run full inventory when the 1.3 Validator becomes available and iterate

**Open Questions:**

1. Should validator compliance be enforced via CI (e.g., GitHub Actions running the validator on PR)?
2. Should generated examples (from tooling like focus-mcp) be the canonical source, or should hand-crafted examples remain?
3. What is the acceptable threshold for "false-negative" child rule failures in OR-composite rules?
4. Should examples target only MUST columns, or also include all RECOMMENDED columns for completeness?

---

_No response_

---

---

## Maintainer Assessment

### Adoption Impact

- [x] **Adoption Accelerator** – This feature would help organizations start using FOCUS
- [ ] **Practitioner Enhancement** – Improves experience for existing FOCUS users
- [ ] **Technical Improvement** – Internal specification refinement

**Justification:**

One organization requested this (FinOps Foundation). Despite low org count, the impact is high: specification examples serve as reference implementations for every data generator adopting FOCUS.

- **Evidence:** The FR documents concrete validation failures across 26 CSV files (46-50 failures per file before remediation in the #1488 proof of concept).
- **Fidelity issue:** When spec examples themselves fail validation, generators cannot use them as authoritative baselines.
- **Adoption position:** The FR #1488 remediation proved the approach works (failures dropped from 46 to 1 per file). This accelerates adoption by giving generators a known-good baseline.
- **Gap:** Unlike purely internal editorial improvements, this FR's output (validated example files) is directly consumed by generators during implementation.

---

**Feature Description:**

This FR supports all existing supported features indirectly by ensuring their worked examples are validator-compliant.

- **Use case impact:** The examples span commitment discount scenarios ("Commit Usage and Under Usage"), SaaS examples ("Billed Cost and Invoice Alignment"), and pricing scenarios ("Verification, Comparison, and Fluctuation Tracking of Unit Prices").
- **Current limitation:** No specific feature is enhanced; rather, the quality of supporting content for multiple features is improved.
- **Fidelity issue:** The "Charge Categorization" feature's example queries depend on correctly populated ChargeCategory, ChargeClass, and ChargeFrequency columns, which current examples populate inconsistently.
- **Design note:** This is an enablement FR that raises the floor for all supported features by making their examples trustworthy.

---

### Implementation Scope

**MVP Definition:**

Remediate all 26 CSV example files to pass the FOCUS Validator for the current spec version:

1. Fix missing MUST columns across all groups (6+ columns per file in Groups 1-2)
2. Correct type mismatches (numeric IDs, decimal formatting)
3. Resolve null violations in conditionally-required columns
4. Fix PricingUnit format violations
5. Correct InvoiceIssuerName/ProviderName inconsistencies

Prioritize by group: Groups 1-2 (most failures, 7+4 files), then Group 4 (BOM + missing columns, 6 files), then Group 3 (validation-only, 9 files).

Add a manual validation checklist to the contribution guidelines so new examples are validated before merge.

Satisfies Success Criteria 1 (zero failures), 2 (all MUST columns), 3 (correct types/formats). Partially satisfies 4 (manual process, not CI). Satisfies 5 (version-aware through Validator version targeting).

Defer CI automation (GitHub Actions integration) to a follow-up. Coordinate with #2012 (automated link validation) if both advance, to share CI infrastructure.

**North Star Vision:**

Every example in the FOCUS specification repository passes the FOCUS Validator automatically on every PR. A CI step in the GitHub Actions workflow runs the Validator against all example files as part of the build pipeline. New examples are generated or validated by tooling before submission. The validation step is version-aware and updates when the spec version changes.

**Phasing Strategy:**

Single phase. The remediation work is systematic and well-scoped (26 files, known failure patterns, proven fix patterns from #1488). CI automation is a natural follow-up but not a blocker for delivering validated examples.

---

### Specification Impact

**Impacted Parties:**
- [ ] Other: [specify]

Existing assessment checked Practitioners, Generators, and Vendors. This assessment narrows to Generators only. Practitioners benefit indirectly but are not directly impacted. Generators are the primary audience: they use examples as implementation references. This FR does not change the spec schema.

**Level of Ambiguity:**

2

The FR includes a detailed inventory of all 26 CSV files across 4 groups, specific failure categories with root causes and fix patterns, and a proven remediation approach from #1488. The work is systematic: run validator, fix failures, re-validate. The main open question is whether to add CI enforcement (GitHub Actions) or maintain manual validation. This is a process decision, not a design question.

Existing assessment rated ambiguity at 2. This assessment agrees.

---

### FinOps Scope Alignment

Existing assessment checked all six scopes. This assessment narrows to Public Cloud and SaaS because those are the scopes the current examples actually cover (commitment discount scenarios, SaaS agreement examples, virtual currency examples).

## Community Support

  - Your **organization**

### LINKED ISSUE #2382: [AI] Draft PR for Validator-Compliant Spec Examples
### Detailed Description

Draft PR that updates existing all worked examples examples to pass the FOCUS Validator.

### Definition of Done

- [x] Draft PR with updated worked examples that cover all MVP requirements outlined in parent issue

### LINKED ISSUE #2561: [AI] Bring the validator-compliant examples PR to TF review
**Assignee(s):** @izadoraluizalabs, @qquigley
**Parent Issue:** #1974
**Target Date:** Thu Aug 27, 2026 (two weeks ahead of the Sep 10 v1.5 gate to start TF review)

**Detailed Description**
PR #2558 is open and carries the substance of this FR: a reusable `validate_examples.py`, a version-scoped exclusion list, corrections to the four commitment discount flexibility example CSVs, and process documentation added to `development-processes.md`. It superseded draft PR #2557. AI #2382, which covered the drafting, is closed, so nothing currently owns moving this from open PR to TF review.

Three things stand between the PR as it is today and a productive TF review.

First, the base branch. PR #2558 targets `main`. Every other open PR in this cycle targets `working_draft`, which is also the repository default and the branch the 1.5 draft is built from. As currently targeted the change would not reach the 1.5 working draft. Retargeting also changes the computed diff, so the CSV corrections need re-verifying against the `working_draft` baseline rather than the published one.

Second, the review feedback. @jpradocueva requested changes on 2026-07-26 with four unresolved threads on `development-processes.md`: code block alignment to the start of the line, the `> **Notes:**` blockquote format, splitting the exclusion list note into separately testable obligations, and a compound noun hyphen on line 247.

Third, the scope statement. The FR covers the specification's examples generally, while this PR covers the commitment discount flexibility set and an exclusion list scoped to v1.2. Reviewers need the PR description to say plainly whether #2558 is intended to close FR #1974 or is the first slice with further example sets to follow, and how the exclusion list advances across spec versions. Without that, TF reviewers cannot tell what they are approving.

One open item this AI does not resolve: FR #1974 has no Task Force assigned on the FOCUS WG board, so the review venue needs confirming before the PR can be scheduled.

**Definition of Done**
- [ ] PR #2558 retargeted from `main` to `working_draft`, with the diff re-verified against that baseline
- [ ] All four @jpradocueva review threads resolved and the changes-requested review cleared
- [ ] Exclusion list version scoping stated on the PR, including which spec versions it applies to and how it advances
- [ ] PR description states whether #2558 closes FR #1974 or is the first slice, naming the remaining example sets if the latter
- [ ] Review venue confirmed and the Task Force field set on FR #1974
- [ ] PR scheduled and in TF review ahead of the Thu Sep 10 v1.5 gate to start TF review

izadoraluizalabs:
## Summary
- add rule exclusion support to `specification/data/validate_examples.py`
- add a version-scoped exclusion list for `commitment_discount_flexibility` (FOCUS 1.2)
- document the adjusted validation workflow in contributor development processes
- update 4 flexibility CSV examples with the latest remediation edits

## Files changed
- `specification/data/validate_examples.py`
- `specification/data/commitment_discount_flexibility_excluded_rules_v1_2.txt`
- `guidelines/contributors/development-processes.md`
- `specification/data/commitment_discount_flexibility/one_hundred_percent_utilization_with_commitment_discount_flexibility_with_1_resource.csv`
- `specification/data/commitment_discount_flexibility/one_hundred_percent_utilization_with_commitment_discount_flexibility_with_2_resources.csv`
- `specification/data/commitment_discount_flexibility/one_hundred_percent_utilization_without_commitment_discount_flexibility.csv`
- `specification/data/commitment_discount_flexibility/zero_percent_utilization_without_commitment_discount_flexibility.csv`

## Context
Related to #1974.

### Full validator (`ALL` criteria)
- `one_hundred_percent_utilization_with_commitment_discount_flexibility_with_1_resource.csv`: `Fail=9`
- `one_hundred_percent_utilization_with_commitment_discount_flexibility_with_2_resources.csv`: `Fail=6`
- `one_hundred_percent_utilization_without_commitment_discount_flexibility.csv`: `Fail=6`
- `zero_percent_utilization_without_commitment_discount_flexibility.csv`: `Fail=9`

### Adjusted validator (approved exclusions for known OR-composite/cascading false-negatives)
- `AdjustedFail=0` for all 4 flexibility files using:
  - `--exclude-rules-file specification/data/commitment_discount_flexibility_excluded_rules_v1_2.txt`


guidelines/contributors/development-processes.md
```diff
@@ -222,6 +222,40 @@ When a contribution requires creating a series of examples, contributors must no
 
 The contributor may only proceed with requesting review of the complete set of examples after the Task Force has formally approved the initial subset. This iterative approach ensures early alignment on formatting, canonical terminology, and technical accuracy, significantly reducing the need for extensive rework later in the review cycle.
 
+#### Validating CSV Examples
+
+Before requesting review for example updates, contributors SHOULD validate the affected CSV examples with the FOCUS Validator.
+
+* Validate all tracked example groups:
+
+  ```bash
```
jpradocueva: @izadoraluizalabs, for this feedback, I am using a new Validation SKILL.

> Rule excerpt: “Code blocks must be aligned to the start of the line (not indented).”

The same applies to the other Bash blocks. They appear indented by two spaces beneath list items. They should be unindented.

guidelines/contributors/development-processes.md
```diff
@@ -222,6 +222,40 @@ When a contribution requires creating a series of examples, contributors must no
...
+  python specification/data/validate_examples.py --group commitment_discount_flexibility --filter-rules CommitmentDiscount --validate-version 1.2 --applicability-criteria ALL
+  ```
+
+* Run adjusted validation when a workstream has approved, documented OR-composite false-negatives:
+
+  ```bash
+  python specification/data/validate_examples.py --group commitment_discount_flexibility --validate-version 1.2 --applicability-criteria ALL --exclude-rules-file specification/data/commitment_discount_flexibility_excluded_rules_v1_2.txt
+  ```
+
+  Notes:
```
jpradocueva:
```suggestion
> **Notes:**
```

jpradocueva: Important notes must use the blockquote format (> **Note:**).

jpradocueva: See an example [here](https://github.com/FinOps-Open-Cost-and-Usage-Spec/FOCUS_Spec/blob/working_draft/guidelines/contributors/normative-requirements-guidelines.md#dataset-normative-requirements-examples).

jpradocueva (changes requested):
@izadoraluizalabs, This review was generated using a new PR-validation skill that is currently under test. It identified two Markdown formatting issues: the added code blocks should not be indented, and the notes should use the required  `> **Note:** ` blockquote format.

guidelines/contributors/development-processes.md
izadoraluizalabs: Thanks, Joaquin — I updated the section to align with the repository’s Markdown conventions:

unindented the Bash code blocks
changed the notes to the required > **Note:** blockquote format
Please take another look when you have a chance.

guidelines/contributors/development-processes.md
```diff
@@ -228,33 +228,34 @@ Before requesting review for example updates, contributors SHOULD validate the a
...
-
-  * `Fail` is the raw validator output.
-  * `AdjustedFail` excludes only the explicitly listed Rule IDs and is the value used for the threshold check.
-  * Exclusion lists MUST be version-scoped, narrowly targeted, and reviewed with the Task Force before merge.
+> **Note:**
+> `Fail` is the raw validator output.
+>
+> `AdjustedFail` excludes only the explicitly listed Rule IDs and is the value used for the threshold check.
+>
+> Exclusion lists MUST be version-scoped, narrowly targeted, and reviewed with the Task Force before merge.
```
jpradocueva: I suggest removing the Exclusion List from the note and breaking it into three different obligations, e.g.,

* Exclusion list MUST be version-scoped.
* Exclusion lists MUST be narrowly targeted.
* Exclusion lists MUST have Task Force approval before merge.

guidelines/contributors/development-processes.md
```diff
@@ -228,33 +228,34 @@ Before requesting review for example updates, contributors SHOULD validate the a
...
 * Validate a specific rule family (for focused remediation):
 
-  ```bash
-  python specification/data/validate_examples.py --group commitment_discount_flexibility --filter-rules CommitmentDiscount --validate-version 1.2 --applicability-criteria ALL
-  ```
+```bash
+python specification/data/validate_examples.py --group commitment_discount_flexibility --filter-rules CommitmentDiscount --validate-version 1.2 --applicability-criteria ALL
+```
 
 * Run adjusted validation when a workstream has approved, documented OR-composite false-negatives:
```
jpradocueva: ### Error: hyphen used in a compound noun
* Line: 247
* Changed content: OR-composite false-negatives
* Rule: editorial-conventions-13: “Use hyphens for compound modifiers...”
false negatives is a plural compound noun in this sentence, not a compound modifier, so it should not be hyphenated.

#### Suggestion

> * Run adjusted validation when a workstream has approved and documented OR-composite false negatives:

izadoraluizalabs (commented):
@jpradocueva — I've addressed all four threads in commit 02392c4. Please take another look when you have a chance.

jpradocueva (commented):

Snapshot totals: 6 conversation comments, 9 reviews, and 7 inline review comments.

@izadoraluizalabs PLEASE WAIT UNTIL I GET FINAL CONFIRMATION FROM THE TF-REQUIREMENTS MODEL ON THESE REPORTED ISSUES. THANKS!

| --- | --- | --- |
| F-001 | Submit inline comment | Normative prose is not formatted as a requirement bullet; the correction does not change schema or normative behavior. |
| F-002 | Submit inline comment | The normative subject is an actor rather than an artifact; behavioral effect requires author clarification because the obligation may change from an activity to a validation result. |
| F-003 | Submit inline comment | The requirement prescribes behavior instead of an artifact state; behavioral effect requires author clarification because passing validation is stronger than running the validator. |

guidelines/contributors/development-processes.md
```diff
@@ -222,6 +222,45 @@ When a contribution requires creating a series of examples, contributors must no
 
 The contributor may only proceed with requesting review of the complete set of examples after the Task Force has formally approved the initial subset. This iterative approach ensures early alignment on formatting, canonical terminology, and technical accuracy, significantly reducing the need for extensive rework later in the review cycle.
 
+#### Validating CSV Examples
+
+Before requesting review for example updates, contributors SHOULD validate the affected CSV examples with the FOCUS Validator.
```
jpradocueva, line 227: ### F-001 — error — [Line 227](https://github.com/izadoraluizalabs/FOCUS_Spec/blob/02392c4841fc2b304372e9973c5e49631661b5c3/guidelines/contributors/development-processes.md#L227)

**Changed content:** `Before requesting review for example updates, contributors SHOULD validate...`

**Rule:** `normative-language-requirements-03` — Writing Specification Content & Review Guidelines → Normative Language & Requirements; position 3. “Use bulleted lists. Each bullet MUST express exactly one verifiable state.”

**Engineering context:** `SHOULD` makes this an enforceable recommendation, which must be structured as an independently testable assertion.

**Problem:** The normative statement is prose rather than a bulleted requirement.

**Impact:** The requirement does not follow the repository’s required structure for normative assertions.

**Suggested correction:** Convert the statement to an asterisk bullet while coordinating its wording with the subject and state requirements that also apply to this line.

jpradocueva, line 227: ### F-002 — error — [Line 227](https://github.com/izadoraluizalabs/FOCUS_Spec/blob/02392c4841fc2b304372e9973c5e49631661b5c3/guidelines/contributors/development-processes.md#L227)

**Changed content:** `contributors SHOULD validate the affected CSV examples`

**Rule:** `normative-language-requirements-08` — Writing Specification Content & Review Guidelines → Normative Language & Requirements; position 8. “Allowed Subjects: MUST be schema-level entities… Actors … MUST NOT be subjects.”

**Engineering context:** A contributor is an actor; an affected CSV example is an artifact comparable to a test fixture.

**Problem:** The normative subject is `contributors` rather than the artifact whose state is constrained.

**Impact:** The statement prescribes human workflow instead of defining an artifact invariant.

**Behavioral effect of the correction:** Behavioral effect requires author clarification: replacing the contributor action with an artifact state may change whether the requirement mandates an activity or a validation result.

**Suggested correction:** Make the affected CSV examples—or another precise dataset artifact—the subject and clarify the intended validation outcome.

jpradocueva, line 227: ### F-003 — error — [Line 227](https://github.com/izadoraluizalabs/FOCUS_Spec/blob/02392c4841fc2b304372e9973c5e49631661b5c3/guidelines/contributors/development-processes.md#L227)

**Changed content:** `contributors SHOULD validate`

**Rule:** `normative-language-requirements-09` — Writing Specification Content & Review Guidelines → Normative Language & Requirements; position 9. “Describe a state, not behavior.”

**Engineering context:** A state-based requirement asserts an observable validation result; `validate` prescribes an action.

**Problem:** The sentence describes validation behavior instead of the required state of the affected examples.

**Impact:** Compliance depends on whether someone performed a process rather than on an objectively testable artifact condition.

**Behavioral effect of the correction:** Behavioral effect requires author clarification: requiring examples to pass validation is stronger than recommending that contributors run the validator.

**Suggested correction:** State the intended observable condition of the examples and retain the commands as informative execution guidance.

jpradocueva: @izadoraluizalabs, the TF-RM agreed at the August 11 meeting that this `Rule` doesn't apply to documents inside the `guidelines` folder. I will modify the AGENTS.md to reflect this change.

jpradocueva: @izadoraluizalabs, the TF-RM agreed at the August 11 meeting that this `Rule` doesn't apply to documents inside the `guidelines` folder. I will modify the AGENTS.md to reflect this change.

jpradocueva: @izadoraluizalabs, the TF-RM agreed at the August 11 meeting that this `Rule` doesn't apply to documents inside the `guidelines` folder. I will modify the AGENTS.md to reflect this change.

jpradocueva (approved):
@izadoraluizalabs, after reviewing the initial changes, the TF-RM agreed that those comments don't apply to normative requirements inside the Guidelines folder.

The content of this PR is valid after applying the focus-pr-validation.SKILL.md

Matt-Cowsert:
adjusted the target branch from main to working_draft

izadoraluizalabs:
@Matt-Cowsert @jpradocueva
Just a quick note for the agenda: my plan is to wrap up the work this week and go for Task Force Review the week starting on August 31.

thecloudman:
@izadoraluizalabs thanks for putting this together. The validation script and the guideline section are a useful addition, and I support the direction. Before this merges I would like to raise a few issues with the CSV changes and the script, and suggest a route to resolving them. I have deliberately not posted these as inline suggestions because the fix is a set of decisions rather than one-line edits.

### 1. The CSV edits change what the examples illustrate

The four flexibility CSVs are linked from the appendix narratives in `specification/appendix/examples/commitment_discount_flexibility_examples/`. After this PR the data and the narratives disagree:

* **`CommitmentDiscountStatus` is now `Active` on every row with a `CommitmentDiscountId`.** The only allowed values are `Used` and `Unused`. The original files used `Used` on covered usage rows and `Unused` on the idle row in the 0% utilization example, which is the whole point of that scenario.
* **`CommitmentDiscountUnit` changed from `Normalized Hour` to `Hours`** in the two flexibility files and from `Hour` to `Hours` in the two non-flexibility files. The overview page states that the two unit names are used specifically to distinguish flexibility from non-flexibility commitments, and each narrative's Outcome section repeats the original unit.
* **`ChargeClass` is set to `Correction` on the on-demand usage rows** in the 1-resource and 0% files. Those rows are not corrections. As a knock-on effect `BilledCost` on those rows dropped from 2.25 and 2.00 to 0.00, while the narratives say the on-demand usage incurs $2.25 and $2.00 in both `BilledCost` and `EffectiveCost`.
* **Smaller inconsistencies:** `ProviderName` is `ExampleCloud` while the narratives name the provider *LatticeScale*; `ConsumedUnit` is `Hour` while `PricingUnit` is `Hours` on the same rows.

Several of the excluded `CommitmentDiscountStatus` rules appear to be failing because of the `Active` value rather than because of an OR-composite defect in the validator, so the exclusion list is partly covering a data problem.

### 2. The script only runs on one machine layout

`run_validator` sets the working directory to `<FOCUS_Spec root>/focus_validator`. That directory does not exist in this repository, so every command in the new guideline section fails with a missing-directory error for anyone who has the validator checked out elsewhere.

### 3. Terminology and placement

* The exclusion file header and the guideline text describe the excluded rules as false negatives. A validator reporting a failure on correct data is a false positive.
* The three `MUST` statements about exclusion lists sit inside the `> **Note:**` block. Repository convention keeps notes informative and expresses obligations as list bullets, which is what @jpradocueva asked for in the earlier thread.
* `CostAndUsage-D-000-M` and `CostAndUsage-D-002-M` are the dataset-level composite rules that cascade from any child failure. Excluding them is defensible, but the header comment should describe them as cascade exclusions so the list stays narrowly targeted as the guideline requires.

### Suggested route

1. Revert the semantic changes in the four CSVs and keep only the structural ones: the added 1.2 columns, the `ChargeClass` column (left empty), and the `SkuPriceid` to `SkuPriceId` header fix. Restore `Used` / `Unused`, `Normalized Hour` / `Hour`, and the original `BilledCost` values.
2. Re-run the validator on the restored data and rebuild the exclusion list from what actually fails on correct data. Any rule that still fails after step 1 is a genuine validator gap and belongs on the list; anything that stops failing was a data issue.
3. Add a `--validator-path` argument or environment variable to the script, default it sensibly, and state the prerequisite in the guideline section.
4. Move the three exclusion-list obligations out of the note into a bulleted list, and change "false negatives" to "false positives" in both the guideline and the exclusion file header.
5. If the intent is to align the CSVs with a corrected 1.2 column set, update the four appendix narrative pages in the same PR so the linked data and prose agree.

Happy to discuss any of these on the next TF call if it is easier.

*Reviewed by thecloudman with AI assistance.*

qquigley:
@thecloudman - I think I've addressed everything you pointed out with my latest commit:

- Reverted the semantic changes. Used/Unused, Normalized Hour/Hour, and the original BilledCost values are restored, ChargeClass=Correction is cleared from the on-demand rows, Replaced ExampleCloud with LatticeScale.
- Added `--validator-path and FOCUS_VALIDATOR_PATH`, with fallback to ./focus_validator then ../focus_validator. The guideline section now states the prerequisite.
-  Changed "false negatives" to "false positives" in both places; moved the three exclusion-list obligations out of the note into bullets; CostAndUsage-D-000-M/D-002-M labeled as cascade exclusions.

> Re-run the validator on the restored data and rebuild the exclusion list from what actually fails on correct data. Any rule that still fails after step 1 is a genuine validator gap and belongs on the list; anything that stops failing was a data issue.

I rebuilt the exclusion list against the validator after the fixes, and all nine entries still fail. The CommitmentDiscountStatus exclusions weren't related to the Active bug.

specification/data/commitment_discount_flexibility/one_hundred_percent_utilization_without_commitment_discount_flexibility.csv
```diff
@@ -1,5 +1,3 @@
...
-
+BillingPeriodStart,BillingPeriodEnd,ChargePeriodStart,ChargePeriodEnd,ChargeCategory,ChargeClass,ChargeFrequency,PricingCategory,ResourceId,PricingQuantity,ListUnitPrice,ListCost,x_CommitmentDiscountUnitPrice,BilledCost,EffectiveCost,ConsumedQuantity,ConsumedUnit,SkuId,SkuPriceId,CommitmentDiscountId,CommitmentDiscountName,CommitmentDiscountType,CommitmentDiscountCategory,CommitmentDiscountQuantity,CommitmentDiscountStatus,CommitmentDiscountUnit,AvailabilityZone,BillingAccountId,BillingAccountName,BillingAccountType,BillingCurrency,CapacityReservationId,CapacityReservationStatus,ChargeDescription,ContractedCost,ContractedUnitPrice,InvoiceId,InvoiceIssuerName,PricingCurrency,PricingCurrencyContractedUnitPrice,PricingCurrencyEffectiveCost,PricingCurrencyListUnitPrice,PricingUnit,ProviderName,PublisherName,RegionId,RegionName,ResourceName,ResourceType,ServiceCategory,ServiceName,ServiceSubcategory,SkuMeter,SkuPriceDetails,SubAccountId,SubAccountName,SubAccountType,Tags
```
flanakin, line 1: 🤖 [AI][Claude] 🚫 **Blocker**

The expanded header adds `ProviderName` and `PublisherName`, and does not include `ServiceProviderName`. Neither of the two added columns exists in the specification these files ship inside.

`specification/datasets/cost_and_usage/columns/serviceprovidername.md` line 47 records it directly:

> `1.3 Introduced as a replacement for the removed ProviderName column`

There is no `providername.md` or `publishername.md` under `specification/datasets/cost_and_usage/columns/` on `working_draft`; the provider columns are `serviceprovidername.md`, `hostprovidername.md`, and `invoiceissuername.md`. And `ServiceProviderName`, which the header omits, is Mandatory with `Allows nulls` = `False`.

These four files are not standalone fixtures — they are linked from the rendered appendix, at `specification/appendix/examples/commitment_discount_flexibility_examples/*.md` line 36 or 37 in each. So the published working draft would ship example data carrying a column its own schema removed three versions ago, and missing a column its own schema requires on every row.

The cause is upstream of the data. The script and every documented command pass `--validate-version 1.2`, so conforming these CSVs to what the validator accepts means conforming them to the 1.2 column set. The 1.2-era columns that came in alongside these two — `AvailabilityZone`, `BillingAccountType`, `CapacityReservationId`, `CapacityReservationStatus`, and the `PricingCurrency*` family — are all still current, which is what makes `ProviderName` and `PublisherName` easy to miss in a 34-column addition.

Two ways out:

1. **Keep 1.2 validation, publish 1.5 data.** Use the current column names in the CSVs and accept that the two renamed columns produce validator findings, recorded on the exclusion list as version-skew rather than as validator defects.
2. **Add only the columns that exist in both.** Drop `ProviderName` and `PublisherName` from the header and add `ServiceProviderName` carrying `LatticeScale`, which is what the narratives already name as the provider.

The same three-column change applies to all four files under `specification/data/commitment_discount_flexibility/`.

No suggestion block, because the header line is over 700 characters and the right fix depends on which of the two routes the Task Force wants.

guidelines/contributors/development-processes.md
```diff
@@ -222,6 +222,51 @@ When a contribution requires creating a series of examples, contributors must no
...
+
+The commands below require a local checkout of the [FOCUS Validator](https://github.com/finopsfoundation/focus_validator) with its dependencies installed. The script looks for the validator in `./focus_validator`, then `../focus_validator`. When the checkout is elsewhere, point at it with `--validator-path` or the `FOCUS_VALIDATOR_PATH` environment variable:
```
flanakin, line 229: 🤖 [AI][Claude] ⚠️ **Should fix**

Every command in this section passes `--validate-version 1.2`, and the section never says why. The working draft is 1.5 — all 27 files under `specification/conditions/` carry `Version Introduced: 1.5` — so a contributor reading this will reasonably assume 1.2 is a typo, or will change it to 1.5 and get a different and more alarming result.

The reason is presumably that 1.2 is the newest version the FOCUS Validator supports, which is a fine reason and worth one sentence. The consequence is worth a second sentence: conforming example data to what a 1.2 validator accepts can pull the data away from the specification the same files publish inside. `ProviderName`, for instance, is valid in 1.2 and was removed in 1.3.

```suggestion
The commands below require a local checkout of the [FOCUS Validator](https://github.com/finopsfoundation/focus_validator) with its dependencies installed. The script looks for the validator in `./focus_validator`, then `../focus_validator`. When the checkout is elsewhere, point at it with `--validator-path` or the `FOCUS_VALIDATOR_PATH` environment variable.

> **Note:** The commands below pass `--validate-version 1.2` because 1.2 is the newest version the FOCUS Validator supports, while the working draft is ahead of it. Findings that arise only from that gap, such as columns renamed or removed after 1.2, are version skew rather than data defects. Resolve them by keeping the example data aligned with the working draft and recording the skew, not by conforming the data to the 1.2 column set.
```

specification/data/validate_examples.py
```diff
@@ -0,0 +1,262 @@
...
+    exit_code: int
+
+
+GROUP_PATTERNS = {
+    "commitment_discount_scenarios": "commitment_discount_scenarios/*.csv",
+    "commitment_discount_flexibility": "commitment_discount_flexibility/*.csv",
+    "saas_simple_agreements": "saas_examples/simple_agreements/*.csv",
+    "saas_spend_agreements": "saas_examples/spend_agreements/*.csv",
+    "saas_virtual_currency": "saas_examples/virtual_currency_pricing_model_*.csv",
+}
```
flanakin, line 40: 🤖 [AI][Claude] ⚠️ **Should fix**

`GROUP_PATTERNS` is a hardcoded allowlist covering three of the eight directories under `specification/data/`. Unlisted today: `ai_model_identity/`, `commitment_eligibility/`, `contract_commitments/`, `correction-handling-examples/`, and `invoice_detail/`.

That would be fine for a one-off script, but this pull request also adds a documented process step telling contributors to "validate the affected CSV examples" before requesting review, and the default with no `--group` is described as "all tracked example groups". A contributor who adds a new example set, runs the documented command, and sees it pass has no way to learn their files were never opened. The failure is silent in the direction that matters.

Discovering groups from the filesystem removes the maintenance burden entirely and makes new directories covered by default:

```suggestion
def discover_groups(data_dir: Path) -> dict[str, str]:
    """Map each directory under data_dir that holds CSVs to its glob pattern."""
    return {
        d.name: f"{d.name}/**/*.csv"
        for d in sorted(data_dir.iterdir())
        if d.is_dir() and any(d.rglob("*.csv"))
    }
```

That needs `--group` to validate against the discovered set rather than a `choices=` list computed at import time, and `iter_files` to use the returned mapping. If an explicit allowlist is preferred — for instance because some directories are known not to validate yet — then the five omitted directories want to be listed with a comment saying why each is excluded, so the gap is visible rather than implicit.

specification/data/validate_examples.py
```diff
@@ -0,0 +1,262 @@
...
+    failed_rules = FAILURE_RULE_RE.findall(process.stdout)
+
+    return ValidationResult(
+        path=csv_path,
+        total=int(match.group("total")),
+        passed=int(match.group("passed")),
+        failed=int(match.group("failed")),
+        skipped=int(match.group("skipped")),
+        failed_rules=failed_rules,
+        exit_code=process.returncode,
```
flanakin, line 206: 🤖 [AI][Claude] ⚠️ **Should fix**

`exit_code` is stored on every `ValidationResult` and never read again. Searching the file, the only other appearances are the field declaration on line 31 and this assignment.

That means the validator’s own exit status is discarded. A run that prints a parseable summary and then exits non-zero — a crash during reporting, an internal error after the counts are emitted, a signal — is indistinguishable from a clean pass, because the threshold check on line 244 looks only at `failed` minus the excluded rules.

Since the field is already carried through, using it costs one condition in `main()`, alongside the existing `worst_failures` check:

```python
    crashed = [r for r in results if r.exit_code not in (0, 1)]
    if crashed:
        print("\nValidator exited abnormally for:")
        for r in crashed:
            print(f"  {r.path.name} (exit {r.exit_code})")
        return 3
```

The tolerated set depends on what the validator returns for “ran fine, found violations” — worth confirming against the validator before picking the values. No suggestion block, since the change belongs in `main()` rather than on this line.

specification/data/validate_examples.py
```diff
@@ -0,0 +1,262 @@
...
+        encoding="utf-8",
+        env=env,
+        check=False,
+    )
+    match = SUMMARY_RE.search(process.stdout)
+    if not match:
+        raise RuntimeError(
+            f"Could not parse validation summary for {csv_path}.\nSTDOUT:\n{process.stdout}\nSTDERR:\n{process.stderr}"
+        )
+    failed_rules = FAILURE_RULE_RE.findall(process.stdout)
```
flanakin, line 197: 🤖 [AI][Claude] ⚠️ **Should fix**

Raising here aborts the whole run. `main()` calls `run_validator` inside the per-file loop with no handler, so one file whose output does not match `SUMMARY_RE` ends the batch with a traceback — after the earlier files have already printed their per-file lines, and before the threshold check produces any verdict.

For an interactive one-off that is acceptable. For something the guidelines direct every contributor to run before requesting review, it means a single malformed or unexpectedly-formatted output hides the status of every file after it, and the partial output on screen looks like a completed run that simply stopped.

Recording the file as unparseable and continuing gives a complete picture, and still fails the run:

```suggestion
    match = SUMMARY_RE.search(process.stdout)
    if not match:
        print(
            f"Could not parse validation summary for {csv_path}.\n"
            f"STDOUT:\n{process.stdout}\nSTDERR:\n{process.stderr}",
            file=sys.stderr,
        )
        return ValidationResult(
            path=csv_path, total=0, passed=0, failed=0, skipped=0,
            failed_rules=[], exit_code=process.returncode or 1,
        )
```

That pairs with reading `exit_code` in the threshold check, so an unparseable file still returns non-zero overall rather than passing as zero failures.

flanakin (commented):
🤖 [AI][Claude] **PR Review**

**Summary:** The validation harness and the guideline section are a real addition and I support the direction. Submitting as a comment rather than an approval because of one issue in the rewritten CSV data: these files publish into the working draft appendix but are being conformed to the FOCUS 1.2 column set, which reintroduces a column removed in 1.3 and drops a Mandatory one. The script findings are all non-blocking robustness points.

### 🚫 Blockers (1)

1. The rewritten flexibility CSVs add `ProviderName` and `PublisherName` and omit `ServiceProviderName`.

### ⚠️ Should fix (4)

1. The guideline section hardcodes `--validate-version 1.2` in five commands without stating why, or what it costs.
2. `GROUP_PATTERNS` covers three of the eight example directories, and the new process step implies full coverage.
3. `ValidationResult.exit_code` is captured and never read.
4. An unparseable summary aborts the whole batch mid-run.

### ✅ What's done well

* `--validator-path` plus `FOCUS_VALIDATOR_PATH` plus two conventional fallbacks, with `_is_validator_checkout` actually probing for `focus_validator/main.py` rather than trusting the path, is a solid resolution chain — and the error message names both ways to fix it.
* Separating `Fail` from `AdjustedFail` in the per-file output, and stating in the guideline note that only `AdjustedFail` feeds the threshold, keeps the exclusion list from hiding what it excludes.
* Version-scoping the exclusion file in its own filename (`..._v1_2.txt`) rather than in a comment makes the scoping enforceable by inspection.
* Distinguishing the child rule exclusions from the dataset-level composite rules that cascade from them, in the file's own comments, keeps the list auditable.

Matt-Cowsert:
## TF2 Action Item - September 9, 2026

PR #2558 passes FOCUS Validator 1.2 as it stands. @qquigley then ran it against 1.3 and found arithmetic that 1.2 does not catch, in how a commitment discount is distributed across the example rows. @Matt-Cowsert set 1.3 as the target the example set should meet and said validator updates already underway make that reachable.

Feature Request (FR) #1974 is scoped to specification examples passing the FOCUS Validator, so an example set that clears only the older version leaves the newer arithmetic checks unexercised. @qquigley carries the pull request through the 1.5 cycle, with @izadoraluizalabs joining again if she returns to it. September 24, 2026 is the next gate in that cycle.

**Scope:**
- [ ] Correct the commitment discount distribution arithmetic that FOCUS Validator 1.3 flags
- [ ] Confirm the example set passes FOCUS Validator 1.3
- [ ] Record any check that cannot pass 1.3, with the reason, on the pull request

**Assignee:** @qquigley

Issue #2588, opened 2026-07-31
[AI] Build a null-value semantics table for the SKU Price effective date columns


Matt-Cowsert:
**Assignee(s):** @Matt-Cowsert
**Parent Issue:** #1057

**Detailed Description**
SkuPriceEffectiveStart and SkuPriceEffectiveEnd carry meaning only as a pair, and the spec does not state what each null combination means. Riley Jenkins set out the four cases: no start and no end means the price applies across all time in both directions; a start with no end means the price applies from that date forward; an end with no start means the price applies from the beginning of the record through that date; both populated means a finite window.

Riley also raised that a reader looking at either column in isolation will query the dataset incorrectly, and that provider-to-provider differences in how the columns get populated make queries non-portable. Matt Cowsert offered to put the scenarios in a table so the group can align on the semantics before the columns are finalized.

**Definition of Done**
- [x] Drafted a table covering the four start and end null combinations and what each means for a price lookup
- [x] Covered the announced-price-change and SKU-deprecation scenarios the columns are meant to serve
- [x] Shared the table with the group for alignment
- [x] Fed the agreed semantics into the SkuPriceEffectiveStart and SkuPriceEffectiveEnd column text on PR #2424

Matt-Cowsert:
Adjusting this due date to align with the next instance of TF-1, where we'll cover this within the appendix that'll supplement the SKU Price dataset

Matt-Cowsert:
#2615 carries the null-value semantics table and the accompanying column text. It is based on `2364-sku-price-dataset`, so it merges into #2424 rather than into `working_draft`.

The appendix gains an Effective Period Boundaries section covering the four combinations of null `SkuPriceEffectiveStart` and null `SkuPriceEffectiveEnd`: what each one means for a price lookup, and where a record with an unbounded side can sit relative to its neighbors. That last point falls out of the non-overlap requirement already in the dataset, since a record unbounded on both sides is the only record its key combination can hold.

The Aura Web rate card gains two SKUs to cover the scenarios these columns exist to serve: a burstable virtual machine whose rate change is published ahead of taking effect, and a legacy object storage tier with no published origin date that is scheduled for withdrawal. The first of those shows that a null end does not mean the record is the one currently in effect.

[...]

Issue #2586, opened 2026-07-31
[AI] Issue a PR removing the IncludesUnitPricing FOCUS condition


Matt-Cowsert:
**Assignee(s):** @ijurica
**Parent Issue:** #2492

**Detailed Description**
The group aligned that no operating model lacks unit pricing. Riley Jenkins and Udam Dewaraja both argued that anything sold has at least one price, negotiated or not, and Michael Flanakin added that a data generator able to issue an invoice can mint an identifier even where the catalog holds a single SKU. On that basis the IncludesUnitPricing condition should come out of the spec rather than be worked around column by column.

PR #2582 already removes IncludesListUnitPrices and re-points its dependents at IncludesUnitPricing. This action item covers the follow-on PR that removes IncludesUnitPricing itself. Irena Jurica will keep it in a separate PR so PR #2582 can proceed on its own merits when the wider change draws pushback.

Removing the condition shifts ListUnitPrice, ContractedUnitPrice, PricingQuantity, PricingUnit, SkuId, SkuMeter, SkuPriceId, and SkuPriceDetails from conditional to mandatory. Per-ChargeCategory nullability requirements for Tax, Credit, and Adjustment are unaffected. The change carries broad impact and goes on the maintainers agenda for Monday, 2026-08-03.

**Definition of Done**
- [ ] Updated `specification/datasets/cost_and_usage/dataset.md` so the affected columns read as mandatory in both the column table and the presence requirements
- [ ] Confirmed the change leaves per-ChargeCategory nullability requirements intact
- [ ] Raised the mandatory-column impact at the maintainers meeting

ijurica:
@Matt-Cowsert, PR #2585 removes both IncludesListUnitPrices and IncudesUnitPricing conditions. Can we change the description of the AI to reflect this?

BTW, I've already created the AI #2584 😄.

Matt-Cowsert:
I believe this is done but will confirm live before closing

ijurica:
PR #2582 issued and later closed in favor of PR #2585.

Issue #2572, opened 2026-07-29
[AI] Update the editorial guidelines for section heading naming and supported column coverage


Matt-Cowsert:
**Assignee(s):** @Matt-Cowsert, @ijurica
**Parent Issue:** #2509

**Detailed Description**
Two conventions surfaced while reviewing PR #2493 that the editorial guidelines do not yet state.

Section headings mix display names and ColumnIds inside a single heading. `## Example SQL Query (Get Total Effective Cost by AllocatedResourceId)` in `data_generator_calculated_split_cost_allocation.md` pairs the display name form with a ColumnId. The guidelines specify display names in prose and entity IDs in normative requirements, and they say nothing about headings, so each author picks differently. @ijurica read headings as prose and recommended display names. @jpradocueva raised the same point on PR #2493, noting the heading is explanatory non-normative text yet carries a ColumnId.

Second, columns listed as supporting columns on a supported features page are not always referenced by any query on that page. @Matt-Cowsert stated the practice going forward: a column listed as a supported column has to appear in at least one query on the same page.

TF2 agreed the guideline gets written first so new content is authored consistently, and that applying it to existing content belongs to the 1.5 consistency review. Guidelines sit outside the specification, so this can land after the August 13, 2026 draft gate, though TF2 targeted the following two weeks.

**Definition of Done**
- [ ] Editorial guidelines state whether section headings use display names or ColumnIds
- [ ] The heading guidance names the display name form (spaces, not PascalCase) so authors do not have to infer it
- [ ] The heading guidance is cross referenced against the existing prose and normative requirement rules so the three cases read as one convention
- [ ] Editorial guidelines state that every column listed as a supporting or directly dependent column appears in at least one query on the same supported features page
- [ ] Both conventions cover supported features pages and appendix example sections

Issue #2565, opened 2026-07-28
[AI] Analyze options for signaling dynamic pricing in the SKU Price dataset


Matt-Cowsert:
**Assignee(s):** @Matt-Cowsert
**Parent Issue:** #1057

**Detailed Description**

Removing the PricingCategory column from the SKU Price dataset dropped the only catalog signal for dynamic pricing, formerly PricingCategory = "Dynamic". It is now the only item left in the Open Questions section of PR #2424.

Deeja Cruz proposed on June 30 extending UnitPriceCategory rather than adding a new column, keeping the signal in a column that is already mandatory and non-nullable. Her proposal covered two values, "Committed" and "Dynamic". The "Committed" half is closed: TF-1 on July 21 decided that whether a rate is subject to a commitment discount does not belong in a price sheet, since at publication time there is no way to know whether a rate will end up covered, and coverage is reflected in EffectiveCost and BilledCost. The "Dynamic" half is still open.

Deeja raised the question that decides the shape of her proposal: whether "Dynamic" is mutually exclusive with "Contracted". A dynamically priced rate under a negotiated contract has a claim on both values, so either one wins by rule or UnitPriceCategory stops being single-valued.

Spot and similar constructs involve availability constraints rather than price alone and were deferred from this MVP, which is part of what the options need to weigh. Matt committed on PR #2424 to bring options to TF-1 with Deeja's proposal as the starting point, and asked her to walk the group through it.

**Definition of Done**

- [ ] Each option is written up with what is gained and lost: extend UnitPriceCategory with "Dynamic", add a separate column for the signal, or ship the MVP with no dynamic pricing signal
- [ ] The mutual-exclusivity question between "Dynamic" and "Contracted" is answered, or deferred with a stated reason
- [ ] The options are presented to TF-1 for a decision
- [ ] The outcome is reflected on PR #2424 and the Open Questions section is updated

Matt-Cowsert:
closing this for now, we will pick this back up once we're in 1.6 development

PR #2595, opened 2026-08-03
AI #2426, #2427: Add SKU Price supported features

### LINKED ISSUE #1057: [FR] Separate dataset for SKUs

### LINKED ISSUE #2426: [AI] Create Supported Feature for Catalog Discovery & Price Estimation
### Detailed Description

Draft the formal documentation for the "Catalog Discovery & Price Estimation" supported feature. This Supported Feature should guide practitioners on how to query the new SKU Price dataset to forecast costs for net-new architectures based on standard list prices. It needs to cover how to filter by `Pricing Category` (e.g., standard vs. dynamic), interpret `SKU Price Eligibility` rules, and calculate projected spend using `Pricing Units` and `Charge Frequencies`.

### Definition of Done

Draft PR with supported feature.

### LINKED ISSUE #2427: [AI] Create Supported Feature for Rate Optimization & Contract Evaluation
### Detailed Description

Draft the formal documentation for the "Rate Optimization & Contract Evaluation" supported feature. This Supported Feature should explain how to join the SKU Price dataset with Cost and Usage data to compare `Standard` versus `Committed` rates. It must detail how to evaluate step-tier pricing using `Volume Tier Minimum` and `Volume Tier Maximum` boundaries, and how to factor in `Contract Commitment Duration Type` and `Contract Commitment Payment Model` to determine the most financially efficient commitment level.

### Definition of Done

Draft PR with Supported Feature.

Matt-Cowsert:
## Summary

Adds two Supported Features for the `SKU Price` dataset, covering [#2426](https://github.com/FinOps-Open-Cost-and-Usage-Spec/FOCUS_Spec/issues/2426) and [#2427](https://github.com/FinOps-Open-Cost-and-Usage-Spec/FOCUS_Spec/issues/2427).

> **Stacked on [#2424](https://github.com/FinOps-Open-Cost-and-Usage-Spec/FOCUS_Spec/pull/2424).** The base branch is `2364-sku-price-dataset`, so this diff shows only the supported features files.

**Catalog Discovery and Price Estimation** (`specification/supported_features/catalog_discovery_and_price_estimation.md`) covers pricing an architecture that does not exist yet: finding the list rates in force at a point in time, estimating cost from a planned quantity expressed in the Pricing Unit, checking which prices a billing account is eligible to receive, separating usage rates from purchase fees and credit values, finding announced price changes and scheduled retirements, and comparing rates across regions.

**Rate Optimization and Contract Evaluation** (`specification/supported_features/rate_optimization_and_contract_evaluation.md`) covers evaluating the rate an organization already pays: measuring what negotiation reduces the rate by, resolving which quantity tier observed consumption falls into, quantifying what reaching the next tier is worth, listing the purchase terms offered for a SKU, and repricing recorded consumption at a contracted rate to size what an agreement is worth.

## Querying the effective date columns

The Open Forum on 2026-07-31 asked that these pages cover how to query against the effective date columns. Both pages state that `SkuPriceEffectiveStart` and `SkuPriceEffectiveEnd` carry meaning only as a pair, that the start is inclusive and the end exclusive, and that either may be null, and every point-in-time query uses the null-tolerant pattern rather than testing one bound. Finding announced price changes is the exception, and says so, because it looks for prices whose applicability changes rather than for prices in force.

This also fixed a correctness problem. Cross-dataset joins now carry the effective date window, so a charge matches the price that applied during its charge period rather than every price ever published under that `SkuPriceId`. Both pages also state plainly that `SkuPriceId` is not a row key, since that is what makes the unconstrained join wrong.

Two queries on Rate Optimization aggregate consumption over a supplied range before resolving the price record, and evaluate the effective window against the earliest charge period start. Both state that the range supplied is one falling within a single effective window, since a range crossing a price change would otherwise price the whole range at the record in force at its start.

The four null combinations are described in prose here rather than as a table, to leave [#2588](https://github.com/FinOps-Open-Cost-and-Usage-Spec/FOCUS_Spec/issues/2588) to place the normative table on the column definitions.

## Column references in the action items

Both action items were written in May, before several columns were renamed or dropped. The pages use the current names.

| Action item text | What the pages use |
| :--- | :--- |
| `Volume Tier Minimum` / `Volume Tier Maximum` | `QuantityTierMinimum` / `QuantityTierMaximum` |
| `Contract Commitment Duration Type` / `Contract Commitment Payment Model` | `PurchaseDurationType` / `PurchasePaymentModel` |
| `Charge Frequencies` | Dropped from `SKU Price` under #2519. The pages use `ChargeCategory` to separate consumption rates, acquisition fees, and credit values. |
| `Pricing Category`, standard versus dynamic | `SKU Price` has no equivalent column. The standard versus negotiated split is now `ListUnitPrice` against `ContractedUnitPrice`. Signaling dynamic pricing is still open under #2565, so nothing here claims to filter for it. |

## Behavior when a conditional column is absent

Principle 7 of the feature-level rubric in [#2552](https://github.com/FinOps-Open-Cost-and-Usage-Spec/FOCUS_Spec/pull/2552) requires a Supported Feature to still hold when a conditional column it lists is not in the dataset instance. Each page has a "Scope When Conditional Columns are Absent" section naming, per condition, which capability narrows and which still holds.

Carrying both prices on one row helps here: `ListUnitPrice` is present in every SKU Price dataset instance, so reading and comparing public rates holds regardless of which conditions a service provider meets, and of the two prices only the negotiated half narrows, with `IncludesContractCommitments`. Five conditions are named across the two pages and each is covered: `IncludesCommitmentDiscounts`, `IncludesContractCommitments`, `IncludesPurchases`, `IncludesQuantityTierPricing`, and `IncludesRegions`.

Neither page names `IncludesUnitPricing`, since [#2586](https://github.com/FinOps-Open-Cost-and-Usage-Spec/FOCUS_Spec/issues/2586) removes it. Both defer to the data model for when the dataset is present. That deferral has already paid off: the base branch re-gated the `SKU Price` dataset on the new `IncludesSkuPrices` condition and neither page needed an edit. The Cost and Usage columns that carry a price reference are still gated on `IncludesUnitPricing`, and Rate Optimization covers the absence of `SkuPriceId` without naming a condition, so the rest of #2586 lands without an edit here either.

## Conventions from #2572

Section headings use display names rather than Column IDs, and every column listed as directly dependent or supporting is referenced by at least one query on the same page. The check runs both ways: no listed column goes unreferenced, and no column referenced in a query is missing from the lists. Catalog Discovery lists 16 columns and Rate Optimization lists 21.

## Not in this PR

`SKU Price` appendix sample data ([#2428](https://github.com/FinOps-Open-Cost-and-Usage-Spec/FOCUS_Spec/issues/2428)) is a separate deliverable. It landed on the base branch and arrives here through the merge rather than as new work in this diff.

## Changelog

**2026-09-08: Review corrections**

Applying the suggestion in `83371202` dropped two things from **Estimate the Cost of a Planned Workload**: the null branches on both effective date bounds, and the closing code fence. `SkuPriceEffectiveEnd` is null while a price is currently active, so the unguarded comparison dropped every live price from the result. The unbalanced fence rendered lines 94 through 143 as one code block and failed `gen_pdf` with `MD022` at `spec.md:18035` in the assembled spec. Both are restored and `make` runs clean.

**Project the Effect of Moving Consumption to a Contracted Rate** subtracted a `PricingCurrency` amount from a `BillingCurrency` amount. `EffectiveCost` is denominated in the `BillingCurrency` while `ContractedUnitPrice` is denominated in the `PricingCurrency`, and nothing established equality between them. `ObservedUsage` now carries `BillingCurrency` through the aggregation, the join constrains it to `SP.PricingCurrency`, and the prose states that consumption billed in a currency the negotiated rate is not quoted in does not return, since `SKU Price` carries no conversion rate.

**Resolve the Quantity Tier That Applies to Observed Consumption** and **Project the Effect of Moving Consumption to a Contracted Rate** both resolve the effective date window against the earliest charge period start, so a range crossing a price change prices the whole range at the opening record. Both pages now state that the range supplied falls within a single effective window. Partitioning by effective window before aggregation was set aside because a tier boundary is evaluated against the quantity accumulated over the pricing period, and a price version boundary is not necessarily a pricing period boundary.

**2026-08-18: Query corrections and base branch alignment**

The base branch is merged in, so both pages read against the current `SKU Price` definitions.

Three queries returned rows that did not resolve to a single `SKU Price` record.

* **Project the Effect of Moving Consumption to a Contracted Rate** joined without a tier predicate, so an agreement that negotiated several tiers returned one row per tier, each repricing the full aggregated quantity at a different rate. Against a three-tier agreement with 100,000 units consumed it returned three rows, and the first row under its own `ORDER BY` overstated the reduction by 21 percent. The join now resolves the tier, and the tier bounds are returned so the resolved tier is visible.
* **Quantify the Effect of Reaching the Next Quantity Tier** matched the next tier on the boundary alone, so a public tier also paired with a negotiated record carrying the same boundary, and the successor record was never bounded in time. Against the rate card in the `SKU Price` examples it returned four rows for two tier transitions. The two records now match on `ServiceProviderName` and `ContractId` as well, and the same point in time bounds both.
* **Resolve the Quantity Tier That Applies to Observed Consumption** returns a row per agreement where a negotiated rate sits on the resolved tier. `ContractId` is now selected so the rows are distinguishable, and the prose states it.

Two base branch changes contradicted the pages. `ChargeCategory` gained "Credit", so the two-value Usage and Purchase framing is replaced on both pages, including the scope bullet that claimed every row carries "Usage" where purchases are absent. `ListUnitPrice` became non-nullable, so the rationale for excluding null list prices and five `ListUnitPrice IS NOT NULL` predicates are removed. `PurchaseDurationType` may also be null on a "Purchase" row, which the purchase term paragraph did not allow for.

**Rate Optimization and Contract Evaluation** now points to Verification, Comparison, and Fluctuation Tracking of Unit Prices, which reads the same two prices as recorded on a charge rather than as published in the catalog.

**2026-08-09: Commitment discount boundary**

Both pages described capabilities that hold only before a commitment discount is applied, without saying so.

* **Catalog Discovery and Price Estimation** now names its estimate a pre-commitment estimate, and points to Cost Comparison for sizing the effect of existing coverage.
* **Rate Optimization and Contract Evaluation** now states that the List minus Contracted difference is the negotiated portion of an agreement rather than its full effect. The section heading and the overview table description change from "what an agreement reduces a rate by" to "what negotiation reduces a rate by."
* **Project the Effect of Moving Consumption to a Contracted Rate** excluded nothing, so for charges a commitment already covered it subtracted an Effective Cost that reflects the commitment from a contracted rate that does not, returning a negative reduction. The CTE now excludes covered consumption. `CommitmentDiscountId` joins the supporting columns, and its absence case is stated.

### Type of Change
* [ ] Bug fix
* [ ] Editorial / Formatting

### Author Checklist


Matt-Cowsert:
## TF1 Action Item - August 11, 2026

PR #2595 bundles the two SKU Price supported features, covering AI #2426 and AI #2427. Matt Cowsert named the reason it is still in draft: the directly dependent and supporting column lists have not been checked against the SQL queries in the same files.

The standard he stated on the call is that a directly dependent column belongs in one of the queries, and a supporting column that appears in no query comes off the list. The purpose is congruency between feature levels and supported features, so that a question about which use case a set of columns serves resolves to a named combination of columns and the query that consumes them.

**Scope:**

- [ ] Check that every directly dependent column in both supported features appears in one of the SQL queries
- [ ] Check that every supporting column appears in a SQL query, and remove the ones that do not
- [ ] Take this PR out of draft once both checks pass

**Assignee:** @Matt-Cowsert

Matt-Cowsert:
discussed in Maintainers on 2026-08-17. This is on track to pull out of draft by EOW.

specification/supported_features/catalog_discovery_and_price_estimation.md
```diff
@@ -0,0 +1,211 @@
...
+  SP.PricingCurrency,
+  SP.PricingCurrencyCategory,
+  PU.PlannedQuantity * SP.ListUnitPrice AS EstimatedAmount
+FROM PlannedUsage PU
+INNER JOIN SkuPrice SP
+  ON PU.SkuPriceId = SP.SkuPriceId
+WHERE (SP.SkuPriceEffectiveStart IS NULL OR SP.SkuPriceEffectiveStart <= ?)
+  AND (SP.SkuPriceEffectiveEnd IS NULL OR SP.SkuPriceEffectiveEnd > ?)
+ORDER BY EstimatedAmount DESC
+```
```
mike-finopsorg:
```suggestion
```sql
WITH PlannedUsage (SkuPriceId, PlannedQuantity, PlannedDate) AS (
  VALUES (?, ?, ?)
)
SELECT
  PU.SkuPriceId,
  SP.SkuPriceDescription,
  SP.PricingUnit,
  PU.PlannedQuantity,
  SP.ListUnitPrice,
  SP.PricingCurrency,
  SP.PricingCurrencyCategory,
  PU.PlannedQuantity * SP.ListUnitPrice AS EstimatedListAmount
FROM PlannedUsage PU
LEFT JOIN SkuPrice SP
  ON SP.SkuPriceId = PU.SkuPriceId
  AND SP.SkuPriceEffectiveStart <= PU.PlannedDate
  AND SP.SkuPriceEffectiveEnd > PU.PlannedDate
ORDER BY SP.PricingCurrencyCategory, SP.PricingCurrency, EstimatedListAmount DESC
```
```

the current query only returns one plannedusage, this simplifying query matches the above text more closely

**Review by mike-finopsorg — APPROVED** [review]

File: specification/supported_features/catalog_discovery_and_price_estimation.md [hunk 2]
```diff
@@ -0,0 +1,210 @@
...
+
+### Scope When Conditional Columns are Absent
+
+This feature applies wherever a *service provider* publishes a SKU Price dataset, and the data model states when that dataset is present. Every column this feature directly depends on is present in every SKU Price dataset instance.
+
+One capability narrows. Pricing Region ID is present when the [*operating model*](#glossary:operating-model) [includes regions](#conditions.includesregions). Where it is absent, prices do not vary by location and a single price stands for every region, so comparing rates across regions does not apply rather than returning an incomplete result.
+
+## Directly Dependent Columns
+
+* [SkuPrice](#datamodel.skuprice)
```
jpradocueva, line 36: ### F-001 — error — [Line 36](https://github.com/FinOps-Open-Cost-and-Usage-Spec/FOCUS_Spec/blob/83371202f5ab795d68d081cafff852d098dbe6de/specification/supported_features/catalog_discovery_and_price_estimation.md#L36)

**Changed content:** `* [SkuPrice](#datamodel.skuprice)`

**Rule:** `linking-rule-01` — Writing Specification Content & Review Guidelines > Editorial Conventions > Linking Rule; position 1. “Link the first remaining occurrence in reading order. Leave all later occurrences unlinked.”

**Engineering context:** Entity links behave like first-use symbol references: link the first occurrence and use plain text thereafter.

**Related locations:** Line 53.

**Problem:** `SKU Price` is already linked at line 5. Lines 36 and 53 repeat the same destination, and no applicable exception covers these lists.

**Impact:** The document violates the selected first-occurrence linking convention.

**Suggested correction:** Render the dataset identifiers at lines 36 and 53 as plain `SkuPrice` text while retaining the list structure.

specification/supported_features/catalog_discovery_and_price_estimation.md
```diff
@@ -0,0 +1,210 @@
...
+  PU.PlannedQuantity,
+  SP.ListUnitPrice,
+  SP.PricingCurrency,
+  SP.PricingCurrencyCategory,
+  PU.PlannedQuantity * SP.ListUnitPrice AS EstimatedListAmount
+FROM PlannedUsage PU
+LEFT JOIN SkuPrice SP
+  ON SP.SkuPriceId = PU.SkuPriceId
+  AND SP.SkuPriceEffectiveStart <= PU.PlannedDate
+  AND SP.SkuPriceEffectiveEnd > PU.PlannedDate
```
jpradocueva, line 111: ### F-003 — error — [Lines 110–111](https://github.com/FinOps-Open-Cost-and-Usage-Spec/FOCUS_Spec/blob/83371202f5ab795d68d081cafff852d098dbe6de/specification/supported_features/catalog_discovery_and_price_estimation.md#L110-L111)

**Changed content:** `AND SP.SkuPriceEffectiveStart <= PU.PlannedDate` / `AND SP.SkuPriceEffectiveEnd > PU.PlannedDate`

**Rule:** `mathematical-schema-accuracy-01` — Writing Specification Content & Review Guidelines > Validation & Schema Accuracy > Mathematical & Schema Accuracy; position 1. “AI reviewers MUST rigorously calculate, parse, and verify all data within examples.”

**Engineering context:** SQL comparisons with `NULL` evaluate to unknown, so an absent date bound needs an explicit `IS NULL` branch to represent an unbounded interval.

**Problem:** The feature defines either effective-date bound as nullable and prescribes `(bound IS NULL OR comparison)`, but this query omits both null branches.

**Impact:** Open-ended prices fail to match the planned workload even when they apply.

**Suggested correction:** Apply the document’s stated null-bound pattern.

```suggestion
  AND (SP.SkuPriceEffectiveStart IS NULL OR SP.SkuPriceEffectiveStart <= PU.PlannedDate)
  AND (SP.SkuPriceEffectiveEnd IS NULL OR SP.SkuPriceEffectiveEnd > PU.PlannedDate)
```

specification/supported_features/catalog_discovery_and_price_estimation.md
```diff
@@ -0,0 +1,210 @@
...
+ORDER BY SP.PricingCurrencyCategory, SP.PricingCurrency, EstimatedListAmount DESC
+
+### Identify the Prices a Billing Account is Eligible For
```
jpradocueva, line 114: ### F-004 — error — [Line 114](https://github.com/FinOps-Open-Cost-and-Usage-Spec/FOCUS_Spec/blob/83371202f5ab795d68d081cafff852d098dbe6de/specification/supported_features/catalog_discovery_and_price_estimation.md#L114)

**Changed content:** `### Identify the Prices a Billing Account is Eligible For`

**Rule:** `writing-specification-content-review-guidelines-01` — Writing Specification Content & Review Guidelines; position 1. “AI agents generating or reviewing content MUST act as strict technical editors enforcing the FOCUS standards. Focus entirely on specification documents, schema definitions, and markdown formatting.”

**Engineering context:** A Markdown fenced block remains open until another matching fence is encountered, so headings inside it render as code.

**Problem:** The SQL fence opened at line 94 is not closed after line 112. The document has an odd number of fence markers, causing this heading and alternating later sections to be parsed incorrectly.

**Impact:** The remainder of the feature is structurally malformed when assembled or rendered.

**Suggested correction:** Insert a closing triple-backtick fence between lines 112 and 114. A suggestion block is omitted because safely representing a literal Markdown fence inside the submission payload requires discussion rather than a direct replacement.

specification/supported_features/rate_optimization_and_contract_evaluation.md
```diff
@@ -0,0 +1,253 @@
...
+* Quantity tier analysis uses Quantity Tier Minimum and Quantity Tier Maximum, present when the *operating model* [includes quantity tier pricing](#conditions.includesquantitytierpricing). Where they are absent, a SKU Price ID carries one rate that applies at any quantity, so consumption cannot sit in the wrong tier and this capability does not apply. Repricing consumption at a contracted rate resolves the tier through the same two columns; where they are absent, an agreement carries one rate per SKU Price ID and the tier predicates drop out.
+* Purchase term evaluation uses Purchase Duration Type and Purchase Payment Model, present when the *operating model* [includes purchases](#conditions.includespurchases). Where they are absent, the catalog publishes no acquisition fees, no row carries a Charge Category of "Purchase", and this capability does not apply.
+
+List Unit Price is present in every SKU Price dataset instance, so reading and comparing public rates over time holds regardless of which of the three conditions a *service provider* meets.
+
+## Directly Dependent Columns
+
+* [SkuPrice](#datamodel.skuprice)
```
jpradocueva, line 37: ### F-005 — error — [Line 37](https://github.com/FinOps-Open-Cost-and-Usage-Spec/FOCUS_Spec/blob/83371202f5ab795d68d081cafff852d098dbe6de/specification/supported_features/rate_optimization_and_contract_evaluation.md#L37)

**Changed content:** `* [SkuPrice](#datamodel.skuprice)`

**Rule:** `linking-rule-01` — Writing Specification Content & Review Guidelines > Editorial Conventions > Linking Rule; position 1. “Link the first remaining occurrence in reading order. Leave all later occurrences unlinked.”

**Engineering context:** Each entity destination should be linked on first use and then referenced as plain text.

**Related locations:** Lines 48 and 58.

**Problem:** `SKU Price` and `Cost and Usage` are already linked at line 5. Lines 37 and 48 repeat the SKU Price link, and line 58 repeats the Cost and Usage link.

**Impact:** The document violates the selected first-occurrence linking convention.

**Suggested correction:** Render the later list roots as plain `SkuPrice` and `CostAndUsage` text, preserving their canonical schema IDs.

specification/supported_features/rate_optimization_and_contract_evaluation.md
```diff
@@ -0,0 +1,253 @@
...
+The join carries the effective date window so that consumption matches the price that applied during the range, not every price ever published under that SKU Price ID. A SKU Price ID that also carries a negotiated rate on the resolved tier returns one row per agreement, and Contract ID is null on the public record.
+
+> **Note:** Whether the resolved rate applies only to the units inside that tier or retroactively to all units consumed is a property of the published pricing terms for the offering rather than of the tier boundaries, so the tier returned here identifies the applicable rate rather than recalculating the charge.
+
+```sql
+WITH PeriodQuantity AS (
+  SELECT
+    SkuPriceId,
+    PricingUnit,
+    MIN(ChargePeriodStart) AS EarliestChargePeriodStart,
```
jpradocueva, line 112: ### F-006 — error — [Line 112](https://github.com/FinOps-Open-Cost-and-Usage-Spec/FOCUS_Spec/blob/83371202f5ab795d68d081cafff852d098dbe6de/specification/supported_features/rate_optimization_and_contract_evaluation.md#L112)

**Changed content:** `MIN(ChargePeriodStart) AS EarliestChargePeriodStart`

**Rule:** `mathematical-schema-accuracy-01` — Writing Specification Content & Review Guidelines > Validation & Schema Accuracy > Mathematical & Schema Accuracy; position 1. “AI reviewers MUST rigorously calculate, parse, and verify all data within examples.”

**Engineering context:** Aggregating records before selecting a time-versioned reference row is safe only when the entire group belongs to one effective window.

**Related locations:** Lines 116 and 137–138.

**Problem:** The query aggregates an arbitrary time range but selects the SKU Price row using only the earliest charge start. If the range crosses a price-effective boundary, later consumption is assigned to the initial price record.

**Impact:** The query can return the wrong tier or rate for part of the aggregated consumption.

**Suggested correction:** Constrain and validate the input as lying inside one effective window, or associate charges with effective SKU Price records before aggregation. The choice requires author intent.

specification/supported_features/rate_optimization_and_contract_evaluation.md
```diff
@@ -0,0 +1,253 @@
...
+Consumption already covered by a *commitment discount* is excluded. Its Effective Cost already reflects that commitment while Contracted Unit Price does not, so including it would subtract the two against different baselines and report the agreement as raising cost rather than lowering it.
+
+The join resolves the quantity tier the aggregated consumption falls into, so an agreement that negotiated several tiers reprices at the one rate that applies rather than once per tier.
+
+```sql
+WITH ObservedUsage AS (
+  SELECT
+    SkuPriceId,
+    PricingUnit,
+    MIN(ChargePeriodStart) AS EarliestChargePeriodStart,
```
jpradocueva, line 215: ### F-007 — error — [Line 215](https://github.com/FinOps-Open-Cost-and-Usage-Spec/FOCUS_Spec/blob/83371202f5ab795d68d081cafff852d098dbe6de/specification/supported_features/rate_optimization_and_contract_evaluation.md#L215)

**Changed content:** `MIN(ChargePeriodStart) AS EarliestChargePeriodStart`

**Rule:** `mathematical-schema-accuracy-01` — Writing Specification Content & Review Guidelines > Validation & Schema Accuracy > Mathematical & Schema Accuracy; position 1. “AI reviewers MUST rigorously calculate, parse, and verify all data within examples.”

**Engineering context:** Repricing a range against a versioned catalog requires every aggregated charge to share the selected price-effective window.

**Related locations:** Lines 219 and 243–244.

**Problem:** The query aggregates an arbitrary range and chooses its SKU Price record using only the earliest charge start. A range spanning a catalog change is repriced entirely with the initial effective record.

**Impact:** `ProjectedContractedAmount` and `ProjectedReduction` can be materially wrong.

**Suggested correction:** Match each charge or time partition to its effective SKU Price record before aggregation, or explicitly constrain the query to one effective window.

specification/supported_features/rate_optimization_and_contract_evaluation.md
```diff
@@ -0,0 +1,253 @@
...
+  OU.SkuPriceId,
+  OU.PricingUnit,
+  OU.TotalPricingQuantity,
+  OU.TotalEffectiveCost,
+  SP.QuantityTierMinimum,
+  SP.QuantityTierMaximum,
+  SP.ContractedUnitPrice,
+  SP.PricingCurrency,
+  OU.TotalPricingQuantity * SP.ContractedUnitPrice AS ProjectedContractedAmount,
+  OU.TotalEffectiveCost - (OU.TotalPricingQuantity * SP.ContractedUnitPrice) AS ProjectedReduction
```
jpradocueva, line 236: ### F-008 — warning — [Lines 235–236](https://github.com/FinOps-Open-Cost-and-Usage-Spec/FOCUS_Spec/blob/83371202f5ab795d68d081cafff852d098dbe6de/specification/supported_features/rate_optimization_and_contract_evaluation.md#L235-L236)

**Changed content:** `OU.TotalEffectiveCost - (OU.TotalPricingQuantity * SP.ContractedUnitPrice)`

**Rule:** `mathematical-schema-accuracy-01` — Writing Specification Content & Review Guidelines > Validation & Schema Accuracy > Mathematical & Schema Accuracy; position 1. “AI reviewers MUST rigorously calculate, parse, and verify all data within examples.”

**Engineering context:** Subtracting monetary values is dimensionally valid only when both operands use the same currency.

**Problem:** The projected contracted amount is denominated by `SP.PricingCurrency`, but the query neither carries the currency associated with `TotalEffectiveCost` nor establishes equality or performs conversion.

**Impact:** The query may calculate a plausible-looking reduction from amounts expressed in different currencies.

**Suggested correction:** Carry the source cost currency through the aggregation and establish currency equality or an explicit conversion before subtraction. If identical denominations are guaranteed, state and enforce that precondition.

jpradocueva (commented):

Snapshot totals: 6 conversation comments, 2 reviews, and 1 inline review comment.

| --- | --- | --- |
| F-001 | Submit inline comment | Repeated SKU Price links violate the first-occurrence convention; correction does not change schema or normative behavior. |
| F-003 | Submit inline comment | Nullable effective bounds are not handled, excluding applicable open-ended prices; correction does not change schema or normative behavior. |
| F-004 | Submit inline comment | An unclosed SQL fence corrupts subsequent Markdown rendering; correction does not change schema or normative behavior. |
| F-005 | Submit inline comment | Repeated dataset links violate the first-occurrence convention; correction does not change schema or normative behavior. |
| F-006 | Submit inline comment | Range aggregation selects a price using only the earliest charge, potentially mispricing later consumption; correction does not change schema or normative behavior. |
| F-007 | Submit inline comment | Contract projection uses only the earliest charge to select a price window, potentially mispricing the range; correction does not change schema or normative behavior. |
| F-008 | Submit inline comment | Projected reduction may subtract amounts in different currencies; correction does not change schema or normative behavior. |

specification/supported_features/catalog_discovery_and_price_estimation.md
Matt-Cowsert: I'd like to keep these linked. The supported features section links the dataset name at every column-list root today: `invoice_reconciliation.md` on `working_draft` links `InvoiceDetail` at lines 5, 11, and 24, and `CostAndUsage` at lines 5, 17, and 29. Unlinking lines 36 and 53 would make this the only page in the section that does not.

The anchor text also differs between the two positions. Line 5 links the display name "SKU Price" in running prose; lines 36 and 53 link the dataset ID `SkuPrice` as the root of a column list. The first-occurrence rule exempts functional links that use different anchor text, and a list root is navigation into the dataset rather than a repeated prose reference.

If the section convention should change, it would be worth doing across all of `supported_features/` in its own PR so the pages stay consistent with each other.

specification/supported_features/catalog_discovery_and_price_estimation.md
Matt-Cowsert: Applied. Both bounds allow nulls, and `SkuPriceEffectiveEnd` MUST be null while a price is currently active, so the unguarded comparison dropped every live price from the result.

This was a regression from applying the suggestion in `83371202`. The query carried the `(bound IS NULL OR ...)` pattern before that commit and lost it there.

specification/supported_features/catalog_discovery_and_price_estimation.md
Matt-Cowsert: Closing fence added after the `ORDER BY` on line 112. Same origin as the null bounds: applying the suggestion in `83371202` consumed the closing fence, so lines 94 through 143 rendered as one code block.

This was breaking the build. `gen_pdf` has failed on this PR since that commit, reporting `MD022` at `spec.md:18035` in the assembled spec, which is the downstream effect of the unbalanced fence. `make` runs clean with the fence restored.

specification/supported_features/rate_optimization_and_contract_evaluation.md
Matt-Cowsert: I'd like to keep these linked. The supported features section links the dataset name at every column-list root today: `invoice_reconciliation.md` on `working_draft` links `InvoiceDetail` at lines 5, 11, and 24, and `CostAndUsage` at lines 5, 17, and 29. Unlinking lines 37, 48, and 58 would make this the only page in the section that does not.

The anchor text also differs between the two positions. Line 5 links the display names "SKU Price" and "Cost and Usage" in running prose; lines 37, 48, and 58 link the dataset IDs `SkuPrice` and `CostAndUsage` as the roots of column lists. The first-occurrence rule exempts functional links that use different anchor text, and a list root is navigation into the dataset rather than a repeated prose reference.

If the section convention should change, it would be worth doing across all of `supported_features/` in its own PR so the pages stay consistent with each other.

specification/supported_features/rate_optimization_and_contract_evaluation.md
Matt-Cowsert: Stated the precondition in prose rather than restructuring the query. The paragraph above the query now says the effective date window is evaluated against the earliest charge period start, that the range supplied is one falling within a single effective window, and that a range crossing a price change resolves the tier against the record in force at the start of the range.

Partitioning consumption by effective window before aggregation would change what the example demonstrates. A tier boundary is evaluated against the quantity accumulated over the pricing period, and a price version boundary is not necessarily a pricing period boundary, so partitioning there would resolve tiers against quantities the pricing terms do not use. Naming the assumption keeps the tier mechanic legible in an illustrative query.

specification/supported_features/rate_optimization_and_contract_evaluation.md
Matt-Cowsert: Stated the precondition in prose rather than restructuring the query. The paragraph above the query now says the effective date window is evaluated against the earliest charge period start and that the range supplied is one falling within a single effective window.

Matching each charge to its effective SKU Price record before aggregation would change what the example demonstrates. The tier is resolved against the quantity accumulated over the pricing period, and a price version boundary is not necessarily a pricing period boundary, so partitioning there would reprice against quantities the negotiated tiers do not use. Naming the assumption keeps the repricing mechanic legible in an illustrative query.

specification/supported_features/rate_optimization_and_contract_evaluation.md
Matt-Cowsert: Carried the currency through the aggregation and constrained the join. `ObservedUsage` now selects and groups by `BillingCurrency`, and the join adds `AND OU.BillingCurrency = SP.PricingCurrency`.

Added prose above the query stating that Effective Cost is denominated in the Billing Currency while Contracted Unit Price is denominated in the Pricing Currency, and that consumption billed in a currency the negotiated rate is not quoted in does not return, because the SKU Price dataset does not carry a conversion rate.

specification/supported_features/catalog_discovery_and_price_estimation.md
Matt-Cowsert: Correction to my reply above: the section does not link every column-list root today. `contract_commitments.md` on `working_draft` carries plain `CostAndUsage` and `ContractCommitment` roots (lines 11 and 16), and `invoice_reconciliation.md` is the only page that links its roots (lines 11, 17, 24, and 29). So the section is already split one page each way. Keeping the links here does not make this page the odd one out, but it does not follow a settled convention either. The anchor-text point stands on its own. My preference is unchanged: settle the root-link convention across `supported_features/` in one pass rather than page by page.

specification/supported_features/rate_optimization_and_contract_evaluation.md
Matt-Cowsert: Correction to my reply above: the section does not link every column-list root today. `contract_commitments.md` on `working_draft` carries plain `CostAndUsage` and `ContractCommitment` roots (lines 11 and 16), and `invoice_reconciliation.md` is the only page that links its roots (lines 11, 17, 24, and 29). So the section is already split one page each way. Keeping the links here does not make this page the odd one out, but it does not follow a settled convention either. The anchor-text point stands on its own. My preference is unchanged: settle the root-link convention across `supported_features/` in one pass rather than page by page.

specification/supported_features/catalog_discovery_and_price_estimation.md
jpradocueva: Thanks, Matt. I logged this in AI #2657 to address it during the consistency review. 

See: https://github.com/FinOps-Open-Cost-and-Usage-Spec/FOCUS_Spec/issues/2657#issuecomment-5595376502

I have closed the other issues as they were addressed. Thanks!

specification/supported_features/rate_optimization_and_contract_evaluation.md
jpradocueva: @Matt-Cowsert, same here. I have logged this issue on AI #2657 and am closing it as resolved. Thanks!

jpradocueva (approved):
The content on this PR has passed the $focus-pr-validation/SKILL.md. Two issues were logged in AI #2657 to address during the consistency review

specification/supported_features/catalog_discovery_and_price_estimation.md
```diff
@@ -0,0 +1,211 @@
...
+  PricingUnit,
+  PricingCurrency,
+  PricingCurrencyCategory,
+  ListUnitPrice
+FROM SkuPrice
+WHERE ServiceProviderName = ?
+  AND PricingServiceName = ?
+  AND ChargeCategory = 'Usage'
+  AND (SkuPriceEffectiveStart IS NULL OR SkuPriceEffectiveStart <= ?)
+  AND (SkuPriceEffectiveEnd IS NULL OR SkuPriceEffectiveEnd > ?)
```
flanakin, line 84: 🤖 [AI][Claude] ⚠️ **Should fix**

These queries treat the effective date pair as the only thing that can multiply rows for one `SkuPriceId`, but `ContractId` does the same and is not filtered anywhere in this file.

`specification/datasets/sku_price/dataset.md` line 79 states the uniqueness rule:

> `SkuPrice MUST NOT contain more than one record for a given ServiceProviderName, SkuPriceId, ContractId, SkuPriceEffectiveStart, and PricingCurrency`

and `sku_price/columns/contractid.md` says "A null Contract ID indicates a public list price that is not tied to a specific contract; a non-null Contract ID associates the price with a specific contract."

So a dataset instance carrying both public and contracted prices — which is the normal case for a customer pulling their own catalogue — returns one row per contract for the same `SkuPriceId`, at the same point in time. Two consequences:

1. **This query** returns duplicate rows per SKU, mixing list rates with contract-specific rates, while the section is titled "Find the **List** Prices in Force at a Point in Time" and the Description defines List Unit Price as "the public rate".
2. **"Estimate the Cost of a Planned Workload"** is worse, because the `LEFT JOIN` on lines 109 to 112 has no aggregation. One planned quantity fans out into one `EstimatedListAmount` row per matching contract, so summing the result over-counts the estimate by the number of contracts, silently.

The file's own Note on line 27 warns about exactly this shape of problem for the time dimension — "Filtering to a point in time rather than assuming one row per SKU Price ID is what makes a query portable" — but point-in-time filtering alone is not sufficient. `rate_optimization_and_contract_evaluation.md` in this same pull request treats `ContractId` as load-bearing throughout, including `CURRENT_TIER.ContractId IS NULL AND NEXT_TIER.ContractId IS NULL`, so the two files currently disagree about whether it matters.

`ContractId` is also absent from this file's Directly Dependent Columns and Supporting Columns lists, so it would need adding to Supporting Columns alongside the query change.

For this query, restricting to public list prices:

```suggestion
  AND ContractId IS NULL
  AND (SkuPriceEffectiveStart IS NULL OR SkuPriceEffectiveStart <= ?)
  AND (SkuPriceEffectiveEnd IS NULL OR SkuPriceEffectiveEnd > ?)
```

The estimate query on lines 109 to 112 needs the same predicate added to its join condition, and the two remaining point-in-time queries in this file are exposed to the same fan-out.

specification/supported_features/catalog_discovery_and_price_estimation.md
```diff
@@ -0,0 +1,211 @@
...
+
+> Note: The following examples are informative and non-normative. They do not define requirements.
+
+SKU Price Eligibility is defined in [*JSON object format*](#attributes.jsonobjectformat), and ANSI SQL does not define a standard for parsing JSON. The eligibility query below uses BigQuery Standard SQL JSON functions (e.g., `JSON_VALUE`, `JSON_EXTRACT_ARRAY`, `JSON_VALUE_ARRAY`, `UNNEST`); similar functions exist in all major SQL engines. Every other query below is ANSI SQL and runs without modification.
+
+> Important Consideration: The following queries assume FOCUS-conformant dataset artifacts. Practitioners should verify provider conformance before relying on these queries. Non-conformant dataset artifacts may produce inaccurate results.
+
+### Find the List Prices in Force at a Point in Time
+
+This query takes inputs of a service provider, a pricing service name, and a point in time, then returns the public rates that apply for that service at that moment. The same point in time is supplied to both bounds, and each bound is tested for null so that an open-ended price is returned rather than filtered out.
```
flanakin, line 68: 🤖 [AI][Claude] 💡 **Suggestion**

The query below carries `AND ChargeCategory = 'Usage'`, but this paragraph describes the result as "the public rates that apply for that service at that moment" without mentioning the filter. A reader comparing the description against the row count will find fewer rows than described, and the Description section three screens up makes the omission material — it says Charge Category separates the rate to consume something ("Usage") from the fee to acquire it ("Purchase") and from the unit value of a granted credit ("Credit"), so acquisition fees for the same service are silently excluded here.

The filter itself is the right default for a consumption rate lookup. It just needs saying.

```suggestion
This query takes inputs of a service provider, a pricing service name, and a point in time, then returns the public consumption rates that apply for that service at that moment. It filters to a Charge Category of "Usage", so acquisition fees and granted credits are excluded; dropping that predicate returns the full set of rates for the service. The same point in time is supplied to both bounds, and each bound is tested for null so that an open-ended price is returned rather than filtered out.
```

flanakin (approved):
🤖 [AI][Claude] **PR Review**

**Summary:** Approving. Two features that read well and explain their own edge cases unusually carefully. One correctness finding on the catalog discovery queries, plus a smaller documentation gap.

### ⚠️ Should fix (1)

1. The catalog discovery queries filter on effective dates but not on `ContractId`, which is also part of the SKU Price uniqueness key.

### 💡 Suggestions (1)

1. The first query's `ChargeCategory = 'Usage'` filter is not mentioned in the prose that describes what it returns.

### ✅ What's done well

* "Reading the Effective Date Columns" is the strongest part of the pull request. Enumerating all four null combinations and then deriving the `(bound IS NULL OR comparison)` pattern from them, rather than just asserting the pattern, is what makes the queries auditable.
* The Note about dataset instances holding history or not, and the conclusion that filtering to a point in time is what makes a query portable, correctly identifies a real portability trap.
* The pre-commitment framing in the Description is an important caveat and it is stated plainly rather than buried — an estimate from list prices is not what a commitment-holding organisation will pay, and pointing at Cost Comparison for that gap is the right handoff.
* `rate_optimization_and_contract_evaluation.md` handles `ContractId` correctly and explicitly, including the `IS NULL AND ... IS NULL` pairing in the tier comparison.

specification/supported_features/catalog_discovery_and_price_estimation.md
```diff
@@ -0,0 +1,211 @@
...
+  AND (SkuPriceEffectiveStart IS NULL OR SkuPriceEffectiveStart <= ?)
+  AND (SkuPriceEffectiveEnd IS NULL OR SkuPriceEffectiveEnd > ?)
+ORDER BY SkuId, ListUnitPrice
+```
+
+### Estimate the Cost of a Planned Workload
+
+This query takes a set of planned quantities, each paired with the SKU Price ID it is priced under and a point in time, and returns the projected cost of each line and the components behind it. The planned quantity is expressed in the Pricing Unit of the matching price, so a rate quoted per `1K Requests` takes a quantity counted in thousands of requests rather than in requests.
+
+Pricing Currency Category is returned alongside the total because a "Consumable" rate produces a balance in a consumption currency rather than a financial amount. Rows carrying different Pricing Currency values, or a mix of "Payable" and "Consumable", are not additive without a conversion step the SKU Price dataset does not carry.
```
ijurica, line 92: The projected cost of each line" suggests one row per planned line, but a line can resolve to one row per pricing currency. A sentence at the end of the existing note could make that explicit

```suggestion
Pricing Currency Category is returned alongside the total because a "Consumable" rate produces a balance in a consumption currency rather than a financial amount. Rows carrying different Pricing Currency values, or a mix of "Payable" and "Consumable", are not additive without a conversion step the SKU Price dataset does not carry. Additionally, a SKU Price ID published in more than one pricing currency returns one row per currency for the same planned line, and those rows are alternative prices for that line rather than parts of it.
```

specification/supported_features/catalog_discovery_and_price_estimation.md
```diff
@@ -0,0 +1,211 @@
...
+  AND PricingServiceName = ?
+  AND ChargeCategory = 'Usage'
+  AND (SkuPriceEffectiveStart IS NULL OR SkuPriceEffectiveStart <= ?)
+  AND (SkuPriceEffectiveEnd IS NULL OR SkuPriceEffectiveEnd > ?)
+ORDER BY SkuId, ListUnitPrice
+```
+
+### Estimate the Cost of a Planned Workload
+
+This query takes a set of planned quantities, each paired with the SKU Price ID it is priced under and a point in time, and returns the projected cost of each line and the components behind it. The planned quantity is expressed in the Pricing Unit of the matching price, so a rate quoted per `1K Requests` takes a quantity counted in thousands of requests rather than in requests.
```
ijurica, line 90: SKU Price ID is an identifier specified by the service provider, and ServiceProviderName is part of the SKU Price uniqueness key. In a dataset instance carrying more than one service provider, the same SKU Price ID value can match records from several providers, and those records need not represent the same thing. Matching on the service provider as well avoids pricing a planned line against another provider's record that happens to share its identifier.

***Note:** A [separate comment](https://github.com/FinOps-Open-Cost-and-Usage-Spec/FOCUS_Spec/pull/2595#discussion_r3981794923) applies the same suggestion to the query.*

```suggestion
This query takes a set of planned quantities, each paired with the service provider and the SKU Price ID it is priced under and a point in time, and returns the projected cost of each line and the components behind it. The planned quantity is expressed in the Pricing Unit of the matching price, so a rate quoted per `1K Requests` takes a quantity counted in thousands of requests rather than in requests.
```

specification/supported_features/catalog_discovery_and_price_estimation.md
```diff
@@ -0,0 +1,211 @@
...
+  SP.ListUnitPrice,
+  SP.PricingCurrency,
+  SP.PricingCurrencyCategory,
+  PU.PlannedQuantity * SP.ListUnitPrice AS EstimatedListAmount
+FROM PlannedUsage PU
+LEFT JOIN SkuPrice SP
+  ON SP.SkuPriceId = PU.SkuPriceId
+  AND (SP.SkuPriceEffectiveStart IS NULL OR SP.SkuPriceEffectiveStart <= PU.PlannedDate)
+  AND (SP.SkuPriceEffectiveEnd IS NULL OR SP.SkuPriceEffectiveEnd > PU.PlannedDate)
+ORDER BY SP.PricingCurrencyCategory, SP.PricingCurrency, EstimatedListAmount DESC
```
ijurica, line 112: - **ServiceProviderName:** This applies the change described in the [comment on line 90](https://github.com/FinOps-Open-Cost-and-Usage-Spec/FOCUS_Spec/pull/2595#discussion_r3980934437): ServiceProviderName is added to `PlannedUsage`, to the join, and to the SELECT.
- **ContractId condition:** The suggestion also adds `SP.ContractId IS NULL` to the join. Without it, a SKU Price ID carrying both a public record and contract records returns one row per record for the same planned line, and since Contract ID is not selected, those rows cannot be told apart. As the query estimates from List Unit Price, the public record is sufficient. A SKU Price ID that exists only under a contract (for example, a custom tier configuration not in the public price list) returns nulls for that line, which keeps it visible rather than dropping it.
- **Alias:** The alias is renamed from `EstimatedListAmount` to `EstimatedPricingCurrencyListCost`. The value is denominated in the Pricing Currency, and the name follows the `PricingCurrency*` naming used in Cost and Usage, where List Cost itself is denominated in the Billing Currency.

```suggestion
WITH PlannedUsage (ServiceProviderName, SkuPriceId, PlannedQuantity, PlannedDate) AS (
  VALUES (?, ?, ?, ?)
)
SELECT
  PU.ServiceProviderName,
  PU.SkuPriceId,
  SP.SkuPriceDescription,
  SP.PricingUnit,
  PU.PlannedQuantity,
  SP.ListUnitPrice,
  SP.PricingCurrency,
  SP.PricingCurrencyCategory,
  PU.PlannedQuantity * SP.ListUnitPrice AS EstimatedPricingCurrencyListCost
FROM PlannedUsage PU
LEFT JOIN SkuPrice SP
  ON SP.ServiceProviderName = PU.ServiceProviderName
  AND SP.SkuPriceId = PU.SkuPriceId
  AND SP.ContractId IS NULL
  AND (SP.SkuPriceEffectiveStart IS NULL OR SP.SkuPriceEffectiveStart <= PU.PlannedDate)
  AND (SP.SkuPriceEffectiveEnd IS NULL OR SP.SkuPriceEffectiveEnd > PU.PlannedDate)
ORDER BY SP.PricingCurrencyCategory, SP.PricingCurrency, EstimatedPricingCurrencyListCost DESC
```

Matt-Cowsert:
## TF1 Action Item - September 15, 2026

Pull request #2424 collapses `ListUnitPrice` and `ContractedUnitPrice` into a single unit price column, which changes every supported feature, query, and use case that reads both prices off one row. Irena Jurica raised that cascade on September 15, 2026 and asked to hold her remaining supported-features feedback until the reshape lands. Matt Cowsert said he would carry the same shape into the supported features before handing them to Shawn Alpay.

Marc Perreaut asked how a practitioner compares a list price and a contracted price when the two carry different effective dates, and different tiers. A worked query answers that, rather than a schema change.

**Scope:**
- [ ] Update the SKU Price supported features for the single unit price column
- [ ] Rewrite the queries that read a list price and a contracted price from one row so they self-join the SKU Price dataset
- [ ] Provide the query that compares a list price and a contracted price whose effective date ranges differ

**Assignee:** @Matt-Cowsert

PR #2596, opened 2026-08-03
FR #2335-B-Revised: Propose alternative feature-level rubric structure (based on Option B)

ijurica:
## Summary

This PR contains an alternative proposal for the feature-level rubric. Rather than incrementally editing the existing draft, it reorganizes the guideline with the goal of making the decision process simpler and easier to follow.

I realize this is a significant restructuring of the current draft, and we may ultimately decide to keep the existing approach. My intention is simply to explore whether a different structure makes the guidance easier to understand and apply.

The proposal aims to:

* simplify the overall decision flow
* separate the feature-level decision from nullability more clearly
* make the decision criteria more explicit and easier to apply consistently
* reduce cross-references and improve readability

@Matt-Cowsert, at this stage, I would especially appreciate your feedback on whether this overall structure is worth pursuing before we spend time discussing the detailed wording. Of course, feedback from everyone is very welcome.

### Type of Change
* [ ] Bug fix
* [ ] Editorial / Formatting

### Author Checklist


guidelines/contributors/feature-level-rubric-guidelines.md
```diff
@@ -0,0 +1,289 @@
...
+
+    classDef level fill:#DDEBF7,stroke:#2E5C8A,color:#10243B
+    classDef nulls fill:#E2EFDA,stroke:#548235,color:#1B3312
+    classDef note fill:#FFF2CC,stroke:#BF8F00,color:#3B2E00
+    classDef out fill:#E4DFEC,stroke:#6B5B95,color:#241C33
+    class M,C level
+    class NF,NT nulls
+    class FL note
+    class S5 out
+```
```
Matt-Cowsert, line 133:
````suggestion
```mermaid
flowchart TD
    A["A column the working group<br/>has decided to add"] --> S1["Step 1<br/>State the concept"]
    S1 --> F{"Step 2 floor<br/>Is the column a Directly Dependent<br/>Column of a Supported Feature?"}
    F -->|"Yes"| FL["Recommended and Optional<br/>are not available"]
    F -->|"No"| S2{"Step 2 Applicability<br/>Does the concept exist in every<br/>operating model, and can a value be<br/>produced without substituting<br/>something else for it?"}
    FL --> S2
    S2 -->|"No to either"| C["Conditional"]
    S2 -->|"Yes to both"| VG{"Does a Condition gate on variance<br/>rather than on absence?"}
    VG -->|"Yes, and another column in the<br/>dataset determines the value"| C
    VG -->|"No, or the value is<br/>not recoverable"| M["Mandatory"]
    C --> S3["Step 3<br/>Identify or propose the<br/>operating model Condition"]
    M --> S4{"Step 4 Nullability<br/>Is a meaningful value available<br/>on every row where present?"}
    S3 --> S4
    S4 -->|"Yes"| NF["Allows nulls = False"]
    S4 -->|"No"| NT["Allows nulls = True"]
    NT --> D{"Does the null rule name a<br/>characteristic of the<br/>operating model?"}
    D -->|"Yes: it is a presence rule"| S2
    D -->|"No"| S5["Step 5<br/>Record the presence requirement,<br/>Content Constraints, and rationale"]
    NF --> S5

    classDef level fill:#DDEBF7,stroke:#2E5C8A,color:#10243B
    classDef nulls fill:#E2EFDA,stroke:#548235,color:#1B3312
    classDef note fill:#FFF2CC,stroke:#BF8F00,color:#3B2E00
    classDef out fill:#E4DFEC,stroke:#6B5B95,color:#241C33
    class M,C level
    class NF,NT nulls
    class FL note
    class S5 out
```

> **Note:** The diagram shows the decision spine. [Applicability Signals](#applicability-signals) and [Tie-Breakers and Defaults](#tie-breakers-and-defaults) govern the cases where a step is not obvious, including the derivation rule in Principle 5 and the `Conditional` default at the boundary.
````

Two edits to the diagram, both where a reader working from it reaches a different level than one working from the prose.

Step 2's node asked only whether the concept exists. Lines 148 to 151 require two properties to both hold for `Mandatory`, the second being that a value can be produced without substituting something else for it, so the node now carries both.

The variance-gate branch is new. The Applicability Signals row at line 242 makes a variance-gated column `Conditional` where its value is recoverable from another column, while the diagram routed every `Yes` straight to `Mandatory`. `PricingCurrency` is the case that separates them: universal concept, recoverable from `BillingCurrency`, so `Conditional` in the prose and `Mandatory` in the diagram as drawn. Round 10 saw raters split on that column for this reason.

The derivation rule and the boundary default stay out of the diagram and move into the note instead, since branching them makes it unreadable, and both are tie-breakers rather than spine.

The layout and the colors are unchanged from yours.

File: guidelines/contributors/feature-level-rubric-guidelines.md [hunk 2]
```diff
@@ -0,0 +1,289 @@
...
+* [Worked Example](#worked-example)
+* [Deferred Topics](#deferred-topics)
+
+## Scope
+
+This revision covers the `Mandatory` and `Conditional` feature levels, and the nullability of columns holding them. It starts once the working group has decided that a column belongs in a dataset.
+
+Two constraints follow:
+
+* **It does not change the meaning of any feature level.** Each level, and the `MUST`, `SHOULD`, or `MAY` wording that expresses it, is defined in [FOCUS Feature Level](#introduction.focusfeaturelevel). This guideline changes how a level is chosen, not what it means.
```
Matt-Cowsert:
```suggestion
* **It does not change the meaning of any feature level.** Each level, and the `MUST`, `SHOULD`, or `MAY` wording that expresses it, is defined in [FOCUS Feature Level](../../specification/overview.md#focus-feature-level). This guideline changes how a level is chosen, not what it means.
```

`#introduction.focusfeaturelevel` does not resolve. The string appears nowhere else in the repository, and the section it points at is `## FOCUS Feature Level` in `specification/overview.md`. Spec-internal anchors work inside the built specification, and this guideline is not part of the build, so a relative path is what resolves for a reader on GitHub.

guidelines/contributors/feature-level-rubric-guidelines.md
```diff
@@ -0,0 +1,289 @@
...
+* **It does not reopen levels the specification already ships.** Adopting it does not oblige a review of published columns.
```
Matt-Cowsert:
```suggestion
* **It applies to decisions made under it, and records the backlog it leaves.** Adopting it does not change the level of any published column. A published column's level is reopened when the working group schedules the backlog, or when the column is revisited for another reason. Columns that predate this guideline and do not meet its criteria are recorded rather than grandfathered.
```

"Does not reopen levels the specification already ships" reads as a permanent carve-out rather than a sequencing statement, and the guideline needs the second reading to be adoptable. Four published columns are Directly Dependent Columns of a Supported Feature at `Recommended` today: `AllocatedMethodDetails`, `AvailabilityZone`, `ChargeFrequency`, and `ServiceSubcategory`. Under the Step 2 floor each of those is either a level that has to move or a dependency list that has to change, so the guideline either says when that happens or it ships a rule that four published columns do not meet.

guidelines/contributors/feature-level-rubric-guidelines.md
```diff
@@ -0,0 +1,289 @@
...
+
+Topics left to a later revision or to the companion guideline are listed in [Deferred Topics](#deferred-topics).
+
+## Terms Used
+
+| Term | Meaning |
+| :--- | :--- |
+| *Operating model* | How a *data generator*'s business works, described through the billing concepts it uses: regions, commitment discounts, unit pricing, and so on. It determines whether a FOCUS concept applies to a *data generator*, and it is not the same thing as the category of company the *data generator* is. |
+| *Operating model* Condition | A verifiable state of an *operating model*, defined in the [Conditions](#conditions) section. Every `Conditional` column references one or more of them through its presence requirement. |
```
Matt-Cowsert:
```suggestion
| *Operating model* Condition | A verifiable state of an *operating model*, defined in the [Conditions](../../specification/conditions/conditions_overview.md) section. Every `Conditional` column references one or more of them through its presence requirement. |
```

`#conditions` has no target. There is no Conditions heading in this file, and the guideline is not part of the built specification, so the anchor resolves to nothing on GitHub. The relative path reaches the Conditions overview directly.

guidelines/contributors/feature-level-rubric-guidelines.md
```diff
@@ -0,0 +1,289 @@
...
+* **The default is `Conditional`.** Where Step 2 does not clearly return `Mandatory`, the column is `Conditional`.
+
+* **Anything at the boundary is `Conditional`** until the companion guideline calibrates the boundary against real *data generator* data.
+
+* **Calling a model an exception requires an explicit decision.** A single unusual *operating model* that lacks the concept may be treated as an exception only when the working group records:
+  * which *operating model* is being treated as exceptional,
+  * why the concept is still considered universal,
+  * why the exception does not represent a broader pattern.
+
+  Until the companion guideline defines where this record belongs, it lives with the leveling decision itself, in the pull request or issue.
```
Matt-Cowsert:
```suggestion
* **Calling a model an exception requires an explicit decision.** A single unusual *operating model* that lacks the concept may be treated as an exception, keeping the column `Mandatory`, only when the working group records:
  * which *operating model* is being treated as exceptional,
  * why the concept is still considered universal,
  * why the exception does not represent a broader pattern.

  A recorded exception takes precedence over the boundary default above. Until the companion guideline defines where this record belongs, it lives with the leveling decision itself, in the pull request or issue.
```

As written the bullet lists three things to record and never says what recording them produces, so a reader reaches the end without a level. It also sits below "Anything at the boundary is `Conditional`", which means a reader working down the list has already landed on `Conditional` before arriving. Round 10 had a rater report the earlier version of this path as unreachable for that ordering reason.

Naming the outcome and giving the exception precedence fixes both without reordering the section.

guidelines/contributors/feature-level-rubric-guidelines.md
```diff
@@ -0,0 +1,289 @@
...
+| Whether a proposed column belongs in the schema | Not part of this revision. Two tests would answer it: whether the data is needed rather than merely useful, and whether a column calculable from other columns earns a place of its own. |
+| What a Supported Feature does about a column that can be absent | Belongs to the Supported Features work. The obligation is prospective: adopting this guideline does not reopen levels of existing columns. What a feature does once a column it depends on is `Conditional`, and therefore absent from some *FOCUS datasets*, is not decided here. Existing Conditional columns that are already dependencies of Supported Features are handled through the Supported Features work. |
+| Applying these criteria to published columns | A separate decision. Conditions that gate on variance rather than on absence should be re-tested against the recoverability rule when that review happens, since such a gate is admissible only where the value is recoverable from another column in the same dataset. |
```
Matt-Cowsert:
```suggestion
| Applying these criteria to published columns | Scheduled, not settled here. A published column enters the backlog when it is a Directly Dependent Column of a Supported Feature and its presence requirement is a `SHOULD` or a `MAY`. Conditions that gate on variance rather than on absence are re-tested against the recoverability rule in the same pass, since such a gate is admissible only where the value is recoverable from another column in the same dataset. The working group sets the release that takes the backlog. |
```

This keeps your variance-gate sentence and gives the row an entry criterion, so a reader can tell whether a given published column is in the backlog without the guideline listing columns that will go stale. The enumerated set belongs in the tracking issue rather than here.

"Are re-tested" rather than "should be re-tested" keeps a lowercase normative keyword out of the table.

Matt-Cowsert (approved):
Approving so this can go to Task Force 1 tomorrow. Step 5 is the part that earns it: it gives an author the presence requirement and the Content Constraints values to write, where the earlier draft only described them.

One suggestion is worth taking before the call. The diagram answers `PricingCurrency` as `Mandatory` while the Applicability Signals table answers `Conditional`, because the diagram carries no variance-gate branch and asks only the first half of Step 2's test. Round 10 recorded the same class of divergence on the earlier diagram, where `PricingCurrency`, `ListCost`, and `BillingAccountType` each read differently from the diagram than from the prose. A diagram that disagrees with the text tends to become the topic of the call.

Two things I would put in front of the group rather than settle here:

* **When the backlog gets addressed.** The floor's practical effect lands outside this procedure, since Step 2 only ever returns `Mandatory` or `Conditional` and the floor cannot change either. Four published columns are Directly Dependent Columns at `Recommended`: `AllocatedMethodDetails`, `AvailabilityZone`, `ChargeFrequency`, and `ServiceSubcategory`. Two are one-word changes, since their presence requirements already name a Condition and are written as `SHOULD`. `ChargeFrequency` has an open issue against the August 13 threshold already. The suggestions on Scope and on Deferred Topics give the group somewhere to put the timing answer.
* **Whether the Supported Features write-back travels with the floor.** The floor moves dependent columns up into `Conditional`, which grows the set of features that need a rule for the column being absent. Adopting the floor by itself leaves that where it is.

Mechanical, and not worth holding anything for: the five `### Step N` headings are not title case (`Step 1: State the Concept`, and so on through Step 5). The build does not lint `guidelines/`, so this is house style rather than a gate.

Matt-Cowsert:
## TF1 Action Item - August 4, 2026

The branch name starts with `origin/`, copied in by mistake when the branch was created. It collides with the remote tracking name, and Irena could not work locally until she traced the cause. Renaming the branch closes this PR and opens a replacement, so Matt took the mechanics, including the alternative of committing the content onto the existing Option B branch and letting that become the current Option B. The title and description also read as an exploratory alternative rather than the Option B revision the group is now working from. Matt committed to revising both after the meeting and to carrying over any comments left in the meantime.

**Scope:**
- [ ] Remove the `origin/` prefix from the branch name, either by renaming the branch or by committing the content onto the existing Option B branch
- [ ] Revise the PR title so it identifies the content as the current Option B rather than an alternative proposal
- [ ] Revise the PR description to match, and carry over any comments left before the cleanup

**Assignee:** @Matt-Cowsert

Matt-Cowsert:
Replaced by #2604, which carries this content unchanged on `2335-feature-level-rubric`. Closing this one.

The diff there is identical to the head here, including the six suggestions applied during review, so none of the work is lost. The new branch has no `origin/` prefix, so it clones and checks out normally.

Picking up the rest of the action item: the title and description on #2604 present the content as the current Option B rather than as an alternative proposal, and the two open questions from my review here are carried into its description.

PR #2599, opened 2026-08-03
[Feedback] #2444: Apply review feedback - new columns and duration format

flanakin:
## Summary

Applies pending review feedback from #2444 that is mechanical or narrowly scoped, so the remaining threads there can stay focused on open design questions. Targets `recommendation-dataset-fr975`.

**New columns:**

| Column | Feature Level | Rationale |
|---|---|---|
| `CommitmentDiscountCategory` | Conditional | [Requested in review](https://github.com/FinOps-Open-Cost-and-Usage-Spec/FOCUS_Spec/pull/2444#discussion_r3470743965) — a programmatically referenceable classification, since `ContractCommitmentType` values differ by service provider. Reuses the Cost and Usage column ID and its `Spend`/`Usage` values so proposed commitments compare directly against existing ones. |
| `RecommendationDetails` | Mandatory, nullable | [Requested in review](https://github.com/FinOps-Open-Cost-and-Usage-Spec/FOCUS_Spec/pull/2444#discussion_r3471002098). Carries detail that is not resource configuration (e.g. pricing properties of a proposed SKU). Keys must begin with `x_` unless FOCUS-defined, following the `SkuPriceDetails` pattern, so they do not collide with FOCUS-defined keys proposed in a follow-on PR. |
| `EvaluationPeriodStart` / `EvaluationPeriodEnd` | Optional, nullable | [Requested in review](https://github.com/FinOps-Open-Cost-and-Usage-Spec/FOCUS_Spec/pull/2444#discussion_r3470996332) as `LookbackPeriod*`. The period a recommendation was derived from, which lets a practitioner judge confidence and compare recommendations across generators. |

**Also:** `ContractCommitmentDurationType` now requires the largest unit of time that expresses the duration as a whole number, so one term has one encoding (matching #2600 for the Contract Commitment dataset). The compound format bullet is split into single-state requirements, per [#2444 review](https://github.com/FinOps-Open-Cost-and-Usage-Spec/FOCUS_Spec/pull/2444#discussion_r3651678953).

Period columns follow the `ChargePeriodStart`/`End` inclusive-start / exclusive-end convention. Cross-column requirements sit on the column whose conformance would fail, per the [normative requirements guidelines](/guidelines/contributors/normative-requirements-guidelines.md).

### Open questions

* `EvaluationPeriod*` is Optional, but an evaluation period arguably applies to every recommendation — an argument for Mandatory and nullable. Tradeoffs in `supporting_content/`.
[...]

PR #2444, opened 2026-06-05
AI #2381: Add `Recommendation` dataset

qquigley:
## Summary

This PR proposes a Recommendation dataset for FOCUS, toward [#975](https://github.com/FinOps-Open-Cost-and-Usage-Spec/FOCUS_Spec/issues/975).

The Recommendation dataset is a Conditional, Reference-type supporting dataset describing cost optimization recommendations (ie:, rightsizing, commitment purchases, idle cleanup, license reclamation) generated for a practitioner by a service provider or other tooling. It normalizes recommendations across providers and vendors so they can be analyzed together, prioritized by estimated savings, and joined to the Cost and Usage dataset via shared identifiers.

### Type of Change
* [ ] Bug fix
* [x] New feature / Specification update
* [ ] Editorial / Formatting

### Author Checklist
* [x] **AI Usage Guidelines:** I have read the [FOCUS AI Usage Guidelines](/guidelines/contributors/ai-usage-guidelines.md).
* [x] **Self Review Attestation:** I have performed a self-review of my own contribution.
* [ ] **AI-Generated Examples:** If this PR includes examples generated by AI, I have manually verified that the data is mathematically accurate, logically consistent, and compliant with the FOCUS schema.

PR #2600, opened 2026-08-03
FR: Require canonical largest-whole-unit form for ContractCommitmentDurationType

flanakin:
## Summary

`ContractCommitmentDurationType` currently allows multiple conformant spellings of the same term — `"1 Year"` and `"12 Months"` are both valid for a one-year commitment, as are `"1 Week"` and `"7 Days"`. Both governing rules are `SHOULD`, and neither picks between units, so providers selling identical terms can emit different strings and cross-provider aggregation fragments.

Adds one `MUST`: use the largest unit of time that expresses the duration as a whole number. Non-breaking for generators already emitting the natural form. `"4 Months"` and `"10 Days"` stay valid — no larger unit divides them evenly.

Matches the same requirement added to the Recommendation dataset in #2599.

**Not chosen:** ISO 8601 durations (`P1Y`). Canonical by construction, but a breaking change to a Mandatory 1.4 column, and FOCUS uses ISO 8601 only for instants today. Rationale in `supporting_content/`.

### Open questions

* `"365 Days"` is excluded — a year isn't exactly 365 days, so it isn't a whole-number conversion. The existing standard-duration `SHOULD` covers it.
* `Quarter` is ambiguous: strictly, `"3 Months"` reduces to `"1 Quarter"`. It's absent from `UnitFormat`'s time units and appears only in this column's local list. Either add it to the ladder or drop it from allowed values.
* No automated check backs the new `MUST` — divisibility isn't expressible with current CheckFunctions.

### Verification

`validate_includes.py` passes; requirements model tests 180 passed / 44 skipped; `build_json.py` builds all 4 versions. **`pandoc`, `wkhtmltopdf`, `markdown-pp`, and `pymarkdownlnt` are not installed in this environment**, so `make` and markdown lint were not run — relying on CI.

### Type of Change
* [ ] Bug fix
* [ ] Editorial / Formatting

### Author Checklist

---


Matt-Cowsert:
Discussed in Maintainers on 2026-08-17. May require update to PR #2424 as well to capture PurchaseDurationType as well. @Matt-Cowsert to sync with @flanakin to align.

specification/datasets/contract_commitment/columns/contractcommitmentdurationtype.md
```diff
@@ -14,6 +14,7 @@ ContractCommitmentDurationType MUST adhere to the following requirements:
 * ContractCommitmentDurationType SHOULD be expressed with a quantity and time unit, where quantity is a positive integer, and time-unit is a standardized unit of time, either singular or plural (e.g., "1 Day", "1 Year", "3 Months", "3 Years").
 * ContractCommitmentDurationType SHOULD present the unit of time as one of the allowed values.
 * ContractCommitmentDurationType SHOULD correspond to the standard duration of the purchased offering (e.g., "1 Year", "3 Years") rather than a precise calculation of days or hours.
+* ContractCommitmentDurationType MUST use the largest unit of time that expresses the duration as a whole number (e.g., "1 Year" rather than "12 Months", "1 Week" rather than "7 Days").
```
kk09v, line 17: If this is valuable, is there a reason we shouldn't include this requirement in unit format (instead)?
https://github.com/FinOps-Open-Cost-and-Usage-Spec/FOCUS_Spec/blob/working_draft/specification/attributes/unit_format.md?plain=1#L93

Matt-Cowsert: We discussed this during Maintainers 2026-08-24. Recommend bringing in the UnitFormat attribute adjustment (and quarter/week) and update the similar column in the SKU price Dataset (PurchaseDurationType) so we can have alignment across all datasets.

Matt-Cowsert:
The new requirement conflicts with the example on the line above it. `Quarter` and `Quarters` are allowed time units, so three months is expressible as "1 Quarter". Under "MUST use the largest unit of time that expresses the duration as a whole number", the value "3 Months" in the preceding bullet becomes non-conformant and would have to be written "1 Quarter", by the same reduction that turns "12 Months" into "1 Year".

Worth settling which way it goes: either the reduction stops at Month and Quarter sits outside it, or the example bullet drops "3 Months".

Flagging it here because PR #2424 adds a `PurchaseDurationType` column to the SKU Price dataset with the same three duration-format bullets and the same allowed values, and it would mirror whatever this settles on.

specification/datasets/contract_commitment/columns/contractcommitmentdurationtype.md
```diff
@@ -14,6 +14,7 @@ ContractCommitmentDurationType MUST adhere to the following requirements:
 * ContractCommitmentDurationType SHOULD be expressed with a quantity and time unit, where quantity is a positive integer, and time-unit is a standardized unit of time, either singular or plural (e.g., "1 Day", "1 Year", "3 Months", "3 Years").
```
Matt-Cowsert:
```suggestion
* ContractCommitmentDurationType SHOULD be expressed with a quantity and time unit, where quantity is a positive integer, and time-unit is a standardized unit of time, either singular or plural (e.g., "1 Day", "1 Year", "4 Months", "3 Years").
```

"4 Months" does not reduce under the new MUST three lines below, so it can stay in this example list where "3 Months" cannot. It keeps a plural-month case in the list, and it matches the example the supporting content already uses for a month value that no larger unit divides evenly.

specification/attributes/unit_format.md
```diff
@@ -21,10 +21,11 @@ Column conforming to UnitFormat attribute MUST adhere to the following requireme
...
-* *FOCUS dataset column* expressing a ratio unit and including a denominator quantity SHOULD use the `<plural-units>/<denominator-quantity> <plural-time-units>` format (e.g., `Units/3 Months`).
+* *FOCUS dataset column* expressing a ratio unit and including a denominator quantity SHOULD use the `<plural-units>/<denominator-quantity> <plural-time-units>` format (e.g., `Units/2 Weeks`).
 * *FOCUS dataset column* expressing a ratio unit with a compound unit numerator SHOULD use the `<compound-unit>/<singular-time-unit>` format (e.g., `Core-Hours/Day`).
 * *FOCUS dataset column* expressing a ratio unit with a time denominator SHOULD use the `<plural-units>/<singular-time-unit>` format (e.g., `GB/Hour`, `PB/Day`).
 * *FOCUS dataset column* expressing a [simple unit](#attributes.unitformat.definitions.simpleunit) SHOULD use the `<plural-units>` format (e.g., `GB`, `Seconds`).
+* *FOCUS dataset column* expressing a time-based unit SHOULD use the largest unit of time that expresses the duration as a whole number (e.g., `1 Year` rather than `12 Months`).
```
Matt-Cowsert, line 28: Is this bullet meant to govern measurement units, or only duration terms?

UnitFormat is the right home for a unit-formatting rule. The open piece is reach: as written the bullet governs both, and the two follow different conventions. The bullet directly below recommends `<unit-quantity> <plural-units>` for any value carrying a unit quantity (`1000 Tokens`), and this bullet's example carries a unit quantity of 1 with a singular unit. Both are SHOULDs, so nothing breaks, but they point opposite ways on the same value.

If you mean a general measurement-unit rule, the example needs a form the plural bullet also accepts, and "duration" needs a referent for a value like `PricingUnit` = `Hours`, where there is nothing to reduce. If you mean duration terms, the duration columns are where the rule applies, and `ContractCommitmentDurationType` already carries it as a MUST. Worth noting that column declares StringHandling conformance and not UnitFormat, so this bullet does not reach it today.

specification/attributes/unit_format.md
```diff
@@ -99,7 +100,9 @@ The table below lists allowed time-based base units.
 | Time-based Unit (Singular) | Time-based Unit (Plural) |
 |:---------------------------|:-------------------------|
 | Year                       | Years                    |
+| Quarter                    | Quarters                 |
```
Matt-Cowsert, line 103: Adding `Quarter` and `Week` here is what I recommended on 2026-08-24, and it does settle the placement question. One consequence worth surfacing before it lands: PR #2424 gives `PurchaseDurationType` its own allowed-values table plus a note reading "These time units express the term of a purchase and are not measurement units. They intentionally differ from the time-based units in Unit Format, which govern how the unit of measure for a quantity is expressed. Purchase Duration Type includes 'Week' and 'Quarter' because they are common purchase terms." The two lists still differ after this change, since UnitFormat carries Second, Millisecond and Microsecond and `PurchaseDurationType` does not. What stops holding is the reason that note gives: Week and Quarter are no longer the units that distinguish the two.

Both PRs are open against `working_draft` and neither closes the question alone: are Week and Quarter measurement units, purchase terms, or both? That belongs at the maintainers meeting on Monday rather than in either PR.

supporting_content/datasets/contract_commitment/columns/contractcommitmentdurationtype.md
```diff
@@ -0,0 +1,69 @@
...
+* It reuses vocabulary already in the column, so no new concepts are introduced to the specification.
+* It resolves the ambiguity at the point where it originates, which is the choice of unit.
+
+Scope decision: reduction applies only across exact conversions between adjacent units (60 minutes to an hour, 24 hours to a day, 7 days to a week, 12 months to a year, 4 quarters to a year). Inexact relationships are deliberately excluded, which is why "365 Days" is not treated as reducible to "1 Year". The existing recommendation that the value reflect the standard duration of the purchased offering already covers that case.
+
+Open question for reviewers: `Quarter` is not treated as a reduction target, so "3 Months" remains valid and is not required to become "1 Quarter". This preserves the "3 Months" example already present in the column and reflects that `Quarter` is absent from the `UnitFormat` allowed time-based units, appearing only in this column's local list. Reviewers may prefer to either add `Quarter` to the reduction ladder or remove it from the allowed values.
```
Matt-Cowsert, line 35:
```suggestion
Scope decision: reduction applies only across exact conversions (60 minutes to an hour, 24 hours to a day, 7 days to a week, 3 months to a quarter, 4 quarters to a year, 12 months to a year). Inexact relationships are deliberately excluded, which is why "365 Days" is not treated as reducible to "1 Year". The existing recommendation that the value reflect the standard duration of the purchased offering already covers that case.

Open question for reviewers: `Quarter` is a reduction target, because `Quarter` and `Quarters` are in the column's allowed values. "3 Months" is therefore required to become "1 Quarter", and "6 Months" to become "2 Quarters". This PR also adds `Quarter` and `Week` to the `UnitFormat` allowed time-based units. Reviewers may prefer to leave `Quarter` out of the ladder, which would keep the month forms and need a stated exception to the largest-whole-unit rule.
```

The head commit adds `Quarter` and `Week` to the `UnitFormat` time-based unit table, so "`Quarter` is absent from the `UnitFormat` allowed time-based units" is no longer true. The reduction consequence never depended on that table, though: `Quarter` and `Quarters` sit in this column's own allowed values, so "3 Months" was already reducible before the head commit. The conversion list above also skipped months-to-quarters while listing quarters-to-years.

supporting_content/datasets/contract_commitment/columns/contractcommitmentdurationtype.md
```diff
@@ -0,0 +1,69 @@
...
+
+Require the largest allowed unit that expresses the purchased term as a whole number.
+
+* "1 Year" is required rather than "12 Months".
+* "1 Week" is required rather than "7 Days".
+* "4 Months" and "10 Days" are unchanged, because no larger allowed unit divides them evenly.
+
+Selected because:
+
+* It is additive. A generator already emitting the natural form of a term is unaffected, so the change is non-breaking in practice.
```
Matt-Cowsert, line 28: Two of the spec's own examples break under this MUST, so the non-breaking claim needs a qualifier. `specification/data/contract_commitments/contract_commitment_scenario_2.csv` line 2 carries `3 Months`, and `specification/appendix/contract_commitments/contract_commitments.md` line 155 carries the same value. Both stop conforming when this MUST lands on a column that has been Mandatory since 1.4.

The qualifier is doing most of the work: a generator already emitting the natural form is unaffected, but that holds for any new constraint, and it does not tell a reviewer how many generators emit "12 Months" or "3 Months" today. Worth stating the classification plainly either way, since breaking-change classification on a Mandatory column is a separate call.

specification/requirements_model/releases/1.5/model_rules/datasets/contract_commitment/columns/contractcommitmentdurationtype.json
```diff
@@ -314,5 +319,28 @@
...
       ]
     }
+  },
+  "CCT-ContractCommitmentDurationType-C-008-M": {
+    "Function": "Validation",
+    "Reference": "ContractCommitmentDurationType",
+    "EntityType": "Column",
+    "EntityName": "Contract Commitment Duration Type",
+    "EntityId": "ContractCommitmentDurationType",
+    "Notes": "Typed Dynamic because confirming the value uses the largest whole unit requires the purchased term from the commercial offering, which is not derivable from the dataset alone.",
```
Matt-Cowsert, line 329:
```suggestion
    "Notes": "Typed Dynamic because the available CheckFunctions provide no divisibility primitive, so reducibility to the next larger unit is not expressible as a static rule. The check itself is derivable from the value alone.",
```

The current reason does not hold. "12 Months" and "3 Months" can be judged from the value by itself, without the purchased term from the commercial offering, because the test is whether the quantity divides evenly into a larger allowed unit. What actually blocks a static rule is the second reason the supporting content gives, the missing divisibility primitive.

supporting_content/datasets/contract_commitment/columns/contractcommitmentdurationtype.md
```diff
@@ -0,0 +1,69 @@
...
+* It does not resolve the ambiguity either. A quantity of 12 with a unit of "Months" and a quantity of 1 with a unit of "Year" would remain equally valid, so a canonical-form rule would still be needed.
+
+## Requirements model
+
+The corresponding rule is `CCT-ContractCommitmentDurationType-C-008-M`.
+
+It is typed `Dynamic` and carries an empty `Requirement`, matching the adjacent `C-006-O` and `C-007-O` rules. Two reasons:
+
+* Confirming that a value uses the largest whole unit requires the purchased term from the commercial offering, which is not derivable from the dataset alone.
+* The underlying check is a divisibility test against the next larger allowed unit. The available `CheckFunctions` provide no divisibility primitive, and expressing divisibility by 7, 12, 24, or 60 as a regular expression over unbounded integers produces patterns tens of thousands of characters long, which are not reviewable.
```
Matt-Cowsert, line 67:
```suggestion
It is typed `Dynamic` and carries an empty `Requirement`, matching the adjacent `C-006-O` and `C-007-O` rules. The underlying check is a divisibility test against the next larger allowed unit, and the available `CheckFunctions` provide no divisibility primitive. Expressing divisibility by 3, 4, 7, 12, 24, or 60 as a regular expression over unbounded integers produces patterns tens of thousands of characters long, which are not reviewable.
```

The first reason does not hold. "12 Months" and "3 Months" can be judged from the value by itself, since the test is whether the quantity divides evenly into a larger allowed unit, so the purchased term from the commercial offering is not needed. Dropping it leaves the missing divisibility primitive as the single reason, which is why the "Two reasons" anchor goes with it. The divisor list also needs 3 and 4, for months-to-quarters and quarters-to-years.

Matt-Cowsert (commented):
**The largest-whole-unit MUST added here makes two of the spec's own contract commitment examples non-conformant.** `specification/data/contract_commitments/contract_commitment_scenario_2.csv` line 2 carries `ContractCommitmentDurationType` = `3 Months`, and `specification/appendix/contract_commitments/contract_commitments.md` line 155 carries the same value in the CC Duration Type row. `Quarter` and `Quarters` are in the column's Allowed Values, so three months reduces to one quarter and both need to read `1 Quarter`. That scenario is a February-to-May reservation described as "H100 GPU Reservation - Q1", so the quarter form reads naturally there.

Three definitions in `unit_format.md` also still carry `Units/3 Months` after line 24 moved to `Units/2 Weeks`: Measurement Unit (line 37), Ratio Unit (line 53), and Denominator Quantity (line 61). Those are illustrations under a SHOULD rather than dataset values, so they are lower stakes than the two above, but the PR changed one instance of the example and left three. Line 61 needs both halves changed, to "(e.g., `2` in `Units/2 Weeks`)", since the bare `3` is the value being illustrated.

The Week and Quarter question deserves its own slot at the maintainers meeting on Monday. PR #2424 takes the opposite position on the same two units, and PR #2444 carries a third copy of this MUST on the Recommendation dataset's own `ContractCommitmentDurationType`, scoped with a "When the unit is a unit of time" clause this copy does not have. No one PR closes the question alone.

Happy to take another pass ahead of that.

specification/attributes/unit_format.md
flanakin: Hmm... I want to say duration only bc dictating Unit or at could result in diff units for the same SKU family. I would not want 100MB vs 1GB vs 1TB for SKUs A, B, and C of the same service. Ideally the unit would be consistent for the SKU family. This does bring my recommendation into question a bit. Not that it's wrong but it's possibly missing something like "for the lowest unit measurement of the SKU family". Of course, that requires us to define what a SKU family is, so we may need to start with SkuId, noting that it's not broad enough today.

Side note: We really need SkuCategory et al.

specification/attributes/unit_format.md
flanakin: Left a comment on the issue, but I do worry quarter might confuse a random sequential 3mos vs calendar quarters vs fiscal quarters. I'd probably avoid to dodge the overlap.

specification/datasets/contract_commitment/columns/contractcommitmentdurationtype.md
flanakin: Good idea, but we should also add a req to keep unit consistent across like SKUs. I can add that here next week...

specification/requirements_model/releases/1.5/model_rules/datasets/contract_commitment/columns/contractcommitmentdurationtype.json
flanakin: If I'm following, I'm good here, but we should simplify and try to avoid jargon, if we can.

supporting_content/datasets/contract_commitment/columns/contractcommitmentdurationtype.md
flanakin: Yes, this says, "anyone following the rule isn't impacted 🙌" but does not acknowledge that anyone nkt following thr rule would need to change their data to abide by the rule, whocb will def impa t qieries people are making. I dont think non-breaking is a valid claim.

Besides, we should avoid "breaking" and confirm that it doesn't change scenario coverage, does simplify scenario enablement, and may require a change to queries if the ubderlying values in older versions were not conformant to the new rule already.

specification/attributes/unit_format.md
kk09v: I disagree that they point opposite directions.

This one states that specifically time-based units should be reduced. Just like how fractions are always reduce to 1/2 from 8/16 or 4/8, we're asking for 1 day instead of 24 hours or 86400 seconds.

The next line states that when you're using a quantity, you should make the unit plural. E.g. the duration of a commitment would be `3 years`, not `3 year`. Note that datepart in SQL is singular e.g. `INTERVAL 3 DAY` so this is specifying that we lean toward human readability in these situations.

Matt-Cowsert:
approving to bring to TF-review, acknowledge we still need to fine tune before merging.

Matt-Cowsert (approved):

Matt-Cowsert:
## TF1 Action Item - September 1, 2026

Task Force 1 (TF1) settled two of the open questions on this pull request (PR) on September 1.

Quarter is dropped. Michael Flanakin's reasoning, which the group accepted: a quarter implies alignment with a calendar or fiscal quarter, and three consecutive months are not one unless they fall on those bounds. Dropping it also settles the "3 Months" versus "1 Quarter" reduction question raised on the PR on August 24.

The unit vocabulary follows UnitFormat. Karl Kraft's thread asked whether this belongs in UnitFormat, and the group agreed the column should reference UnitFormat for its time-unit vocabulary, keeping the unit quantity followed by the unit form (for example "3 Years").

**Scope:**
- [ ] Remove Quarter and Quarters from the ContractCommitmentDurationType allowed values and from the time-based units this PR adds to UnitFormat
- [ ] Reference the UnitFormat time-based units as the unit vocabulary for ContractCommitmentDurationType

**Assignee:** @flanakin

specification/datasets/contract_commitment/columns/contractcommitmentdurationtype.md
```diff
@@ -11,9 +11,10 @@ ContractCommitmentDurationType MUST adhere to the following requirements:
 * ContractCommitmentDurationType MUST be of type String.
 * ContractCommitmentDurationType MUST conform to [StringHandling](#attributes.stringhandling) requirements.
 * ContractCommitmentDurationType MUST NOT be null.
-* ContractCommitmentDurationType SHOULD be expressed with a quantity and time unit, where quantity is a positive integer, and time-unit is a standardized unit of time, either singular or plural (e.g., "1 Day", "1 Year", "3 Months", "3 Years").
+* ContractCommitmentDurationType SHOULD be expressed with a quantity and time unit, where quantity is a positive integer, and time-unit is a standardized unit of time, either singular or plural (e.g., "1 Day", "1 Year", "4 Months", "3 Years").
 * ContractCommitmentDurationType SHOULD present the unit of time as one of the allowed values.
 * ContractCommitmentDurationType SHOULD correspond to the standard duration of the purchased offering (e.g., "1 Year", "3 Years") rather than a precise calculation of days or hours.
+* ContractCommitmentDurationType MUST use the largest unit of time that expresses the duration as a whole number (e.g., "1 Year" rather than "12 Months", "1 Week" rather than "7 Days").
```
Matt-Cowsert, line 17: ## TF1 Action Item - September 1, 2026

Michael Flanakin raised a scenario in Task Force 1 (TF1) on September 1 that this pull request (PR) does not cover yet and said he still needs to incorporate it. Two SKUs in the same service may differ only in term, such as a one-month option and a twelve-month option. Both should be expressed in the same time unit, so the longer one stays "12 Months" rather than switching to "1 Year". The point is comparability: a practitioner evaluating which SKU to use within one service should see the terms measured the same way. The expectation is scoped to related SKUs within a service, not across providers, which Matt Cowsert agreed is workable where a cross-provider list would not be.

Ben Olson argued the same from the other direction: choose an appropriate measure and use it across all like things, which matters more to him than the largest whole unit. Michael Flanakin's counterpoint stands too: the smaller the unit, the larger and harder to work with the number, so reservations measured in seconds are not the goal. How the consistency guidance sits next to the largest-whole-unit MUST, which would otherwise turn "12 Months" into "1 Year", was not resolved on the call.

**Scope:**
- [ ] Add guidance that SKUs within one service that differ only in term keep the same time unit

**Assignee:** @flanakin

specification/datasets/contract_commitment/columns/contractcommitmentdurationtype.md
```diff
@@ -11,9 +11,10 @@ ContractCommitmentDurationType MUST adhere to the following requirements:
 * ContractCommitmentDurationType MUST be of type String.
 * ContractCommitmentDurationType MUST conform to [StringHandling](#attributes.stringhandling) requirements.
 * ContractCommitmentDurationType MUST NOT be null.
-* ContractCommitmentDurationType SHOULD be expressed with a quantity and time unit, where quantity is a positive integer, and time-unit is a standardized unit of time, either singular or plural (e.g., "1 Day", "1 Year", "3 Months", "3 Years").
+* ContractCommitmentDurationType SHOULD be expressed with a quantity and time unit, where quantity is a positive integer, and time-unit is a standardized unit of time, either singular or plural (e.g., "1 Day", "1 Year", "4 Months", "3 Years").
```
Matt-Cowsert, line 14: ## TF1 Action Item - September 8, 2026

This pull request no longer merges into working_draft, and one line accounts for the whole conflict on this file. Both sides rewrote the quantity-and-time-unit bullet: pull request #2549 shortened it to the `"[Numeric Value] [Unit]"` form when it landed on August 27, and this branch keeps the longer wording while changing the example from "3 Months" to "4 Months". The `Operating Model Conditions` row and the Allowed Values note that working_draft has gained since this branch was cut both merge in without a conflict.

That note is worth a second look once the rebase brings it in. Pull request #2549 added it, and it says the column's time units intentionally differ from the time-based units in Unit Format, naming "Week" and "Quarter" as the reason. This pull request adds Week and Quarter to Unit Format's allowed time-based units, so once both sit on one branch the note describes a divergence that is no longer there.

Matt Cowsert committed to working through this with @flanakin after the September 8 Task Force 1 session. The September 1 items on this pull request, dropping Quarter and adding the related-SKU unit consistency case, are still open.

**Scope:**
- [ ] Resolve the conflict on this bullet, choosing between #2549's `"[Numeric Value] [Unit]"` wording and this branch's longer form
- [ ] Reconcile the Allowed Values note from #2549 with this branch's Unit Format additions, which give Unit Format the Week and Quarter the note says it lacks

**Assignee:** @Matt-Cowsert

specification/datasets/contract_commitment/columns/contractcommitmentdurationtype.md
flanakin: 🤖 [AI][Claude Code] ✅ **Implemented**

Task Force 1 (2026-09-02) agreed: `ContractCommitmentDurationType` now references `UnitFormat`'s allowed time-based unit names instead of keeping its own duplicate list. `PurchaseDurationType` alignment is tracked separately in #2424.

specification/datasets/contract_commitment/columns/contractcommitmentdurationtype.md
flanakin: 🤖 [AI][Claude Code] ✅ **Implemented**

Resolved differently than suggested: adopted #2549's short `"[Numeric Value] [Unit]"` wording (already merged to `working_draft`) to resolve the merge conflict between the two PRs, keeping the `"3 Months"` example rather than changing it to `"4 Months"`. Safe now that `Quarter` is being dropped from this column's allowed values (see thread below), so `"3 Months"` no longer reduces to `"1 Quarter"`.

specification/attributes/unit_format.md
flanakin: 🤖 [AI][Claude Code] ✅ **Implemented**

Task Force 1 (2026-09-02) settled this: the column references `UnitFormat`'s time-based unit vocabulary directly rather than duplicating it locally. Michael's SKU-family unit-consistency idea is tracked as a separate first-pass requirement (see the September 1 action item thread) since it needs a "SKU family" concept this dataset doesn't formally have yet.

specification/attributes/unit_format.md
flanakin: 🤖 [AI][Claude Code] ✅ **Implemented**

Task Force 1 (2026-09-02) resolved this: `Quarter`/`Quarters` are dropped from the Quarter/Week addition this PR made to `UnitFormat` (`Week` stays). That also resolves the tension with #2424's note, since `Quarter` is no longer a `UnitFormat` time-based unit either.

supporting_content/datasets/contract_commitment/columns/contractcommitmentdurationtype.md
flanakin: 🤖 [AI][Claude Code] ✅ **Implemented**

Resolved differently than suggested: rather than adding "3 months to a quarter" and "4 quarters to a year" to the conversion ladder, `Quarter`/`Quarters` are dropped from the column entirely per Task Force 1's 2026-09-02 decision, so there's no quarter reduction to define. Updated the scope-decision paragraph and removed the now-resolved open question.

supporting_content/datasets/contract_commitment/columns/contractcommitmentdurationtype.md
flanakin: 🤖 [AI][Claude Code] ✅ **Implemented**

Reworded the "non-breaking" claim per Michael's comment below: it now says the change doesn't affect scenario coverage and simplifies scenario enablement, but generators/consumers not already emitting the natural form may need to update. Separately, dropping `Quarter` means `contract_commitment_scenario_2.csv` and `contract_commitments.md` no longer break at all — `"3 Months"` stays valid since no larger unit divides it evenly.

specification/requirements_model/releases/1.5/model_rules/datasets/contract_commitment/columns/contractcommitmentdurationtype.json
flanakin: 🤖 [AI][Claude Code] ✅ **Implemented**

Agreed the "purchased term" reasoning didn't hold, and Michael asked to simplify and avoid jargon. `Notes` now reads: "Typed Dynamic because no CheckFunction can test whether a quantity divides evenly into a larger unit, so this cannot be enforced as a static rule."

supporting_content/datasets/contract_commitment/columns/contractcommitmentdurationtype.md
flanakin: 🤖 [AI][Claude Code] ✅ **Implemented**

Same simplification as the adjacent RM-json `Notes` field: dropped the "purchased term" reasoning and reworded to plain language, avoiding jargon per Michael's feedback. The supporting_content Requirements model section now gives the divisibility-primitive reason directly, without the "Two reasons" framing.

specification/datasets/contract_commitment/columns/contractcommitmentdurationtype.md
flanakin: 🤖 [AI][Claude Code] 🤔 **Needs discussion**

On reflection this doesn't belong on this column. The scenario ("which SKU to use within one service") is about comparing purchase options — that's `PurchaseDurationType` / SKU Price dataset territory (#2424), which carries `SkuId` to formally scope "same offering" against. `ContractCommitmentDurationType` lives on Contract Commitment, which records commitments already entered into, not a catalog of options, and has no comparable identifier; "offering" also isn't itself a defined FOCUS term. Not adding a requirement here — flagging that this scenario should be tracked against #2424 instead. Noted as an open item in supporting_content.

specification/datasets/contract_commitment/columns/contractcommitmentdurationtype.md
flanakin: 🤖 [AI][Claude Code] ✅ **Implemented**

Rebased this branch onto current `working_draft` (not merged) and resolved the conflict by adopting #2549's short `"[Numeric Value] [Unit]"` wording. The Allowed Values note from #2549 ("Week and Quarter... intentionally differ from... Unit Format") is reconciled by removing the local allowed-values list entirely and referencing `UnitFormat`'s time-based unit names instead, per Task Force 1's 2026-09-02 decision — so there's no more divergence for the note to describe.

flanakin:
🤖 [AI][Claude Code] Addressing @Matt-Cowsert's 2026-08-25 review:

* Dropping `Quarter` (Task Force 1, 2026-09-02) means `contract_commitment_scenario_2.csv` and `contract_commitments.md` no longer break — `"3 Months"` stays valid.
* Fixed the three leftover `Units/3 Months` examples in `unit_format.md` (Measurement Unit, Ratio Unit, Denominator Quantity) to match the `Units/2 Weeks` example already used in the requirement bullet.
* Week/Quarter placement is resolved per the thread on line 103; the `PurchaseDurationType` (#2424) alignment stays tracked there separately.

flanakin:
🤖 [AI][Claude Code] **PR Update Summary**

**Addressed:** 11 threads
- ✅ Implemented: 9
- 🤔 Needs discussion: 1 (SKU/offering term consistency — redirected to #2424, doesn't fit this dataset)
- Also replied to @Matt-Cowsert's 2026-08-25 review directly

**Key changes:**
- Rebased onto current `working_draft`, resolving the merge conflict with #2549 by adopting its short `"[Numeric Value] [Unit]"` wording
- Dropped `Quarter`/`Quarters` (TF1, 2026-09-02); `ContractCommitmentDurationType` now references `UnitFormat`'s time-based unit names instead of a local duplicate list
- Fixed the `non-breaking` claim and three stale `Units/3 Months` examples in `unit_format.md`
- Simplified the requirements-model `Notes` wording per feedback to avoid jargon
- Did **not** add a SKU-consistency requirement to this column — the dataset has no SkuId/OfferingId and "offering" isn't a defined FOCUS term; that scenario belongs on `PurchaseDurationType` (#2424)

specification/requirements_model/releases/1.5/model_rules/datasets/contract_commitment/columns/contractcommitmentdurationtype.json
```diff
@@ -158,12 +163,12 @@
     "DatasetId": "ContractCommitment",
     "DatasetName": "Contract Commitment",
     "ValidationCriteria": {
-      "MustSatisfy": "ContractCommitmentDurationType SHOULD be expressed with a quantity and time unit, where quantity is a positive integer, and time-unit is a standardized unit of time, either singular or plural (e.g., \"1 Day\", \"1 Year\", \"3 Months\", \"3 Years\").",
+      "MustSatisfy": "ContractCommitmentDurationType SHOULD use the \"[Numeric Value] [Unit]\" format (e.g., \"1 Day\", \"1 Year\", \"3 Months\", \"3 Years\").",
```
Matt-Cowsert, line 166: The gen_pdf check fails because of this rule. This line now uses the word "format", and the requirements model tests require any rule whose MustSatisfy says "format" to have `"Function": "Format"` (`tests/test_format_function.py`). This rule's `Function` is still `"Validation"` on line 151, which is outside the diff, so it needs a manual edit rather than a suggestion:

```diff
-    "Function": "Validation",
+    "Function": "Format",
```

With that change, the full requirements model test suite passes.

specification/datasets/contract_commitment/columns/contractcommitmentdurationtype.md
```diff
@@ -12,33 +12,11 @@ ContractCommitmentDurationType MUST adhere to the following requirements:
 * ContractCommitmentDurationType MUST conform to [StringHandling](#attributes.stringhandling) requirements.
 * ContractCommitmentDurationType MUST NOT be null.
 * ContractCommitmentDurationType SHOULD use the "[Numeric Value] [Unit]" format (e.g., "1 Day", "1 Year", "3 Months", "3 Years").
-* ContractCommitmentDurationType SHOULD present the unit of time as one of the allowed values.
+* ContractCommitmentDurationType SHOULD present the unit of time as one of the allowed time-based unit names defined by [UnitFormat](#attributes.unitformat).
```
Matt-Cowsert:
```suggestion
* ContractCommitmentDurationType SHOULD present the unit of time as one of the allowed time-based unit names defined by [UnitFormat](#attributes.unitformat), from Minute through Year.
```

This line now points at UnitFormat's list, which adds Second, Millisecond, and Microsecond to this column's units, but the C-004-O pattern and the C-005-O list in the requirements model still stop at Minute. A value like "30 Seconds" meets this line and fails both checks. If commitment terms should stop at Minute, this suggestion says so and the checks already match. If the intent is every UnitFormat time unit, keep the line and add those three units, singular and plural, to both checks.

specification/requirements_model/releases/1.5/model_rules/datasets/contract_commitment/columns/contractcommitmentdurationtype.json
```diff
@@ -185,7 +190,7 @@
     "DatasetId": "ContractCommitment",
     "DatasetName": "Contract Commitment",
     "ValidationCriteria": {
-      "MustSatisfy": "ContractCommitmentDurationType SHOULD present the unit of time as one of the allowed values.",
+      "MustSatisfy": "ContractCommitmentDurationType SHOULD present the unit of time as one of the allowed time-based unit names defined by UnitFormat.",
```
Matt-Cowsert:
```suggestion
      "MustSatisfy": "ContractCommitmentDurationType SHOULD present the unit of time as one of the allowed time-based unit names defined by UnitFormat, from Minute through Year.",
```

This rule's list of units stops at Minute, so if the column requirement is narrowed to "from Minute through Year", this is the matching MustSatisfy text. If the requirement keeps every UnitFormat time unit, leave this line as it is and add Second, Millisecond, and Microsecond, singular and plural, to the list below and to the C-004-O pattern.

specification/datasets/contract_commitment/columns/contractcommitmentdurationtype.md
```diff
@@ -12,33 +12,11 @@ ContractCommitmentDurationType MUST adhere to the following requirements:
...
 * ContractCommitmentDurationType SHOULD correspond to the standard duration of the purchased offering (e.g., "1 Year", "3 Years") rather than a precise calculation of days or hours.
+* ContractCommitmentDurationType MUST use the largest unit of time that expresses the duration as a whole number (e.g., "1 Year" rather than "12 Months", "1 Week" rather than "7 Days").
```
Matt-Cowsert:
```suggestion
* ContractCommitmentDurationType MUST use the largest time-based unit defined by UnitFormat that expresses the duration as a whole number (e.g., "1 Year" rather than "12 Months", "1 Week" rather than "7 Days").
```

As written, "the largest unit of time" isn't tied to a list, and a quarter is still a unit of time, so "3 Months" can be read as having to become "1 Quarter", the outcome Task Force 1 avoided on September 1 by dropping Quarter. Naming UnitFormat's units keeps "3 Months" valid and makes "1 Quarter" nonconforming. The supporting content's "largest allowed unit" already reads it this way.

specification/requirements_model/releases/1.5/model_rules/datasets/contract_commitment/columns/contractcommitmentdurationtype.json
```diff
@@ -314,5 +309,28 @@
...
+    "ModelVersionIntroduced": "1.5",
+    "Status": "Active",
+    "Conditions": [],
+    "Type": "Dynamic",
+    "Order": 80,
+    "DatasetType": "CCT",
+    "DatasetId": "ContractCommitment",
+    "DatasetName": "Contract Commitment",
+    "ValidationCriteria": {
+      "MustSatisfy": "ContractCommitmentDurationType MUST use the largest unit of time that expresses the duration as a whole number (e.g., \"1 Year\" rather than \"12 Months\", \"1 Week\" rather than \"7 Days\").",
```
Matt-Cowsert:
```suggestion
      "MustSatisfy": "ContractCommitmentDurationType MUST use the largest time-based unit defined by UnitFormat that expresses the duration as a whole number (e.g., \"1 Year\" rather than \"12 Months\", \"1 Week\" rather than \"7 Days\").",
```

If the column's MUST is tied to UnitFormat's units, so that "3 Months" can't be read as "1 Quarter", this is the matching MustSatisfy text. If the MUST keeps its current wording, leave this line as it is.

supporting_content/datasets/contract_commitment/columns/contractcommitmentdurationtype.md
```diff
@@ -0,0 +1,66 @@
...
+Require the largest allowed unit that expresses the purchased term as a whole number.
+
+* "1 Year" is required rather than "12 Months".
+* "1 Week" is required rather than "7 Days".
+* "4 Months" and "10 Days" are unchanged, because no larger allowed unit divides them evenly.
+
+Selected because:
+
+* It does not change scenario coverage and simplifies scenario enablement: a generator already emitting the natural form (the common case in practice) needs no change. A generator emitting a non-natural encoding (e.g., "12 Months") must update to the canonical form, and a consumer filtering or grouping on a value that was not already conformant may need to update its queries. "Non-breaking" is not an accurate description of that impact.
+* It preserves the existing string format, the allowed-value list, the data type, and nullability. No consumer parsing logic changes.
```
Matt-Cowsert:
```suggestion
* It preserves the existing string format, the data type, and nullability. No consumer parsing logic changes.
```

The allowed values do change in this PR: Quarter and Quarters are dropped, and the local list gives way to UnitFormat's time-based unit names.

supporting_content/datasets/contract_commitment/columns/contractcommitmentdurationtype.md
```diff
@@ -0,0 +1,66 @@
...
+* It resolves the ambiguity at the point where it originates, which is the choice of unit.
+
+Scope decision: reduction applies only across exact conversions between adjacent units (60 minutes to an hour, 24 hours to a day, 7 days to a week, 12 months to a year). Inexact relationships are deliberately excluded, which is why "365 Days" is not treated as reducible to "1 Year". The existing recommendation that the value reflect the standard duration of the purchased offering already covers that case.
+
+Resolved by Task Force 1 (2026-09-02): `Quarter`/`Quarters` are dropped entirely, both from this column's allowed values and from the Quarter/Week addition this PR otherwise made to `UnitFormat` (Week stays). A quarter implies alignment with a calendar or fiscal quarter, and three consecutive months do not necessarily fall on those bounds, so treating "3 Months" as reducible to "1 Quarter" would have been incorrect. Dropping `Quarter` also removes the need for a reduction ladder decision on it. The column's unit vocabulary now references UnitFormat's allowed time-based unit names directly instead of duplicating the list locally, so the two lists cannot drift apart again.
```
Matt-Cowsert:
```suggestion
Resolved by Task Force 1 (2026-09-01): `Quarter`/`Quarters` are dropped entirely, both from this column's allowed values and from the Quarter/Week addition this PR otherwise made to `UnitFormat` (Week stays). A quarter implies alignment with a calendar or fiscal quarter, and three consecutive months do not necessarily fall on those bounds, so treating "3 Months" as reducible to "1 Quarter" would have been incorrect. Dropping `Quarter` also removes the need for a reduction ladder decision on it. The column's unit vocabulary now references UnitFormat's allowed time-based unit names directly instead of duplicating the list locally, so the two lists cannot drift apart again.
```

Task Force 1 met on September 1, the date the Open items section below also gives.

supporting_content/datasets/contract_commitment/columns/contractcommitmentdurationtype.md
```diff
@@ -0,0 +1,66 @@
...
+
+## Open items
+
+* **SKU/offering term consistency is out of scope for this column.** Task Force 1 raised a scenario on 2026-09-01: two terms for the same purchased offering (e.g., a 1-month and a 12-month option) should stay in the same time unit so the largest-whole-unit MUST does not make them harder to compare. That scenario is about comparing purchase options across SKUs, which is `PurchaseDurationType` / SKU Price dataset territory (#2424) — that dataset carries `SkuId` to formally scope "same offering" against. The Contract Commitment dataset records commitments already entered into, not a catalog of options, and has no comparable identifier, and "offering" is not itself a defined FOCUS term. This PR does not add a requirement here; the scenario should be tracked against #2424 instead.
```
Matt-Cowsert, line 66: Nothing on #2424 tracks the SKU term consistency case yet, so deferring it there still needs a tracked home: a thread on #2424, or a FOCUS 1.6 item if it waits for a later release. Without one, the deferral has no next step once this PR merges.

Matt-Cowsert (commented):
The PR description for this ContractCommitmentDurationType change dates from August 3, and four parts of it are out of date for member review starting September 24:

* The description lists Quarter as an open question, but the column now drops it.
* The description does not mention UnitFormat, which gains Week and a new SHOULD to use the largest whole time unit for every UnitFormat column that expresses a time-based unit.
* The description calls the change non-breaking for generators already emitting the natural form, but the supporting content now says "non-breaking" is not accurate.
* The description says the MUST matches the Recommendation dataset in #2599, but #2444 removed that column on September 18.

Removing the Allowed Values section also settles #2262, the question about that section's introductory sentence, so the description can close it.

Once this merges, PurchaseDurationType in #2424 still carries Quarter in its own unit list, and its note describes Week as a unit UnitFormat lacks, so both need the same alignment on #2424. Whether it also takes the largest-whole-unit MUST depends on the SKU term consistency case this PR defers.

We can walk through the line comments on the Task Force 1 call on September 22.

specification/requirements_model/releases/1.5/model_rules/datasets/contract_commitment/columns/contractcommitmentdurationtype.json
flanakin: 🤖 [AI][Claude Code] ✅ **Implemented**

Confirmed against `tests/test_format_function.py`: any rule whose `MustSatisfy` contains the word "format" must have `Function: "Format"`. Changed C-004-O's `Function` from `Validation` to `Format`.

specification/datasets/contract_commitment/columns/contractcommitmentdurationtype.md
flanakin: 🤖 [AI][Claude Code] ✅ **Implemented**

Confirmed C-004-O's regex and C-005-O's OR-list both stop at Minute — neither includes Second, Millisecond, or Microsecond. Applied your suggestion to scope the wording to Minute–Year rather than expanding the checks.

specification/requirements_model/releases/1.5/model_rules/datasets/contract_commitment/columns/contractcommitmentdurationtype.json
flanakin: 🤖 [AI][Claude Code] ✅ **Implemented**

Applied — kept in sync with the column bullet's Minute–Year scoping fix (same gap, same resolution as the linked MD comment).

specification/datasets/contract_commitment/columns/contractcommitmentdurationtype.md
flanakin: 🤖 [AI][Claude Code] ✅ **Implemented**

Agreed — "unit of time" wasn't tied to a defined list, so "1 Quarter" was arguably still valid under this MUST even after Quarter was dropped from the column's own vocabulary elsewhere in this file. Tied it to UnitFormat's units to close that.

specification/requirements_model/releases/1.5/model_rules/datasets/contract_commitment/columns/contractcommitmentdurationtype.json
flanakin: 🤖 [AI][Claude Code] ✅ **Implemented**

Applied — matches the column MUST's UnitFormat tie-in (same gap, same resolution as the linked MD comment).

supporting_content/datasets/contract_commitment/columns/contractcommitmentdurationtype.md
flanakin: 🤖 [AI][Claude Code] ✅ **Implemented**

Confirmed: this PR drops Quarter/Quarters from the allowed values, so "preserves ... the allowed-value list" was inaccurate. Removed that clause.

supporting_content/datasets/contract_commitment/columns/contractcommitmentdurationtype.md
flanakin: 🤖 [AI][Claude Code] ✅ **Implemented**

Confirmed against the TF1 decision comment (issue comment posted 2026-09-02, but stating the meeting itself was held "on September 1"). 2026-09-02 was this file mistaking the comment's post-date for the meeting date. Fixed to 2026-09-01.

supporting_content/datasets/contract_commitment/columns/contractcommitmentdurationtype.md
flanakin: 🤖 [AI][Claude Code] ✅ **Implemented**

Posted the scenario on [#2424](https://github.com/FinOps-Open-Cost-and-Usage-Spec/FOCUS_Spec/pull/2424#issuecomment-5781505360) so it has a tracked home before this PR merges.

flanakin:
🤖 [AI][Claude Code] **PR Update Summary**

**Addressed:** 8 thread(s)
- ✅ Implemented: 8

Fixed a CI-breaking `Function`/`MustSatisfy` mismatch on C-004-O, scoped the UnitFormat references (bullets + RM rules) to Minute–Year to close the sub-minute and "1 Quarter" gaps, corrected a stale allowed-value-list claim, and fixed the TF1 meeting date. Also posted the SKU/offering term-consistency scenario on #2424 so it has a tracked home.

rileyjenk (approved):

rileyjenk:
Though I wish we use a duration format like [ISO 8601](https://en.wikipedia.org/wiki/ISO_8601)

PR #2395, opened 2026-05-18
FR #2315: Add AllocatedServiceName column

thecloudman:
## Summary

Adds `AllocatedServiceName` as a new Conditional column to the Cost and Usage dataset to address a gap created by the origin-preservation requirement in `DataGeneratorCalculatedSplitCostAllocationHandling`.

### The Problem

`DataGeneratorCalculatedSplitCostAllocationHandling` requires that `ServiceName` and `ServiceCategory` on allocated charge rows reflect the **origin charge** service — not the consuming service. This is correct and necessary: it ensures `SUM(EffectiveCost) WHERE ServiceName = 'AmazonEC2'` always returns the correct invoice-reconcilable EC2 total, inclusive of all allocated rows.

The side-effect is that the consuming service identity is absent from allocated rows. The EKS workload cost sits on those rows but carries no queryable service label — `SUM(EffectiveCost) WHERE ServiceName = 'AmazonEKS'` returns zero. To break down cost by consumer service a practitioner must filter on `AllocatedResourceId IS NOT NULL` and parse `AllocatedTags`, or join `AllocatedResourceId` to an external cluster/pod inventory.

### Concrete Example: AWS CURv2 → FOCUS 1.4 → FOCUS 1.5

AWS CURv2 natively emits both a parent EC2 row and separate EKS pod child rows for each split. Practitioners can `GROUP BY line_item_product_code` to see both EC2 infrastructure cost and EKS workload cost at pod grain simultaneously. FOCUS 1.4 origin-preservation collapses both to `ServiceName = 'AmazonEC2'` — the EKS service label is gone and the consumer service cost is not directly queryable. FOCUS 1.5 `AllocatedServiceName` restores it.

**CURv2 — parent EC2 row + child EKS pod row for the same instance:**

| CURv2 field | Parent row (EC2) | Child row (EKS pod) |
| :--- | :--- | :--- |
| `line_item_product_code` | `AmazonEC2` | `AmazonEKS` |
| `line_item_resource_id` | `i-0a1b2c3d4e5f67890` | `arn:aws:eks:us-east-1:123456789012:pod/prod-k8s-cluster-01/observability/metrics-agent-7d4f9b-xk2lp/…` |
| `split_line_item_parent_resource_id` | *(null — this is the parent)* | `i-0a1b2c3d4e5f67890` |
| `line_item_usage_type` | `USE1-BoxUsage:m5.4xlarge` | `USE1-EKS-EC2-vCPU-Hours` |
[...]

Issue #2315, opened 2026-04-28
[FR] Improve split cost allocation guidance for data generators and practitioners


Matt-Cowsert:
### Problem Statement

The FOCUS specification defines columns and normative requirements for split cost allocation (e.g., `AllocatedResourceId`, `AllocatedMethodId`, `AllocatedMethodDetails`, and the `DataGeneratorCalculatedSplitCostAllocationHandling` attribute), but practitioners and data generators report that the spec lacks sufficient guidance to implement and consume split allocation data correctly.

Two specific gaps surfaced in the FinOps Foundation `#chat-focus` Slack channel ([thread](https://finopsfoundation.slack.com/archives/C034GBRQGGY/p1776763266729129), April 2026):

**Gap 1 — Data generator transformation is non-obvious.** When mapping native billing data (e.g., AWS CUR v2 split cost allocation columns) into FOCUS, generators cannot simply copy provider augmentation columns into FOCUS allocated columns. The FOCUS approach requires *splitting the underlying rows* such that allocated rows sum back to the origin charge. This structural difference is implicit in the spec but not clearly stated, and there are no worked examples contrasting native input data against the resulting FOCUS rows.

> Graham Murphy (TechnologyOne): *"I had huge difficulty in designing the split Data. It's not a simple use cur solution columns into new focus allocated columns, you actually need to split the main data."*

**Gap 2 — Practitioner queries become fragmented.** Once data is split per FOCUS rules, common questions like "what is the total cost or running hours of EC2 instance X?" require combining allocated rows with origin/unallocated portions. Practitioners must understand that the original full-resource view is no longer expressed as a single row. The spec's Supported Features document provides 5 short SQL snippets, but does not document the full set of common aggregation patterns or warn against pitfalls (double-counting, missing unallocated remainders, treating `ResourceId` and `AllocatedResourceId` interchangeably).

> Graham Murphy (TechnologyOne): *"What happens though is you lose ec2 data so to view full ec2 hours you have to change queries to look at multiple columns now. We prob need a few more columns to make split cost work well."*

> Mike Fuller (FinOps Foundation): *"I think we need to write up a detailed doc on how split costs handling is defined to work in the FOCUS spec to make it easier and clearer for data generators and practitioners. Looking in the spec and supporting_content I do not think we have enough information to make it simple."*

[...]

PR #2611, opened 2026-08-06
AI #2610: Carry open AllocatedServiceName requirement threads out of PR #2395

### LINKED ISSUE #2610: [AI] Submit a PR against PR #2395 covering the open AllocatedServiceName requirement threads
**Assignee(s):** @kk09v
**Parent Issue:** #2315

**Detailed Description**

**Definition of Done**

Matt-Cowsert:
## Summary

Carries the two open review threads on PR #2395 out of that PR, so PR #2395 can be delivered as approved. Both threads sit on the `AllocatedServiceName` requirements block, and this PR changes only that block. No other file is touched.

Opened against the PR #2395 branch (`2315-allocated-service-columns`). It re-points to `working_draft` on its own once PR #2395 merges.

### Nullability bullets state the condition directly

The bullets on PR #2395 anchor `AllocatedServiceName`'s nullability to `AllocatedResourceId`'s. That stands in for the condition this column cares about, which is whether the *charge* is the allocated portion of a split. The three cases here state that condition directly, in the wording `AllocatedResourceId` already uses for itself. Behavior is identical on every row where both columns are populated.

It also covers the case @pydi-aws raised on PR #2395. When a data generator's configuration suppresses resource identifiers on split rows, as with AWS Data Exports where `INCLUDE_RESOURCES` is disabled, the inherited wording requires `AllocatedServiceName` to be null even where the consuming *service* is known. Keying on the *charge's* role lets a generator surface it, without softening the rule to SHOULD.

### Value-consistency requirement added

From @kk09v's suggestion on PR #2395: a data generator that publishes a *service* under one `ServiceName` should not surface a permuted form of that name in `AllocatedServiceName`, so that "K8S" does not appear where the dataset elsewhere carries "Kubernetes".

The `when` condition follows @flanakin's proposal from TF-2 on 2026-08-05. It scopes the requirement to the *service* the column names, so the column can still carry a *service* different from the one in `ServiceName` on the same row, which is the reason the column exists. SHOULD rather than MUST for the reason @kk09v gave: an allocated *service* is not always offered on its own, so a data generator may have no stand-alone `ServiceName` to align to.

The bullet sits as a sibling of the nullability anchor rather than a child of it. The anchor reads "MUST adhere to the following nullability requirements:", which scopes its children to nullability, and this is a value-consistency expectation.

### Resulting block

```markdown
  * AllocatedServiceName MUST be null when a [*charge*](#glossary:charge) is not related to a data generator-calculated split cost allocation.
  * AllocatedServiceName MUST be null when a *charge* represents the unallocated portion of the origin *charge* after split cost allocation.
  * AllocatedServiceName MUST NOT be null when a *charge* represents the allocated portion of the origin *charge*.
* AllocatedServiceName SHOULD match the ServiceName of the *service* to which cost is allocated when that *service* is represented by ServiceName in the FOCUS dataset.
```

### Verification

`enhanced_markdown_lint.py` passes on the changed file. The three nullability bullets are word-for-word the cases `AllocatedResourceId` states for itself, other than the column name and the first-occurrence glossary link on *charge*.

### Type of Change

### Author Checklist

* [ ] **AI-Generated Examples:** Not applicable. This PR contains no examples.

### Changelog

* 2026-08-06: Opened against the PR #2395 branch carrying both open threads.

Closes #2610
Part of #2315

### DIFF
--- a/specification/datasets/cost_and_usage/columns/allocatedservicename.md
+++ b/specification/datasets/cost_and_usage/columns/allocatedservicename.md
@@ -9,8 +9,10 @@ AllocatedServiceName MUST adhere to the following requirements:
 * AllocatedServiceName MUST be of type String.
 * AllocatedServiceName MUST conform to [StringHandling](#attributes.stringhandling) requirements.
 * AllocatedServiceName MUST adhere to the following nullability requirements:
-  * AllocatedServiceName MUST be null when [AllocatedResourceId](#datasets.costandusage.allocatedresourceid) is null.
-  * AllocatedServiceName MUST NOT be null when AllocatedResourceId is not null.
+  * AllocatedServiceName MUST be null when a [*charge*](#glossary:charge) is not related to a data generator-calculated split cost allocation.
+  * AllocatedServiceName MUST be null when a *charge* represents the unallocated portion of the origin *charge* after split cost allocation.
+  * AllocatedServiceName MUST NOT be null when a *charge* represents the allocated portion of the origin *charge*.
+* AllocatedServiceName SHOULD match the ServiceName of the *service* to which cost is allocated when that *service* is represented by ServiceName in the FOCUS dataset.
 
 ## Column ID
 


Matt-Cowsert:
closing in favor of #2612

Issue #2515, opened 2026-07-06
[Feedback] DataGeneratorCalculatedSplitCostAllocationHandling: dimension-matching wording, missing double-counting rule, and undefined origin-row model


thecloudman:
## Feedback Summary

Reviewing PR #2395 (part of FR #2315) surfaced three pre-existing gaps in `DataGeneratorCalculatedSplitCostAllocationHandling` (introduced 1.3, current wording since 1.4). None were introduced by #2395 or its `AllocatedServiceName` column; they predate it, and the same gaps exist in the 1.3 text. Full credit to @ijurica, who identified all three in [PR #2395's review thread](https://github.com/FinOps-Open-Cost-and-Usage-Spec/FOCUS_Spec/pull/2395#issuecomment-4856938846).

**1. "When present" wording is unsatisfiable for `Allocated*` columns**

Current text:

> * *FOCUS dataset column* representing a dimension MUST match the corresponding value in the *origin charge* when present in an *allocated charge*.
> * *FOCUS dataset column* representing a non-summable *metric* (e.g., unit prices) MUST match the corresponding value in the *origin charge* when present in an *allocated charge*.

"Present" is a dataset-level concept (§1.7: Mandatory/Conditional/Optional): once a column is part of the schema, it is present on every row, including `Allocated*` columns on allocated charge rows. Read literally, this requires `AllocatedResourceId`, `AllocatedResourceName`, `AllocatedTags`, and now `AllocatedServiceName` to match the origin charge, which is impossible by design, since those columns exist specifically to carry a value the origin charge doesn't have.

Proposed fix:

> * *FOCUS dataset column* representing a dimension MUST match the corresponding value in the *origin charge* when the column is not null in the *origin charge*.
> * *FOCUS dataset column* representing a non-summable *metric* (e.g., unit prices) MUST match the corresponding value in the *origin charge* when the column is not null in the *origin charge*.

Since `Allocated*` columns are always null on the origin charge (by their own nullability requirements), the rule simply never triggers for them, so there is no exclusion list to maintain as new `Allocated*` columns are added.

**2. No requirement prevents double-counting when a charge is split**

The only summable-metric constraint today is:

> * The sum of *FOCUS dataset column* across *allocated charges* MUST match the *FOCUS dataset column* in the corresponding *origin charge* when the *FOCUS dataset column* represents a summable metric (e.g., costs and quantities).

[...]

PR #2612, opened 2026-08-06
FR #2315: Enhance normative requirements for AllocatedServiceName

### LINKED ISSUE #2610: [AI] Submit a PR against PR #2395 covering the open AllocatedServiceName requirement threads

kk09v:
Enhance normative requirements for AllocatedServiceName.

## Summary

Decision was made to separate these suggestions out of #2395 to be debated and reviewed separately from the original PR.
- Improving nullability requirements
- Adding requirement for 

### Type of Change
* [X] Bug fix
* [X] Editorial / Formatting

### Author Checklist
* [N/A] **AI-Generated Examples:** If this PR includes examples generated by AI, I have manually verified that the data is mathematically accurate, logically consistent, and compliant with the FOCUS schema.


kk09v:
Decision was made to separate these comments from the original PR
https://github.com/FinOps-Open-Cost-and-Usage-Spec/FOCUS_Spec/pull/2395/changes#r3680105803
https://github.com/FinOps-Open-Cost-and-Usage-Spec/FOCUS_Spec/pull/2395/changes#r3718226458

kk09v:
Per Matt Cowsert:
On the nullability bullets: anchoring them to AllocatedResourceId makes AllocatedServiceName inherit that column's nullability rather than stating the condition this column cares about, which is whether the row is the allocated portion of a split. Keying on the charge's role states it directly, matches how AllocatedResourceId words its own three cases, and keeps behavior identical on every row. It also lets a data generator populate AllocatedServiceName on split rows where the consuming service is known and the resource identifier is suppressed, for example AWS Data Exports with INCLUDE_RESOURCES disabled, without softening the rule to SHOULD.

kk09v:
Commentary on inclusion of `AllocatedServiceName SHOULD match the ServiceName used by the data generator for the equivalent stand-alone service.` requirement:

Karl (originator):
- I'm trying to make the expectations explicit in that if the provider offers a "Kubernetes" service, that they don't decide to send "K8S" as the Allocated Service Name.
- I left this as SHOULD because I could see a situation where the Allocated Service is not something that's offered as a stand-alone service. E.g. it could be a platform that's only offered on top of Kubernetes.

Matt Cowsert:
The consistency expectation is worth having, and SHOULD is the right strength for the reason you gave: an allocated service is not always offered on its own, so a data generator may have no stand-alone ServiceName to match.

specification/datasets/cost_and_usage/columns/allocatedservicename.md
```diff
@@ -9,8 +9,10 @@ AllocatedServiceName MUST adhere to the following requirements:
 * AllocatedServiceName MUST be of type String.
 * AllocatedServiceName MUST conform to [StringHandling](#attributes.stringhandling) requirements.
 * AllocatedServiceName MUST adhere to the following nullability requirements:
-  * AllocatedServiceName MUST be null when [AllocatedResourceId](#datasets.costandusage.allocatedresourceid) is null.
-  * AllocatedServiceName MUST NOT be null when AllocatedResourceId is not null.
+  * AllocatedServiceName MUST be null when a [*charge*](#glossary:charge) is not related to a data generator-calculated split cost allocation.
+  * AllocatedServiceName MUST be null when a *charge* represents the unallocated portion of the origin *charge* after split cost allocation.
```
kk09v, line 13: Do we want this to be NULL or to be the same as ServiceName?

If NULL, we will need to change one of the queries because it filters out NULLs.
(Get Total Effective Cost by Allocated Service Name) in
specification/supported_features/data_generator_calculated_split_cost_allocation.md

specification/datasets/cost_and_usage/columns/allocatedservicename.md
```diff
@@ -9,8 +9,10 @@ AllocatedServiceName MUST adhere to the following requirements:
 * AllocatedServiceName MUST be of type String.
 * AllocatedServiceName MUST conform to [StringHandling](#attributes.stringhandling) requirements.
 * AllocatedServiceName MUST adhere to the following nullability requirements:
-  * AllocatedServiceName MUST be null when [AllocatedResourceId](#datasets.costandusage.allocatedresourceid) is null.
-  * AllocatedServiceName MUST NOT be null when AllocatedResourceId is not null.
+  * AllocatedServiceName MUST be null when a [*charge*](#glossary:charge) is not related to a data generator-calculated split cost allocation.
```
davidearney, line 12: These requirements define nullability in charge-state terms (not related, unallocated portion, allocated portion). To make downstream validation deterministic, can we add one explicit mapping sentence to existing row signals, or state that this is intentionally semantic-only and validated outside static/null checks?

Matt-Cowsert:
Worth also looking at #2515 to ensure we're covering our bases

Matt-Cowsert:
## TF2 Action Item - August 12, 2026

This PR targets 1.6 rather than 1.5. @kk09v confirmed he will drive the two open threads to answers and said the existing AllocatedServiceName definition is not problematic, which is why the feedback went in as a suggestion rather than a change request. @ijurica asked whether the work is aimed at 1.6 and @kk09v agreed, adding there is no urgency to push it forward. @matt-cowsert confirmed 1.5 shipping with the merged definition is acceptable.

Two threads are open. The first is wording that communicates the intended outcome rather than what is easy to test for. @ijurica offered the commitment programs eligibility wording as the closest prior art, where the requirement asks for values that are generally used. The second is @matt-cowsert's suggested change to the nullability requirements, which states the conditions under which the value is null rather than chaining nullability off another column. @ijurica leaned toward that suggestion for now, while noting a later pass over the attributes could support more static phrasing.

Two items of context settled during the meeting. The unexpected files in this diff trace to PR #2395 merging while this PR was built on it. Separately, Feedback #2515 covers dimension-matching wording, a missing double-counting rule, and the undefined origin-row model on the split cost allocation attribute; @ijurica asked that it be visible here so the two efforts can be coordinated if this work lands in 1.6. It is assigned to @matt-cowsert.

**Scope:**
- [ ] Answer the open thread on requirement wording that communicates intent over testability
- [ ] Answer the open thread on stating null conditions directly rather than chaining off another column's nullability
- [ ] Obtain @thecloudman's review as the author of the merged AllocatedServiceName column

**Assignee:** @kk09v

specification/datasets/cost_and_usage/columns/allocatedservicename.md
thecloudman: AllocatedServiceName cannot have a value if the charge row is not spit costs. This is meant to outline the fact these columns are not populated when the row does not have split cost. I anchored on AllocatedResourceID as it made sense since this one would be populated when split costs exist.

kk09v: I think that's the above line. This line is separate (but related) and describes what happens when it's the unallocated portion of a split cost allocation. This line wouldn't have an AllocatedResourceID but is part of a split cost allocation.

Here's a hyper-simplified example that illustrates this.

Origin charge
| resourceid/allocatedresourceid | effectivecost | servicename/allocatedservicename |
| :-------------- | :-------------- | :-------------- |
| VM1234 | $10 |  Virtual Machines |

Split charges
| resourceid/allocatedresourceid | effectivecost | servicename/allocatedservicename |
| :-------------- | :-------------- | :-------------- |
| Pod1 | $5 | Kubernetes | 
| Pod2 | $3 | Kubernetes |
| NULL (representing unallocated portion) | $2 | [NULL or Virtual Machines?] |

Matt-Cowsert:
## TF2 Action Item - August 19, 2026

Draft status closes on August 27, 2026 for anything carried in the FOCUS 1.5 scope. @kk09v confirmed PR #2612 stays in the FOCUS 1.5 scope.

Two changes sit in the pull request. The first moves the nullability of AllocatedServiceName off AllocatedResourceId. Hanging it there forces AllocatedServiceName to be null on the unallocated portion of a split cost allocation, and @kk09v expects a value on that record when the sibling allocated records carry one. @ijurica raised the same concern from another angle: charge records exist with no resource identifier and a populated service name, so the resource identifier may be the wrong anchor. The second change requires AllocatedServiceName to match the service provider's existing ServiceName for that offering where one exists, so a virtual machine split to Kubernetes reports the provider's own name for its Kubernetes service rather than a variant of it.

@kk09v has open review comments to answer, including one from @thecloudman he has replied to and one from @matt-cowsert he was unsure how to answer. PR #2395 has merged since this pull request was opened, and @matt-cowsert flagged the merge state as something to scan.

**Scope:**
- [ ] Answer the open review comments, including @matt-cowsert's question on the nullability change
- [ ] Reconcile the file against PR #2395 now that it has merged
- [ ] Mark PR #2612 ready for review

**Assignee:** @kk09v

Matt-Cowsert:
## TF2 Action Item - August 19, 2026

@kk09v and @ijurica both asked for input from a data generator already producing AllocatedServiceName, because the open question is what the column should carry on the unallocated portion of a split cost allocation rather than how the requirement should be worded. @thecloudman runs the column internally and committed to running scenarios against it and reporting back.

@thecloudman's opening position is that gating nullability on another column is used elsewhere in the specification, so the pattern itself is not the concern. @kk09v's response is that the pattern forces a null on the unallocated portion, which may not be what a consumer of that record wants, and that coalescing is one alternative. @thecloudman named a second case worth covering: practitioners have been tripped up assessing total EffectiveCost against split cost allocation under the current definition.

**Scope:**
- [ ] Run internal scenarios covering AllocatedServiceName on the unallocated portion of a split cost allocation
- [ ] Report what the column should carry on that record, and whether its nullability should hang off AllocatedResourceId
- [ ] Report how the change reads when assessing total EffectiveCost with split cost allocation

**Assignee:** @thecloudman

thecloudman:
Thanks Karl — that aligns with what the production data actually supports once you weight it the way the WG is leaning.

**Q1:** Agreed — populate AllocatedServiceName with the origin ServiceName on the unallocated-remainder row rather than NULL. Our internal build (an EKS split-cost implementation, referred to below as our "Model C" — that's our own naming, not AWS/CUR terminology) already preserves origin ServiceName on every split row by construction, so this costs us nothing and gives a consistent "same service, allocated vs. unallocated" view without needing a COALESCE in every query. Worth calling out as a lesson: an earlier version of our build didn't preserve the origin ServiceName on split rows at all (it took the consumer's identity instead), and that under-reported EC2 totals by ~$60K over 14 days — a much costlier mistake than any NULL-vs-populated choice on this column. Reinforces that origin-dimension preservation is the thing to get right; this consensus direction is compatible with it either way.

**Q2:** Good catch, and it matches something we found: our AWS-based build never actually emits an unallocated-remainder row in production (its EC2→EKS split method fully allocates parent cost to pods, folding idle capacity into pod rows as memo detail) — but that's specific to that data generator's method, not a property of split-cost allocation in general. Google's data confirms the spec needs to handle the case as real and common, not conditional. Support the nullability change.

**Q3:** Agreed this is cleaner. We independently found the same gap — `AllocatedResourceId IS NULL` isn't sufficient to disambiguate an unallocated-remainder row from an origin-anchor row, since both can be NULL there for different reasons. Anchoring on `AllocatedMethodId` (which already has its own nullability requirement) is a simpler, more directly testable proxy for "this charge is part of a data-generator-calculated split-cost-allocation method" than adding new logic around AllocatedResourceId.

kk09v:
This PR will be closed and replaced with #2655

PR #2606, opened 2026-08-06
FR #2099: Distinguish cached from uncached token costs

Matt-Cowsert:
## Summary

Implements FR #2099, the feature request to distinguish cached from uncached token costs, by adding two FOCUS-defined `SkuPriceDetails` properties, `TokenCacheAction` and `TokenDirection`, each with allowed values, plus a worked appendix example and two example queries under the Resource Usage supported feature. No new columns. The structure follows what pull request #2442 established for model identity: rows in the FOCUS-Defined Properties table and an appendix example backed by CSV datasets.

Task Force 2 decided on September 2, 2026 to replace the single `TokenType` property this pull request first proposed with two normalized properties scoped to tokens, one for caching and one for direction, each with allowed values. The two properties were built on PR #2682, stacked on this branch, and merged in on September 15, 2026.

**What changes**

* `SkuPriceDetails`: two FOCUS-defined properties. `TokenCacheAction` carries "Uncached", "Read", "Write", or "Other" and applies to SKU prices that meter tokens consumed from a request. `TokenDirection` carries "Input" or "Output" and applies to SKU prices metered in tokens whose tokens share a direction, so a SKU that bills input and output on one meter omits it. Inclusion is SHOULD, as for every FOCUS-defined property; a SKU price the property does not apply to MUST NOT carry it; the value rules are MUST, in the form `PricingCategory` and `ChargeCategory` use, with inline examples.
* One requirement makes allowed values normative for any FOCUS-defined property that lists them: the property value MUST be one of the allowed values when allowed values are specified. The properties table's fourth column is retitled from "example values" to "values" so it can hold either kind. These are the first FOCUS-defined properties to carry allowed values.
* One requirement stops a cached token being counted twice: `ConsumedQuantity` on an "Uncached" row MUST NOT include tokens counted on a "Read" or "Write" row. OpenAI's usage report counts cached tokens inside the prompt token total, so a generator that maps that total to "Uncached" and the cached count to "Read" would double count without it. Both example queries rest on it.
* Footnote 4 keeps the informative context: the model developer defines which cache interactions and which directions a model distinguishes; which of those a service provider meters as its own charge, and the meter name it uses, vary by service provider and model version; reasoning tokens carry `TokenDirection` "Output"; token modality, reasoning effort, cache retention duration, and context window size are not values of these properties.
[...]

PR #2594, opened 2026-07-31
Feedback #2543: Add ModelMesh to Fictitious Data Generator Reference and sort alphabetically

Matt-Cowsert:
## Summary

Closes #2543.

`ModelMesh` is used as an example value for the `ModelDeveloper`, `ModelFamily`, and `ModelId` properties in the [SkuPriceDetails](/specification/datasets/cost_and_usage/columns/skupricedetails.md) column definition, but it was never defined in the Fictitious Data Generator Reference table in [appendix_overview.md](/specification/appendix/appendix_overview.md). This PR adds it and sorts all three fictitious entity tables in that file alphabetically.

**1. Added a ModelMesh row.**

| Fictitious Data Generator | Service Offering | Fictitious Data Generator Description | Similar Real-World Examples |
| :--- | :--- | :--- | :--- |
| **ModelMesh** | Foundation Model Developer | An AI lab that publishes open-weight foundation models, which are self-hosted or served by cloud providers and inference platforms. | Meta (Llama), Mistral AI |

ModelMesh is described as an open-weight model developer. That distinguishes it from Solora AI, which develops proprietary models "sold directly and served by cloud providers as a first-party offering", and it fits the parameter count carried in the `modelmesh-general-7b` example model identifier. The original intent for ModelMesh was not recorded anywhere, so this is an inference from how the name is used. Reviewers who know the intent should redirect the framing.

**2. Sorted the table alphabetically by fictitious data generator name.** The table previously listed the three cloud service providers first and then appended later entries in the order they were added, which is what the issue's Additional Notes flagged. The sibling Fictitious Customer Reference table and the Appendix Entries table are both alphabetical. Row content is unchanged, and the reorder is in its own commit so it can be reviewed or dropped independently of the ModelMesh addition.

**3. Sorted the Fictitious Commitment Program Reference table alphabetically** by program name, which also puts it in acronym order (ARC, BCC, DCC, ESA, FSP, ISC, RR). It had the same append-order problem as the data generator table. Row content is unchanged. The Fictitious Customer Reference table in the same file was already alphabetical and its rows are untouched.

**4. Stated the alphabetical convention in each section's introductory sentence,** so all three tables carry the same rule and later additions have a stated insertion point rather than defaulting to the bottom of the table again.

[...]

PR #2442, opened 2026-06-03
FR #2018: AI model identity as SkuPriceDetails properties

Matt-Cowsert:
## Summary

Implements Approach A for FR #2018 (AI model identity and token consumption), tracked under AI #2417. It defines the model-identity dimensions as FOCUS-defined properties of `SkuPriceDetails`, and adds a worked appendix example. The naming and scoping questions raised in review are settled per the 2026-06-17 TF2 decision (see Naming and placement below), and the PR is out of draft.

**What changes**
* `SkuPriceDetails`: four FOCUS-defined properties added to the FOCUS-Defined Properties table: `ModelDeveloper`, `ModelFamily`, `ModelId`, `ModelVersion`. No new columns, and no change to the `SkuPriceDetails` requirements.
* Appendix: a new "AI Model Identity" example with two scenarios (a model purchased directly, and the same model served by a cloud provider as first-party), backed by two CSV datasets.

**Scope and design notes**
* Scope follows the TF2 direction: bare-bones model identity plus the input vs. output token split. Adjacent dimensions are tracked in parallel FRs: cached tokens (#2099), provisioned throughput (#2371), regional serving (#2414), processing tier and token modality (#2416), and token-unit denomination (#1941).
* The input vs. output token split is carried structurally (separate SKUs and SkuPriceIds, distinguished by `SkuMeter`), so no token-type property is added.
* FR #2018's third use case, surfacing per-request max token limits alongside cost, is out of this bare-bones scope. It is request-level configuration that sits closer to observability (in line with the FR's note excluding telemetry), so it is not addressed here and is not routed to one of the parallel FRs above.
* The worked modeling that informed this PR, including two placement approaches and a practitioner analysis, is in the FR #2018 modeling sheet: https://docs.google.com/spreadsheets/d/1YREIONAKGDikt1iAO1cL7xX7Uw0BG8EM3GrOCl3iOlY/edit

**Naming and placement**
[...]

Issue #2371, opened 2026-05-13
[FR] Standardize AI provisioned throughput commitments


Matt-Cowsert:
### Problem Statement

Practitioners managing AI workloads with prepaid throughput commitments cannot perform consistent commitment analysis using FOCUS data today. Provisioned throughput (token quantity commitments and capacity-based throughput allocations) functions as a prepaid commitment with the same analytical needs as other FOCUS commitments (utilization, savings, renewal), but FOCUS guidance does not currently address how AI commitment forms align with the specification. Practitioners must build per-provider logic to interpret AI commitment data, breaking the consistency FOCUS provides for non-AI commitments.

This FR was carved out of FR #2018 (Surface AI model identity and token consumption) on 2026-05-11 to maintain ticket discipline and enable champion-led delivery.

---

### Use Case / User Story

**Use Case 1:**
**As a** FinOps practitioner with prepaid AI throughput commitments, **I need to** analyze AI commitments alongside my other commitment types, **so that** commitment utilization and savings reporting is consistent across AI and non-AI workloads.

**Use Case 2:**
**As a** FinOps practitioner evaluating AI commitment effectiveness, **I need to** relate AI consumption to its underlying commitment (token quantity or capacity allocation), **so that** I can measure utilization, identify unused throughput, and inform renewal decisions.

---

### Success Criteria

- [ ] Practitioners can identify AI provisioned throughput commitments (token-based and capacity-based) in FOCUS data without per-provider interpretation logic.
- [ ] AI provisioned throughput commitments are consistently represented across data generators that support this capability.
- [ ] Practitioners can apply existing commitment utilization and savings workflows to AI workloads without modifying their analysis approach.
- [ ] Data generators have clear guidance on how AI provisioned throughput maps to FOCUS commitment constructs.

---

### Type of Request

- [x] **Standardization** - Data already exists across data generators but is not yet normalized in FOCUS
- [ ] **Enhancement** - Refinement or extension of an existing FOCUS column, metric, or capability
- [ ] **Net New** - A concept not yet represented in FOCUS or widely available from data generators
- [ ] **Enablement/Supporting Content** - Non-normative items like appendix explanations, examples, or supporting documentation
- [ ] **Tech Debt** - Addressing technical debt or maintenance items

---

### Organizations Requesting This Feature

[...]

PR #2613, opened 2026-08-10
FR #2414: Region scope as a SkuPriceDetails property

### LINKED ISSUE #2414: [FR] Distinguish global from regional serving scope
### Problem Statement

Service providers price the same offering differently depending on how broad a geographic area a request may be served from. A globally-routed deployment spreads serving across regions; a region-pinned deployment constrains it to one. The two are distinct price points, and FOCUS data cannot currently tell them apart.

AI model serving is where providers price this dimension most distinctly today, and it is the clearest case. A globally-routed deployment (e.g., `global.anthropic.claude`) routes inference across regions for availability, while a region-pinned deployment (e.g., `us.anthropic.claude`) constrains inference to a specific region. The TF2 discussion cited a delta of roughly 10-20% between the two for the same model.

* **AWS** Bedrock cross-region (global) inference profiles vs. regional inference profiles, surfaced in AWS Marketplace SKU identifiers as `global.anthropic.*` vs. `us.anthropic.*`.
* **Azure** global, data-zone, and regional deployment types for AI models.
* **GCP** multi-region vs. regional scopes.

The dimension itself is not specific to AI. Other multi-region offerings, including storage and content delivery, distinguish a multi-region footprint from a single-region one. In several of those, serving breadth is currently bundled with replication in provider pricing, which is why the existing Redundancy property may already cover part of that ground. Scoping the concept generally now means those cases can arrive as values under one key rather than as competing per-service properties.

Practitioners cannot currently:

* Distinguish globally-routed from region-pinned spend in a standardized, cross-provider way.
* Quantify the cost premium (or discount) of one serving scope over another.
* Identify which spend ran under a region-pinned scope using billing data alone.

Without guidance, data generators carry serving scope inconsistently. Some encode it inside provider-specific SKU identifiers, others omit it, and others may invent ad hoc properties. The result is the cross-provider normalization gap FOCUS exists to close.

This FR was carved out of FR #2018 (Surface AI model identity and token consumption) during TF2 on 2026-05-20, continuing the carve-out pattern established with FR #2371 (provisioned throughput commitments). AI remains the driving case; the concept is scoped generally so the property does not need replacing when other services price the same dimension.

---

### Use Case / User Story

**Use Case 1:**
**As a** FinOps practitioner managing AI costs across providers, **I need to** distinguish globally-routed from region-pinned model serving in billing data, **so that** I can report cost by serving scope without writing per-provider parsing logic.

**Use Case 2:**
**As a** FinOps practitioner subject to data-residency requirements, **I need to** identify which spend ran under a region-pinned serving scope, **so that** I can produce the region-pinned share of spend as evidence for a residency review, without per-provider parsing.

**Use Case 3:**
**As a** FinOps practitioner optimizing AI spend, **I need to** quantify the price difference between global and regional serving of the same model, **so that** I can size the savings available from moving a workload between scopes.

---

### Success Criteria

* Practitioners can filter and group costs by serving scope across providers without provider-specific parsing logic.
* Practitioners can quantify the price difference between global and regional serving of the same underlying offering using FOCUS data alone.
* Serving scope is consistently represented across data generators that price global and regional serving distinctly.
* Data generators have clear guidance on how to represent serving scope in FOCUS data, including its example values and its boundary with existing replication properties.

---

### Type of Request

* [x] **Standardization** - Data already exists across data generators but is not yet normalized in FOCUS
* [ ] **Enhancement** - Refinement or extension of an existing FOCUS column, metric, or capability

---

* **FinOps Foundation** - Nice to Have - Carved out from FR #2018 during TF2 on 2026-05-20. Standardizing global vs. regional serving scope extends the AI cost coverage begun by FR #2018 (model identity and token consumption), alongside FR #2099 (cached vs. uncached tokens) and FR #2371 (provisioned throughput commitments), which sit at different stages of the 1.5 cycle, and scopes the dimension generally so other regionally-served offerings can use the same key. Supporting organizations welcomed via Community Support comments below.

---

* [x] **Partially** - Some data generators support this, others don't

Both major AI providers and the cloud platforms that host them expose serving scope in SKU identifiers or deployment configuration: AWS Bedrock (global vs. regional inference profiles, surfaced as `global.anthropic.*` / `us.anthropic.*`), Azure (global, data-zone, and regional deployment types), and GCP (multi-region vs. regional scopes). Coverage and the exposed billing dimension vary by provider, and the scope is often embedded in a composite SKU identifier rather than a discrete field. Outside AI, multi-region and single-region footprints are priced distinctly in several storage and delivery offerings, though there the scope is typically bundled with replication rather than priced as routing alone.

---

### Data Generator Documentation Links (Optional)

* AWS Bedrock model/region compatibility: https://docs.aws.amazon.com/bedrock/latest/userguide/models-region-compatibility.html

---

* FR #2018 (Surface AI model identity and token consumption) - parent FR from which this was carved out
* FR #2099 (Distinguish cached from uncached token costs) - adjacent AI SkuPriceDetails pricing-dimension FR
* FR #2371 (Standardize AI provisioned throughput commitments) - adjacent AI carve-out
* TF1 SKU Price dataset working sheet (Pricing Region ID, SKU Price Eligibility), overlapping scope to coordinate (see Open Dependencies): https://docs.google.com/spreadsheets/d/1vtkK5cNCwZIXM0LbZOs9WTw9LN1pZWUU4DZXe_FXPoc/edit?gid=2001743556#gid=2001743556
* TF2 meeting 2026-05-20, carve-out decision and scope discussion (FR #2018 status update): https://github.com/FinOps-Open-Cost-and-Usage-Spec/FOCUS_Spec/issues/2018#issuecomment-4503797831

---

This is a candidate direction for discovery, not a settled design, and is subject to TF1 coordination (see Open Dependencies). It follows the pattern proposed for adjacent AI pricing dimensions in FR #2099 (cache status) and FR #2371 (AI commitment attributes).

* **Serving scope as a SkuPriceDetails property.** A FOCUS-defined property (name pending standardization; see MVP Definition) with a controlled value set; see MVP Definition for the values the implementation takes. Region-pinned rows would continue to carry the specific region through existing region columns; the property would capture the routing topology, not the region itself.
* **Separate rows per serving scope.** Global and regional serving are distinct price points and would be represented as separate rows, each with a distinct SkuPriceId, following the same pattern as cache-variant pricing in FR #2099.
* **ListUnitPrice reflects the resolved scope price.** The price delta between global and regional serving would be captured as the difference in ListUnitPrice across the two rows, not as a multiplier or formula.

Open design questions for discovery:

* Whether the value set stays binary ("Global"/"Regional") or admits additional values (e.g., "DataZone", "Sovereign") as provider offerings evolve. The MVP Definition takes three values; broader enumeration remains a discovery decision.
* Where the boundary sits between serving scope and the existing Redundancy property, which carries "Local", "Zonal", and "Global" and may already cover offerings that bundle replication with multi-region serving.
* Coordination with TF1's SKU Price dataset work so serving scope does not duplicate or conflict with Pricing Region ID or SKU Price Eligibility (see Open Dependencies).
* Whether SkuPriceDetails should remain a flat key-value column or evolve toward a structured (JSON-object) representation like ContractApplied.

---

**Current Workaround:** Practitioners parse provider-specific SKU identifiers (e.g., the `global.` vs. `us.` prefix in Bedrock and AWS Marketplace SKU ids) to infer serving scope, with custom logic per provider.

**Open Dependencies:**

* **TF1 coordination (open).** TF1's SKU Price dataset work defines Pricing Region ID and SKU Price Eligibility. Serving scope (routing topology) appears related to but distinct from the region a price is quoted for. The boundary needs to be settled with TF1 before column or property decisions, to avoid duplicate scoping. Flagged as an open coordination item, not resolved.
* **SkuPriceDetails structure (tracking).** Irena Jurica (Neos) raised during TF2 whether SkuPriceDetails, currently a flat key-value column, needs to evolve toward a structured JSON-object representation (like ContractApplied). Relevant to how this property is carried, but not a prerequisite for opening this FR.

**Related Requests:** Carved out from FR #2018 during TF2 (2026-05-20). Adjacent AI carve-outs: FR #2099 (caching) and FR #2371 (provisioned throughput). A single FOCUS row may carry properties from more than one of these FRs; model identity, cache status, and serving scope are orthogonal pricing dimensions.

---

### Adoption Impact

- [ ] **Adoption Blocker** - Organizations cannot adopt FOCUS without this feature
- [ ] **Adoption Accelerator** - This feature would help organizations start using FOCUS
- [x] **Practitioner Enhancement** - Improves experience for existing FOCUS users
- [ ] **Technical Improvement** - Internal specification refinement

**Justification:** One organization supports this, the FinOps Foundation, as Nice to Have. It was carved out of FR #2018 during TF2 rather than raised by a practitioner or vendor with a blocked workflow. A workaround exists: practitioners parse the `global.` versus `us.` prefix in provider SKU identifiers, with custom logic per provider. Organizations can adopt FOCUS today without this, so the gap is comparison fidelity rather than a capability barrier.

This FR enhances existing supported features and does not define a new one. The closest is **Verification, Comparison, and Fluctuation Tracking of Unit Prices**, which depends on SkuPriceId, SkuPriceDetails, and ListUnitPrice. That feature lets a practitioner compare two unit prices for the same SKU but carries nothing that explains why they differ, so a global and a region-pinned price for the same model appear as two unexplained rates. The **AI Billing** feature arriving in 1.5 through open PR #2605 lists SkuPriceDetails as a supporting column and groups token cost by SkuPriceId; serving scope would become a grouping dimension there alongside model identity and token type. **Location** is adjacent but does not cover this, since RegionId, RegionName, and AvailabilityZone describe where a resource runs and a globally routed price has no single region to record. No new supported feature is needed. Which column carries the property is settled, because PR #2442 established SkuPriceDetails as where AI pricing dimensions live. That is a separate question from the open TF1 boundary, which concerns whether the property duplicates SKU Price dataset constructs rather than where it sits. Neither feature's example query can separate serving scopes today. The unit price query filters to a single SkuPriceId and selects SkuPriceDetails, so a practitioner comparing scopes must already know which SkuPriceId corresponds to which. The AI Billing query groups by SkuPriceId, so scopes fall into separate rows with nothing naming what separates them. Scoping the property to the general dimension rather than the AI case means offerings outside AI that price serving breadth land on the same key and the same features, without new supported-feature work.

### Implementation Scope

**MVP:** A FOCUS-defined serving scope property on SkuPriceDetails, of type String, with example values "Global", "Regional", and "DataZone", plus a footnote separating it from the existing Redundancy property. Global and regional serving are separate price points carrying separate SkuPriceId values, so ListUnitPrice reflects the resolved scope price and the delta is the difference between rows. An appendix example with CSV data demonstrates both scopes on the same model. No new columns and no change to the SkuPriceDetails requirements. Drafted in PR #2613.

This refines the requester proposal in two places. The proposal carried `x_ModelServingScope` as a working key, which was internally inconsistent, since the `x_` prefix marks provider-defined extensions and cannot appear on a FOCUS-defined property. The MVP uses an unprefixed FOCUS-defined key, following the precedent PR #2442 set for model identity. The proposal scopes the value set to "Global" and "Regional"; the MVP adds "DataZone" because Azure prices data-zone deployments today and a binary set would be incomplete on arrival. The proposal framed the concept as AI-specific; the MVP scopes it to serving scope as a general priced dimension, so the key does not need replacing when other multi-region offerings price the same thing. AI remains the documented instance.

Covers all four Success Criteria, with two caveats. Criterion 3, consistent representation across data generators, is satisfied to the extent the specification can reach: the property definition and worked example set the convention, and actual consistency follows adoption. Criterion 4 asks for example values rather than a normative enumeration, consistent with how every other SkuPriceDetails property works; whether FOCUS-defined properties should carry normative value sets is an open question shared with FR #2099. Its boundary clause is met by a footnote separating serving scope from Redundancy, which is adequate for the AI case and thin for offerings that bundle replication with multi-region serving.

Excluded from the MVP: "Sovereign" and further scope values until a provider prices them as a distinct point; processing tier (standard, batch, provisioned), which qualifies a charge without describing geography; latency and data-residency guarantees, which are contract terms rather than billing dimensions.

**North Star:** Practitioners can group and compare spend by serving scope across every data generator that prices scope distinctly, with no provider-specific parsing, and can identify region-pinned spend from billing data alone. The property is carried on Cost and Usage through SkuPriceDetails and, once the SKU Price dataset lands, on rate-card rows under the same key, so scope-based price comparison works before consumption as well as after. The AI Billing supported feature carries serving scope as a grouping dimension alongside model identity and token type, and non-AI offerings that price serving breadth use the same key rather than service-specific properties. The value set has grown to cover the scope tiers providers actually price rather than the three the MVP names.

**Phasing:** Single phase. The MVP satisfies all four Success Criteria within the Cost and Usage dataset, which is where the FR asks its questions, so nothing is blocked on work that does not yet exist.

Two sequencing dependencies, neither of which splits delivery:

* PR #2424 (Add `SKU Price` dataset) defines PricingRegionId and SkuPriceEligibility. The boundary argument in PR #2613 is written against that PR as drafted. If TF1 changes it, the argument gets revisited rather than the phasing.
* PR #2606 (FR #2099) also adds a row and a footnote to SkuPriceDetails. Whichever merges second renumbers the footnote.

Carrying the same property onto SKU Price dataset rows is a follow-on once that dataset merges, not a second phase of this FR.

### Impacted Parties

- [x] FinOps Practitioner - end users who analyze or act on the data
- [x] FOCUS Data Generator - data generators producing output aligned to the spec
- [ ] Vendor Supporting FOCUS - vendors or tools ingesting the spec or using the spec language in their UI
- [ ] Other

### Level of Ambiguity

**3.** Reducing ambiguity: the gap is documented at every named provider (AWS Bedrock inference profiles, Azure deployment types, GCP multi-region scopes), so the data exists and is not hypothetical. PR #2442 settled the mechanism for carrying AI pricing dimensions, so there is no open question about where this lands. Spec impact is one property row plus an appendix example, with no new columns, and a draft implementation exists in PR #2613.

Remaining complexity: the boundary against the SKU Price dataset is argued but not settled. PricingRegionId as drafted in PR #2424 MUST represent the provider's boundary "even when this represents a global or macro-region scope" and MAY be null "when the unit price applies globally or is not regionally scoped", so two generators can encode the same globally routed price differently and both conform. SkuPriceEligibility.IsGlobalScope uses "global" for a third, non-geographic meaning. Scoping the FR to the general dimension rather than the AI case settles the property name, but raises the boundary against the existing Redundancy property, which carries "Local", "Zonal", and "Global" and may already cover offerings that bundle replication with multi-region serving. PR #2613 draws that line in a footnote, which is thinner than the question now warrants. TF2 has taken no design decisions on this FR. The 2026-05-20 session decided to carve it out of FR #2018; the 2026-05-27 session received it for awareness and recorded no new decisions.

### FinOps Scope Alignment

- [x] Public Cloud - e.g., AWS, Azure, GCP, OCI
- [x] Software-as-a-Service (SaaS) - e.g., Salesforce, Snowflake
- [ ] Data Center - on-prem compute and infrastructure
- [ ] Licensing - subscription or usage-based licensing models *(under development)*
- [x] AI - cost and usage for AI models and platforms *(under development)*
- [ ] Custom - internal tooling, specialized infra *(under development)*

Matt-Cowsert:
## Summary

Implements FR #2414 by adding a FOCUS-defined `RegionScope` property to `SkuPriceDetails`, plus a worked appendix example. This follows the shape PR #2442 established for model identity: a row in the FOCUS-Defined Properties table and an appendix example backed by CSV datasets. No new columns.

> **Note:** The property was drafted as `ServingScope`. The rename to `RegionScope`, the appendix rewrite in pricing terms, and its supporting content land through #2667, which is stacked on this branch. This body describes the merged result.

**What changes**
* `SkuPriceDetails`: one FOCUS-defined property, `RegionScope`, added to the FOCUS-Defined Properties table with example values "Global", "Regional", and "DataZone", plus a footnote separating it from `Redundancy`. No change to the `SkuPriceDetails` requirements.
* Appendix: a new "Region Scope" example with two scenarios, backed by two CSV datasets.

## The boundary problem this has to solve first

FR #2414 parked itself on TF1 coordination. That coordination is now possible, because PR #2424 (Add `SKU Price` dataset) is open and defines both columns the FR named. Reading it changes the shape of the problem: there are now four unrelated uses of the word "global" in or near the specification, and any region scope proposal has to place itself among them.

| Construct | Where | What "Global" means there |
|---|---|---|
| `SkuPriceEligibility.IsGlobalScope` | SKU Price dataset (PR #2424) | The price applies to all entities without restriction. Not geographic at all. |
| `PricingRegionId` | SKU Price dataset (PR #2424) | The geographic boundary the rate card is valid for. |
| `Redundancy` | `SkuPriceDetails`, since v1.1 | How widely the SKU is replicated for durability or availability. |
| `RegionScope` | This PR | How broad an area the SKU price applies to. |

`IsGlobalScope` and `Redundancy` are separable on their definitions, and the footnote in this PR draws the `Redundancy` line explicitly. `PricingRegionId` is the one that needs an argument.

**Why `PricingRegionId` does not close the gap.** Two reasons, and the second is the stronger one.

First, placement. `PricingRegionId` is a SKU Price dataset column. Every success criterion in FR #2414 is asked of Cost and Usage data: filter and group cost by region scope, quantify the delta between scopes. Answering those from Cost and Usage alone is the point, and `PricingRegionId` is not available there.

Second, and independent of placement, `PricingRegionId` is not a reliable discriminator even after a join. Its requirements say it MUST represent the boundary the provider defines "even when this represents a global or macro-region scope", and its nullability rule says it MAY be null "when the unit price applies globally or is not regionally scoped". Both representations are conformant for the same globally priced rate. Two generators can therefore encode identical pricing differently and both pass. That is the cross-provider normalization gap FR #2414 set out to close, not a fix for it.

The same argument rules out inferring scope from `RegionId`. A null `RegionId` is also produced by services that are simply not regionalized, so absence cannot separate a globally priced charge from an unregionalized service, and presence cannot separate a global price from a regional one. Scenario A shows the second half: all four rows carry a `RegionId` of "us-east-1", the region the service provider anchors the offering to, and only `RegionScope` shows that two of them were priced at the global rate.

`RegionScope` and the region columns are orthogonal rather than competing. Scenario B makes that concrete: the data zone row carries a `RegionId` of "northeurope" and the regional row carries "westeurope". Both are single regions inside the area the service provider prices as one data zone, and the scope value is what says whether the price applies across that whole area or is pinned to one region inside it.

## Naming

The property is named for the dimension rather than for the AI case, so the value set can grow and other multi-region offerings can use the same key. This follows the reasoning applied in PR #2606, where naming the dimension meant future variants arrive as values rather than as new properties.

FR #2414 has since been generalized to match: its title, problem statement, use cases, and success criteria describe the dimension rather than an AI-only concern.

The generic name costs a wider boundary to defend, which is what the table above and the `Redundancy` footnote are for.

**Settled 2026-09-17: `RegionScope`.** @flanakin and @marc-perreaut flagged `ServingScope` on September 1, and the name went to Task Force 2 on September 2 and then to a naming research action item, #2672. The name is `RegionScope`, which that research recommends and which @marc-perreaut proposed independently on September 3. "Pricing" is not in the key because none of the other 17 `SkuPriceDetails` properties carries a pricing qualifier: `Redundancy`, `StorageClass`, and `OperatingSystem` all describe the SKU price without saying so, because the column already sets that context. A bare `Scope` was ruled out for a location property, and "geo" is reserved for a later geographic aspect of resource regions.

The rename, the appendix rewrite in pricing terms, and supporting content covering the full naming history and the relationship to `PricingRegionId` are staged on #2667, stacked on this branch, so the group reviews one diff and merges it in place. It goes to the Open Forum on September 18 for confirmation.

## Value set

Three example values rather than the binary pair FR #2414 originally scoped itself to. Azure ships data-zone deployments today, so a binary set would be visibly incomplete on arrival, and the appendix scenario needs a third value to show that scope is a spectrum rather than a switch. These are example values in the same sense as every other row in the table, not a normative enumeration, so adding one does not change how the table works. FR #2414's success criteria were updated to ask for example values rather than allowed values, so the FR and the table now agree on this too. Sovereign and similar scopes are deliberately left out until a provider prices them as a distinct point.

## Scope boundaries

* **Processing tier** (standard, batch, provisioned) qualifies a charge without describing geography. Azure bundles tier and scope into a single deployment-type string, which is an argument for keeping them as separate FOCUS properties rather than reproducing the composite.
* **Token type** is the subject of PR #2606. A row can carry both properties, but the two are not fully independent in practice: see the limitation noted under Verification below.
* **Provisioned throughput commitments** (#2371) are a purchasing construct rather than a routing one.
* **Latency and data-residency guarantees** are contract terms, not billing dimensions, and are not represented here. `RegionScope` supports the residency question by making pinned spend identifiable; it does not assert a guarantee.

## A limitation worth reviewing

Each of FR #2414's three use cases was run as a query against the example CSVs. Two are satisfied by this PR on its own. The third is not, and the appendix originally claimed otherwise.

**Reporting cost by region scope works.** Grouping the two CSVs by `RegionScope` alone yields $46.50 global, $22.14 regional, $15.84 data zone, across two providers, touching no other field.

**Identifying region-pinned spend works**, and the example now shows the case that makes it non-trivial. Scenario B carries a fourth row for a model priced without a scope distinction, so it has no `RegionScope`. Treating that row as unclassified gives a region-pinned share of 26.2%; folding it into the denominator gives 25.3%. The appendix states which is correct, and that an absent `RegionScope` does not mean a charge was globally priced.

**Comparing prices across providers needs one thing this PR does not deliver.** The appendix previously said the $2.50 and $2.40 global input token prices could be compared "with no parsing of service-provider-specific SKU identifiers". That was wrong. `RegionScope` and the model-identity properties establish that both rows describe global pricing of the same model, but nothing FOCUS-defined establishes that both are *input* token prices. `ConsumedUnit` is "Tokens" on every row, `PricingUnit` is "1000000 Tokens" on every row, and the `SkuPriceDetails` keys are identical between the input and output rows. `SkuMeter` is the only discriminator, and it is provider-specific free text ("Global Input Tokens" against "Glbl Std Inp Tokens").

The bullet has been corrected to state what the properties do establish and to name the metered dimension as the remaining interpretation step. The property that would close it is `TokenType` in PR #2606. That makes the relationship between these two PRs stronger than "orthogonal": for the cross-provider unit price comparison this appendix demonstrates, both properties are needed. Reviewers may reasonably prefer to sequence the two, or to accept the corrected wording and let `TokenType` land separately.

## Open questions

1. **Whether the boundary against `PricingRegionId` holds once PR #2424 settles.** The argument above is made against PR #2424 as currently drafted. If TF1 changes the nullability rule so that global pricing has one conformant representation, part of the second argument weakens; the placement argument does not.
2. **Whether the value set should ever become normative.** This PR treats the values as examples, consistent with every other row in the table. FR #2099 carries the same question for `TokenType`, and it is worth settling once for both rather than per property.
3. **Whether `SkuPriceDetails` should stay a flat key-value column.** FR #2414 records this as a tracking dependency, raised by Irena Jurica during TF2: whether the column needs to evolve toward a structured JSON-object representation like `ContractApplied`. It affects how this property is carried but is not a prerequisite for it.
## Sequencing and conflicts

* This is written against PR #2424, which is not merged. If that PR changes, question 1 above reopens.
* PR #2606 also adds a row and a footnote to `SkuPriceDetails`. The table rows do not collide (`RegionScope` sorts before `StorageClass`, `TokenType` after), but both PRs add a footnote numbered 4. Whichever merges second will need to renumber.

> Note: Requirements-model rules are not included; those are updated downstream per release.

## Verification

* `validate_includes.py`, the code-block alignment check, `markdown-pp` assembly, and the linter all completed without error. The appendix anchor resolves into the generated table of contents; the rename on #2667 takes it to `#appendix.examples:regionscope`, verified by `gen_pdf` on that PR.
* The 134 spec tests pass, including the `test_spec_links.py` rendered-anchor checks, which run against a locally built `spec.html`.
* Both CSVs were checked programmatically: every row satisfies `PricingQuantity` x `ListUnitPrice` = `ListCost`; `ListCost`, `ContractedCost`, `BilledCost`, and `EffectiveCost` agree on every row; `ConsumedQuantity` divided by 1,000,000 equals `PricingQuantity` on every row; `SkuPriceDetails` keys are alphabetical on every row; and `SkuPriceId` values are unique. Scenario totals are $36.00 and $51.48, the latter comprising $48.48 across the three scoped rows plus a $3.00 unscoped row.
* The price relationships stated in the appendix prose were checked against the data: 20% for region pinning in Scenario A, and 10% and 20% above global in Scenario B.
* Both CSVs use the same column set and ordering as the existing AI Model Identity examples, with `RegionId` and `RegionName` added because the region columns carry part of the modelling point.
* Both scenarios reuse existing fictitious data generators (Aura Web, CrestNode, Solora AI) and the existing fictitious customer (Acme Corp), so no addition to the appendix reference tables is needed and there is no overlap with PR #2594.

### Changelog
* **2026-09-17** (review): Settled the property name as `RegionScope` on #2667, per the Action Item #2672 naming research and @marc-perreaut's independent proposal of 2026-09-03. That PR also rewrites the appendix in pricing terms, gives the Scenario B data zone row a single-region `RegionId` of "northeurope" in place of the macro-region "eu", anchors the Scenario A globally priced rows to "us-east-1", qualifies the model-identity claim in the cross-provider bullet, and adds supporting content covering the naming history, the relationship to `PricingRegionId`, and what the two dimensions would enable on a rate card. This body now describes the merged result, and open question 4 is closed.
* **2026-09-02** (review): Proposed renaming `ServingScope` to `PricingRegionScope` on the naming threads from @flanakin and @marc-perreaut, pending the Task Force 2 discussion. Answered @ijurica's operational-framing thread with the same proposal; the footnote reorder is the fallback if the group keeps the operational frame. Closed @qquigley's thread, since the 2026-08-26 revision landed the applicability clause he asked for. Applied @jpradocueva's F-001 in 7527218c (second service provider link on the appendix's first bullet dropped to italics). Opened #2667 with the rename staged on top of that commit.
* **2026-08-26**: Moved the `ServingScope` population guidance out of the appendix and into footnote 4 on the property, following the shape footnote 3 already uses for `OperatingSystem`, after @qquigley flagged that the appendix line read as a rule with nothing defining it. The appendix line now describes what the scenarios show. Reconciled the `RegionId` convention between the two scenarios, which left the globally routed rows null in Scenario A and gave the data zone row an identifier in Scenario B without saying why. Linked `SKU Price` and `SKU` on their first occurrence in the appendix file.
* **2026-08-10**: Corrected the cross-provider comparison bullet, which claimed a parse-free comparison that the data does not support. Added an unscoped row to Scenario B so the example demonstrates the unclassified case rather than only describing it; Scenario B total moves from $48.48 to $51.48.
* **2026-08-10**: Generalized the appendix beyond AI to match the `ServingScope` property name and the generalized FR. Renamed `ai_serving_scope_examples.md` to `serving_scope_examples.md` and the data directory and CSVs to match, retitled the section to "Examples: Serving Scope", and moved it to its alphabetical position after "Examples: SaaS". Added guidance that an absent `ServingScope` does not imply global routing.
* **2026-08-09** (initial draft): `ServingScope` property on `SkuPriceDetails` with values "Global", "Regional", and "DataZone", plus the appendix example with two scenarios and two CSVs.

### Type of Change
* [ ] Bug fix
* [ ] Editorial / Formatting

### Author Checklist


Matt-Cowsert:
## TF2 Action Item - August 19, 2026

Draft status closes on August 27, 2026 for anything carried in the FOCUS 1.5 scope. @matt-cowsert said PR #2613 is not waiting on content and committed to marking it ready within the week.

ServingScope separates a request pinned to a region, which can carry a higher price, from a globally served request the service provider routes at its own discretion. Some service providers publish that distinction and some do not, and some customers treat geographic isolation as a requirement of their own. The change is scoped to the SkuPriceDetails properties, with the key defined by FOCUS and the values defined by each service provider.

@kk09v had the pull request open during the meeting.

**Scope:**
- [ ] Mark PR #2613 ready for review

**Assignee:** @matt-cowsert

specification/appendix/serving_scope_examples.md
```diff
@@ -0,0 +1,54 @@
...
+
+Serving scope is a property of the *SKU Price* rather than of the individual charge, because a [*service provider*](#glossary:service-provider) prices each scope separately. The dimension is not specific to AI; the same distinction applies wherever a *service provider* offers the same capability at more than one geographic breadth. AI model serving is where *service providers* price it most distinctly today, which is why both scenarios below use AI *SKUs*. Other offerings that vary geographic breadth commonly bundle it with replication, which the Redundancy property describes instead.
+
+ServingScope is populated when a *service provider* prices more than one serving scope for the same capability. Its absence does not indicate that a [*charge*](#glossary:charge) was globally routed, so an analysis measuring the share of spend at a given scope treats rows without ServingScope as unclassified rather than folding them into either scope.
```
qquigley: "ServingScope is populated when" reads as a rule, not an example. If it's a rule, I can't find it defined anywhere. I double checked 2442 to be sure.

Matt-Cowsert: Good catch Andrew. That sentence does state a population rule, and no such rule exists. The nearest thing in `skupricedetails.md` is the pair of requirements that turn on applicability, one MUST NOT for properties that are not applicable and one SHOULD for those that are, and neither defines what makes a property applicable.

Footnote 3 already puts an applicability condition on OperatingSystem, so the shape exists in that file. I'd add the same kind of clause to footnote 4 and rewrite the appendix line to describe what the scenarios show.

That still leaves one gap. A service provider pricing a single scope could populate "Global" and be conformant either way, so counting unscoped rows as unclassified is not guaranteed across data generators.

specification/datasets/cost_and_usage/columns/skupricedetails.md
```diff
@@ -58,12 +58,14 @@ The following keys should be used when applicable to facilitate cross-SKU and cr
...
 | StorageClass             | Class or tier of storage provided                                        | String           | Examples: "Hot", "Archive", "Nearline"                |
 
 Notes
 <br><sup>1</sup> In the case of "burstable" SKUs offering variable levels of performance, the baseline or guaranteed value should be used.
 <br><sup>2</sup> Memory manufacturers still commonly uses "GB" to refer to 2<sup>30</sup> bytes, which is known as GiB in other contexts.
 <br><sup>3</sup> This is the operating system family of the SKU, if it's included with the SKU or the SKU only supports one type of operating system.
+<br><sup>4</sup> ServingScope applies when a service provider prices more than one serving scope for the same capability. ServingScope describes the breadth of the geographic area a request may be served from, which is distinct from Redundancy. Redundancy describes how widely a SKU is replicated for durability or availability, whereas ServingScope describes where the work is carried out. A SKU may carry both.
```
ijurica, line 68: ServingScope is described in routing/execution terms throughout the footnote and appendix ("where the work is carried out", "a request may be served from", "routed across a broad geographic area"), but it also introduces a pricing-based trigger ("applies/populated when a service provider prices more than one serving scope").

Consider dropping the pricing framing and keeping the property purely operational — practitioners choose a serving scope based on data residency, latency, or compliance (Azure OpenAI's Global / Data Zone / Regional deployment types being the canonical AI example). 

This also avoids reopening the unresolved resource-region vs. billed/priced-region discussion in the RegionId supporting content (MeterRegion vs. ResourceLocation).

specification/appendix/serving_scope_examples.md
```diff
@@ -0,0 +1,54 @@
...
+
+In the scenarios below, ServingScope appears only on rows for a capability the *service provider* prices at more than one serving scope. An absent ServingScope does not indicate that a [*charge*](#glossary:charge) was globally routed, so an analysis measuring the share of spend at a given scope treats rows without ServingScope as unclassified rather than folding them into either scope.
+
+## Baseline Scenario
+
+The following conditions apply to the scenarios below:
+
+* Acme Corp runs the same foundation model, Solora Reasoning Pro, on two [*service providers*](#glossary:service-provider).
```
jpradocueva, line 13: ### F-001 — error — [Line 13](https://github.com/FinOps-Open-Cost-and-Usage-Spec/FOCUS_Spec/blob/f46659389ef837d33a9e9026439ba0de8481f85f/specification/appendix/serving_scope_examples.md#L13)

**Changed content:** `[*service providers*](#glossary:service-provider)`

**Rule:** `editorial-conventions-06` — Writing Specification Content & Review Guidelines › Editorial Conventions; bullet 6 of 13 ("Linking Rule"). "Link the first remaining occurrence in reading order. Leave all later occurrences unlinked."

**Engineering context:** Each glossary destination should be hyperlinked once per file — like a "no duplicate anchor" lint rule — with every later mention left as plain emphasized text.

**Problem:** Line 5 already links `[*service provider*](#glossary:service-provider)`, the first occurrence in reading order. This line links `[*service providers*](#glossary:service-provider)` again, to the same destination, with no applicable exception (no normative requirement bullets exist in this file).

**Impact:** Produces a redundant hyperlink in the built HTML/PDF output that this file's own convention otherwise follows consistently for every other repeated term.

**Suggested correction:** Drop the link, keep the italic emphasis.

```suggestion
* Acme Corp runs the same foundation model, Solora Reasoning Pro, on two *service providers*.
```

jpradocueva (commented):

Snapshot totals: 3 conversation comments, 3 reviews, and 3 inline review comments.

| --- | --- | --- |
| F-001 | Submit inline comment | Duplicate glossary link to an already-linked anchor (#glossary:service-provider); removing the duplicate link does not change normative behavior. |

specification/appendix/serving_scope_examples.md
```diff
@@ -0,0 +1,54 @@
+# Examples: Serving Scope
+
+The following examples illustrate how a Cost and Usage [*FOCUS dataset*](#glossary:FOCUS-dataset) distinguishes serving that is routed across a broad geographic area from serving that is pinned to a narrower one, using the FOCUS-defined ServingScope property of [SkuPriceDetails](#datamodel.costandusage.skupricedetails). Provider, model, and region names below are illustrative.
```
flanakin, line 3: I'm finding "serving" as a strange word to use in this case and it doesn't feel exceedingly clear to me here.

specification/appendix/serving_scope_examples.md
```diff
@@ -0,0 +1,54 @@
...
+
+Serving scope is a property of the [*SKU Price*](#glossary:sku-price) rather than of the individual charge, because a [*service provider*](#glossary:service-provider) prices each scope separately. The dimension is not specific to AI; the same distinction applies wherever a *service provider* offers the same capability at more than one geographic breadth. AI model serving is where *service providers* price it most distinctly today, which is why both scenarios below use AI [*SKUs*](#glossary:sku). Other offerings that vary geographic breadth commonly bundle it with replication, which the Redundancy property describes instead.
```
flanakin, line 5: nit: Are we concerned with the use of "scope" here relative to how FinOps Framework defines scopes? It's a valid word, but just not used the way the Framework uses it, so figured I'd ask.

marc-perreaut, line 5: The term "serving scope" might be too generic.
When reading the full PR, I understand that serving scope is a routing property, which can be either global or regional.
What about using simply the name "routing", instead of "serving scope"?

ijurica (approved):

specification/appendix/serving_scope_examples.md
Matt-Cowsert: Andrew, this landed in f4665938 on August 26: footnote 4 on the property now carries the applicability clause, and this line describes what the scenarios show rather than stating a rule. Resolving on that basis; reopen if the footnote wording does not do it for you.

specification/datasets/cost_and_usage/columns/skupricedetails.md
Matt-Cowsert: Irena, the pricing trigger stays, and the proposed revision makes the name match it: PricingRegionScope, with the appendix rewritten in pricing terms. A rate card shows which price point applied and not where the request ran, so the pricing trigger is checkable against the price list and an operational name is not. On RegionId, the scope says how broad the priced area is, not which region did the work, so the MeterRegion versus ResourceLocation question stays where it was. On the Task Force 2 agenda today; if the group prefers the operational frame, the footnote reorder you asked for is the fallback.

specification/appendix/serving_scope_examples.md
Matt-Cowsert: Agreed, accepting the suggestion. Line 5 carries the first service provider link, so this one drops to plain emphasis.

specification/appendix/serving_scope_examples.md
Matt-Cowsert: Michael, scope stays in the proposed revision. The specification already uses it for geographic breadth: the PricingRegionId requirements on #2424 say "global or macro-region scope" and "regionally scoped", and commitment applicability carries IsGlobalScope. The Framework's Scopes are spending segments aligned to business constructs, a different sense of the word rather than a collision.

specification/appendix/serving_scope_examples.md
Matt-Cowsert: Michael, proposed revision: PricingRegionScope. The property sits on the SKU price and populates only when a provider prices more than one scope, and billing data shows which price point applied rather than where the work ran, so the name follows the data. It pairs with PricingRegionId on #2424: which region, versus how broad. On the Task Force 2 agenda today; the rename and the appendix rewrite land as one commit if the group agrees.

specification/appendix/serving_scope_examples.md
Matt-Cowsert: Marc, proposed revision: PricingRegionScope rather than Routing. Billing data shows which price point applied, not where a request was routed, so the name follows the price, and the appendix gets rewritten in those terms. A bare Routing key would also read as a network property on a networking SKU. RoutingScope is the runner-up if the group prefers the operational frame. Both go to the Task Force 2 call today.

Matt-Cowsert:
The rename is staged as #2667, stacked on this branch, so the Task Force 2 call can review it as one diff and merge it with one click if PricingRegionScope carries. If the room prefers the operational frame, the same branch takes RoutingScope with a smaller prose pass.

specification/appendix/serving_scope_examples.md
marc-perreaut: Yes, thanks, the term 'PricingRegionScope' clarifies better.
What about simply 'RegionScope' or just 'Region', as this property is already in the context of pricing?

jpradocueva (approved):
The content passes AGENTS validation.

specification/appendix/serving_scope_examples.md
udam-stitcher: If someone looks at a key (or even a column) and has no idea what it or the values represent, I think we should go back and reevaluate the naming and whether the key/column is appropriate in the spec. I think we're at risk of that with either name. Serving and scope are such broad terms that they could mean so many things. That said, I don't have a great naming suggestion here either. Seems we're talking about the functionality and specifically as it relates to geography. How about GeographicScope?

specification/datasets/cost_and_usage/columns/skupricedetails.md
udam-stitcher: Feels like we should avoid the 'request'-related language here. A service that you pay for doesn't have to be request based.. IAM and CloudFront, for example, are your standard 'infra & services' type SKUs that are 'global'-scoped.

specification/appendix/serving_scope_examples.md
mike-finopsorg: The fact we use geographic area as the descriptor makes me think @udam-stitcher has it right here " that is routed across a broad geographic area from serving that is pinned to a narrower one," GeographicScope feels like it pairs with that description.

specification/appendix/serving_scope_examples.md
mike-finopsorg: Good question to ask ourselves, I would be cautious of a "naked" scope (i.e. Scope), but a named scope feels acceptable (i.e. GeographicScope).

Matt-Cowsert: RegionScope and Region are both on the candidate list for the property name, Marc, and they go to the September 9 Task Force 2 call with the rest of it: PricingRegionScope, PricingRegion, PricingRegionId, RegionScope, Region, and GeographicScope, which Udam Dewaraja raised on September 7 and Mike Fuller backed on September 8.

Two points from the September 2 call bear on the short forms. Michael Flanakin wants "region" paired with a word saying what the region is for, so a bare Region sitting next to RegionId would have to carry that distinction on its own. Dave Moreau ruled out a bare "scope" for a location property, which lands on Scope alone rather than on RegionScope.

Michael holds the naming research as action item #2672.

specification/datasets/cost_and_usage/columns/skupricedetails.md
Matt-Cowsert: Agreed on the request framing in footnote 4, and the rename staged on #2667 already drops it. The footnote here says the property describes the breadth of the geographic area a request may be served from. The staged version says the breadth of the geographic area a SKU price applies to, which covers IAM and CloudFront, since neither is metered per request. The appendix prose moves the same way: the Scenario A bullet on RegionId now reads in terms of the price rather than the request.

The last routing language was the ChargeDescription text in the Scenario A CSV ("Globally routed input tokens" and "Globally routed output tokens"). Those are illustrative descriptions rather than FOCUS-defined values, and they now read "Globally priced" on #2667.

One other point in this footnote for the Task Force 2 call. It says the property applies when a service provider prices more than one scope for the same capability. The September 2 call recorded that population follows the same SHOULD every FOCUS-defined property carries, so a service provider may populate it where both scopes carry the same price. #2667 drops that clause so the footnote matches the record.

specification/appendix/serving_scope_examples.md
Matt-Cowsert: GeographicScope goes on the candidate list for the property name, and the current wording supports it. The property row in skupricedetails.md reads "Geographic scope across which the SKU is served", and footnote 4 describes the breadth of a geographic area.

The room has to reconcile one thing first. On the September 2 call Irena Jurica ruled out "geo" in the name, on the grounds that a later geographic aspect of resource regions may need the word, and that is the word GeographicScope takes. Either the reservation moves or the name does, and the group makes that call rather than this thread.

The full list going to the September 9 Task Force 2 call is PricingRegionScope, PricingRegion, PricingRegionId, RegionScope, Region, and GeographicScope. Michael Flanakin holds the naming research as action item #2672.

Udam, on the broader point about a reader not knowing what a key represents: the values are "Global", "Regional", and "DataZone", so a reader who lands on the key sees the axis from the values. What the name has to do is put that axis in the key itself, and region and geographic both do.

specification/appendix/serving_scope_examples.md
Matt-Cowsert: A named scope rather than a bare one is where the September 2 Task Force 2 call landed too, Mike. Dave Moreau ruled out a bare "scope" for a location property, and no candidate on the list uses Scope on its own: PricingRegionScope, PricingRegion, PricingRegionId, RegionScope, Region, and GeographicScope.

On the Framework question Michael Flanakin opened this thread with, the specification already uses scope for geographic breadth. The PricingRegionId requirements on #2424 say the column represents the provider's boundary "even when this represents a global or macro-region scope", and SkuPriceEligibility on the same pull request carries IsGlobalScope for a price that applies to all entities. The Framework's Scopes are spending segments aligned to business constructs, so the two senses sit apart rather than collide.

Michael holds the naming research as action item #2672.

flanakin:
🤖 [AI][Claude] **Naming research for the pricing region property (AI #2672)**

Recommendation: **`RegionScope`**. Alternatives and the reasoning behind each ruling are below, along with two findings that sit underneath the name question.

## Vendor terminology

The call asked for a check of the words vendors actually use for this axis.

| Source | Term for the axis | Values |
|---|---|---|
| [AWS Bedrock](https://docs.aws.amazon.com/bedrock/latest/userguide/inference-profiles-support.html) | "Geo scope" | Global, US, EU (profiles: In-Region, Geo, Global) |
| [Azure AI Foundry](https://learn.microsoft.com/azure/ai-foundry/openai/how-to/deployment-types) | "deployment type" | Global, Data zone, Azure geography |
| [Google Cloud](https://cloud.google.com/storage/docs/locations) | "location type" | region, dual-region, multi-region, zone |
| FOCUS, `supporting_content/datasets/cost_and_usage/columns/regionid.md` | "Region Category" | single-region, multi-region, global |

Three of the four name the axis with a classifier word (scope, type, category). None uses "serving". FOCUS sketched this dimension during the RegionId work and called it Region Category, which is the closest thing to prior art in the repository.

## Why `RegionScope`

* **"Scope" is the word AWS uses for this axis**, and #2424 already uses it geographically: PricingRegionId MUST represent the boundary "even when this represents a global or macro-region scope".
* **It survives a null `RegionId`.** Half the rows in Scenario A carry no region. `RegionScope` = "Global" reads correctly with nothing in `RegionId`; a name built on "type" or "category" implies a region exists to classify.
* **It pairs with `RegionId` rather than competing with it.** The value's job is to say what kind of thing `RegionId` holds: "eu" is a data zone, "westeurope" is a region. The shared stem signals that the two are read together.
* **It answers the criterion recorded on the September 2 call** — region paired with a word saying what the region is for — without putting `Region` in the value-holder position.
* **It is a named scope**, not a bare one, so the ruling against a bare "Scope" for a location property is satisfied.

## Ruled out

**`PricingRegionId`** — this is a collision rather than a weaker option. #2424 defines `PricingRegionId` as a service-provider-assigned identifier holding values like "us-east-1". A `SkuPriceDetails` key of the same name holding "Regional" puts two different semantics on one name across a joinable row. Recommend removing it from the list.

**`PricingRegion`, `PricingRegionScope`** — none of the 17 existing `SkuPriceDetails` keys carries a pricing qualifier. `Redundancy`, `StorageClass`, and `OperatingSystem` all describe the SKU price without saying so, because the column already does. The prefix is redundant inside the column and misleading outside it, and `PricingRegion` additionally reads as the display-name pair to `PricingRegionId`.

**`GeographicScope`, `GeoScope`** — three problems, and the reservation raised on the September 2 call is the smallest of them.

1. In AWS's own taxonomy "Geo" is the **middle tier**, not the axis: the docs list "In-Region, Geo, and Global inference profiles". Azure does the same — its narrow tier is the "Azure geography" deployment type. Naming the axis after one of its own values means `GeoScope` = "Global" and `GeoScope` = "Regional" are both non-geo geo-scopes.
2. AWS's Geo scope values are Global, US, and EU — named geographies, not breadth levels. Borrowing the name imports a value expectation this property does not meet: a practitioner who knows Bedrock would expect "EU" and get "DataZone".
3. `specification/attributes/focus_column_handling.md` requires that a FOCUS column "MUST NOT use abbreviations other than `Id` in the Column ID". "Geo" is a truncation. `Iops` and `Gpu` in the key table are established acronyms, which is a different case.

**`Region`** — collides with `RegionId` and `RegionName` on the same row and carries no indication of what it holds.

**`ServingScope`** — no vendor uses "serving" for this axis, which is consistent with the reaction on the review threads.

## Alternatives worth keeping on the list

**`CoverageRegion`** and **`OperatingRegion`** — both are explicit about what the region is for, which is the right instinct, but both put `Region` in the value-holder position while the values are breadth levels. `CoverageRegion` = "DataZone" does not parse, since a data zone is not a region, and "coverage" is a load-bearing FinOps term already (`CommitmentCoverageRate`, coverage rate, uncovered spend) on the same dataset. `OperatingRegion` additionally restates RegionId's own definition, which says RegionId is non-null "when a resource or service is operated in or managed from a distinct region".

Both become viable if the value set changes to named geographies — see the first finding below.

**`RegionGranularity`** — no FinOps or Framework overload at all, and literally what the property measures. This is the answer if the concern about "scope" being an overloaded word is the deciding factor. Less natural to say.

**`RegionType`** — matches Google Cloud's "location type" directly, and `*Type` correctly signals an open value set in FOCUS (`DiskType`, `InstanceType`). Softer than `RegionScope` on rows where `RegionId` is null.

**`RegionCategory`** — matches FOCUS's own prior art and the strongest grammar precedent, but every `*Category` column in FOCUS carries "MUST be one of the allowed values". The name would promise a closed enumeration this PR deliberately does not define (open question 2). This becomes the right name if the value set ever becomes normative.

## Two findings underneath the name

**1. The value set is Azure-shaped, and that decision should precede the name.**

"DataZone" is Azure's word. AWS's middle tier is `us.` / `eu.`, which are geographies, and Google Cloud's is `multi-region` / `dual-region`. Mapping `eu.anthropic.*` onto "DataZone" uses a term AWS never uses and discards the fact that the geography is specifically the EU.

There are two coherent value sets, and they lead to different names:

| Shape | Values | Fits | Name follows |
|---|---|---|---|
| Breadth levels (this PR) | Global, Regional, DataZone | Azure | `RegionScope` |
| Named geographies | Global, US, EU | AWS | a region-valued name |

The name is currently being chosen against a value set that has not been stress-tested against AWS data. Worth settling first.

**2. The value overlap with `Redundancy` is sharper than the definitional overlap.**

`Redundancy` example values are "Local", "Zonal", "Global". This property's are "Global", "Regional", "DataZone". Two keys on one row can both read "Global" meaning different things. Footnote 4 separates the definitions but not the values. The two ladders should either align or differ visibly.

## An adjacent gap this surfaced

`RegionId` is defined as where a resource is "operated in or managed from". Nothing in FOCUS carries where a service *delivers* — CDN edge geography, or Azure's meter region (`x_SkuMeterRegion`), which the action item named as an input. That is a genuine gap and a separate FR.

It is also the reason to keep `*Region` names free: `CoverageRegion` and `OperatingRegion` are the natural names for that property, and spending one of them on a breadth classifier leaves the real region-valued property without a name.

specification/datasets/cost_and_usage/columns/skupricedetails.md
Matt-Cowsert: Irena, you flagged footnote 4 for describing the property in routing terms while triggering it on pricing. The version merged from #2667 on September 23 is pricing only: it describes the breadth of the geographic area a price applies to, the trigger clause is gone, and no routing or request language remains in the footnote, the appendix, or the CSVs. RegionId keeps its definition, so the MeterRegion versus ResourceLocation question stays where it was.

specification/appendix/serving_scope_examples.md
Matt-Cowsert: Michael, you found "serving" a strange word for this property, and the name no longer uses it. #2667 renamed the property RegionScope, as your #2672 research recommended, rewrote the appendix in pricing terms, and merged into this branch on September 23. The new supporting content records GeographicScope, which Udam raised here and Mike backed, as set aside because "geo" is reserved for a later geographic aspect of resource regions.

specification/appendix/serving_scope_examples.md
Matt-Cowsert: Michael, you asked whether "scope" collides with how the FinOps Framework uses the word. The name merged from #2667 on September 23, RegionScope, uses it only as a named scope, the form Mike found acceptable here. The two senses stay apart: RegionScope names geographic breadth, and the Framework's Scopes are spending segments.

Matt-Cowsert: Marc, you flagged serving scope as too generic, and the property now carries the name you proposed on September 3, RegionScope. #2667 merged the rename into this branch on September 23, and the #2672 naming research recommended the same name. The appendix and footnote 4 now describe pricing rather than routing, and a bare Region was set aside because it would sit beside RegionId and RegionName on the same row without saying what it holds.

jpradocueva (approved):

Validation completed for PR head `4e5cfa883c1652d23b262f5b9d8cbdc6b4b62fb7` using `updates-agents` at `3c12b80af414838dfaa315fe4532cbea88a64d3b`.

No errors or warnings were identified within the supported Markdown validation scope.

Approval is not recommended because 2 changed file(s) were unsupported.

Content-validated documents: 5
Unsupported files: 2
Incomplete documents: 0
Errors: 0
Warnings: 0
Suggestions: 0
Unresolved claims: 0
Evidence complete: true

### Pre-existing Issues are reported separately to AI #2657

PR #2615, opened 2026-08-11
AI #2588: Define null semantics for the SKU Price effective date columns

### LINKED ISSUE #1057: [FR] Separate dataset for SKUs

### LINKED ISSUE #2588: [AI] Build a null-value semantics table for the SKU Price effective date columns

Matt-Cowsert:
## Summary

Defines what a null `SkuPriceEffectiveStart` and a null `SkuPriceEffectiveEnd` mean for a price lookup, and adds the worked data behind them. Addresses #2588. Based on `2364-sku-price-dataset`, so it merges into #2424 rather than into `working_draft`.

The two columns only carry meaning as a pair, and the spec did not state what each null combination meant. Issue #2588 records the concern that a reader looking at either column in isolation will query the dataset incorrectly, and that provider-to-provider differences in how the columns get populated make those queries non-portable.

### New appendix section: Effective Period Boundaries

A table covering the four combinations of null start and null end. For each one it states the validity period, when a charge resolves to that record, and where the record can sit relative to its neighbors.

That last point is a consequence of the non-overlap requirement already in `dataset.md`. A record unbounded on both sides is the only record its key combination can hold, since any second record would overlap it. A record unbounded at the start must be the earliest, and one unbounded at the end must be the latest. The null pattern is therefore not a free modeling choice; it fixes the record's position.

The section also states the lookup rule. Comparing a charge period start against a null boundary yields an undefined result rather than a match, so writing the interval test as a single pair of comparisons silently drops every record that is unbounded on either side. Each side needs its own test. `NullHandling` rules out placeholder timestamps, so this has to be handled in the query rather than in the data.

### Four new SKUs on the Aura Web rate card

The rate card goes from 17 records across 6 SKUs to 25 across 10, and `sku_price_boundaries.csv` and `sku_price_promotional.csv` extracts are added. Every added record carries a null `ContractId`, so its `ContractedUnitPrice` equals its `ListUnitPrice`.

* A burstable virtual machine whose rate change is announced ahead of taking effect. The record in effect today carries a null start and a populated end, and the record carrying a null end does not begin until 2027. Selecting the row with a null end to read the current price returns the wrong rate.
* A legacy object storage tier with no published origin date, scheduled for withdrawal. Its earlier record carries a null start, and its final record ends with nothing following it.
* A shared-core virtual machine offered at one flat rate since the catalog began, with no start date and no end date. Its single record is unbounded on both sides, the one combination the other SKUs do not exercise.
* A managed database with a promotional rate for the first quarter of 2026. Three records describe one price point: the standing rate, the bounded promotional rate, and the resumed standing rate. The promotional record is closed on both sides and sits between two records that are each open on one side.

Together these cover the announced-price-change, deprecation, temporary-price, and always-applied scenarios, and put all four combinations of null start and null end into the data, including the record unbounded on both sides.

### New appendix section: Temporary Pricing

A three-record chain showing a promotional rate that applies for a bounded interval and then reverts. It separates two shapes that are easy to conflate: a closed interval is a temporary price, while a null start is a price that has always applied.

### Column requirement changes

`SkuPriceEffectiveEnd` anchored its nullability on whether a price was currently active. A price published ahead of taking effect is not currently active, and has not expired, been superseded, or been scheduled for termination, so it fell outside both bullets and its end was unconstrained. Nullability is now anchored on whether an upper boundary exists, which covers the future-dated case and separates a null end from the question of which price is currently in effect.

`SkuPriceEffectiveStart` described null two ways: as a date the provider did not specify in its requirement, and as unbounded applicability in its description. Those read differently and produce different query results. Its nullability is now a MUST pair matching `SkuPriceEffectiveEnd`, with null meaning no lower boundary.

The chronology requirement is now scoped to records where both timestamps are populated. It previously fired whenever the end was populated, including against a null start, which is not a comparison that can be evaluated.

### Note for reviewers

`SkuPriceEffectiveStart` moves from `MAY be null` to a MUST/MUST NOT pair. This is a strengthening rather than a clarification. Under the current text a data generator can populate or omit the column at its discretion; under the proposed text, omission is required when the price has no lower boundary. This is the piece to settle in the next Task Force 1 session.

### Changelog

* 2026-08-24: Merged `2364-sku-price-dataset` again to pick up seven further commits on the base branch. The only conflict was the record-count sentence in the examples overview, which both branches had revised: this branch to 25 records across ten SKUs, the base branch to 18 across six after adding a Credit charge category record. The merged rate card carries 26 records across ten SKUs, because the credit record reuses the `AURAWEB-PLATFORM-CREDITS` SkuId and so adds a record without adding a SKU. The scenario bullet list merged without conflict, keeping this branch's four new bullets and the base branch's revised consumption-currency bullet. The base branch also corrected the flexible spend plan recurring fee from 16877.28 to 17183.04; that record is not among the eight added here, and no text in this pull request quotes the figure.
* 2026-08-24: Merged `2364-sku-price-dataset` to pick up the `working_draft` catch-up and the base branch's review-suggestion commit. None of the seven files this pull request changes were touched, so the appendix text, the rate card, and the two extracts are unchanged.
* 2026-08-19: Merged `2364-sku-price-dataset` to clear the rate card conflict. The base branch made `ContractedUnitPrice` non-nullable and equal to `ListUnitPrice` on records with a null `ContractId` under #2619, which populated the column on the 15 uncontracted records it already had. The eight records added here now carry it the same way, so the rate card is 25 records with `ContractedUnitPrice` populated on the 23 that carry no contract. `sku_price_boundaries.csv` and `sku_price_promotional.csv` were updated to match; they did not surface as conflicts because the base branch never saw them. All nine extracts remain exact subsets of the full rate card. No specification text conflicted.
* 2026-08-18: Added a both-null example (shared-core virtual machine: one record, null start and null end) so all four combinations in the null value semantics table appear in the sample data. Rate card is now 25 records across ten SKUs.
* 2026-08-18: Related the non-overlap combination to the composite key (the same members without SkuPriceEffectiveStart) in the Effective Period Boundaries note.
* 2026-08-18: Conditioned SkuPriceEffectiveStart nullability on the record state rather than provider behavior, matching the SkuPriceEffectiveEnd pair. Set the both-sided boundaries-table cell to "Any position", since a closed interval can be the last record for its SKU (the legacy tier).
* 2026-08-17: Added the three-record temporary-pricing example (managed database: standing, promotional, resumed), reconciled the null-start narrative across the boundaries extract and dependent text, and applied code-review suggestions.

### Type of Change
* [ ] Bug fix
* [ ] Editorial / Formatting

### Author Checklist


Matt-Cowsert:
## TF1 Action Item - August 11, 2026

The SKU Price effective date columns moved from a MAY-be-null formulation to explicit MUST and MUST NOT requirements. Riley Jenkins raised the case that motivated the change: distinguishing a SKU that has existed for the life of the catalog from one introduced at a known point in time.

Matt Cowsert noted on the call that the sample data already exists but has not been read against the tightened language, so the published examples may not exercise every nullability case the requirements now describe. He committed to taking this PR out of draft before the next TF1 on 18 August 2026.

**Scope:**

- [ ] Read the SKU Price effective date sample data against each nullability requirement in the null value semantics table
- [ ] Add or adjust sample rows so every MUST and MUST NOT case is represented
- [ ] Take this PR out of draft

**Assignee:** @Matt-Cowsert

specification/appendix/sku_price_examples/price_effectivity_and_currency.md
```diff
@@ -16,6 +16,32 @@ Note the following details in the example dataset:
 * Both records carry the same [SkuPriceCreated](#datamodel.skuprice.skupricecreated) of 2025-06-15T09:12:00Z and the same [SkuPriceLastUpdated](#datamodel.skuprice.skupricelastupdated) of 2025-12-18T16:40:00Z. These describe when the catalog record was written and last modified, which is a different question from when the price applies. The 2026 rate was published in December 2025, ahead of taking effect.
 * A charge in the [Cost and Usage](#datamodel.costandusage) dataset resolves to one of these records by comparing its charge period start against the effective interval. Retaining superseded records is what makes a historical charge resolvable after a price change.
 
+## Effective Period Boundaries
+
+The two effectivity columns carry meaning as a pair. Either may be null, and a null states that the interval is unbounded on that side, not that the boundary is unknown. Reading either column on its own, or treating a null as missing data, resolves a charge to the wrong price.
```
rileyjenk: I am not sure what that last phrase is saying. 
```suggestion
The two effectivity columns carry meaning as a pair. Either may be null, and a null states that the interval is unbounded on that side, not that the boundary is unknown. Reading either column on its own, or treating a null as missing data, may result in misinterpretation of the row.
```

specification/appendix/sku_price_examples/price_effectivity_and_currency.md
```diff
@@ -16,6 +16,32 @@ Note the following details in the example dataset:
...
+| Null | Populated | Up to the end timestamp | The charge period start is before the end timestamp | The first record |
+| Populated | Populated | A closed interval | The charge period start is at or after the start timestamp and before the end timestamp | An interior record |
+
+Aura Web offers a burstable virtual machine whose rate changes at the start of 2027, and a legacy object storage tier it is withdrawing.
+
+[**CSV Example**](/specification/data/sku_price_examples/sku_price_boundaries.csv)
+
+Note the following details in the example dataset:
+
+* The burstable virtual machine carries two records. The one in effect today runs from 2026-01-01 to 2027-01-01 and carries a populated SkuPriceEffectiveEnd. The one carrying a null SkuPriceEffectiveEnd does not begin until 2027-01-01. Selecting the record with a null end to read the current price returns the announced rate of 0.088000 rather than the rate of 0.096000 actually in effect. The current price is the record whose interval contains the moment being asked about, which is a different test.
```
rileyjenk:
```suggestion
* The burstable virtual machine carries two records. One record specifies the rate prior to the rate change and therefore has a SkuPriceEffectiveStart of null and a SkuPriceEffectiveEnd of 2027-01-01. This row has a rate of 0.096000.  The second record specifies the rate  after the price change. It's SkuPriceEffectiveStart is 2027-01-01 and its SkuPriceEffectiveEnd is null. Its rate is set to 0.088000. 
```

specification/data/sku_price_examples/aura_web_rate_card_full.csv
```diff
@@ -6,6 +6,8 @@ Usage,auraweb:contract::ACCT-123456789012:agreements/ctr-7f3a91b2c4d5,0.326400,0
...
 Usage,,,0.142000,USD,Payable,us-east-1,Aura Web Compute,Hours,,,,0,Aura Web,AURAWEB-COMPUTE-VM-STD-8X32,2025-12-18T16:40:00Z,"Standard VM 8 vCPU 32 GB, Linux, 1-year reservation rate (partial upfront)",,2026-01-01T00:00:00Z,"{""IsGlobalScope"":true}",AURAWEB-USEAST1-COMPUTE-USAGE-COMMITTED-PARTIAL-UPFRONT,2025-12-18T16:40:00Z
 Usage,,,0.000000,USD,Payable,us-east-1,Aura Web Compute,Hours,,,,0,Aura Web,AURAWEB-COMPUTE-VM-STD-8X32,2025-12-18T16:40:00Z,"Standard VM 8 vCPU 32 GB, Linux, 1-year reservation rate (all upfront)",,2026-01-01T00:00:00Z,"{""IsGlobalScope"":true}",AURAWEB-USEAST1-COMPUTE-USAGE-COMMITTED-ALL-UPFRONT,2025-12-18T16:40:00Z
+Usage,,,0.096000,USD,Payable,us-east-1,Aura Web Compute,Hours,,,,0,Aura Web,AURAWEB-COMPUTE-VM-BURST-2X8,2025-12-18T16:40:00Z,"Burstable VM 2 vCPU 8 GB, Linux, on-demand",2027-01-01T00:00:00Z,2026-01-01T00:00:00Z,"{""IsGlobalScope"":true}",AURAWEB-USEAST1-COMPUTE-ONDEMAND-BURST,2026-07-15T10:22:00Z
```
rileyjenk:
```suggestion
Usage,,,0.096000,USD,Payable,us-east-1,Aura Web Compute,Hours,,,,0,Aura Web,AURAWEB-COMPUTE-VM-BURST-2X8,2025-12-18T16:40:00Z,"Burstable VM 2 vCPU 8 GB, Linux, on-demand",2027-01-01T00:00:00Z,,"{""IsGlobalScope"":true}",AURAWEB-USEAST1-COMPUTE-ONDEMAND-BURST,2026-07-15T10:22:00Z
```

rileyjenk: Corresponding update to the example csv.

specification/appendix/sku_price_examples/price_effectivity_and_currency.md
rileyjenk: I think that we likely should add another example that is both the beginning rate, a rate that applied for a discreet amount of time and then a rate that applied after the discreet period. That would explain how a temporary price would look.

rileyjenk (changes requested):

Matt-Cowsert:
@rileyjenk , the temporary-price example you asked for is in, as a new Temporary Pricing section. A managed database carries three records: a standing rate, a promotional rate closed on both sides for the first quarter of 2026, and the resumed standing rate. It also separates the two shapes you raised on the call, since a rate that has always applied carries a null start while a temporary one is a closed interval.

Tracing the rest of your review:

* Your burstable CSV suggestion went into the full rate card, but the `sku_price_boundaries.csv` extract had not picked it up, so the two files disagreed on the start value. Both match now.
* Your burstable bullet is in, with the trailing space removed and "therefore" dropped, since being the pre-change rate does not on its own imply a null start.
* Your rewrite of the paired-columns paragraph is in as written.

Two things changed outside your comments. The boundaries table said a record populated on both sides is an interior record, which the legacy tier's final record contradicts, so that cell now reads "Any position". And the `SkuPriceEffectiveStart` MUST NOT bullet now conditions on the record rather than on the *service provider* designating a timestamp, so it reads as a pair with `SkuPriceEffectiveEnd`.

I rebased onto `2364-sku-price-dataset`, so your threads point at old anchors.

The `SkuPriceEffectiveStart` move from MAY-be-null to a MUST pair is still the piece I want to settle at Task Force 1. That one is a strengthening rather than a clarification.

specification/datasets/sku_price/columns/skupriceeffectiveend.md
```diff
@@ -11,10 +11,9 @@ SkuPriceEffectiveEnd MUST adhere to the following requirements:
 * SkuPriceEffectiveEnd MUST be of type Date/Time.
 * SkuPriceEffectiveEnd MUST conform to [DateTimeFormat](#attributes.date/timeformat) requirements (e.g., UTC).
 * SkuPriceEffectiveEnd MUST adhere to the following nullability requirements:
-  * SkuPriceEffectiveEnd MUST be null when the unit price is currently active and does not have a defined expiration or deprecation timestamp.
-  * SkuPriceEffectiveEnd MUST NOT be null when the unit price has expired, been superseded, or has a scheduled termination timestamp.
-* When SkuPriceEffectiveEnd is not null, SkuPriceEffectiveEnd MUST adhere to the following requirements:
-  * SkuPriceEffectiveEnd MUST be chronologically greater than [SkuPriceEffectiveStart](#datamodel.skuprice.skupriceeffectivestart).
+  * SkuPriceEffectiveEnd MUST be null when the unit price is applicable without an upper time boundary.
+  * SkuPriceEffectiveEnd MUST NOT be null when the unit price has a designated timestamp beyond which it is no longer applicable.
+* When SkuPriceEffectiveStart and SkuPriceEffectiveEnd are both not null, SkuPriceEffectiveEnd MUST be chronologically greater than [SkuPriceEffectiveStart](#datamodel.skuprice.skupriceeffectivestart).
```
Matt-Cowsert, line 16: ## TF1 Action Item - August 25, 2026

This pull request defines what each combination of a populated or null `SkuPriceEffectiveStart` and `SkuPriceEffectiveEnd` means, covering both null, either one null, and both populated. Riley Jenkins asked for that full range and Matt Cowsert said he has applied nearly all of it.

Nothing in the pull request prevents two rows sharing the same grouping columns from carrying overlapping date ranges. The greater-than and less-than requirement constrains a single row, not the relationship between rows. Irena Jurica raised the gap in the Task Force 1 meeting on 2026-08-25, said she tried to write a static normative requirement for it and could not, and said she would propose one expressed dynamically instead.

The constraint spans rows rather than sitting inside one, which the requirements model does not represent today and which the FOCUS Validator does not check. Matt Cowsert noted the 1.4 precedent: the group agreed then to keep queries that detect inconsistent data out of the specification body, and he is open to that kind of check living in supported features or in training material rather than the appendix.

**Scope:**
- [ ] Propose a dynamic normative requirement preventing overlapping effective date ranges for rows sharing the same grouping columns
- [ ] Name the column set the non-overlap constraint groups by

**Assignee:** @ijurica

flanakin (approved):

specification/datasets/sku_price/columns/skupriceeffectiveend.md
mike-finopsorg:
```suggestion
* When SkuPriceEffectiveStart is not null and SkuPriceEffectiveEnd is not null, SkuPriceEffectiveEnd MUST be chronologically greater than [SkuPriceEffectiveStart](#datamodel.skuprice.skupriceeffectivestart).
```

I don't like the contracted condition, I think it is clearer to have both not null conditions stated to make it clean and clear what the situation is.

shawnalpay:
@Matt-Cowsert How shall this PR be revised (or closed), based on yesterday's TF-1 discussion?

ijurica:
> @Matt-Cowsert How shall this PR be revised (or closed), based on yesterday's TF-1 discussion?

Following up on @shawnalpay's question. I suggest merging this PR into the 2364-sku-price-dataset branch and continuing the review of the effective date semantics on #2424 as part of the SKU Price dataset, so that the discussion is not split across two PRs touching the same requirements.

For reference, the minutes of the previous TF-1 session record the following: "Riley: confirm PR #2615 folds into the #2424 branch rather than going to Members".

Matt-Cowsert:
Shawn, on revising or closing this pull request: it merged into the #2424 branch on September 15. Task Force 1 planned on September 1 to fold it in, and Irena suggested the same on September 10. The effective date null semantics hold against the current #2424 records and under the separate-rows model that action item #2679 is weighing, because each record carries its own SkuPriceEffectiveStart and SkuPriceEffectiveEnd either way. If the price columns change, the eight records added here get reshaped with the other example records.

@rileyjenk, the merge went ahead because your August 12 suggestions and the temporary-price example are in. If anything from your review still isn't covered, #2424 is the place for it now.

PR #2564, opened 2026-07-28
AI #2563: Add tiered-pricing-related glossary terms

ijurica:
## Summary

This PR improves domain terminology across the specification to reduce ambiguity between distinct concepts that use `tier` or `tiered` terminology.

### What Changed

- Added and refined glossary terms to explicitly distinguish related but different pricing and commitment concepts, including:
  - `Threshold-Based Tiered Pricing`
  - `Threshold-Based Pricing Tier`
  - `Quantity-Based Tiered Pricing`
  - `Tiered Commitment Discount`
  - `Commitment Tier`
- Added an editorial guideline requiring sufficiently domain-qualified terminology for terms containing `tier` or `tiered`, while allowing abbreviated terminology in schema and column names where the schema context makes the intended concept unambiguous.
- Applied the terminology consistently across the specification, including replacing generic or potentially ambiguous uses of `tier` and `tiered` terminology.

### Key Glossary Definitions

The following glossary terms establish the terminology introduced by this PR:

> **Threshold-Based Tiered Pricing**
> A pricing model in which the unit price applied to a charge is determined by the applicable [*threshold-based pricing tier*](#glossary:threshold-based-pricing-tier).
> The applicable *threshold-based pricing tier* is determined based on quantity, duration, or spend within a defined aggregation scope and aggregation interval according to the threshold-based tier application model, without requiring a prior [*contract commitment*](#glossary:contract-commitment). The threshold-based tier application model defines whether the applicable *threshold-based pricing tier* is determined independently for each threshold range (graduated pricing) or whether reaching a higher threshold causes a higher *threshold-based pricing tier* to be applied retroactively to all charges within the aggregation interval (retroactive pricing).
> Threshold-based tiered pricing should not be confused with a [*tiered commitment discount*](#glossary:tiered-commitment-discount), which requires a prior *contract commitment*.
> [*Quantity-based tiered pricing*](#glossary:quantity-based-tiered-pricing) is a common form of threshold-based tiered pricing.

> **Threshold-Based Pricing Tier**
> A predefined pricing level within a [*threshold-based tiered pricing model*](#glossary:threshold-based-tiered-pricing), consisting of a threshold range and an associated unit price. The threshold range may be based on quantity, duration, or spend depending on the *threshold-based tiered pricing* model.

> **Quantity-Based Tiered Pricing**
[...]

PR #2494, opened 2026-06-29
Part 3 of #2315: Add split cost allocation field mapping

thecloudman:
## Summary

Adds a new appendix section - **Examples: Split Cost Allocation Field Mapping** - providing a complete column-by-column reference implementation for the row-splitting transformation a data generator must perform when mapping native billing data into FOCUS split allocation rows.

### The Problem

FR #2315 identified that the FOCUS spec does not provide a concrete, column-level worked example showing how provider-native split allocation data (which typically uses a column-augmentation approach — all consumers on one row) maps to the FOCUS multi-row output. Data generators have independently reverse-engineered this transformation at significant cost.

### What This PR Adds

* **Native input table** — A realistic Aura Web billing export row (fictitious equivalent of AWS CUR v2 with `split_line_item_*`-style columns), including three consuming workloads with allocation ratios and measured utilization in extension column sets.
* **Four FOCUS output rows** — One *origin charge* row and three *allocated charge* rows, each with every relevant FOCUS column populated and a source annotation tracing each value back to the native input column or transformation rule applied.
* **Metric reconciliation table** — Verifies that EffectiveCost, BilledCost, and ConsumedQuantity sum correctly across all four rows to the pre-split origin totals.
* **Cross-provider consumer note** — One consuming workload (LatticeScale ML Runtime) is from a different provider than the host compute instance (Aura Web), demonstrating that `AllocatedServiceName` carries the consuming service identity regardless of provider, while origin dimensions are preserved for invoice reconciliation.

### Depends On

* PR #2395 (Part 1) - introduces `AllocatedServiceName`, referenced in the field mapping tables.

Part 3 of #2315.

## Type of Change

- [x] New feature (non-breaking change which adds functionality)

## Changes

* `specification/appendix/split_cost_allocation_field_mapping.md` - new appendix section.
* `specification/appendix/appendix.mdpp` - `!INCLUDE` entry in alphabetical position.
* `specification/appendix/appendix_overview.md` - table row in alphabetical position.

## Test plan

- [ ] CI build passes (`make` in `specification/`).
- [ ] Markdown lint passes.
- [ ] Metric reconciliation arithmetic verified: $0.00 + $120.00 + $72.00 + $48.00 = $240.00 EffectiveCost; $0.00 + $144.00 + $86.40 + $57.60 = $288.00 BilledCost; 0 + 12 + 7.2 + 4.8 = 24 ConsumedQuantity.

## Author Checklist

[...]

PR #2617, opened 2026-08-11
AI #2550: Scope the `CurrencyFormat` virtual currency allowance to consumption currency

### LINKED ISSUE #1057: [FR] Separate dataset for SKUs

### LINKED ISSUE #2550: [AI] Review virtual currency treatment of tokens, credits, and crypto in the CurrencyFormat attribute
**Assignee(s):** @Matt-Cowsert, @ijurica
**Parent Issue:** #1057

**Detailed Description**

PR #2424 updates the virtual currency glossary entry and adds cryptocurrency as an example. Reviewing that change surfaced a problem with the two existing examples. "Tokens" has become overloaded since the term was first used: in 1.5 it functions primarily as a unit of measure for AI consumption rather than as a store of value. "Credits" collides with the "Credit" ChargeCategory, so a reader cannot tell whether a provider-issued store of value or a billing adjustment is meant.

Separately, the CurrencyFormat attribute constrains values to ISO currency codes, which covers national currencies only, yet it still references virtual currencies. Two questions follow: whether cryptocurrency belongs in the allowed currency formats, and whether the virtual currency references in that attribute should stay at all. Irena noted this predates PR #2424 and has been deferred before, because there is no standard to point to for cryptocurrency codes and an open-ended format weakens the attribute.

Matt proposed a first week of August due date. The review outcome decides whether a draft PR lands ahead of the 1.5 draft deadline or the work moves to the 1.5 consistency review.

**Definition of Done**
- [ ] The virtual currency glossary examples are assessed against current usage of "tokens" and "credits", with a recommendation on whether to replace them
- [ ] A position is recorded on whether cryptocurrency belongs in the CurrencyFormat attribute allowed values
- [ ] A position is recorded on whether the virtual currency references in the CurrencyFormat attribute stay or are removed
- [ ] The work is either scoped into a draft PR ahead of the 1.5 draft deadline or moved to the 1.5 consistency review

### LINKED ISSUE #2687: [AI] Restore the consumption currency glossary links unwrapped on PR #2617
## Action Item Details

**Detailed Description**

PR #2424 adds the `Consumption Currency` glossary entry. PR #2617 carries that term into the CurrencyFormat attribute, the PricingCurrency columns of the Cost and Usage and Contract Commitment datasets, PricingCurrencyEffectiveCost, and the SaaS appendix example, linking it to its glossary entry on first occurrence in each of those five files.

Until #2424 merges to `working_draft`, `#glossary:consumption-currency` resolves to nothing. The `test_internal_anchor_links_resolve` check added in #2510 fails on that missing anchor, which fails `make test`, which fails the `Build PDF` step. The `gen_pdf` check on #2617's branch has been red on every run since 2026-08-11 for this reason alone. The PDF itself builds. The failure is the link validation that runs after it.

On 2026-09-10 the five links were unwrapped to plain italics on #2617 so the branch carries a passing build while the dependency is outstanding. The italics were kept, so the glossary-term styling is unchanged, and the eight other italic occurrences of the term across those files were untouched. Only the five first-occurrence links were removed:

* `specification/attributes/currency_format.md` line 8
* `specification/appendix/saas_examples/virtual_currency_pricing_model.md` line 3
* `specification/datasets/contract_commitment/columns/pricingcurrency.md` line 3
* `specification/datasets/cost_and_usage/columns/pricingcurrency.md` line 3
* `specification/datasets/cost_and_usage/columns/pricingcurrencyeffectivecost.md` line 5

`guidelines/contributors/editorial-guidelines.md` requires glossary terms to be linked on first occurrence within each source markdown file. Those five files do not meet that requirement for `consumption currency` while this item is open.

This item exists because nothing enforces the requirement. The linter is pymarkdownlnt, and the four checks under `specification/tests/` cover JSON blocks, JSON schemas, spec links, and SQL blocks. No rule verifies that an italicized glossary term carries a link on first occurrence. The unresolved anchor was the only automated signal that these links were owed, and unwrapping them removed it.

Grouped under #2509. Related to #2653, which covers the remaining consumption currency alignment work on the SaaS appendix example. This item is narrower and covers the link restoration only.

**Definition of Done**

* #2424 has merged to `working_draft` and #2617 has rebased onto it.
* The five links listed above are restored to `[*consumption currency*](#glossary:consumption-currency)`, preserving the existing capitalization at `currency_format.md` line 8, which reads `Consumption currency`.
* `gen_pdf` passes on the branch carrying the restored links, confirming the anchor resolves.
* If #2617 merges before this is done, the restoration lands as a separate pull request against `working_draft` ahead of the 1.5 release gate.

Matt-Cowsert:
## Summary

Resolves the contradiction in the `CurrencyFormat` attribute between its stated currency types and its normative requirements, and implements the outcome of the review in [AI #2550](https://github.com/FinOps-Open-Cost-and-Usage-Spec/FOCUS_Spec/issues/2550).

`CurrencyFormat` lists virtual currency as one of two allowed currency types in its preamble, while its requirements mandate an ISO 4217:2015 three-letter alphabetic code for every value. ISO 4217:2015 defines no code for a provider-issued currency, so no virtual currency value can satisfy both halves of the attribute.

The contradiction is residue from the v1.4 attribute refactor. Through v1.3 the attribute carried a conditional pair: `CurrencyFormat-A-001-C` required the ISO 4217:2015 code only when the value was presented in national currency, and `CurrencyFormat-A-002-C` sent the value to `StringHandling` instead when it was presented in virtual currency. v1.4 dropped the condition from the first rule, removed the second, and added a new unconditional `ATT-CurrencyFormat-A-004-M` requiring ISO 4217:2015 conformance. The preamble bullet advertising virtual currency stayed in place.

The gap has practical reach. `PricingCurrency` binds to `CurrencyFormat` in Cost and Usage, Contract Commitment, and the SKU Price dataset proposed in #2424. Nine rows across three SaaS appendix CSVs carry a provider-issued `PricingCurrency` ("Token" before this PR, "OmniBucks" after it), and #2424 adds `sku_price_consumable_currency.csv` with a `PricingCurrency` of "Credit" against a `PricingCurrencyCategory` of "Consumable". A consumption currency pricing model cannot be expressed conformantly today, even though the `PricingCurrency` column family and the `IncludesVirtualCurrency` condition exist for that case.

### Key Changes

**Requirements.** The v1.3 conditional pair is restored, narrowed to *consumption currency* rather than *virtual currency* generally, and written as two condition grouping bullets, one per branch:

* Unless the value is presented in a consumption currency, the column conforms to ISO 4217:2015 and uses its three-letter alphabetic code. These are the two existing requirements, nested under a shared condition. The condition is `unless ... consumption currency` rather than `when ... national currency` so that a value that is neither, a cryptocurrency for instance, still falls under the ISO requirement instead of passing by omission.
* When the value is presented in a consumption currency, the column conforms to `StringHandling` and does not use a code defined in ISO 4217:2015. The `StringHandling` rule constrains nothing new, since every column that binds to `CurrencyFormat` today already declares that conformance on its own; it is here so the attribute is self-contained and the v1.3 pair is restored intact. The MUST NOT has no v1.3 equivalent and keeps the two branches from being confused.

Narrowing the exception from virtual currency to consumption currency is what keeps cryptocurrency outside the attribute, which is the position recorded on #2550.

**Preamble.** The currency type list now reads *National currency* and *Consumption currency*, with the examples "platform credits" and "normalized billing units" drawn from the glossary entry carried here from #2424. The prior examples "tokens" and "credits" are dropped. "Tokens" now reads primarily as a unit of measure in FOCUS, appearing as `ConsumedUnit` and `PricingUnit` values in the AI model identity appendix, and unqualified "credits" collides with the `ChargeCategory` value "Credit".

**SaaS appendix example.** The example at `specification/appendix/saas_examples/consumption_currency_pricing_model.md` (renamed from `virtual_currency_pricing_model.md`, together with its six CSV data files, the `saas_examples.mdpp` include, and the six CSV links) is the only worked illustration of the pricing model this attribute now permits, and it described that model as a *virtual currency* throughout. Its section title, its eight scenario headings, and its body prose now read *consumption currency*, with the term linked to its glossary entry on first use alongside *national currency*. Its introductory example list, which offered "credits, tokens, or points", now uses the same qualified forms as the preamble above. The SaaS row of `appendix_overview.md` matches. No internal link anchors any of the renamed headings.

OmniQuery's denomination is now "OmniBucks" rather than "Token", in the prose and in the data, following the Slack discussion Shawn summarized in review: "token" is a unit of measure in FOCUS, not a currency. The name is provider-specific on purpose. A generic "Credit" would collide with the `ChargeCategory` value "Credit", and with the Credit-Based Consumption example in `billing_scenario_examples.md`, where the same provider, OmniQuery, consumes credits as a unit priced in USD. A currency that reads as OmniQuery's own keeps the currency columns and the unit columns apart. In the CSVs the change touches `PricingCurrency`, `PricingUnit`, `ChargeDescription`, `ResourceName`, and `ResourceType` cells only; no amount changes. The three purchase rows carry "OmniBucks" in `PricingUnit`, since a purchase buys the currency as a unit, and the usage rows carry it in `PricingCurrency`. Two prose lines that had to change anyway were also corrected, both previously recorded on [AI #2653](https://github.com/FinOps-Open-Cost-and-Usage-Spec/FOCUS_Spec/issues/2653): the Scenario A and B contract terms now name OmniQuery, not Acme Corp, as the party offering the discount, and the Scenario A2 note now describes `PricingCurrency` as the denomination the usage is priced in rather than as the per usage price.

**Implementation Context.** A new section records that a consumption currency is one subtype of virtual currency and that other subtypes, cryptocurrency among them, have no code in ISO 4217:2015 and no allowed value under this attribute. That matches the note in #2424 stating that a cryptocurrency-denominated price does not by itself satisfy `IncludesVirtualCurrency`. The section links the glossary rather than restating what a consumption currency is, per the Separation of Concerns guidance in the normative requirements guidelines. It also records that the exception does not loosen columns separately constrained to national currency, since `BillingCurrency` also binds to `CurrencyFormat` in three datasets.

**Consuming column descriptions.** `PricingCurrency` in Cost and Usage and Contract Commitment, and `PricingCurrencyEffectiveCost` in Cost and Usage, described their denomination as "national or virtual currency". Those columns bind to `CurrencyFormat`, so they are narrowed to "consumption currency" to match. #2424's SKU Price `PricingCurrency` already carries this wording, so no follow-up edit is needed there on rebase.

**Glossary.** The `Consumption Currency` entry and the revised `Virtual Currency` definition are carried here verbatim from #2424, so the PR reads on its own and `#glossary:consumption-currency` resolves. The five first-occurrence links to that anchor, unwrapped earlier on 2026-09-10 while the dependency was outstanding, are restored, which closes [AI #2687](https://github.com/FinOps-Open-Cost-and-Usage-Spec/FOCUS_Spec/issues/2687).

### Known open point

The four rules still admit any non-ISO string, because a value that is not an ISO 4217:2015 code is by construction presented in a consumption currency. The new MUST NOT closes the collision case, not the broader concern that an open-ended format weakens the attribute. Closing that needs a decidable signal for the condition. SKU Price has one in `PricingCurrencyCategory`; Cost and Usage and Contract Commitment do not, and the nearest candidate is the `IncludesVirtualCurrency` condition that #2424 rescopes to consumption currency. Flagged for TF-1 rather than resolved here.

### Dependency and sequencing

This PR no longer waits on #2424. The two glossary hunks it relies on, the `consumption currency` entry and the rewritten `virtual currency` entry, are carried here verbatim (see Glossary above), and `gen_pdf` passes on the branch. The Implementation Context section reads consumption currency as a *virtual currency* subtype, which holds under the rewritten entry and is circular under the current one, so if #2424's revision of that entry changes during its review, the same change lands here.

`working_draft` is merged into this branch through 9de5e741. The two later commits on `working_draft` (#2360 and #2624) also edit `glossary.md`, in entries that do not touch these, and a dry-run merge is clean.

#2424 and this PR overlap in two places. The glossary hunks are identical, so they merge without conflict in either order. `appendix_overview.md` conflicts: #2424 adds its SKU Price row directly below the SaaS row edited here. Whichever merges second resolves to this PR's SaaS row plus #2424's SKU Price row; taking both sides verbatim reinstates "virtual currency pricing models", because #2424's side carries the unedited SaaS row.

Three other open PRs touch files changed here and will conflict with whichever merges second:

* **#2606** edits `pricingcurrencyeffectivecost.md` line 5, the same line this PR rewrites, and moves it the other way: it keeps `[*virtual currency*](#glossary:virtual-currency)` and changes the example from "tokens to USD" to "prepaid credits to USD". This is a terminology disagreement, not just a text collision, and is worth settling between the two PRs rather than at resolution time.
* **#2627** and **#2564** edit the SaaS example at its old path, `virtual_currency_pricing_model.md`. Git rates the renamed file at 48% similar to the old one, under its default 50% rename threshold, so the merge presents a modify/delete conflict on that file rather than a line conflict. #2564's two lines are untouched here and apply to `consumption_currency_pricing_model.md` as they are. #2627's one line is the denomination line, which this PR rewrites, so its backtick change needs re-applying to `"Credit"` by hand.

### Out of scope

* Requirements model JSON, handled downstream per release version. The work is larger than reinstating `ATT-CurrencyFormat-A-002-C`: the four atomic requirements now resolve from two condition grouping bullets, the reinstated rule reads "consumption currency" rather than "virtual currency", and the new MUST NOT has no existing rule to map to.
* The "includes virtual currency" link text inside three normative bullets of `cost_and_usage/dataset.md`. It describes the `IncludesVirtualCurrency` condition that #2424 rescopes, and it is the one item left on #2653.
* `unit_format.md`. Its eight Token references are all units of measure (a base unit, the compound and unit-quantity examples, and the count-based unit table row), the sense the word now carries in FOCUS, so none change. This closes the TF-1 action item recorded on this PR on 2026-08-25.

### Changelog

* 2026-08-10: Initial draft.
* 2026-08-24: Marked ready for review. Recorded the #2424 merge-order dependency and the expected `gen_pdf` failure until it lands. Removed a stale note asking for a follow-up edit to #2424's SKU Price `PricingCurrency`, which already uses this wording.
* 2026-08-25: Brought the SaaS pricing example and the `appendix_overview.md` SaaS row into the PR, closing the terminology portion of #2653. Corrected the rebase note: #2424 does conflict, on the `appendix_overview.md` SaaS row. File and CSV renames and the "Token" denomination question stay on #2653.
* 2026-08-25: Corrected three defects found in pre-review grading. The appendix intro offered "credits, tokens, or points" as consumption currency examples while the preamble above dropped exactly those words; the denomination was written "tokens" where the CSV data carries "Token"; and a second `#glossary:FOCUS-dataset` link broke the first-occurrence rule. Recorded the #2606, #2627, and #2564 conflicts, corrected the #2424 resolution note, and stated the second half of the #2424 dependency.
* 2026-08-25: Broadened the new MUST NOT from the ISO 4217:2015 three-letter alphabetic code to any code defined in the standard, closing the numeric-code case. Trimmed the Implementation Context so it links the glossary definition instead of restating it. Recorded why the `StringHandling` rule is present when every consuming column already declares that conformance.
* 2026-08-31: Merged `working_draft` to clear the `appendix_overview.md` conflict created by #2494, keeping this PR's SaaS row and taking #2494's Split Cost Allocation Field Mapping row. No specification content in this PR changed. The #2424 dependency and the expected `gen_pdf` failure are unaffected.
* 2026-09-10: Unwrapped the five first-occurrence `#glossary:consumption-currency` links to plain italics so `gen_pdf` passes while #2424 is outstanding. The PDF always built; the red check was `test_internal_anchor_links_resolve` failing on the unresolved anchor. The italics and the eight other italic occurrences of the term are unchanged, as are the adjacent glossary links on the same lines. No normative content changed. Restoration on rebase is tracked on AI #2687 and added as step 3 of the merge order. The #2424 dependency itself is unaffected.
* 2026-09-10: Addressed Shawn's review. The four requirement bullets are regrouped as two condition grouping bullets, one per currency type, with the ISO branch conditioned on `unless ... consumption currency` so a value that is neither national nor consumption currency still falls under the ISO requirement. The SaaS example and its six CSV data files are renamed to `consumption_currency_pricing_model*`, and OmniQuery's denomination changes from "Token" to "OmniBucks" in the prose and the data, correcting the Scenario A and B discount party and the Scenario A2 `PricingCurrency` note on the way. The `Consumption Currency` and revised `Virtual Currency` glossary entries are carried from #2424 and the five glossary links are restored, so the PR no longer waits on the merge order and AI #2687 closes. `unit_format.md` is confirmed unchanged.

### Type of Change
* [ ] Bug fix
* [ ] Editorial / Formatting

### Author Checklist


Matt-Cowsert:
## TF1 Action Item - August 25, 2026

This pull request narrows the `CurrencyFormat` attribute so the virtual currency allowance covers a provider's own consumption currency, and it stops using tokens as the illustrative example. The change touches `currency_format.md`, three `PricingCurrency` column files, and the SaaS appendix example.

The `UnitFormat` attribute still names Token and Tokens in eight places, as a base unit, in compound and unit-quantity examples, and as a row in the count-based unit table. Irena Jurica raised this in the Task Force 1 meeting on 2026-08-25 and asked whether some of those should come out given the scoping change. Matt Cowsert agreed to look at how `UnitFormat` is structured, make a recommendation, and fold the outcome into this pull request so the currency and unit treatment of tokens land together.

**Scope:**
- [ ] Record whether each `UnitFormat` token reference stands as a unit of measure or should change given the consumption currency scoping
- [ ] Apply the resulting changes to `unit_format.md` in this pull request

**Assignee:** @matt-cowsert

Matt-Cowsert:
Pull request #2600 already edits `specification/attributes/unit_format.md`, which changes how the action item in my earlier comment should land. Its diff covers the compound unit bullet and the unit quantity bullet, two of the eight lines in that file carrying a Token reference. Folding token changes into this pull request would put them on the same lines pull request #2600 is changing, so it is worth settling which pull request carries the `UnitFormat` token treatment before either one moves.

Two corrections to the same comment. The file list has grown to six: `specification/appendix/appendix_overview.md` and `specification/appendix/saas_examples/virtual_currency_pricing_model.md` now carry the action item #2550 appendix alignment, so the earlier description of `currency_format.md` plus three `PricingCurrency` column files no longer matches. And `specification/datasets/cost_and_usage/columns/pricingcurrencyeffectivecost.md` is the `PricingCurrencyEffectiveCost` column rather than a `PricingCurrency` column.

flanakin (approved):

shawnalpay (changes requested):
@Matt-Cowsert Can you please repair the PDF build on this PR before we proceed with reviewing it further?

Arjun-Trident (approved):

specification/appendix/saas_examples/virtual_currency_pricing_model.md
shawnalpay, line 1: Rename this file to use the word "consumption" instead of "virtual".

specification/appendix/saas_examples/virtual_currency_pricing_model.md
```diff
@@ -9,7 +9,7 @@ The scenarios described below illustrate how a Cost and Usage [*FOCUS dataset*](
 The following baseline conditions apply to the scenarios described below:
 
 * Acme Corp has signed an agreement with SaaS service provider OmniQuery to use their services.
-* OmniQuery offers a virtual currency pricing model for their services and requires a purchase of virtual currency in advance of usage. Their denomination of virtual currency is called "tokens".
+* OmniQuery offers a consumption currency pricing model for their services and requires a purchase of consumption currency in advance of usage. Their denomination of consumption currency is called "Token".
```
shawnalpay: Discussed via Slack with @Matt-Cowsert.  All examples in this file need to avoid use of the word "token".  Especially with the rise of Tokenomics, we need to align the term "token" with a unit of measure and not a currency.

specification/attributes/currency_format.md
```diff
@@ -4,15 +4,23 @@ Columns that contain currency information in cost data following a consistent fo
...
+* *Consumption currency* (e.g., platform credits, normalized billing units).
 
 ## Requirements
 
 Column conforming to CurrencyFormat attribute MUST adhere to the following requirements:
 
-* [*FOCUS dataset column*](#glossary:FOCUS-dataset-column) MUST conform to [ISO 4217:2015](https://www.iso.org/standard/64758.html) standard.
-* *FOCUS dataset column* MUST use the three-letter alphabetic code defined in ISO 4217:2015 (e.g., USD, EUR).
+* [*FOCUS dataset column*](#glossary:FOCUS-dataset-column) MUST conform to [ISO 4217:2015](https://www.iso.org/standard/64758.html) standard, except when the value is presented in a *consumption currency*.
+* *FOCUS dataset column* MUST use the three-letter alphabetic code defined in ISO 4217:2015 (e.g., USD, EUR), except when the value is presented in a *consumption currency*.
```
shawnalpay: These could be wrapped in a composite such as, "When the value is presented in a national currency, FOCUS dataset column adheres to the following requirements"...

specification/appendix/saas_examples/virtual_currency_pricing_model.md
```diff
@@ -1,6 +1,6 @@
-# Virtual Currency Pricing Model
+# Consumption Currency Pricing Model
 
-Many SaaS service providers support pricing models that utilize virtual currencies such as credits, tokens, or points. Charges may be provided using a virtual currency, which can subsequently be converted to a national currency such as USD or EUR at an advertised or agreed-upon conversion rate.
+Many SaaS service providers support pricing models that utilize a *consumption currency* such as platform credits or normalized billing units. Charges may be provided using a consumption currency, which can subsequently be converted to a [*national currency*](#glossary:national-currency) such as USD or EUR at an advertised or agreed-upon conversion rate.
```
shawnalpay: Consumption currency doesn't have an entry in the glossary.  The Virtual currency entry will need to be revised.

shawnalpay (changes requested):
Left a comment to summarize my Slack discussion with @Matt-Cowsert, plus left a few other comments.

specification/appendix/saas_examples/virtual_currency_pricing_model.md
Matt-Cowsert: Renamed to `consumption_currency_pricing_model.md`, together with the six CSV data files, the `saas_examples.mdpp` include, and the six CSV links. #2627 and #2564 both edit the old path, so they will need the rename at merge time; noted in the body.

specification/appendix/saas_examples/virtual_currency_pricing_model.md
Matt-Cowsert: Done. OmniQuery's denomination is now "OmniBucks" in the prose and in all six CSVs. Provider-specific on purpose: a generic "Credit" would collide with the ChargeCategory value and with the Credit-Based Consumption example, where OmniQuery consumes credits as a unit priced in USD. In the data that touches PricingCurrency, PricingUnit, ChargeDescription, ResourceName, and ResourceType only; no amount changes. Two lines that had to change anyway were also corrected, both open on #2653: the Scenario A and B contract terms now name OmniQuery rather than Acme Corp as the party offering the discount, and the A2 note now describes PricingCurrency as the denomination the usage is priced in rather than the per usage price.

specification/attributes/currency_format.md
Matt-Cowsert: Restructured as two condition grouping bullets, one per branch. For the ISO branch I kept the condition as "unless the value is presented in a consumption currency" rather than "when the value is presented in a national currency". With the national currency phrasing, a value that is neither, a cryptocurrency for instance, would fall outside both branches and pass by omission. The unless form keeps ISO 4217 as the default, which is the position the Implementation Context states.

specification/appendix/saas_examples/virtual_currency_pricing_model.md
Matt-Cowsert: It does now. Both entries are on #2424, which adds Consumption Currency and rewrites Virtual Currency to "a digital representation of value that is independent of government regulation or central banks (e.g., cryptocurrency, digital credits)". This PR carries both hunks verbatim so it reads on its own and the glossary links resolve, which also closes #2687. The text is identical on both branches, so whichever merges second rebases clean. If the Virtual Currency wording needs more than #2424's revision, I can take that here.

Matt-Cowsert:
Pushed three commits for the review threads; each is answered in its thread.

* `currency_format.md` requirements regrouped as two condition grouping bullets, one per currency type.
* SaaS example and its six CSVs renamed to `consumption_currency_pricing_model*`, with OmniQuery's denomination changed from "Token" to "OmniBucks" throughout.
* Consumption Currency and Virtual Currency glossary entries carried from #2424 and the five glossary links restored, so `gen_pdf` passes without waiting on the merge order.

On the UnitFormat action item from the August 25 TF-1: all eight Token references in `unit_format.md` are units of measure (a base unit, the compound and unit-quantity examples, and the count-based unit table row), which is the sense the word now carries in FOCUS. None change, and `unit_format.md` stays out of this PR.

ljadvey (approved):

specification/glossary.md
```diff
@@ -326,4 +332,4 @@ A billing concept where the cost of a [*charge*](#glossary:charge) is determinis
 
 <a name="glossary:virtual-currency"><b>Virtual Currency</b></a>
 
-A proprietary currency (e.g., credits, tokens) issued by service providers and independent of government regulation.
+A digital representation of value that is independent of government regulation or central banks (e.g., cryptocurrency, digital credits).
```
jpradocueva, line 335: ### F-001 — warning — [Line 335](https://github.com/FinOps-Open-Cost-and-Usage-Spec/FOCUS_Spec/blob/438d3ea443e9cec1fc9ea6c20a6d1c9f134f0626/specification/glossary.md#L335)

**Changed content:** `A digital representation of value that is independent of government regulation or central banks (e.g., cryptocurrency, digital credits).`

**Rule:** `review-diff-scope-01` — Writing Specification Content & Review Guidelines > Review Conduct > Diff-scope discipline; position 3. “Pre-existing problems are out of scope unless they create a direct inconsistency with new content in the same PR.”

**Semantic evidence:**

- `head` `specification/glossary.md` lines 333–335 at `438d3ea443e9cec1fc9ea6c20a6d1c9f134f0626` — Shows the PR's new Virtual Currency definition, which now covers cryptocurrency. Excerpt: `<a name="glossary:virtual-currency"><b>Virtual Currency</b></a> A digital representation of value that is independent of government regulation or central banks (e.g., cryptocurrency, digital credits).`
- `base` `specification/glossary.md` lines 327–329 at `9de5e7418d66c86160cf35ccd24470be4d74ba81` — Shows the definition before the PR: a proprietary currency issued by service providers (credits, tokens), which did not cover cryptocurrency. Excerpt: `<a name="glossary:virtual-currency"><b>Virtual Currency</b></a> A proprietary currency (e.g., credits, tokens) issued by service providers and independent of government regulation.`
- `head` `specification/operating_model_conditions/includesvirtualcurrency.md` lines 9–9 at `438d3ea443e9cec1fc9ea6c20a6d1c9f134f0626` — The unchanged IncludesVirtualCurrency condition turns on whenever prices are in virtual currency, so it inherits the broadened definition. Excerpt: `* IncludesVirtualCurrency MUST evaluate to true when the *operating model* includes prices in *virtual currency*.`
- `head` `specification/attributes/currency_format.md` lines 23–23 at `438d3ea443e9cec1fc9ea6c20a6d1c9f134f0626` — The PR's own CurrencyFormat text says non-consumption virtual currencies such as cryptocurrency have no allowed value. Excerpt: `A *consumption currency* is one subtype of [*virtual currency*](#glossary:virtual-currency). Other subtypes, such as a cryptocurrency, have no code defined in ISO 4217:2015 and no allowed value under this attribute.`
- `head` `specification/datasets/cost_and_usage/dataset.md` lines 143–143 at `438d3ea443e9cec1fc9ea6c20a6d1c9f134f0626` — Shows the unchanged obligation triggered by IncludesVirtualCurrency: a PricingCurrency-denominated metric must be present. Excerpt: `* CostAndUsage MUST include PricingCurrencyEffectiveCost when the *operating model* [includes virtual currency](#operatingmodelconditions.includesvirtualcurrency) and [includes list unit prices](#operatingmodelconditions.includeslistunitprices).`
- `head` `specification/datasets/cost_and_usage/columns/pricingcurrency.md` lines 11–11 at `438d3ea443e9cec1fc9ea6c20a6d1c9f134f0626` — PricingCurrency values must satisfy CurrencyFormat, which has no allowed value for a cryptocurrency. Excerpt: `* PricingCurrency MUST conform to [CurrencyFormat](#attributes.currencyformat) requirements.`

**Engineering context:** In FOCUS, an operating model condition works like a feature flag: when it is true, certain columns become mandatory. IncludesVirtualCurrency is true when prices are in a virtual currency, and it makes the PricingCurrency cost and price columns mandatory. Those columns must hold a currency value that CurrencyFormat allows. This PR splits currencies into national currencies (ISO 4217 codes), consumption currencies (a provider's own credits), and other virtual currencies such as cryptocurrency, which CurrencyFormat gives no allowed value.

**Problem:** Broadening Virtual Currency should not require data that nobody can produce, but the new definition now includes cryptocurrency. That turns on the unchanged IncludesVirtualCurrency condition for crypto-priced models, which then requires PricingCurrency columns that, per this PR's CurrencyFormat text, have no allowed value for a cryptocurrency.

**Impact:** A data generator that prices in a cryptocurrency would be required to include PricingCurrencyEffectiveCost, PricingCurrencyListUnitPrice, and PricingCurrencyContractedUnitPrice, but could not fill PricingCurrency with a conforming value, so no conforming dataset exists for that case. Before this PR the definition covered only provider-issued credits and tokens, so cryptocurrency did not trigger the condition.

**Behavioral effect of the correction:** Behavioral effect requires author clarification: decide whether IncludesVirtualCurrency (and the columns it makes mandatory) should apply to every virtual currency or only to consumption currency. The author has offered in the glossary thread to adjust the Virtual Currency wording here if more than the #2424 revision is needed.

**Suggested correction:** Either scope the IncludesVirtualCurrency condition (or the three dataset.md obligations that use it) to consumption currency, or keep the narrower Virtual Currency definition until a later change allows other virtual currencies in CurrencyFormat. Coordinate the choice with PR #2424, which carries the same glossary text.

specification/appendix/saas_examples/consumption_currency_pricing_model.md
```diff
@@ -0,0 +1,132 @@
+# Consumption Currency Pricing Model
+
+Many SaaS service providers support pricing models that utilize a [*consumption currency*](#glossary:consumption-currency) such as platform credits or normalized billing units. Charges may be provided using a consumption currency, which can subsequently be converted to a [*national currency*](#glossary:national-currency) such as USD or EUR at an advertised or agreed-upon conversion rate.
+
+The scenarios described below illustrate how a Cost and Usage [*FOCUS dataset*](#glossary:FOCUS-dataset) should look for various scenarios where a provider utilizes this pricing model.
```
jpradocueva, line 5: ### F-001 — error — [Line 5](https://github.com/FinOps-Open-Cost-and-Usage-Spec/FOCUS_Spec/blob/438d3ea443e9cec1fc9ea6c20a6d1c9f134f0626/specification/appendix/saas_examples/consumption_currency_pricing_model.md#L5)

**Changed content:** `The scenarios described below illustrate how a Cost and Usage [*FOCUS dataset*](#glossary:FOCUS-dataset) should look for various scenarios where a provider utilizes this pricing model.`

**Rule:** `editorial-links-01` — Writing Specification Content & Review Guidelines > Editorial Conventions > Linking Rule; position 9. “Link the first remaining occurrence in reading order.”

**Engineering context:** FOCUS entity links act like symbol references in API documentation: the first use gives a reader a direct route to the entity's definition.

**Related locations:** Lines 15, 37–42, 58, 132.

**Problem:** The first non-heading occurrence of each distinct FOCUS entity should link to its definition, but the new file leaves Cost and Usage and the first occurrences of numerous column entities as plain text.

**Impact:** Readers cannot navigate directly from the example's first schema references to the definitions needed to interpret the example, and the file does not meet the repository's deterministic linking convention.

**Suggested correction:** Link Cost and Usage at line 5 and link the first non-heading occurrence of each distinct column entity in the file, leaving later occurrences unlinked.

jpradocueva (commented):

Snapshot totals: 9 conversation comments, 14 reviews, and 9 inline review comments.

| --- | --- | --- |
| F-001 | Submit inline comment | Missing first-occurrence entity links; correcting navigation does not change schema or normative behavior. |
| F-002 | Submit inline comment | Incomplete price phrases obscure two scenario inputs; correcting the prose does not change schema or normative behavior. |
| F-003 | Submit inline comment | Scenario B3 names billing-currency fields and wrong amounts; correcting the informative example does not change schema or normative behavior. |
| F-004 | Submit inline comment | A double negative reverses Scenario C's stated scope; correcting it does not change schema or normative behavior. |

specification/appendix/saas_examples/consumption_currency_pricing_model.md
```diff
@@ -0,0 +1,132 @@
...
+
+For this scenario, contract terms include the following terms in addition to the baseline scenario mentioned above:
+
+* OmniQuery offers no discount for purchased OmniBucks.
+
+## Scenario A1: Purchase of Consumption Currency Without a Discount
+
+For this scenario, the initial purchase of consumption currency is executed as follows:
+
+* On April 1, 2025, Acme Corp agrees to purchase 100,000 OmniBucks at $2 per OmniBuck for a total spend $200,000. These OmniBucks are only valid for 12 months.
```
jpradocueva, line 31: ### F-002 — error — [Line 31](https://github.com/FinOps-Open-Cost-and-Usage-Spec/FOCUS_Spec/blob/438d3ea443e9cec1fc9ea6c20a6d1c9f134f0626/specification/appendix/saas_examples/consumption_currency_pricing_model.md#L31)

**Changed content:** `* On April 1, 2025, Acme Corp agrees to purchase 100,000 OmniBucks at $2 per OmniBuck for a total spend $200,000. These OmniBucks are only valid for 12 months.`

**Rule:** `editorial-tone-01` — Writing Specification Content & Review Guidelines > Normative Language & Requirements > Tone; position 7. “Use formal language.”

**Engineering context:** A specification example should state commercial inputs as complete, unambiguous prose so implementers can map each amount to its role.

**Related locations:** Lines 71.

**Problem:** The purchase terms should use complete formal price phrases, but line 31 omits “of” before the total and line 71 omits both “a” before “discounted cost” and “of” before its total.

**Impact:** The omissions make both scenario inputs read as editing artifacts and slow interpretation of the purchase price and total.

**Suggested correction:** Use “for a total spend of $200,000” on line 31, and “at a discounted cost of $1 per OmniBuck for a total spend of $100,000” on line 71.

specification/appendix/saas_examples/consumption_currency_pricing_model.md
```diff
@@ -0,0 +1,132 @@
...
+
+Additionally, OmniQuery offers a modified usage to OmniBuck ratio for one of their services as follows:
+
+* 1 Workflow Operation = 2 OmniBucks
+
+[**CSV Example**](/specification/data/saas_examples/consumption_currency_pricing_model_b3.csv)
+
+Note the following details in the example dataset:
+
+* Because of the modified rate for Workflow Operations, the PricingCurrencyContractedUnitPrice and PricingCurrencyListUnitPrice are different for this charge. The ContractedUnitPrice is set to $1 and the ListUnitPrice is set to $2.
```
jpradocueva, line 115: ### F-003 — error — [Line 115](https://github.com/FinOps-Open-Cost-and-Usage-Spec/FOCUS_Spec/blob/438d3ea443e9cec1fc9ea6c20a6d1c9f134f0626/specification/appendix/saas_examples/consumption_currency_pricing_model.md#L115)

**Changed content:** `* Because of the modified rate for Workflow Operations, the PricingCurrencyContractedUnitPrice and PricingCurrencyListUnitPrice are different for this charge. The ContractedUnitPrice is set to $1 and the ListUnitPrice is set to $2.`

**Rule:** `schema-accuracy-01` — Writing Specification Content & Review Guidelines > Validation & Schema Accuracy > Mathematical & Schema Accuracy; position 14. “AI reviewers MUST rigorously calculate, parse, and verify all data within examples (especially JSON snippets and tables).”

**Semantic evidence:**

- `head` `specification/appendix/saas_examples/consumption_currency_pricing_model.md` lines 19–19 at `438d3ea443e9cec1fc9ea6c20a6d1c9f134f0626` — Establishes the baseline published list rate for one Workflow Operation. Excerpt: `* 1 Workflow Operation = 3 OmniBucks`
- `head` `specification/appendix/saas_examples/consumption_currency_pricing_model.md` lines 58–58 at `438d3ea443e9cec1fc9ea6c20a6d1c9f134f0626` — Establishes that the example prices the consumed operation quantity rather than an OmniBuck purchase quantity. Excerpt: `* PricingQuantity reflects the amount of usage of the PricingUnit for each charge and is equivalent to ConsumedQuantity. While relevant to this example, there are scenarios including tiered pricing where ConsumedQuantity and PricingQuantity may not be the same.`
- `head` `specification/appendix/saas_examples/consumption_currency_pricing_model.md` lines 109–109 at `438d3ea443e9cec1fc9ea6c20a6d1c9f134f0626` — Establishes the modified contracted rate used in Scenario B3. Excerpt: `* 1 Workflow Operation = 2 OmniBucks`
- `head` `specification/datasets/cost_and_usage/columns/pricingcurrencycontractedunitprice.md` lines 3–3 at `438d3ea443e9cec1fc9ea6c20a6d1c9f134f0626` — Defines the contracted field as one PricingUnit denominated in PricingCurrency. Excerpt: `The Pricing Currency Contracted Unit Price represents the agreed-upon unit price for a single [Pricing Unit](#datamodel.costandusage.pricingunit) of the associated [*SKU*](#glossary:sku), inclusive of [*negotiated discounts*](#glossary:negotiated-discount), if present, while excluding negotiated [*commitment discounts*](#glossary:commitment-discount) or any other discounts. This price is denominated in the [Pricing Currency](#datamodel.costandusage.pricingcurrency). When negotiated discounts do not apply to unit prices and instead are applied to exchange rates, the Pricing Currency Contracted Unit Price defaults to the [Pricing Currency List Unit Price](#datamodel.costandusage.pricingcurrencylistunitprice). The Pricing Currency Contracted Unit Price is commonly used to calculate savings based on negotiation activities.`
- `head` `specification/datasets/cost_and_usage/columns/pricingcurrencylistunitprice.md` lines 3–3 at `438d3ea443e9cec1fc9ea6c20a6d1c9f134f0626` — Defines the list field as one PricingUnit denominated in PricingCurrency. Excerpt: `The Pricing Currency List Unit Price represents the suggested service-provider-published unit price for a single [Pricing Unit](#datamodel.costandusage.pricingunit) of the associated [*SKU*](#glossary:sku), exclusive of any discounts. This price is denominated in the [Pricing Currency](#datamodel.costandusage.pricingcurrency). The Pricing Currency List Unit Price is commonly used for calculating savings based on various rate optimization activities.`

**Engineering context:** PricingCurrencyContractedUnitPrice and PricingCurrencyListUnitPrice price one PricingUnit in PricingCurrency. ContractedUnitPrice and ListUnitPrice are separate fields denominated in BillingCurrency.

**Problem:** Scenario B3 should describe the changed Workflow Operation rate in the PricingCurrency unit-price fields, but the explanation switches to the billing-currency fields and gives the $1/$2 OmniBuck purchase prices instead of the scenario's 2/3 OmniBucks per operation rates.

**Impact:** An implementer following the prose can populate the wrong columns with amounts for the wrong unit and denomination, contradicting both the example's stated rates and the column definitions.

**Behavioral effect of the correction:** No schema or normative behavior change. The informative example changes to describe the existing schema correctly.

**Suggested correction:** State that PricingCurrencyContractedUnitPrice is 2 and PricingCurrencyListUnitPrice is 3 for the Workflow Operation charge.

```suggestion
* Because of the modified rate for Workflow Operations, PricingCurrencyContractedUnitPrice is set to 2 and PricingCurrencyListUnitPrice is set to 3 for this charge.
```

specification/appendix/saas_examples/consumption_currency_pricing_model.md
```diff
@@ -0,0 +1,132 @@
...
+For this scenario, Acme Corp has exceeded their purchased OmniBucks on October 1st 2025 by 1,500 OmniBucks and OmniQuery has charged them for the overage. The following conditions apply:
+
+* OmniQuery has charged Acme Corp for the cost of OmniBucks at the list price of $2 per OmniBuck, and this purchase is effective from April 1st 2025 to the date of the purchase, October 1st 2025.
+* Acme Corp purchases an additional 25,000 OmniBucks to facilitate usage to the end of their contract. These OmniBucks are valid from October 1st 2025 to April 1st 2026.
+
+[**CSV Example**](/specification/data/saas_examples/consumption_currency_pricing_model_c.csv)
+
+Note the following details in the example dataset:
+
+* This example focuses on the purchases records only for the overage and additional purchases. Neither usage charges nor earlier purchases are not included in this example.
```
jpradocueva, line 130: ### F-004 — error — [Line 130](https://github.com/FinOps-Open-Cost-and-Usage-Spec/FOCUS_Spec/blob/438d3ea443e9cec1fc9ea6c20a6d1c9f134f0626/specification/appendix/saas_examples/consumption_currency_pricing_model.md#L130)

**Changed content:** `* This example focuses on the purchases records only for the overage and additional purchases. Neither usage charges nor earlier purchases are not included in this example.`

**Rule:** `editorial-tone-01` — Writing Specification Content & Review Guidelines > Normative Language & Requirements > Tone; position 7. “Use formal language.”

**Engineering context:** The scope sentence is the reader's contract for which records the linked Scenario C dataset contains.

**Problem:** The scope note should say that usage charges and earlier purchases are excluded, but “Neither … nor … are not included” is a double negative that says the opposite and conflicts with the preceding sentence.

**Impact:** Readers can infer the wrong expected contents for the Scenario C CSV and misdiagnose an implementation that contains only the purchase records.

**Suggested correction:** Remove the double negative and the “purchases records” typo while preserving the stated scope.

```suggestion
* This example includes only purchase records for the overage and additional purchases. Usage charges and earlier purchases are not included.
```

Matt-Cowsert:
## TF1 Action Item - September 22, 2026

The Includes Virtual Currency operating model condition gates PricingCurrencyContractedUnitPrice, PricingCurrencyEffectiveCost, and PricingCurrencyListUnitPrice. Pull request #2424 had narrowed it to consumption currency while keeping its name, with a note explaining the difference, and Deeja Cruz flagged the mismatch between the name and what the condition evaluates. Task Force 1 agreed on September 22, 2026 to rename it Includes Consumption Currency and drop the note. Shawn Alpay has since reverted the condition on #2424 to its working_draft text, and the rename moves to this pull request, which scopes the CurrencyFormat attribute to consumption currency. Shawn Alpay's re-review of this pull request waits for the rename.

**Scope:**
- [ ] Rename the condition's file, ID, and display name to IncludesConsumptionCurrency and Includes Consumption Currency
- [ ] Scope both of its requirements to prices in a consumption currency, with no explanatory note
- [ ] Update the references in the three gated columns, the Cost and Usage dataset, the operating model conditions overview, and the include file

**Assignee:** @Matt-Cowsert

specification/attributes/currency_format.md
```diff
@@ -4,15 +4,25 @@ Columns that contain currency information in cost data following a consistent fo
...
 Column conforming to CurrencyFormat attribute MUST adhere to the following requirements:
 
-* [*FOCUS dataset column*](#glossary:FOCUS-dataset-column) MUST conform to [ISO 4217:2015](https://www.iso.org/standard/64758.html) standard.
-* *FOCUS dataset column* MUST use the three-letter alphabetic code defined in ISO 4217:2015 (e.g., USD, EUR).
+* Unless the value is presented in a *consumption currency*, [*FOCUS dataset column*](#glossary:FOCUS-dataset-column) MUST adhere to the following requirements:
+  * *FOCUS dataset column* MUST conform to [ISO 4217:2015](https://www.iso.org/standard/64758.html) standard.
+  * *FOCUS dataset column* MUST use the three-letter alphabetic code defined in ISO 4217:2015 (e.g., USD, EUR).
+* When the value is presented in a *consumption currency*, *FOCUS dataset column* MUST adhere to the following requirements:
+  * *FOCUS dataset column* MUST conform to [StringHandling](#attributes.stringhandling) requirements.
+  * *FOCUS dataset column* MUST NOT use a code defined in ISO 4217:2015.
```
ijurica, line 19: By design, we generally avoid referencing one attribute from another. For example, `UnitFormat` could theoretically reference the `StringHandling` requirements, but we chose not to add that requirement there. Instead, we prefer each FOCUS column to explicitly reference each applicable attribute requirement.

`CustomColumnHandling` is an exception to this pattern: it typically references other attributes (e.g., `NullHandling`, `DateTimeFormat`, and `NumericFormat`).

```suggestion
* *FOCUS dataset column* MUST NOT use a code defined in ISO 4217:2015 when the value is presented in a *consumption currency*.
```

---

On that note, now that we've added that condition to the ISO 4217:2015-related requirements, I suggest we replace the below requirement in `specification\attributes\custom_column_handling.md`:

> * *Custom column* representing a national currency SHOULD conform to [CurrencyFormat](#attributes.currencyformat) requirements.

With this:

> * *Custom column* representing a currency SHOULD conform to [CurrencyFormat](#attributes.currencyformat) requirements.

I can't submit this as an inline suggestion from my side, so @Matt-Cowsert, if you agree with the suggestion, could you please apply this change directly in `specification\attributes\custom_column_handling.md`?

ijurica: ### Column representing A/B vs Value presented in A/B

I think there is a potential **circularity** here: the `when the value is presented in ...` condition determines which requirements apply, but it is not clear how we independently establish that the value is presented in a consumption currency in the first place.

For example, this works for `BillingCurrency` because the column itself is independently defined as representing a national currency:

> `BillingCurrency MUST be expressed in national currency.`

So we know the currency type before applying the `CurrencyFormat` requirements.

For `PricingCurrency`, however, there is no equivalent requirement establishing whether its values represent a national or consumption currency. `CurrencyFormat` allows both, but the subsequent requirements use the currency type as the condition for determining which requirements apply. This leaves us without an independent point at which the value is classified as national or consumption currency.

If `PricingCurrency` can contain both national and consumption currencies, the mental model seems to become something like:

* Determine whether the value represents a national currency.

  * If the value is a three-letter ISO 4217 code, this evaluates to true.
  * Otherwise, it evaluates to false.
* If true, the value MUST be a three-letter ISO 4217 code.

That is circular: the requirement used to validate the value is also being used to determine which category the value belongs to in the first place.

I think this could be resolved by **declaring** the currency type at the `PricingCurrency` column level, rather than determining it from individual values. We could add a `PricingCurrency` requirement analogous to `BillingCurrency`, requiring the provider to specify in its documentation (or the Metadata 😄)  whether `PricingCurrency` represents a national currency or a consumption currency.

The `CurrencyFormat` requirements could then be reworded to make the conditions refer to **what the column represents**, rather than what an individual record's value happens to contain. We could also provide an alternative path for when the currency type is supplied through metadata, once such metadata is available.

#### Suggestions:

##### PricingCurrency

Add the following requirement to all PricingCurrency columns:

> * PricingCurrency MUST represent either a *national currency* or a *consumption currency*, as specified in the provider's documentation.

In the future, the Metadata should provide this instead of provider documentation.

##### CurrencyFormat

Replace the following `CurrencyFormat` requirements:

> * Unless the value is presented in a *consumption currency*, [*FOCUS dataset column*](#glossary:FOCUS-dataset-column) MUST adhere to the following requirements:
>   * *FOCUS dataset column* MUST conform to [ISO 4217:2015](https://www.iso.org/standard/64758.html) standard.
>   * *FOCUS dataset column* MUST use the three-letter alphabetic code defined in ISO 4217:2015 (e.g., USD, EUR).
> * When the value is presented in a *consumption currency*, *FOCUS dataset column* MUST adhere to the following requirements:
>   * *FOCUS dataset column* MUST conform to [StringHandling](#attributes.stringhandling) requirements.
>   * *FOCUS dataset column* MUST NOT use a code defined in ISO 4217:2015.

With this:

> * [*FOCUS dataset column*](#glossary:FOCUS-dataset-column) representing a *national currency*  MUST adhere to the following requirements:
>   * *FOCUS dataset column* MUST conform to [ISO 4217:2015](https://www.iso.org/standard/64758.html) standard.
>   * *FOCUS dataset column* MUST use the three-letter alphabetic code defined in ISO 4217:2015 (e.g., USD, EUR).
> * *FOCUS dataset column* representing a *consumption currency* MUST NOT use a code defined in ISO 4217:2015.

```suggestion
* [*FOCUS dataset column*](#glossary:FOCUS-dataset-column) representing a *national currency*  MUST adhere to the following requirements:
  * *FOCUS dataset column* MUST conform to [ISO 4217:2015](https://www.iso.org/standard/64758.html) standard.
  * *FOCUS dataset column* MUST use the three-letter alphabetic code defined in ISO 4217:2015 (e.g., USD, EUR).
* *FOCUS dataset column* representing a *consumption currency* MUST NOT use a code defined in ISO 4217:2015.
```

---

### Conformance to String Handling

By design, we generally avoid referencing one attribute from another. For example, `UnitFormat` could theoretically reference the `StringHandling` requirements, but we chose not to add that requirement there. Instead, we prefer each FOCUS column to explicitly reference each applicable attribute requirement.

`CustomColumnHandling` is an exception to this pattern: it typically references other attributes (e.g., `NullHandling`, `DateTimeFormat`, and `NumericFormat`).

On that note, I suggest (see the above suggestion) we remove conformance to StringHandling from the CurrencyFormat attribute.
In addition to that, if we add that condition to the ISO 4217:2015-related requirements, I suggest we replace:

> * *Custom column* representing a national currency SHOULD conform to [CurrencyFormat](#attributes.currencyformat) requirements.

with:

> * *Custom column* representing a currency SHOULD conform to [CurrencyFormat](#attributes.currencyformat) requirements.

I can't submit this as an inline suggestion from my side, so Matt, could you please apply this change directly?

---

@mike-finopsorg, I'd appreciate your thoughts on all of this as well, especially on the circularity part.

specification/appendix/saas_examples/virtual_currency_pricing_model.md
shawnalpay: Confirmed.

specification/attributes/currency_format.md
shawnalpay: I will close this thread in favor of [this thread](https://github.com/FinOps-Open-Cost-and-Usage-Spec/FOCUS_Spec/pull/2617/files#r4082437436) that @ijurica opened on that new section.

specification/glossary.md
shawnalpay: Good catch, @jpradocueva.

@Matt-Cowsert 
I pulled the `Includes Virtual Currency` change out of #2424 as it is ultimately not directly relevant to that PR.

I think the solution is to make the swap from `Includes Virtual Currency` to `Includes Consumption Currency` in this PR.

specification/attributes/currency_format.md
shawnalpay: Agree with your points, @ijurica.  So how about:

```suggestion
* When the value represents a [*national currency*](#glossary:national-currency), [*FOCUS dataset column*](#glossary:FOCUS-dataset-column) MUST adhere to the following requirements:
  * *FOCUS dataset column* MUST conform to [ISO 4217:2015](https://www.iso.org/standard/64758.html) standard.
  * *FOCUS dataset column* MUST use the three-letter alphabetic code defined in ISO 4217:2015 (e.g., USD, EUR).
* When the value represents a [*virtual currency*](#glossary:virtual-currency), *FOCUS dataset column* MUST adhere to the following requirements:
  * *FOCUS dataset column* MUST conform to [StringHandling](#attributes.stringhandling) requirements.
  * *FOCUS dataset column* SHOULD NOT use a code defined in ISO 4217:2015.
```

shawnalpay (changes requested):
Reviewed and resolved older changes.

Left comments or 👍 on @jpradocueva and @ijurica's comments, which need to be addressed before I can approve.

specification/glossary.md
Matt-Cowsert: Joaquin, Shawn, 6c837375 renames `IncludesVirtualCurrency` to `IncludesConsumptionCurrency`, per the September 22 Task Force 1 decision, and both of its requirements now say consumption currency. A cryptocurrency-priced operating model no longer requires `PricingCurrencyContractedUnitPrice`, `PricingCurrencyEffectiveCost`, or `PricingCurrencyListUnitPrice`. The Cost and Usage dataset table and bullets, the three columns' Content Constraints, and the operating model conditions overview and include follow the new name.

Shawn, #2424 still rewords the old Includes Virtual Currency row in `operating_model_conditions_overview.md`. Dropping that edit there removes its conflict with this rename.

specification/appendix/saas_examples/consumption_currency_pricing_model.md
Matt-Cowsert: Joaquin, linked in 2c121c4c: Cost and Usage on line 5, and the first mention of each column ID outside a heading, on lines 15, 38, 40 to 42, 58, and 132. Later mentions stay unlinked.

specification/appendix/saas_examples/consumption_currency_pricing_model.md
Matt-Cowsert: Joaquin, fixed in 60c36dfb: "for a total spend of" on lines 31 and 71, and "at a discounted cost of" on line 71.

specification/attributes/currency_format.md
Matt-Cowsert: Irena, Shawn, 08773ee1 classifies each value rather than the column: `PricingCurrency` in Cost and Usage and Contract Commitment now requires a national or a consumption currency, the `CurrencyFormat` branches say "represents", and the Implementation Context points to the provider's pricing documentation or contract terms for the classification. Per value, because the SaaS example mixes USD purchase rows with OmniBucks usage rows. SKU Price in #2424 already classifies each row through `PricingCurrencyCategory`.

Shawn, the second branch stays consumption rather than virtual currency. Virtual currency now includes cryptocurrency, which the #2550 review kept out of `CurrencyFormat`, and the rename to `IncludesConsumptionCurrency` does the same for the three PricingCurrency columns it gates. The "unless" keeps a value that is neither national nor consumption currency on the ISO branch.

The MUST NOT applies only to consumption currency values, so mixed columns still conform, and it keeps a consumption currency from reusing an ISO code. `StringHandling` is gone, and the custom column line names national or consumption currency, since cryptocurrency has no allowed value. The description covers the metadata follow-up under Known open point, and how this merges with #2567 and #2564.

specification/glossary.md
jpradocueva: @Matt-Cowsert, @shawnalpay, the changes are OK with me. The contradiction is removed. I am resolving the issue. Thanks.

jpradocueva (approved):

Validation completed for PR head `2c121c4c8372bee041cd56153043282b5624060c` using `updates-agents` at `3c12b80af414838dfaa315fe4532cbea88a64d3b`.

No errors or warnings were identified within the supported Markdown validation scope.

Approval is not recommended because 9 changed file(s) were unsupported; 1 contributor claim(s) remain unresolved.

Content-validated documents: 15
Unsupported files: 9
Incomplete documents: 0
Errors: 0
Warnings: 0
Suggestions: 0
Unresolved claims: 1
Evidence complete: true

specification/datasets/cost_and_usage/columns/pricingcurrency.md
```diff
@@ -9,6 +9,7 @@ PricingCurrency MUST adhere to the following requirements:
 * PricingCurrency MUST be of type String.
 * PricingCurrency MUST conform to [StringHandling](#attributes.stringhandling) requirements.
 * PricingCurrency MUST conform to [CurrencyFormat](#attributes.currencyformat) requirements.
+* PricingCurrency MUST represent a *national currency* or a *consumption currency*.
 * PricingCurrency MUST NOT be null.
```
ijurica:
```suggestion
* PricingCurrency MUST NOT be null.
* PricingCurrency MUST represent a *national currency* or a *consumption currency*.
* PricingCurrency documentation MUST adhere to the following requirements:
  * PricingCurrency documentation MUST specify how to determine whether a value represents a *national currency*.
  * PricingCurrency documentation MUST be accessible to practitioners.
```

specification/attributes/currency_format.md
```diff
@@ -4,15 +4,25 @@ Columns that contain currency information in cost data following a consistent fo
...
+  * *FOCUS dataset column* MUST use the three-letter alphabetic code defined in ISO 4217:2015 (e.g., USD, EUR).
+* When the value represents a *consumption currency*, *FOCUS dataset column* MUST NOT use a code defined in ISO 4217:2015.
+
+## Implementation Context
+
+A *consumption currency* is one subtype of [*virtual currency*](#glossary:virtual-currency). Other subtypes, such as a cryptocurrency, have no code defined in ISO 4217:2015 and no allowed value under this attribute.
+
+Whether a value represents a *consumption currency* follows from the [*service provider*](#glossary:service-provider)'s pricing documentation or the applicable contract terms, not from the format of the value. A single column can carry both currency types (e.g., a *national currency* on purchase rows and a *consumption currency* on usage rows).
+
+Columns that carry a *national currency* by their own definition are unaffected by the *consumption currency* exception. Billing Currency, for example, separately constrains its values to *national currency* in each dataset that includes it.
```
ijurica:
```suggestion
* When the value represents a *national currency*, [*FOCUS dataset column*](#glossary:FOCUS-dataset-column) MUST adhere to the following requirements:
  * *FOCUS dataset column* MUST conform to [ISO 4217:2015](https://www.iso.org/standard/64758.html) standard.
  * *FOCUS dataset column* MUST use the three-letter alphabetic code defined in ISO 4217:2015 (e.g., USD, EUR).
* *FOCUS dataset column* MUST NOT use a code defined in ISO 4217:2015 when the value does not represent a *national currency*.

## Implementation Context

A *consumption currency* is one subtype of [*virtual currency*](#glossary:virtual-currency). Which currency types a column allows, and whether other subtypes such as a cryptocurrency are excluded, is set by that column's own requirements, not by this attribute.

Whether a value represents a *national currency* follows from the column's own definition or from the column's documentation, not from the format of the value. A single column can carry both currency types (e.g., a *national currency* on purchase rows and a *consumption currency* on usage rows).

Columns that carry a *national currency* by their own definition take their currency type from that definition. Billing Currency, for example, separately constrains its values to *national currency* in each dataset that includes it.
```

specification/attributes/currency_format.md
```diff
@@ -4,15 +4,25 @@ Columns that contain currency information in cost data following a consistent fo
...
 ## Requirements
 
 Column conforming to CurrencyFormat attribute MUST adhere to the following requirements:
 
-* [*FOCUS dataset column*](#glossary:FOCUS-dataset-column) MUST conform to [ISO 4217:2015](https://www.iso.org/standard/64758.html) standard.
-* *FOCUS dataset column* MUST use the three-letter alphabetic code defined in ISO 4217:2015 (e.g., USD, EUR).
+* Unless the value represents a *consumption currency*, [*FOCUS dataset column*](#glossary:FOCUS-dataset-column) MUST adhere to the following requirements:
+  * *FOCUS dataset column* MUST conform to [ISO 4217:2015](https://www.iso.org/standard/64758.html) standard.
+  * *FOCUS dataset column* MUST use the three-letter alphabetic code defined in ISO 4217:2015 (e.g., USD, EUR).
+* When the value represents a *consumption currency*, *FOCUS dataset column* MUST NOT use a code defined in ISO 4217:2015.
```
ijurica: ### Rationale — Requirements

**1. Requirement ordering (`PricingCurrency`)**
Per [Logical Grouping of Column Requirements](guidelines/contributors/normative-requirements-guidelines.md#logical-grouping-of-column-requirements), Nullability precedes Values and Value Ranges, and requirements on external documentation (Relationships Outside the Spec / Other) come last. The new "MUST represent a *national currency* or a *consumption currency*" line therefore moves below "MUST NOT be null", and the documentation requirement goes at the end.

**2. Conditioning on *national currency* rather than *consumption currency***
The purpose of `CurrencyFormat` is to bind *national currency* to ISO 4217. Keying the branches to *national currency* expresses that directly. It also makes the condition answerable from a single classification (national or not), which the column documentation now provides. Excluding cryptocurrency moves to the column level, where `PricingCurrency` already constrains allowed types ("MUST represent a *national currency* or a *consumption currency*"), in the same way `BillingCurrency` constrains itself to *national currency*.

**3. `when … does not` instead of `unless`**
Both are permitted, but `when … not` is the established form. Across `specification/`, a trailing `when … not` condition appears 88 times, including five `MUST NOT … when … not` (e.g., "PricingUnit MUST NOT be null when PricingQuantity is not null."). A trailing `unless` appears 4 times, never with `MUST NOT`. `MUST NOT … unless` also reads as a double negative.

**4. Documentation as a normative requirement, on the column**
- *Why normative:* the branches depend on whether a value represents a *national currency*. The only source for that is currently the Implementation Context, which is non-normative, so no requirement makes the classification exist or be accessible.
- *Why on the column:* only `PricingCurrency` can carry more than one currency type. `BillingCurrency` and `PaymentCurrency` are *national currency* by definition and need nothing further. Placing the requirement on the column keeps `CurrencyFormat` about format only and follows the column-level precedent in `PrincipalId` ("PrincipalId documentation MUST include the use cases in which PrincipalId is not null.").
- *Why not SKU Price's `PricingCurrencyCategory` or contract terms:* Cost and Usage has to be self-sufficient and cannot rely on the SKU Price dataset for classification, and contract terms are not a FOCUS artifact. The requirement therefore belongs on every `PricingCurrency` column.

**Notes:** These changes also require adjusting 
- `CurrencyFormat` Implementation Context
-  Cost and Usage `PricingCurrency` (this PR)
- Contract Commitment `PricingCurrency` (this PR)
- SKU Price `PricingCurrency` (#2424, not part of this PR; to be applied there)


```suggestion
* Unless the value represents a *consumption currency*, [*FOCUS dataset column*](#glossary:FOCUS-dataset-column) MUST adhere to the following requirements:
  * *FOCUS dataset column* MUST conform to [ISO 4217:2015](https://www.iso.org/standard/64758.html) standard.
  * *FOCUS dataset column* MUST use the three-letter alphabetic code defined in ISO 4217:2015 (e.g., USD, EUR).
* When the value represents a *consumption currency*, *FOCUS dataset column* MUST NOT use a code defined in ISO 4217:2015.
```

specification/attributes/currency_format.md
ijurica: ### Rationale — Implementation Context

**Note:** This change comes paired with the proposed changes to the `CurrencyFormat` requirements and to the `PricingCurrency` columns. It should be applied only together with them.

With the `CurrencyFormat` branches keyed to *national currency* ("When the value represents a *national currency*, …" / "… MUST NOT use a code defined in ISO 4217:2015 when the value does not represent a *national currency*.") and the classification moved to `PricingCurrency` documentation, three statements in the Implementation Context no longer hold. Only what the change makes necessary is adjusted:

- **¶1:** "and no allowed value under this attribute" is no longer true. A cryptocurrency value is not a *national currency* and uses no ISO code, so the attribute permits it. It is excluded at the column level ("PricingCurrency MUST represent a *national currency* or a *consumption currency*.").
- **¶2:** the requirements now rely on whether a value represents a *national currency*, established by the column's own definition or its documentation. They no longer rely on *consumption currency* established by provider pricing documentation or contract terms. The second sentence is unchanged.
- **¶3:** "the *consumption currency* exception" no longer exists. The second sentence is unchanged.

```suggestion
A *consumption currency* is one subtype of [*virtual currency*](#glossary:virtual-currency). Which currency types a column allows, and whether other subtypes such as a cryptocurrency are excluded, is set by that column's own requirements, not by this attribute.

Whether a value represents a *national currency* follows from the column's own definition or from the column's documentation, not from the format of the value. A single column can carry both currency types (e.g., a *national currency* on purchase rows and a *consumption currency* on usage rows).

Columns that carry a *national currency* by their own definition take their currency type from that definition. Billing Currency, for example, separately constrains its values to *national currency* in each dataset that includes it.
```

specification/attributes/custom_column_handling.md
```diff
@@ -36,7 +36,7 @@ Column conforming to CustomColumnHandling attribute MUST adhere to the following
 * *Custom column* containing numeric values MUST contain a single numeric value.
 * *Custom column* containing numeric values SHOULD conform to [NumericFormat](#attributes.numericformat) requirements.
 * *Custom column* containing string values SHOULD conform to [StringHandling](#attributes.stringhandling) requirements.
-* *Custom column* representing a national currency SHOULD conform to [CurrencyFormat](#attributes.currencyformat) requirements.
+* *Custom column* representing a [*national currency*](#glossary:national-currency) or a [*consumption currency*](#glossary:consumption-currency) SHOULD conform to [CurrencyFormat](#attributes.currencyformat) requirements.
 * *Custom column* representing a measurement unit SHOULD conform to [UnitFormat](#attributes.unitformat) requirements.
```
ijurica: **Rationale**
- **Qualifier:** "representing a *national currency* or a *consumption currency*" cannot be evaluated for a custom column, because telling a *consumption currency* from another *virtual currency* (e.g., cryptocurrency) needs a classification that custom columns do not have. With the `CurrencyFormat` branches keyed to *national currency* (see <LINK>), the attribute no longer needs to exclude cryptocurrency, so "representing a currency" is sufficient.
- **Documentation:** the `CurrencyFormat` branches depend on whether a value represents a *national currency*, and a custom column has no definition that establishes this. `SHOULD` matches the `SHOULD conform to CurrencyFormat` it serves.
  - Skipped the accessibility requirement since `DatasetCompleteness` already requires "*FOCUS dataset* MUST have all included *custom columns* documented in publicly-available documentation, including description, purpose, and relationship to *native dataset* columns."
- **Placement:** the new line is placed last, in line with the Documentation group in the attribute ordering guidelines. The `UnitFormat` line is unchanged and included only to anchor the placement.
- **#2550:** the cryptocurrency exclusion recorded there concerned FOCUS columns gated by the operating model condition. It still holds where FOCUS columns declare their allowed currency types (`BillingCurrency`, `PaymentCurrency`, `PricingCurrency`).


```suggestion
* *Custom column* representing a currency SHOULD conform to [CurrencyFormat](#attributes.currencyformat) requirements.
* *Custom column* representing a measurement unit SHOULD conform to [UnitFormat](#attributes.unitformat) requirements.
* *Custom column* documentation SHOULD specify how to determine whether a value represents a [*national currency*](#glossary:national-currency) when the *custom column* represents a currency.
```

specification/attributes/currency_format.md
ijurica: Thanks @Matt-Cowsert!

A status update against original points:

**Addressed**
- `StringHandling` reference removed from `CurrencyFormat`. ✔
- Open question on `SHOULD NOT` vs `MUST NOT`: agreed, since the `MUST NOT` applies only to values that do not represent a *national currency*, so mixed columns still conform. ✔
- Per-value rather than per-column classification: agreed, given the SaaS example mixes USD purchase rows with OmniBucks usage rows. ✔

**Still open**
I have left a few new comments on this PR with concrete proposals for the following:
- **Circularity.** The classification rests on the Implementation Context, which is non-normative, so nothing requires it to exist or be accessible. The Known open point in the PR description acknowledges this.
- **Branch conditions.** The branches say "represents", but remain keyed to *consumption currency* ("Unless …"). Since the purpose of `CurrencyFormat` is to bind *national currency* to ISO 4217, I propose keying both branches to *national currency*.
- **Documentation part of the column-level suggestion.** `PricingCurrency` gained "MUST represent a *national currency* or a *consumption currency*", but not the documentation that establishes whether a value is a *national currency*. I propose a `PricingCurrency` documentation group in Cost and Usage and Contract Commitment (and on SKU Price in #2424).
- **Implementation Context.** Adjusted only where the above makes current statements no longer hold.
- **Custom column qualifier.** With the attribute keyed to *national currency*, it no longer needs to exclude cryptocurrency, and "national or consumption currency" cannot be evaluated for custom columns anyway. I propose "representing a currency", as originally suggested, plus a classification line in the documentation.

**Follow-up**
The documentation requirement makes the classification normative, but it still cannot be checked from the dataset itself. I will likely add a backlog item to revisit this and consider a column that carries the classification per row. I considered `PricingCurrencyCategory` but as currently defined in the SKU Price dataset (#2424) it is not suitable. It distinguishes "Payable" from "Consumable" by what the rate yields, and "Payable" is defined as "typically a *national currency*", so it does not establish whether a value is a *national currency*.

---

@Matt-Cowsert, feel free to resolve this thread in favor of the new comments.

PR #2609, opened 2026-08-06
AI #2536: Correct for Circular dependency found in 1.2, 1.3, 1.4, 1.5 model

mike-finopsorg:
Signed-off-by: Mike Fuller <mike@finops.org>

## Summary

Fixes the circular rule dependency that stops the FOCUS Validator from loading the 1.4 Cost and Usage requirements model. Tracked in #2536; the same defect is reported in #2447 and in finopsfoundation/focus_validator#149. Resolves #2447.

`CAU-CommitmentDiscountUnit-C-005-C` and `CAU-CommitmentDiscountUnit-C-006-C` declared a dependency on `CAU-CommitmentDiscountQuantity-C-000-C`, the composite that aggregates every `CommitmentDiscountQuantity` requirement. Both rules only check whether `CommitmentDiscountQuantity` is null, so the requirement they actually rely on is the nullability composite `CAU-CommitmentDiscountQuantity-C-003-C`. In the 1.4 and 1.5 models, four `CommitmentDiscountQuantity` validation rules (`-C-011-C`, `-C-012-C`, `-C-014-C`, and `-C-015-C`) depend on `CAU-CommitmentDiscountUnit-C-004-C`, the parent composite of the two rules above, so the over-broad dependency closed a cycle between the two rule families. The 1.2 and 1.3 models carry the same over-broad dependency but not the four return dependencies, so they contain no cycle.

### What changed

**Rule dependency narrowed in all four release models.** In the 1.2, 1.3, 1.4, and 1.5 `commitmentdiscountunit.json` files, the `Dependencies` entry of the two rules now names `CommitmentDiscountQuantity-C-003-C` (with the `CAU-` prefix from 1.3 on) instead of `CommitmentDiscountQuantity-C-000-C`. Nothing else in the rule bodies changes. In 1.4 and 1.5 this breaks the cycle. In 1.2 and 1.3 it aligns the declared dependency with what the rule logic requires and keeps the four models consistent.

**Model versions bumped for the published releases.** 1.2.0.3 to 1.2.0.4, 1.3.0.3 to 1.3.0.4, and 1.4 to 1.4.0.1. The 1.5 model keeps `ModelVersion` 1.5 because that release is not yet published.

[...]

PR #1956, opened 2026-02-05
AI #1906: Hotfixes to Requirements Model 1.2

mike-finopsorg:
Corrections that have been identified in the 1.2.0 version of the Requirements Model. 
TF-RM discussed the versioning approach and has decided its better to use a slightly unorthodox 4 decimal model so 1.2.0.1 as if FOCUS releases a 1.2.1 version of the spec the Model version will end up in conflict. So 1.2.0.1 is what this PR implements. 
Signed-off-by: Mike Fuller <mike@finops.org>

PR #2422, opened 2026-05-28
Feedback #2418: Errata for CommitmentDiscountStatus allowed values in 1.2 and 1.3 requirements model

Matt-Cowsert:
## Summary

The Requirements Model rule for `CommitmentDiscountStatus` allowed values (`CommitmentDiscountStatus-C-005-M` in 1.2, `CAU-CommitmentDiscountStatus-C-005-M` in 1.3) lists `Active`, `Expired`, and `Pending`. None of these values appear in the FOCUS specification.

`CommitmentDiscountStatus` is a utilization field with exactly two allowed values, `Used` and `Unused`:

* The column "Allowed Values" table and normative requirements in `commitmentdiscountstatus.md` ("MUST be 'Used' when...", "MUST be 'Unused' when...")
* The canonical commitment-discount SQL in `spec.md` filters `WHERE CommitmentDiscountStatus = 'Unused'`

The specification text for both 1.2 and 1.3 is correct; only the Requirements Model is wrong. This is a Requirements Model Correction (`ERRATA.md` category D), the same class as the existing 1.2 `#2` and 1.3 `#7`/`#8` entries. The defect originated in the 1.2 model (`ModelVersionIntroduced: 1.2`) and persists in 1.3.

This PR:

* Corrects `C-005-M` allowed values to `Used` / `Unused` in the 1.2 and 1.3 Requirements Models, matching the spec text and the sibling `CapacityReservationStatus` rule
* Bumps the Requirements Model versions: 1.2 `1.2.0.2 -> 1.2.0.3`, 1.3 `1.3.0.2 -> 1.3.0.3`
* Documents the correction in `ERRATA.md` (Version 1.2 `#3`, Version 1.3 `#11`), noting the 1.2 origin on the 1.3 entry and revising the 1.3 section preamble to drop the claim that every issue originated in 1.3

The same defect in the unratified 1.4 / `latest` model is corrected as a non-errata draft fix in #2421.

Refs #2418

### Type of Change
* [x] Bug fix
* [ ] New feature / Specification update
* [ ] Editorial / Formatting

### Author Checklist
* [x] **AI Usage Guidelines:** I have read the [FOCUS AI Usage Guidelines](/guidelines/contributors/ai-usage-guidelines.md).
* [x] **Self Review Attestation:** I have performed a self-review of my own contribution.
* [x] **AI-Generated Examples:** N/A (no example data in this PR).

PR #2622, opened 2026-08-11
AI #2369: Backport fixes in Requirements Model 1.4 to RM 1.3 and 1.2

### LINKED ISSUE #2369: [AI] Backport fixes in RM 1.4 to RM 1.3 and RM 1.2
### Detailed Description

We need to back port some fixes to bugs identified by @ruben567 to the 1.3 release.

### Definition of Done

Bugs have been addressed and 1.3.0.2 released. We should also assess if these are relevant to 1.2.

mike-finopsorg:
## Summary

Backports three Requirements Model defects from the 1.4 model into the 1.3 and 1.2 models. All three sit in the `CapacityReservationStatus` and `CommitmentDiscountStatus` rules. No specification text changes.

* **Malformed conditions on `CapacityReservationStatus` (1.2 and 1.3).** `CapacityReservationStatus-C-003-C` wrapped its single `CheckValue` clause in an `"AND"` key, and `CapacityReservationStatus-C-004-C` used `{"AND": [...]}`. `AND` is a check function that takes an `Items` argument, so the model's form is `{"CheckFunction": "AND", "Items": [...]}`. Neither condition matched it, so neither gated its rule and both requirements were evaluated on every row, which put the two rules in direct contradiction (one requires the column to be null, the other requires it not to be). C-003-C now carries the bare `CheckValue` clause; C-004-C uses the composite form.

* **Inverted null requirement on `CommitmentDiscountStatus` (1.3 only).** `CAU-CommitmentDiscountStatus-C-003-C` states that `CommitmentDiscountStatus` MUST be null when `CommitmentDiscountId` is null, but used `CheckNotValue` against `null`, asserting the opposite. Rows that satisfy the requirement raised a violation. Corrected to `CheckValue`. The 1.2 model already carries this correction from #1956.

* **Inverted and incomplete condition on `CommitmentDiscountStatus` (1.2 and 1.3).** `CommitmentDiscountStatus-C-004-C` conditioned on `CommitmentDiscountId` being null, the inverse of the stated requirement, and omitted the `ChargeCategory` is `"Usage"` clause. Replaced with an `AND` composite over `CheckNotValue` on `CommitmentDiscountId` and `CheckValue` of `"Usage"` on `ChargeCategory`, with `ChargeCategory-C-002-M` added to the rule's dependencies.

Both columns now match the 1.4 and 1.5 models, which already carry the corrected forms. `ERRATA.md` gains three entries in the 1.3 Requirements Model Corrections table and two in the 1.2 table.

This PR targets the `2536-ai-circular-dependency-found-in-14-model` branch, so #2609 merges first.

### Type of Change
* [X] Bug fix
* [ ] Editorial / Formatting

### Author Checklist


jpradocueva (approved):
$focus-pr-validation/SKILL.md has not identified any validation issue.

specification/requirements_model/releases/1.2/model_rules/columns/commitmentdiscountstatus.json
```diff
@@ -136,9 +136,19 @@
...
+          },
+          {
+            "CheckFunction": "CheckValue",
+            "ColumnName": "ChargeCategory",
+            "Value": "Usage"
+          }
+        ]
       },
       "Dependencies": [
         "CommitmentDiscountId-C-000-C"
```
Matt-Cowsert:
```suggestion
      "Dependencies": [
        "CommitmentDiscountId-C-000-C",
        "ChargeCategory-C-002-M"
```

The new condition reads `ChargeCategory`, but the rule doesn't declare it, so nothing tells the dependency graph to check `ChargeCategory` first. 1.4 declares `CAU-ChargeCategory-C-002-M` on this rule, and `CapacityReservationStatus-C-004-C` here already declares its ChargeCategory dependency.

Matt-Cowsert (commented):
The 1.2 and 1.3 model versions still read `1.2.0.4` and `1.3.0.4`, so anyone using either one cannot tell whether they have the `CapacityReservationStatus` and `CommitmentDiscountStatus` fixes. The two errata changes before this one both bumped the version: #2422 took both to .0.3, and #2609 took both to .0.4. #2369 also asks for a released 1.3 model. These should go to `1.2.0.5` and `1.3.0.5`.

Two items on `CommitmentDiscountStatus-C-004-C`:

* The new condition reads `ChargeCategory`, but `Dependencies` still lists only `CommitmentDiscountId`. Suggestions inline for both releases.
* That rule's `MustSatisfy` says "Charge Category". Every other reference in 1.2 and 1.3 uses the column ID, 1.4 already reads `ChargeCategory`, and the new ERRATA rows quote it that way too. I can't suggest on that line, so it needs a manual edit at 1.2 line 131 and 1.3 line 161:

```diff
-      "MustSatisfy": "CommitmentDiscountStatus MUST NOT be null when CommitmentDiscountId is not null and Charge Category is \"Usage\".",
+      "MustSatisfy": "CommitmentDiscountStatus MUST NOT be null when CommitmentDiscountId is not null and ChargeCategory is \"Usage\".",
```

Separately: `test_checkfunction_required_args` never would have caught these malformed conditions, in any version, so 1.5 could pick up the same bug. The test skips any node without a `CheckFunction` key (`specification/requirements_model/tests/test_checkfunction_required_args.py:25`), and `{"AND": {...}}` has none. It finds nothing on the unfixed 1.2 and 1.3 files even without the 1.4 version limit. I can log an action item for a check that a non-empty `Condition` names a `CheckFunction`.

specification/requirements_model/releases/1.3/model_rules/datasets/cost_and_usage/columns/commitmentdiscountstatus.json
```diff
@@ -166,9 +166,19 @@
...
+          },
+          {
+            "CheckFunction": "CheckValue",
+            "ColumnName": "ChargeCategory",
+            "Value": "Usage"
+          }
+        ]
       },
       "Dependencies": [
         "CAU-CommitmentDiscountId-C-000-C"
```
Matt-Cowsert:
```suggestion
      "Dependencies": [
        "CAU-CommitmentDiscountId-C-000-C",
        "CAU-ChargeCategory-C-002-M"
```

The new condition reads `ChargeCategory`, but the rule doesn't declare it, so nothing tells the dependency graph to check `ChargeCategory` first. 1.4 declares `CAU-ChargeCategory-C-002-M` on this rule, and `CAU-CapacityReservationStatus-C-004-C` here already declares its ChargeCategory dependency.

ijurica:
> The 1.2 and 1.3 model versions still read `1.2.0.4` and `1.3.0.4`, so anyone using either one cannot tell whether they have the `CapacityReservationStatus` and `CommitmentDiscountStatus` fixes. The two errata changes before this one both bumped the version: #2422 took both to .0.3, and #2609 took both to .0.4. #2369 also asks for a released 1.3 model. These should go to `1.2.0.5` and `1.3.0.5`.
> 
> Two items on `CommitmentDiscountStatus-C-004-C`:
> 
> * The new condition reads `ChargeCategory`, but `Dependencies` still lists only `CommitmentDiscountId`. Suggestions inline for both releases.
> * That rule's `MustSatisfy` says "Charge Category". Every other reference in 1.2 and 1.3 uses the column ID, 1.4 already reads `ChargeCategory`, and the new ERRATA rows quote it that way too. I can't suggest on that line, so it needs a manual edit at 1.2 line 131 and 1.3 line 161:
> 
> ```diff
> -      "MustSatisfy": "CommitmentDiscountStatus MUST NOT be null when CommitmentDiscountId is not null and Charge Category is \"Usage\".",
> +      "MustSatisfy": "CommitmentDiscountStatus MUST NOT be null when CommitmentDiscountId is not null and ChargeCategory is \"Usage\".",
> ```
> 
> Separately: `test_checkfunction_required_args` never would have caught these malformed conditions, in any version, so 1.5 could pick up the same bug. The test skips any node without a `CheckFunction` key (`specification/requirements_model/tests/test_checkfunction_required_args.py:25`), and `{"AND": {...}}` has none. It finds nothing on the unfixed 1.2 and 1.3 files even without the 1.4 version limit. I can log an action item for a check that a non-empty `Condition` names a `CheckFunction`.

Note that the 1.2/1.3 spec uses `Charge Category` (i.e., Column Display Name instead of Column ID) in that particular normative requirement, so MustSatisfy was a verbatim quote.

ijurica (approved):

Matt-Cowsert (approved):
The 1.2 and 1.3 model versions still read `1.2.0.4` and `1.3.0.4`, so a consumer holding either model cannot tell whether it carries the `CapacityReservationStatus` and `CommitmentDiscountStatus` corrections in this PR. Both prior errata PRs bumped: #2422 took them .0.2 to .0.3, #2609 took them .0.3 to .0.4. Neither file is in the diff, so it would need a manual edit:

```diff
--- a/specification/requirements_model/releases/1.2/model_details.json
-    "ModelVersion": "1.2.0.4",
+    "ModelVersion": "1.2.0.5",

--- a/specification/requirements_model/releases/1.3/model_details.json
-    "ModelVersion": "1.3.0.4",
+    "ModelVersion": "1.3.0.5",
```

@ijurica, you are right on `MustSatisfy` and I withdraw that one. The published v1.2 and v1.3 spec text both read `[Charge Category](#chargecategory)` in that nullability bullet, so the verbatim quote is correct as it stands. 1.4 reads `ChargeCategory` because the spec text itself changed in 1.4, not because the model diverged.

The test gap I raised is wider than I described. Putting the broken condition back into 1.5 still passes all 179 tests, so the release we are building now is as open to this as 1.2 and 1.3 were. The tests only inspect conditions that already name a check function, and the defect here is that the name is missing, so they skip the broken ones entirely (`tests/test_checkfunction_required_args.py:25`). Closing it takes a test that every condition names a check function in the first place.

One more that can ride on a later PR. The `CommitmentDiscountStatus` nullability rule still depends on the broad parent rule for `CommitmentDiscountId`, where 1.4 and the `CapacityReservationStatus` rule next to it both depend on the narrower nullability rule instead. Nothing breaks either way and no cycle forms. It just leaves this one rule pointing somewhere different from the 1.4 version it was backported from.

Approving. How you want to account for the version bump is your call.

specification/requirements_model/releases/1.2/model_rules/columns/commitmentdiscountstatus.json
Matt-Cowsert: Applied in 642369c, and `ChargeCategory-C-002-M` resolves in the 1.2 rule set. Nothing further from me on this one.

specification/requirements_model/releases/1.3/model_rules/datasets/cost_and_usage/columns/commitmentdiscountstatus.json
Matt-Cowsert: Applied in 642369c, and `CAU-ChargeCategory-C-002-M` resolves in the 1.3 rule set. Nothing further from me on this one.

jpradocueva:
Approved by the group without objections or abstentions during the August 27 Members' call.

flanakin (approved):

Issue #2547, opened 2026-07-21
[AI] Draft a companion details column PR for PrincipalId


Matt-Cowsert:
**Assignee(s):** @Matt-Cowsert
**Parent Issue:** #2358

**Detailed Description**
Maintainers aligned on Option A for principal identity: PrincipalId is defined as the IAM principal, and richer identity attributes (e.g., email, key names, other provider-specific values) live in a companion details column following the existing details-column pattern. The details column needs its own PR so the group can work through the column name (principal details vs. identity details), the value structure (flat key-value vs. nested), and which keys to define first, without holding up review of the PrincipalId column in PR #2360. The stated goal is for both columns to land within the 1.5 draft window. See AI #2539 for the analysis that led to this direction.

**Definition of Done**
- [ ] Draft PR adding the details column to the Cost and Usage dataset under FR #2358
- [ ] Open questions on naming and value structure framed for group discussion in the PR
- [ ] Relationship to PrincipalId (PR #2360) and AI #2539 referenced in the PR description

Matt-Cowsert:
The draft PR exists, though we're still working through the shape of the column within TF-2 this week

PR #2553, opened 2026-07-21
AI #2547: Add `Requester Details` and `Credential ID` columns to `Cost and Usage` dataset

Matt-Cowsert:
## Summary

This PR adds `RequesterDetails` and `CredentialId` to the Cost and Usage dataset, completing the three-column shape Maintainers converged on 2026-08-03. `PrincipalId`, which merged on PR #2360 on 2026-09-10, carries the principal to which access is granted. `CredentialId` carries the credential presented on the request that produced the charge. `RequesterDetails` carries the descriptive attributes for both, and for any further level a data generator represents. It also amends `PrincipalId`, which merged with #2360, and the `JsonObjectFormat` attribute.

The three columns describe one requester at different levels. The principal is the durable entity, the credential is what the request actually carried, and a service provider may sit levels between the two. `RequesterDetails` is named for that umbrella rather than for either level, because a column holding principal attributes should not be named after the credential. Earlier drafts on this branch named it `PrincipalDetails`, `IdentityDetails`, and `CredentialDetails`.

### Key Changes

* **New Column:** `CredentialId`, Conditional on `IncludesRequesterAttribution`, nullable, String. Null when no credential is associated with the charge and when the credential has no published identifier. Its requirements settle Michael's 2026-08-05 question on identifier versus name: the value is an identifier that references a credential and never a credential value that authenticates a request.
* **New Column:** `RequesterDetails`, Conditional on the same condition, nullable, JSON. A JSON array of key-value entries, each carrying a `key` naming the requester level and a `value` of descriptive attributes (`Type`, `Name`, `Email`). The `Principal` and `Credential` entries are FOCUS-defined; further levels (e.g., a delegating identity) are custom `x_` entries of the same shape, so a level named by a data generator and a level named by a later version are read the same way.
* **Nullability Coupling:** `RequesterDetails` MUST NOT be null when `PrincipalId` or `CredentialId` is not null, and MAY be null when both are null, per TF-2 on 2026-08-26. The column follows the identifiers rather than determinability, and the MAY preserves publishing requester detail on a row where both identifiers are absent. In the other direction, `RequesterDetails` includes no `Principal` entry when `PrincipalId` is null and no `Credential` entry when `CredentialId` is null, so a principal or credential with no published identifier has no entry, and a row with neither identifier can carry only custom entries.
[...]

Issue #1079, opened 2025-05-21
[FR] Standardize tag export format across cloud provider datasets


Matt-Cowsert:
### Summary

AWS FOCUS export stores TAG in MAP type. Azure fabric does not support this. Azure exports tags to maybe a string of some kind. Can the AWS export be changed so that the tag is exported in the same format as Azure in the parquet file.  The workaround is not to export tags field in AWS focus export.

### Which area does this issue relate to?

A: Missing FinOps feature, use case, or scenario

### Detailed Description

AWS FOCUS export stores TAG in MAP type. Azure fabric does not support this. Azure exports tags to maybe a string of some kind. Can the AWS export be changed so that the tag is exported in the same format as Azure in the parquet file.  The workaround is not to export tags field in AWS focus export.

### What use cases, FinOps or others, can't be performed with the existing specification unless this issue is addressed?

Tag imports into AWS from Parquet using Fabric.

### Which FinOps personas does this issue relate to?

Engineering.

### Which provider groups or specific providers does this issue relate to?

AWS

### On a scale of 1 - 4, how critical is this for your organization?

2: Important for adoption in the next 3-6 months

### Objective

Fabric able to ingest tagging from parquet export.

---

### 🧪 Use Case

TBD

---

### 🎯 Desired Outcome / Practitioner Impact

TBD

---

### 📨 Type of Request

- [ ] Refinement of an existing FOCUS attribute
- [ ] Addition of a provider-supported field not yet in FOCUS
- [ ] Net-new concept (not currently supported by providers or FOCUS)
- [ ] Supporting Content (e.g., appendices, use cases)

---

### 🏛️ Organizations Requesting or Supporting This Feature

TBD

---

### 🌐 FinOps Scope Alignment (Optional)

- [ ] Public Cloud – e.g., AWS, Azure, GCP, OCI

- [ ] Software-as-a-Service (SaaS) – e.g., Salesforce, Snowflake

- [ ] Data Center – on-prem compute and infrastructure

- [ ] Licensing – subscription or usage-based licensing models *(under development)*

- [ ] AI – cost and usage for AI models and platforms *(under development)*

- [ ] Custom – internal tooling, specialized infra *(under development)*

---

### 📊 Impacted Parties

- [ ] FinOps Practitioner – end users who analyze or act on the data

- [ ] FOCUS Data Generator – providers generating output aligned to the spec

[...]

PR #2634, opened 2026-08-12
AI #2547: Restructure `Requester Details` as a JSON array of key-value entries

### LINKED ISSUE #2358: [FR] Add Actor dimensions to Cost and Usage

udam-stitcher:
## Summary

Restructures `RequesterDetails` from a single nested JSON object into a JSON array of key-value entries, and updates the supporting schema and attribute to match.

## What changed

* **`RequesterDetails` is now a JSON array of key-value entries.** Each entry has a `key` (`Principal`, `Credential`, or an `x_`-prefixed custom key) and a `value` object of descriptive attributes (`Type`, `Name`, `Email`, plus `x_`-prefixed custom properties). Every level shares one shape, so a newly defined level adds an entry instead of widening the structure.
* **Top-level construct renamed `RequesterDetailsObject` → `RequesterDetailsValue`.** The construct is an array, not an object, so its ID, headings, display name, and schema (`requesterdetailsvalueschema.json`, `$id`, `title`) now say "Value". References to genuine objects (each entry's `value`) are unchanged.
* **`JsonObjectFormat` relaxed to permit a top-level array.** The ECMA-404 "object" allowance becomes "object or an array". This only relaxes the attribute, so existing object-valued columns (`ContractApplied`, `AllocatedMethodDetails`, `CommitmentProgramEligibilityDetails`, `ContractCommitmentApplicability`) remain conformant unchanged.
* **Schema and examples updated** to the array shape; each entry validates against one uniform value schema under `additionalProperties: false`.

## Note

`Entry.key MUST be unique within RequesterDetailsValue` is normative but not expressible in JSON Schema (`uniqueItems` only rejects wholly identical entries), so uniqueness requires a build-time check.

### Type of Change
* [ ] Bug fix
* [ ] Editorial / Formatting

### Author Checklist

### DIFF
--- a/specification/datasets/cost_and_usage/columns/requesterdetails.md
+++ b/specification/datasets/cost_and_usage/columns/requesterdetails.md
@@ -1,10 +1,10 @@
 # Requester Details
 
-Requester Details represents a set of properties describing the [*requester*](#glossary:requester) on whose behalf a request that produced a [*charge*](#glossary:charge) was made. A [*service provider*](#glossary:service-provider) commonly represents a *requester* at more than one level: the [*principal*](#glossary:principal) to which access to a [*resource*](#glossary:resource) or [*service*](#glossary:service) is granted, and the [*credential*](#glossary:credential) that *principal* presented on the request. Requester Details describes the *principal* identified by [Principal ID](#datamodel.costandusage.principalid) through its top-level properties, and describes the *credential* identified by [Credential ID](#datamodel.costandusage.credentialid) in a nested `Credential` property.
+Requester Details represents a set of properties describing the [*requester*](#glossary:requester) on whose behalf a request that produced a [*charge*](#glossary:charge) was made. A *requester* is commonly represented at more than one level: the [*principal*](#glossary:principal) to which access to a [*resource*](#glossary:resource) or [*service*](#glossary:service) is granted, and the [*credential*](#glossary:credential) that *principal* presented on the request. Requester Details carries these as a collection of key-value entries, where the `Principal` entry describes the *principal* identified by [Principal ID](#datamodel.costandusage.principalid) and the `Credential` entry describes the *credential* identified by [Credential ID](#datamodel.costandusage.credentialid).
 
-Where Principal ID and Credential ID provide opaque identifiers suited to grouping and joining, Requester Details carries the descriptive attributes a *service provider* publishes alongside those identifiers, such as a display name, an email address, and the kind of *principal* or *credential* the identifier represents. A request may be authenticated with an API key acting under a named user, or with a workload identity acting under a service account; in each case the named user or service account is the *principal*, and the API key or workload identity is the *credential*. Keeping the descriptive attributes of the *principal* at the top level means the most common analyses (e.g., grouping *charges* by principal name or email) read directly from those properties.
+Where Principal ID and Credential ID provide opaque identifiers suited to grouping and joining, Requester Details carries the descriptive attributes published alongside those identifiers, such as a display name, an email address, and the kind of *principal* or *credential* the identifier represents. A request may be authenticated with an API key acting under a named user, or with a workload identity acting under a service account; in each case the named user or service account is the *principal*, and the API key or workload identity is the *credential*.
 
-Requester Details helps practitioners attribute cost to a recognizable *requester* while preserving Principal ID and Credential ID as stable, opaque keys, and gives *service providers* a defined location for identity attributes that would otherwise appear as custom columns. A *service provider* that represents a *requester* at levels beyond the *principal* and the *credential* describes each additional level in its own nested object, so the levels this specification defines and the levels a *service provider* defines share one shape.
+Requester Details helps practitioners attribute cost to a recognizable *requester* while preserving Principal ID and Credential ID as stable, opaque keys, and provides a defined location for identity attributes that would otherwise appear as custom columns.
 
 ## Requirements
 
@@ -22,7 +22,7 @@ RequesterDetails MUST adhere to the following requirements:
 
 ## Requester Details Object
 
-Requester Details consists of a valid JSON object whose top-level properties describe the *principal* identified by PrincipalId. An optional `Credential` property describes the *credential* identified by CredentialId.
+Requester Details consists of an array of entries, where each entry is an object carrying a `key` property naming what the entry describes and a `value` property carrying its attributes. The `Principal` entry describes the *principal* identified by PrincipalId, and the `Credential` entry describes the *credential* identified by CredentialId.
 
 The following section details the normative requirements for the RequesterDetailsObject and its nested properties. For a logical overview of the expected content, see the [Schema Structure](#datamodel.costandusage.requesterdetails.requesterdetailsobject.objectschemastructure) and [Object Example](#datamodel.costandusage.requesterdetails.requesterdetailsobject.objectexample) sections.
 
@@ -31,80 +31,93 @@ The following section details the normative requirements for the RequesterDetail
 RequesterDetailsObject MUST adhere to the following requirements:
 
 * RequesterDetailsObject MUST conform to the [RequesterDetailsObjectSchema](#schemas.costandusage.requesterdetailsobjectschema) JSON Schema.
-* RequesterDetailsObject MUST NOT include properties that are not applicable to the *requester* associated with the *charge*.
-* RequesterDetailsObject.Name MUST represent a readable display name for the *principal* identified by PrincipalId.
-* RequesterDetailsObject.Email MUST represent an email address associated with the *principal* identified by PrincipalId.
-* RequesterDetailsObject.Type MUST represent the kind of *principal* identified by PrincipalId.
-* RequesterDetailsObject.Type MUST be a consistent, readable display value.
-* RequesterDetailsObject SHOULD include Type when the *service provider* distinguishes kinds of *principals*.
-* RequesterDetailsObject.Credential MUST represent the *credential* identified by CredentialId.
-* RequesterDetailsObject.Credential.Id MUST match CredentialId.
-* RequesterDetailsObject.Credential.Type MUST represent the kind of *credential* identified by CredentialId.
-* RequesterDetailsObject.Credential.Type MUST be a consistent, readable display value.
-* RequesterDetailsObject.Credential.Name MUST represent a readable display name for the *credential* identified by CredentialId.
-* RequesterDetailsObject SHOULD represent a level of the *requester* that is neither the *principal* nor the *credential* as a custom property whose value is an object.
-* RequesterDetailsObject custom properties representing a level of the *requester* SHOULD include the Id, Type, and Name entries defined for RequesterDetailsObject.Credential.
+* RequesterDetailsObject MUST be an array of entries.
+* RequesterDetailsObject MUST include at least one entry.
+* RequesterDetailsObject MUST NOT include entries that are not applicable to the *requester* associated with the *charge*.
+
+RequesterDetailsObject entries MUST adhere to the following requirements:
+
+* Entry in RequesterDetailsObject MUST include a `key` property.
+* Entry in RequesterDetailsObject MUST include a `value` property.
+* Entry.key in RequesterDetailsObject MUST be "Principal", "Credential", or a custom key.
+* Entry.key in RequesterDetailsObject MUST be unique within RequesterDetailsObject.
+* Entry.value in RequesterDetailsObject MUST be an object carrying the descriptive attributes of the *requester* level named by Entry.key.
+* Entry.value.Type MUST represent the kind of entity described by the entry.
+* Entry.value.Name MUST represent a readable display name for the entity described by the entry.
+* Entry.value.Email MUST represent an email address associated with the entity described by the entry.
+* Custom entry in RequesterDetailsObject MUST represent a level of the *requester* that is neither the *principal* nor the *credential*.
 
 ### Object Schema Structure
 
-RequesterDetails contains a structured JSON object describing the *principal* identified by PrincipalId and the *credential* identified by CredentialId.
+RequesterDetails contains an array of key-value entries describing the *principal* identified by PrincipalId and the *credential* identified by CredentialId.
 
-<div class="h7-nonindex">Top-Level Properties</div>
+<div class="h7-nonindex">Entry Properties</div>
+
+Each entry in the array contains the following properties:
 
 | Property | Type | Required | Description |
 | :--- | :--- | :--- | :--- |
-| `Name` | String | False | Readable display name for the *principal* identified by Principal ID. |
-| `Email` | String | False | Email address associated with the *principal* identified by Principal ID. |
-| `Type` | String | False | The kind of *principal* identified by Principal ID (e.g., "User", "Service Account", "Application"). |
-| `Credential` | Object | False | The *credential* identified by Credential ID. |
+| `key` | String | True | Names what the entry describes. One of the FOCUS-defined keys below, or a custom key. |
+| `value` | Object | True | The attributes applicable to `key`. |
 
-<div class="h7-nonindex">Credential Object</div>
+<div class="h7-nonindex">FOCUS-Defined Keys</div>
 
-The `Credential` property contains an object with the following entries:
+| Key | Required | Description |
+| :--- | :--- | :--- |
+| `Principal` | False | The *principal* identified by Principal ID. |
+| `Credential` | False | The *credential* identified by Credential ID. |
 
-| Key | Type | Required | Description |
+<div class="h7-nonindex">Value Properties</div>
+
+The `value` of each entry is an object. FOCUS supports the properties below for describing the entity; a data generator uses them as necessary and may include additional custom properties.
+
+| Property | Type | Required | Description |
 | :--- | :--- | :--- | :--- |
-| `Id` | String | False | The value of Credential ID for the *credential* described by this object. |
-| `Type` | String | True | The kind of *credential* identified by Credential ID (e.g., "API Key", "Session", "Delegated Role"). |
-| `Name` | String | False | Readable display name for the *credential* identified by Credential ID. |
+| `Type` | String | False | The kind of entity the entry describes (e.g., "User", "Service Account", "API Key", "Session"). |
+| `Name` | String | False | Readable display name for the entity the entry describes. |
+| `Email` | String | False | Email address associated with the entity the entry describes. |
 
 ### Object Implementation Guidance
 
-<div class="h7-nonindex">Custom Properties</div>
+<div class="h7-nonindex">Custom Keys and Properties</div>
 
-To facilitate querying data across *principals* and across *service providers*, a data generator may include one or more custom properties. These may be placed at the top level of the object (alongside `Name`) or nested within the `Credential` object. Custom keys must be prefixed with "x_" followed by PascalCase format (e.g., `x_MyCustomKey`) to make them easy to identify as well as prevent collisions with FOCUS-defined keys.
+To facilitate querying data across *principals* and across [*service providers*](#glossary:service-provider), a data generator may include one or more custom entries, or custom properties within the `value` of a FOCUS-defined entry. Custom entry keys and custom property names MUST be prefixed with "x_" followed by PascalCase format (e.g., `x_MyCustomKey`) to make them easy to identify as well as prevent collisions with FOCUS-defined keys.
 
 <div class="h7-nonindex">Additional Levels</div>
 
-The `Credential` object is the shape a level of the *requester* takes: `Id` for the identifier the *service provider* publishes for that level, `Type` for the kind of entity the level represents, and `Name` for a readable display name. A *service provider* that represents a *requester* at further levels (e.g., a tenant, a delegating identity, an intermediate role) describes each one as a custom property holding an object of that same shape.
+Every entry value uses the same set of properties, so the *principal*, the *credential*, and any additional level share one shape. A *requester* represented at further levels (e.g., a tenant, a delegating identity, an intermediate role) carries each one as a custom entry whose value uses those same properties.
 
 This specification defines the *principal* and the *credential* because they are the two levels that appear across *service providers*. It does not enumerate the levels between them, which vary by authorization model. Reusing one shape for every level means a level named by a data generator and a level named by a later version of this specification are read the same way.
 
-<div class="h7-nonindex">Property Placement</div>
+<div class="h7-nonindex">Entry Placement</div>
 
-Attributes describing the *principal* identified by PrincipalId belong at the top level of the object. Attributes describing the *credential* identified by CredentialId belong in the `Credential` object. A top-level property whose value is an object describes a level of the *requester* rather than an attribute of the *principal*. Placing the descriptive attributes of the *principal* at the top level allows analyses that group or filter on those attributes to read them directly.
-
-`Credential.Id` repeats the value carried in CredentialId. A *service provider* that publishes the *credential* identifier alongside its descriptive attributes has a FOCUS-defined key available for it rather than a custom property. Analyses that group or filter on the *credential* read CredentialId, which is available without parsing the object.
-
-Where a *service provider* publishes no identifier for the *credential* presented, CredentialId is null and the `Credential` property is omitted. The attributes describing the *principal* remain at the top level.
+Attributes describing the *principal* identified by PrincipalId belong in the `value` of the `Principal` entry. Attributes describing the *credential* identified by CredentialId belong in the `value` of the `Credential` entry. An entry is present only when attributes for it are published, so a *charge* with a *principal* and no identified *credential* carries a `Principal` entry alone.
 
 ### Object Example
 
-Here is a basic example of the object format.
+Here is an example of the object format.
 
 * For more detailed examples, please see this column's entry in the JSON Object Examples appendix entry [here](#appendix.examples:jsonobject.examples:requesterdetails).
 * For the JSON schema, please see [Requester Details Object Schema](#schemas.costandusage.requesterdetailsobjectschema).
 
 ```json
-{
-  "Name" : "Alex Rivera",
-  "Email" : "alex.rivera@example.com",
-  "Type" : "User",
-  "Credential" : {
-    "Type" : "API Key",
-    "Name" : "prod-ingest-key"
+[
+  {
+    "key" : "Principal",
+    "value" : {
+      "Name" : "Alex Rivera",
+      "Email" : "alex.rivera@example.com",
+      "Type" : "User"
+    }
+  },
+  {
+    "key" : "Credential",
+    "value" : {
+      "Type" : "API Key",
+      "Name" : "prod-ingest-key"
+    }
   }
-}
+]
 ```
 
 ### Object ID
--- a/specification/datasets/cost_and_usage/columns/requesterdetails.md
+++ b/specification/datasets/cost_and_usage/columns/requesterdetails.md
@@ -46,6 +46,8 @@ RequesterDetailsObject entries MUST adhere to the following requirements:
 * Entry.value.Name MUST represent a readable display name for the entity described by the entry.
 * Entry.value.Email MUST represent an email address associated with the entity described by the entry.
 * Custom entry in RequesterDetailsObject MUST represent a level of the *requester* that is neither the *principal* nor the *credential*.
+* Custom Entry.key in RequesterDetailsObject MUST begin with the string "x_" followed by PascalCase.
+* Custom property name in Entry.value in RequesterDetailsObject MUST begin with the string "x_" followed by PascalCase.
 
 ### Object Schema Structure
 
--- a/specification/datasets/cost_and_usage/columns/requesterdetails.md
+++ b/specification/datasets/cost_and_usage/columns/requesterdetails.md
@@ -12,44 +12,44 @@ Requester Details helps practitioners attribute cost to a recognizable *requeste
 
 RequesterDetails MUST adhere to the following requirements:
 
-* RequesterDetails MUST be of type JSON Object (serialized as a String where necessary).
+* RequesterDetails MUST be a valid JSON value (serialized as a String where necessary).
 * RequesterDetails MUST conform to [StringHandling](#attributes.stringhandling) requirements.
 * RequesterDetails MUST conform to [JsonObjectFormat](#attributes.jsonobjectformat) requirements.
 * RequesterDetails MUST adhere to the following nullability requirements:
   * RequesterDetails MAY be null when PrincipalId is null and CredentialId is null.
   * RequesterDetails MUST NOT be null when PrincipalId is not null or CredentialId is not null.
-* RequesterDetails MUST conform to [RequesterDetailsObject](#datamodel.costandusage.requesterdetails.requesterdetailsobject) requirements when RequesterDetails is not null.
+* RequesterDetails MUST conform to [RequesterDetailsValue](#datamodel.costandusage.requesterdetails.requesterdetailsvalue) requirements when RequesterDetails is not null.
 
-## Requester Details Object
+## Requester Details Value
 
-Requester Details consists of an array of entries, where each entry is an object carrying a `key` property naming what the entry describes and a `value` property carrying its attributes. The `Principal` entry describes the *principal* identified by PrincipalId, and the `Credential` entry describes the *credential* identified by CredentialId.
+Requester Details consists of a JSON array of entries, where each entry is an object carrying a `key` property naming what the entry describes and a `value` property carrying its attributes. The `Principal` entry describes the *principal* identified by PrincipalId, and the `Credential` entry describes the *credential* identified by CredentialId.
 
-The following section details the normative requirements for the RequesterDetailsObject and its nested properties. For a logical overview of the expected content, see the [Schema Structure](#datamodel.costandusage.requesterdetails.requesterdetailsobject.objectschemastructure) and [Object Example](#datamodel.costandusage.requesterdetails.requesterdetailsobject.objectexample) sections.
+The following section details the normative requirements for the RequesterDetailsValue and its nested properties. For a logical overview of the expected content, see the [Schema Structure](#datamodel.costandusage.requesterdetails.requesterdetailsvalue.valueschemastructure) and [Value Example](#datamodel.costandusage.requesterdetails.requesterdetailsvalue.valueexample) sections.
 
-### Object Requirements
+### Value Requirements
 
-RequesterDetailsObject MUST adhere to the following requirements:
+RequesterDetailsValue MUST adhere to the following requirements:
 
-* RequesterDetailsObject MUST conform to the [RequesterDetailsObjectSchema](#schemas.costandusage.requesterdetailsobjectschema) JSON Schema.
-* RequesterDetailsObject MUST be an array of entries.
-* RequesterDetailsObject MUST include at least one entry.
-* RequesterDetailsObject MUST NOT include entries that are not applicable to the *requester* associated with the *charge*.
+* RequesterDetailsValue MUST conform to the [RequesterDetailsValueSchema](#schemas.costandusage.requesterdetailsvalueschema) JSON Schema.
+* RequesterDetailsValue MUST be an array of entries.
+* RequesterDetailsValue MUST include at least one entry.
+* RequesterDetailsValue MUST NOT include entries that are not applicable to the *requester* associated with the *charge*.
 
-RequesterDetailsObject entries MUST adhere to the following requirements:
+RequesterDetailsValue entries MUST adhere to the following requirements:
 
-* Entry in RequesterDetailsObject MUST include a `key` property.
-* Entry in RequesterDetailsObject MUST include a `value` property.
-* Entry.key in RequesterDetailsObject MUST be "Principal", "Credential", or a custom key.
-* Entry.key in RequesterDetailsObject MUST be unique within RequesterDetailsObject.
-* Entry.value in RequesterDetailsObject MUST be an object carrying the descriptive attributes of the *requester* level named by Entry.key.
+* Entry in RequesterDetailsValue MUST include a `key` property.
+* Entry in RequesterDetailsValue MUST include a `value` property.
+* Entry.key in RequesterDetailsValue MUST be "Principal", "Credential", or a custom key.
+* Entry.key in RequesterDetailsValue MUST be unique within RequesterDetailsValue.
+* Entry.value in RequesterDetailsValue MUST be an object carrying the descriptive attributes of the *requester* level named by Entry.key.
 * Entry.value.Type MUST represent the kind of entity described by the entry.
 * Entry.value.Name MUST represent a readable display name for the entity described by the entry.
 * Entry.value.Email MUST represent an email address associated with the entity described by the entry.
-* Custom entry in RequesterDetailsObject MUST represent a level of the *requester* that is neither the *principal* nor the *credential*.
-* Custom Entry.key in RequesterDetailsObject MUST begin with the string "x_" followed by PascalCase.
-* Custom property name in Entry.value in RequesterDetailsObject MUST begin with the string "x_" followed by PascalCase.
+* Custom entry in RequesterDetailsValue MUST represent a level of the *requester* that is neither the *principal* nor the *credential*.
+* Custom Entry.key in RequesterDetailsValue MUST begin with the string "x_" followed by PascalCase.
+* Custom property name in Entry.value in RequesterDetailsValue MUST begin with the string "x_" followed by PascalCase.
 
-### Object Schema Structure
+### Value Schema Structure
 
 RequesterDetails contains an array of key-value entries describing the *principal* identified by PrincipalId and the *credential* identified by CredentialId.
 
@@ -79,7 +79,7 @@ The `value` of each entry is an object. FOCUS supports the properties below for 
 | `Name` | String | False | Readable display name for the entity the entry describes. |
 | `Email` | String | False | Email address associated with the entity the entry describes. |
 
-### Object Implementation Guidance
+### Value Implementation Guidance
 
 <div class="h7-nonindex">Custom Keys and Properties</div>
 
@@ -95,12 +95,12 @@ This specification defines the *principal* and the *credential* because they are
 
 Attributes describing the *principal* identified by PrincipalId belong in the `value` of the `Principal` entry. Attributes describing the *credential* identified by CredentialId belong in the `value` of the `Credential` entry. An entry is present only when attributes for it are published, so a *charge* with a *principal* and no identified *credential* carries a `Principal` entry alone.
 
-### Object Example
+### Value Example
 
-Here is an example of the object format.
+Here is an example of the value format.
 
 * For more detailed examples, please see this column's entry in the JSON Object Examples appendix entry [here](#appendix.examples:jsonobject.examples:requesterdetails).
-* For the JSON schema, please see [Requester Details Object Schema](#schemas.costandusage.requesterdetailsobjectschema).
+* For the JSON schema, please see [Requester Details Value Schema](#schemas.costandusage.requesterdetailsvalueschema).
 
 ```json
 [
@@ -122,13 +122,13 @@ Here is an example of the object format.
 ]
 ```
 
-### Object ID
+### Value ID
 
-RequesterDetailsObject
+RequesterDetailsValue
 
-### Object Display Name
+### Value Display Name
 
-Requester Details Object
+Requester Details Value
 
 ## Column ID
 
@@ -152,7 +152,7 @@ A set of properties describing the *requester* on whose behalf a request that pr
 | Allows nulls | True |
 | Data type | JSON |
 | Value format | [JSON Object Format](#attributes.jsonobjectformat) |
-| Object | [RequesterDetailsObject](#datamodel.costandusage.requesterdetails.requesterdetailsobject) |
+| Value | [RequesterDetailsValue](#datamodel.costandusage.requesterdetails.requesterdetailsvalue) |
 
 ## Version Introduced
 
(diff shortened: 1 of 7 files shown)


specification/appendix/json_object_examples/credential_details_examples.md
```diff
@@ -4,51 +4,51 @@ The examples below are not exhaustive and may change over time. Service provider
...
+Scenario: A platform subscription billed at the account level, with no entity in the StackLens identity and access management model associated with it and no *credential* presented. PrincipalId, CredentialId, and CredentialDetails are all null.
 
 | ServiceProviderName | ServiceName | PrincipalId | CredentialId | CredentialDetails |
 |---------------------|-------------|-------------|--------------|-------------------|
 | StackLens | Observability | null | null | null |
 
 ## Aura Web (Provider-Specific Attributes)
 
-Scenario: A *service provider* publishes an identity attribute that has no FOCUS-defined property. The attribute is carried as a custom property prefixed with "x_".
+Scenario: Aura Web publishes an identity attribute that has no FOCUS-defined property. The attribute is carried as a custom property prefixed with "x_" inside the `Principal` entry. An attribute that describes neither the *principal* nor the *credential* is carried as a custom entry whose `key` is prefixed with "x_".
```
jpradocueva: ### F-001 — error — [Line 47](https://github.com/FinOps-Open-Cost-and-Usage-Spec/FOCUS_Spec/blob/2e439d47a78c3db88812dc5648b16675b3ef7ccf/specification/appendix/json_object_examples/credential_details_examples.md#L47)

**Changed content:** `An attribute that describes neither the principal nor the credential is carried as a custom entry...`

**Rule:** `validation-schema-accuracy-01` — Writing Specification Content & Review Guidelines → Validation & Schema Accuracy → Mathematical & Schema Accuracy; position 1. “AI reviewers MUST rigorously calculate, parse, and verify all data within examples...”

**Engineering context:** The example permits an extension entry independent of the predefined `Principal` and `Credential` entries.

**Related locations:** `credentialdetails.md` lines 38, 43, and 108.

**Problem:** This example permits an entry describing neither the principal nor credential, but `credentialdetails.md` line 38 prohibits entries inapplicable to the corresponding `PrincipalId` or `CredentialId`. That document also permits custom keys and custom entries. These statements cannot all be applied consistently.

**Impact:** Producers and validators may disagree about whether charge-level custom entries are conformant.

**Behavioral effect of the correction:** Behavioral effect requires author clarification: decide whether independent custom entries are permitted and align the requirement, guidance, schema, and example.

**Suggested correction:** If independent custom entries are allowed, narrow the line 38 prohibition so it does not exclude them. Otherwise, remove the independent custom-entry guidance and example.

specification/datasets/cost_and_usage/columns/credentialdetails.md
```diff
@@ -31,55 +33,91 @@ The following section details the normative requirements for the CredentialDetai
...
-* CredentialDetailsObject.Credential.Type MUST represent the kind of *credential* identified by CredentialId.
-* CredentialDetailsObject.Credential.Type MUST be a consistent, readable display value.
-* CredentialDetailsObject.Credential.Name MUST represent a readable display name for the *credential* identified by CredentialId.
+* CredentialDetailsObject MUST be an array of entries.
+* CredentialDetailsObject MUST include at least one entry.
+* CredentialDetailsObject MUST NOT include entries that are not applicable to the corresponding PrincipalId or CredentialId.
+
+CredentialDetailsObject entries MUST adhere to the following requirements:
+
+* Entry in CredentialDetailsObject MUST include a `key` property and a `value` property.
```
jpradocueva: ### F-002 — error — [Line 42](https://github.com/FinOps-Open-Cost-and-Usage-Spec/FOCUS_Spec/blob/2e439d47a78c3db88812dc5648b16675b3ef7ccf/specification/datasets/cost_and_usage/columns/credentialdetails.md#L42)

**Changed content:** `Entry in CredentialDetailsObject MUST include a key property and a value property.`

**Rule:** `normative-language-requirements-03` — Writing Specification Content & Review Guidelines → Normative Language & Requirements → Structure; position 3. “Each bullet MUST express exactly one verifiable state. Split bullets that combine multiple obligations.”

**Engineering context:** Presence of `key` and presence of `value` are separately testable schema constraints.

**Problem:** This bullet combines two independently verifiable property-presence obligations.

**Impact:** The requirement cannot be represented as one atomic validation assertion and violates the required normative structure.

**Suggested correction:** Split the requirement into two bullets.

```suggestion
* Entry in CredentialDetailsObject MUST include a `key` property.
* Entry in CredentialDetailsObject MUST include a `value` property.
```

specification/datasets/cost_and_usage/columns/credentialdetails.md
```diff
@@ -31,55 +33,91 @@ The following section details the normative requirements for the CredentialDetai
...
+| `Id` | String | False | The value of Credential ID for the *credential* described by this entry. |
 | `Type` | String | True | The kind of *credential* identified by Credential ID (e.g., "API Key", "Session", "Delegated Role"). |
 | `Name` | String | False | Readable display name for the *credential* identified by Credential ID. |
 
 ### Object Implementation Guidance
 
-<div class="h7-nonindex">Custom Properties</div>
+<div class="h7-nonindex">Custom Keys and Properties</div>
+
+To facilitate querying data across *principals* and across [*service providers*](#glossary:service-provider), a data generator may include one or more custom entries, or custom properties within the `value` of a FOCUS-defined entry. Custom entry keys and custom property names must be prefixed with "x_" followed by PascalCase format (e.g., `x_MyCustomKey`) to make them easy to identify as well as prevent collisions with FOCUS-defined keys.
```
jpradocueva: ### F-003 — error — [Line 108](https://github.com/FinOps-Open-Cost-and-Usage-Spec/FOCUS_Spec/blob/2e439d47a78c3db88812dc5648b16675b3ef7ccf/specification/datasets/cost_and_usage/columns/credentialdetails.md#L108)

**Changed content:** `Custom entry keys and custom property names must be prefixed with "x_"...`

**Rule:** `normative-language-requirements-01` — Writing Specification Content & Review Guidelines → Normative Language & Requirements → BCP-14 Keywords; position 1. “Use MUST, MUST NOT, SHOULD, SHOULD NOT, MAY (uppercase).”

**Engineering context:** The `x_` prefix is a namespace rule that prevents extensions from colliding with future FOCUS-defined names.

**Problem:** The sentence imposes a mandatory constraint using lowercase “must” in implementation guidance. The normative entry requirements permit custom keys but do not define their required prefix.

**Impact:** Validators cannot reliably treat the extension namespace constraint as required.

**Behavioral effect of the correction:** Normative behavior changes as described: the `x_` prefix becomes an explicit conformance requirement for custom entry keys and custom property names.

**Suggested correction:** Move the prefix constraints into the applicable Requirements section, use uppercase `MUST`, and express the custom-entry-key and custom-property-name constraints as separate bullets.

jpradocueva (commented):

Snapshot totals: 1 conversation comment, 0 reviews, and 0 inline review comments.

| --- | --- | --- |
| F-001 | Submit inline comment | Conflicting custom-entry rules require author clarification. Behavioral effect requires author clarification. |
| F-002 | Submit inline comment | One bullet combines two testable obligations; splitting it does not change schema or normative behavior. |
| F-003 | Submit inline comment | The extension-prefix constraint is non-normative; correcting it changes normative behavior by making the `x_` prefix an explicit conformance requirement. |

specification/datasets/cost_and_usage/columns/credentialdetails.md
```diff
@@ -31,55 +33,92 @@ The following section details the normative requirements for the CredentialDetai
...
-* CredentialDetailsObject.Type MUST be a consistent, readable display value.
-* CredentialDetailsObject SHOULD include Type when the *service provider* distinguishes kinds of *principals*.
-* CredentialDetailsObject.Credential MUST represent the *credential* identified by CredentialId.
-* CredentialDetailsObject.Credential.Id MUST match CredentialId.
-* CredentialDetailsObject.Credential.Type MUST represent the kind of *credential* identified by CredentialId.
-* CredentialDetailsObject.Credential.Type MUST be a consistent, readable display value.
-* CredentialDetailsObject.Credential.Name MUST represent a readable display name for the *credential* identified by CredentialId.
+* CredentialDetailsObject MUST be an array of entries.
+* CredentialDetailsObject MUST include at least one entry.
+* CredentialDetailsObject MUST NOT include entries that are not applicable to the corresponding PrincipalId or CredentialId.
```
udam-stitcher:
```suggestion
* CredentialDetailsObject MUST NOT include entries that are not applicable to the corresponding PrincipalId, CredentialId, or requester-related details.
```

specification/appendix/json_object_examples/credential_details_examples.md
udam-stitcher:
```suggestion
Scenario: Aura Web publishes a Principal-related attribute that has no FOCUS-defined property. The attribute is carried as a custom property prefixed with "x_" inside the `Principal` entry in CredentialDetails. An attribute that describes neither the *principal* nor the *credential* is carried as a custom entry whose `key` is prefixed with "x_".
```

Matt-Cowsert:
I support this pull request landing in 1.5, both the CredentialDetails entries structure and the one-line `JsonObjectFormat` change it depends on. This is the written assessment for action item #2637, following today's Open Forum. Two points where I would change the proposal, and two corrections to the record, are below.

## What I agree with

**The entries structure, adopted in 1.5.** CredentialDetails becomes an array of `{key, value}` entries, one for `Principal` and one for `Credential`, rather than a single object with named properties. Adopting it now is part of adding a new column, which the [change guidelines](https://github.com/FinOps-Open-Cost-and-Usage-Spec/FOCUS_Spec/blob/working_draft/guidelines/contributors/spec-change-guidelines.md) classify as a Compatible Change: no consumer modifies anything. Converting a published CredentialDetails object to entries in 1.6 would instead be an Improvement carrying a Migration Compatible Change, since the column and its supported feature survive but every consumer rewrites the queries and ingestion that read it. FOCUS has made this change once already: the 1.4 changelog records the revised `ContractApplied` format as a Migration Compatible Change, and states that queries parsing the `ContractApplied` JSON structure need updating. CredentialDetails has not shipped in any release, since its own pull request, #2553, is still open, so that option is available until 1.5 publishes. Neither structure costs consumers anything to adopt today, so the question is not which is cheaper now but which serves practitioners over the life of the column. FOCUS would rather not make a later change at all, whatever its impact class, and on that test the entries structure holds up better, because newly defined keys land as entries rather than as structural changes. Adopting it here also leaves the existing JSON object columns alone. This pull request does not touch ContractApplied, AllocatedMethodDetails, CommitmentProgramEligibilityDetails, or ContractCommitmentApplicability, and the trailing requirements of the attribute stay scoped to objects, so those four keep the definitions, schemas, and requirements they have today.

**The `JsonObjectFormat` relaxation.** It is required by the structure, and on its own it is a Compatible Change: all four columns that already conform to the attribute stay conformant unchanged and no consumer modifies anything.

## Where I disagree, with alternatives

**The attribute change should land here rather than separately.** This pull request offers to split it out if the group prefers. I would not. Adopting entries while deferring the attribute that permits them leaves the specification self-contradictory for a release, and the relaxation creates no consumer cost that would justify handling it on its own timeline.

**Entry-key uniqueness belongs in the requirements model, not in a build-time check.** Open question 1 here states that catching a duplicate `Principal` entry requires a build-time check. A build check runs against the specification's own examples, which are correct today; the violation it needs to catch happens in generator output, which the build never sees. The requirements model is the artifact that reaches generator output, through the validators built on it.

Encoding it costs more than a rule addition. The model has no path-scoped uniqueness function today. `JSONCheckPathDistinctParent` asserts referential cardinality, that each child value has one parent value, and because it counts distinct parents, two identical entries collapse to one and pass. `CheckNoDuplicates` operates on a column rather than a path, and none of the remaining check functions covers distinctness over a path. That means a new check function in the catalogue and a validator that implements it.

Action item #2635 is easy to read as covering this and does not: its Definition of Done compares property key spellings across column prose, entry tables, examples, and type definitions against the model, which is a different defect class.

## Two corrections to the record

**What the attribute change does.** The description on #2637 says the change lets a data generator deliver "a serialized or a non-serialized representation." The diff is narrower. It leaves "MUST contain a serialized JSON string" intact and edits only what follows, from "the ECMA 404 definition of an object" to "an object or an array." The column must still contain a serialized string, and the one change is that the string may serialize an array as well as an object. Whether these columns can be delivered as native types rather than serialized strings is open in #979, and #1079 asks for one standardized export format. This edit does not reach either.

**My framing of the consistency cost.** On #2637 I said that carrying the new structure on one column in 1.5 is defensible given the inconsistencies already present. I said that from assumption rather than from the schemas, and it is weaker than I made it sound. Three of the four columns that already conform to `JsonObjectFormat` share one top-level structure, an object with a single required array property, so there is a dominant pattern and the entries array does not follow it. The table below has the detail. My conclusion does not change, since timing decides it, but the consistency cost is real rather than nominal.

## How I ranked the arguments

Three criteria set the order below. The first is a judgment call rather than a given. Change Impact Classification matters more than day-to-day query and storage cost, because a change of impact class forces every consumer to act once and cannot be worked around, while query cost is continuous and a practitioner can mitigate it in how they materialize the column. Someone who weighted recurring query cost first would order these differently, and that reading has the better evidence behind it, since the query measurement is the best-supported item on either list. A structural property matters more than a count of what generators ship today, since the specification is forward-looking. An argument that applies equally to both options does not move the decision at all. Both lists are ordered from most to least decisive.

**Pros**

* Adopting entries now is a Compatible Change and converting later is a Migration Compatible Change.
* A newly defined FOCUS key adds an entry rather than widening a structure practitioners have already materialized, whether that is a Parquet `list<struct>` or a relational key-value table. FOCUS has not yet added a key to a published JSON object column, so this benefit is prospective rather than demonstrated.
* Storage and scan cost favor a fixed-schema array of key-value objects over an object whose keys are open-ended, which is @udam-stitcher's argument. It applies to native materialization rather than to the serialized string the attribute requires: serialized, the entries form of this pull request's own example runs 173 bytes against 127 for the object, and 100 against 70 for a charge carrying a principal alone. Query cost is a separate matter, covered in the cons.
* Entries carry attributes attached to a single key without a further nesting level. Google's FOCUS export puts `x_Inherited` and `x_Namespace` alongside each key and value in its `x_Tags` field. A map whose values are objects expresses the same thing, at the cost of one more level.
* The structure has production precedent, though a minority of it. Of the generators publishing their tag representation, two are entries-shaped, Tencent in its bill exports and Google in its BigQuery export, against four publishing object maps.
* Value-level validation is preserved. The schema constrains the `Principal` and `Credential` values with `oneOf` and `const` under `additionalProperties: false`. That removes an objection rather than adding a reason.

**Cons**

* The measured query cost favors the object, not the array. The DuckDB benchmark in #2553 recorded an array shape at 144 ms against 52 ms for the top-level attribute object on cost by principal at 2 million rows, and 214 ms against 45 ms on filtering to one principal. That benchmark used a chain-of-principals array rather than the two-key entries array proposed here, so it is the closest available measurement rather than a measurement of this structure. Nobody has benchmarked the proposed structure.
* The array imposes fan-out on data that never repeats, since a charge carries at most one principal and at most one credential. The 1.60x figure cited on #2553 comes from a constructed mix in which most charges reach a user through an API key, not from any generator's output. The exposure is bounded: FOCUS defines two entry keys, so fan-out from FOCUS-defined entries reaches 2x at most, and this pull request documents the single-`key` restriction that removes it from aggregate queries. If entries turn out to be the wrong structure, unwinding them is itself a Migration Compatible Change, which is the cost of being wrong.
* Entries stabilize the key set, not the values inside it. The `Principal` value carries `Name`, `Email`, and `Type`; the `Credential` value carries `Id`, `Type`, and `Name`, both under `additionalProperties: false`. A new FOCUS-defined property inside either still changes a three-field structure a practitioner has materialized.
* The array forfeits a guarantee the object had. `JsonObjectFormat` already requires that a key in an object be unique, which JSON object semantics enforce at no cost. As an array, entry-key uniqueness becomes normative prose with nothing enforcing it, since JSON Schema cannot express it and a duplicate `Principal` entry validates while violating the requirement. The attribute's existing rule that an object in an array must not be repeated catches byte-identical entries, not two entries sharing a key with different values.
* A top-level array does not follow the dominant pattern among the four columns that already conform to `JsonObjectFormat`. The table below has the detail.
* Array to entry to value reaches three levels, so entry values have to stay flat if the nesting guidance applies to them. Whether it does is unsettled: that bullet reads "Object in FOCUS dataset column SHOULD NOT exceed 3 levels of nesting," and this pull request relaxes only the serialization bullet, so the nesting bullet still takes an object as its subject.
* Entries give up the flexibility a plain object carries across layers, which @kk09v raised. He also noted that an undeclared key set is hard to extract from. That half does not distinguish the two options, since FOCUS fixes the keys either way.
* Entries are not yet the norm. Microsoft, Oracle Cloud Infrastructure, Redis, and Vercel all publish tags as a JSON object map, so the case for entries depends on how storage formats are expected to develop rather than on what most generators ship today.

## `JsonObjectFormat` columns today, and what this pull request proposes

| Column | Top-level structure | Difference from the proposed entries array |
|---|---|---|
| ContractApplied | Object wrapping a required `Elements` array | Object rather than array at the top level; typed domain objects rather than `{key, value}` entries |
| AllocatedMethodDetails | Object wrapping a required `Elements` array | Object rather than array at the top level; typed domain objects rather than `{key, value}` entries |
| CommitmentProgramEligibilityDetails | Object wrapping a required `CommitmentPrograms` array | Object rather than array at the top level; typed domain objects rather than `{key, value}` entries, under a second array name already in use |
| ContractCommitmentApplicability | Object with seven fixed properties, including `Inclusions` and `Exclusions` arrays of rule objects | No single wrapper array; arrays appear alongside scalar and object properties instead of holding the whole value |
| CredentialDetails, as proposed here | Top-level array of `{key, value}` entries, with keys fixed to `Principal`, `Credential`, and `x_` extensions | The structure the four rows above are compared against |

Three of the four existing columns share one top-level structure: an object with a single required array property, named `Elements` in two of them and `CommitmentPrograms` in the third. ContractCommitmentApplicability does not follow that pattern. The entries array is therefore a new top-level structure for a set that is otherwise consistent.

## Where this stands after the Open Forum

I agreed at the Open Forum to expanding `JsonObjectFormat` to admit an object or an array. On where that change lands, my position is that it belongs in this pull request rather than on its own timeline. Open question 1 on entry-key uniqueness is still open, and it belongs in the requirements model rather than in a build check. Open question 2, on nesting, is one instance of a wider gap. This pull request relaxes the first bullet of `JsonObjectFormat` to admit an array, and the bullets after it still take an object as their subject, including the nesting guidance and the requirement that a key in an object be unique. As drafted it is undefined which of them reach an array-rooted column. Either those bullets take the same treatment or the group records which of them apply. Open question 3, on the column name, this pull request records as settled by Maintainers on 2026-08-03.

Separately, when the 1.6 uniformity work decides map versus entries for the four columns above, it should decide each on its own merits. Each of those conversions would be a Migration Compatible Change on a published column, so the uniformity has a cost, and adopting entries here is not the default answer for them. Tags, AllocatedTags, SkuPriceDetails, and InvoiceDetailGrain conform to `KeyValueFormat` rather than `JsonObjectFormat`, so those are a separate question against a different attribute.

specification/datasets/cost_and_usage/columns/credentialdetails.md
Matt-Cowsert: "Or other identity-related details" introduces a term the specification does not define, so the bullet stops being testable. A validator cannot decide whether a given entry qualifies, and neither can a data generator.

The rules below already bound this. `Entry.key` MUST be "Principal", "Credential", or a custom key, and a custom key carries the `x_` prefix. That is control enough without characterizing what a custom entry holds. Scoping the applicability test to FOCUS-defined entries closes the same gap and leaves custom entries to the extension rule.

It also matches how the specification treats every other extension point. A custom column carries naming and documentation requirements rather than content requirements, and the model rule behind the documentation requirement has no check function attached. An applicability test on a custom entry would fail the same way the phrase above does: nothing can check it.

```suggestion
* CredentialDetailsObject MUST NOT include FOCUS-defined entries that are not applicable to the corresponding PrincipalId or CredentialId.
```

specification/appendix/json_object_examples/credential_details_examples.md
Matt-Cowsert: This resolves it. The scenario now describes both extension paths, a custom property inside a FOCUS-defined entry and a custom entry standing alone.

One case is missing. The example covers a Principal-related attribute and an attribute related to neither, but not a Credential-related attribute with no FOCUS-defined property. That is the case a data generator publishing API key metadata reaches first, since the `Credential` value carries only `Id`, `Type`, and `Name`.

```suggestion
Scenario: Aura Web publishes a Principal-related attribute that has no FOCUS-defined property. The attribute is carried as a custom property prefixed with "x_" inside the `Principal` entry in CredentialDetails. A Credential-related attribute with no FOCUS-defined property is carried the same way, inside the `Credential` entry. An attribute that describes neither the *principal* nor the *credential* is carried as a custom entry whose `key` is prefixed with "x_".
```

specification/datasets/cost_and_usage/columns/credentialdetails.md
Matt-Cowsert: Agreed on promoting this, and the promotion matters more than the casing. The `x_` prefix is the only control on a custom entry, so it needs normative force rather than guidance.

It is also worth splitting into two requirements rather than one. The sentence carries two obligations, one on custom entry keys and one on custom property names, and the authoring rules ask for a bullet to express one verifiable state. Split, each is assessed on its own; merged, the property-name half is easy to lose. Both would be Static rules, since each can be validated from the dataset alone.

```text
* Custom Entry.key in CredentialDetailsObject MUST be prefixed with "x_" followed by PascalCase.
* Custom property name in Entry.value in CredentialDetailsObject MUST be prefixed with "x_" followed by PascalCase.
```

These sit with the existing `Entry.key` bullets rather than in the Custom Keys and Properties section, so the change is a move as well as a rewording.

specification/datasets/cost_and_usage/columns/credentialdetails.md
udam-stitcher: That wasn't the intention here. Ideally, I think providers can specify any number of authentication levels they want to create a key for and add details to, right? I get the issue you're trying to solve, but I think we need a way to allow the necessary flexibility while still limiting this to credentials-related information. Since we've used the name 'Credential' to specifically be the 'key that opens the door' we can't really say credential related info.

specification/appendix/json_object_examples/credential_details_examples.md
udam-stitcher: these are in different PRs

specification/datasets/cost_and_usage/columns/credentialdetails.md
```diff
@@ -31,55 +33,91 @@ The following section details the normative requirements for the CredentialDetai
...
+* Principal.value.Type MUST be a consistent, readable display value.
+* Principal.value SHOULD include Type when a *FOCUS dataset* represents more than one kind of *principal*.
+
+The `Credential` entry MUST adhere to the following requirements:
+
+* Credential.value MUST represent the *credential* identified by CredentialId.
+* Credential.value.Id MUST match CredentialId.
+* Credential.value.Type MUST represent the kind of *credential* identified by CredentialId.
+* Credential.value.Type MUST be a consistent, readable display value.
+* Credential.value.Name MUST represent a readable display name for the *credential* identified by CredentialId.
```
udam-stitcher: I would like to remove these (these are formatted from the existing requirements. We don't really need to mandate the types of values that should be in the value object IMO. Yes, that could mean the value object has slightly different keys that describe the email, but we also can't force it to include emails, names, etc. or to say that a credential can't be associated with an email. 

Maybe a middle-ground approach would be to list a set of keys commonly used to describe properties of these credential entries.

specification/datasets/cost_and_usage/columns/credentialdetails.md
```diff
@@ -31,55 +33,91 @@ The following section details the normative requirements for the CredentialDetai
...
-The `Credential` property contains an object with the following entries:
+The `value` of the `Credential` entry contains an object with the following properties:
 
-| Key | Type | Required | Description |
+| Property | Type | Required | Description |
 | :--- | :--- | :--- | :--- |
-| `Id` | String | False | The value of Credential ID for the *credential* described by this object. |
+| `Id` | String | False | The value of Credential ID for the *credential* described by this entry. |
 | `Type` | String | True | The kind of *credential* identified by Credential ID (e.g., "API Key", "Session", "Delegated Role"). |
 | `Name` | String | False | Readable display name for the *credential* identified by Credential ID. |
```
udam-stitcher: Consider deleting this (same reason as [above](https://github.com/FinOps-Open-Cost-and-Usage-Spec/FOCUS_Spec/pull/2634/changes/BASE..2e439d47a78c3db88812dc5648b16675b3ef7ccf#r3804606962))

specification/datasets/cost_and_usage/columns/credentialdetails.md
```diff
@@ -31,55 +33,91 @@ The following section details the normative requirements for the CredentialDetai
...
 
-<div class="h7-nonindex">Property Placement</div>
+`Credential.value.Id` repeats the value carried in CredentialId. Where the *credential* identifier is published alongside its descriptive attributes, a FOCUS-defined property is available for it rather than a custom property. Analyses that group or filter on the *credential* read CredentialId, which is available without parsing the array.
 
-Attributes describing the *principal* identified by PrincipalId belong at the top level of the object. Attributes describing the *credential* identified by CredentialId belong in the `Credential` object. Placing the descriptive attributes of the *principal* at the top level allows analyses that group or filter on those attributes to read them directly.
+Where the *credential* presented has no published identifier, CredentialId is null and the `Credential` entry is omitted, while the `Principal` entry still describes the *principal*. The `Credential` entry describes a *credential* that CredentialId identifies; it is not a place to record a *credential* that has no published identifier.
```
udam-stitcher: Consider deleting this (same reason as [above](https://github.com/FinOps-Open-Cost-and-Usage-Spec/FOCUS_Spec/pull/2634/changes/BASE..2e439d47a78c3db88812dc5648b16675b3ef7ccf#r3804606962))

specification/datasets/cost_and_usage/columns/credentialdetails.md
```diff
@@ -31,55 +33,92 @@ The following section details the normative requirements for the CredentialDetai
...
+| `Id` | String | False | The value of Credential ID for the *credential* described by this entry. |
 | `Type` | String | True | The kind of *credential* identified by Credential ID (e.g., "API Key", "Session", "Delegated Role"). |
 | `Name` | String | False | Readable display name for the *credential* identified by Credential ID. |
 
 ### Object Implementation Guidance
 
-<div class="h7-nonindex">Custom Properties</div>
+<div class="h7-nonindex">Custom Keys and Properties</div>
+
+To facilitate querying data across *principals* and across [*service providers*](#glossary:service-provider), a data generator may include one or more custom entries, or custom properties within the `value` of a FOCUS-defined entry. Custom entry keys and custom property names must be prefixed with "x_" followed by PascalCase format (e.g., `x_MyCustomKey`) to make them easy to identify as well as prevent collisions with FOCUS-defined keys.
```
udam-stitcher:
```suggestion
To facilitate querying data across *principals* and across [*service providers*](#glossary:service-provider), a data generator may include one or more custom entries, or custom properties within the `value` of a FOCUS-defined entry. Custom entry keys and custom property names MUST be prefixed with "x_" followed by PascalCase format (e.g., `x_MyCustomKey`) to make them easy to identify as well as prevent collisions with FOCUS-defined keys.
```

specification/datasets/cost_and_usage/columns/credentialdetails.md
jpradocueva: changes applied by Udam

specification/datasets/cost_and_usage/columns/credentialdetails.md
```diff
@@ -31,55 +33,92 @@ The following section details the normative requirements for the CredentialDetai
...
-* CredentialDetailsObject.Email MUST represent an email address associated with the *principal* identified by PrincipalId.
-* CredentialDetailsObject.Type MUST represent the kind of *principal* identified by PrincipalId.
-* CredentialDetailsObject.Type MUST be a consistent, readable display value.
-* CredentialDetailsObject SHOULD include Type when the *service provider* distinguishes kinds of *principals*.
-* CredentialDetailsObject.Credential MUST represent the *credential* identified by CredentialId.
-* CredentialDetailsObject.Credential.Id MUST match CredentialId.
-* CredentialDetailsObject.Credential.Type MUST represent the kind of *credential* identified by CredentialId.
-* CredentialDetailsObject.Credential.Type MUST be a consistent, readable display value.
-* CredentialDetailsObject.Credential.Name MUST represent a readable display name for the *credential* identified by CredentialId.
+* CredentialDetailsObject MUST be an array of entries.
```
jpradocueva: @udam-stitcher let me know if you disagree with this suggestion.

### F-003 — error — [Line 36](https://github.com/FinOps-Open-Cost-and-Usage-Spec/FOCUS_Spec/blob/4a9b5561f005961c258e14760b4985a24e1718ef/specification/datasets/cost_and_usage/columns/credentialdetails.md#L36)

**Changed content:** `CredentialDetailsObject MUST be an array of entries.`

**Rule:** `validation-schema-accuracy-01` — Writing Specification Content & Review Guidelines → Validation & Schema Accuracy → Mathematical & Schema Accuracy; position 1. “AI reviewers MUST rigorously calculate, parse, and verify all data within examples...”

**Engineering context:** In JSON, an array and an object are distinct root types, comparable to distinct variants in a type system.

**Problem:** This array requirement contradicts the same document’s line 17 requirement that `CredentialDetails MUST be of type JSON Object`. The new array examples cannot satisfy both statements.

**Impact:** A generator or validator cannot determine the required root JSON type from the normative column definition.

**Behavioral effect of the correction:** No schema or normative behavior change. The correction should align the stale type wording with the array behavior introduced by this PR.

**Suggested correction:** Revise the line 17 requirement to identify the intended serialized JSON array, or use another precise type term consistent with this array requirement and the column’s content constraints.

jpradocueva (commented):

Snapshot totals: 4 conversation comments, 10 reviews, and 15 inline review comments.

| --- | --- | --- |
| F-003 | Submit inline comment | Conflicting JSON root types make the normative column definition ambiguous; aligning the stale object wording with the intended array does not change schema or normative behavior. |

Matt-Cowsert:
@udam-stitcher #2553 renamed `CredentialDetails` to `RequesterDetails`, and that landed after this PR was cut, so this one is showing conflicts now. The rename is the whole cause. Nothing about the entries structure changed.

The mapping, if you want to take the rebase:

* `specification/datasets/cost_and_usage/columns/credentialdetails.md` becomes `requesterdetails.md`
* `specification/schemas/datasets/cost_and_usage/credentialdetailsobjectschema.json` becomes `requesterdetailsobjectschema.json`
* `specification/appendix/json_object_examples/credential_details_examples.md` becomes `requester_details_examples.md`
* `specification/attributes/json_object_format.md` merges without conflict

Two things that will not appear as conflicts but change what you are rebasing onto. The column file was renamed and then edited again on #2553, moving the coverage documentation requirement from the column to the dataset level, so that block is no longer in the target file. And #2553 took a merge from #2360 today, which brought in `AllocatedServiceName`, the Tags key preservation change, and the normative requirements guidelines refactor.

On the mechanics: the rename is followed automatically for the examples file and the schema, leaving content conflicts in both, seven hunks and two. It is not followed for the column file, so those 86 added and 39 removed lines need re-homing by hand. That is the bulk of the work.

I am glad to do the re-homing myself if that is easier, with you reviewing the result. The reason I am asking first is that it would outdate the five open threads here, including Jaime's suggestion from the 19th that is still with you, and I would rather not move those anchors while that conversation is live.

No change to where this is headed. I still support the entries structure and the `JsonObjectFormat` change landing in 1.5.

specification/datasets/cost_and_usage/columns/credentialdetails.md
Matt-Cowsert: I agree, we need provider flexibility. Do you have a recommendation on refined wording or are you recommending we leave as is with "identity-related details"?

Matt-Cowsert:
## TF2 Action Item - August 26, 2026

@udam-stitcher committed in Task Force 2 (TF-2) on 2026-08-26 to rebase PR #2634 and carry the RequesterDetails naming through. The rename landed after the call: `requesterdetails.md`, `requester_details_examples.md`, and `requesterdetailsobjectschema.json` now carry the new term, so that half is done.

The rebase is still open. PR #2634 has diverged from the work it targets on PR #2553 and does not merge cleanly. What it conflicts with now is the nullability change @matt-cowsert applied during the call, where RequesterDetails moved to MAY be null when PrincipalId is null and CredentialId is null, and MUST NOT be null when either one is not null.

The same discussion settled where the deletions go. @udam-stitcher proposed merging first and removing the passages that spell out keys for the principal and the credential afterward, and @matt-cowsert agreed that order is easier. Those removals are therefore not part of this pull request. RequesterDetails covers any requester-related detail, so naming the principal and the credential and pointing at the proposed keys is enough, and anything past those two levels is left to the data generator.

**Scope:**
- [ ] Rebase PR #2634 onto the current state of PR #2553 so the nullability change and the key-value restructure both survive

**Assignee:** @udam-stitcher

specification/datasets/cost_and_usage/columns/credentialdetails.md
Matt-Cowsert: The JSON Object versus array collision you flagged is still open at 7198e302, Joaquin, and this thread's anchor is outdated now, so here is the current shape.

Line 15 reads `RequesterDetails MUST be of type JSON Object (serialized as a String where necessary)` and line 34 reads `RequesterDetailsObject MUST be an array of entries`. They sit in different sections, Column Requirements and Object Requirements, and collide through line 21, which binds RequesterDetails to RequesterDetailsObject whenever it is not null. A data generator following both cannot produce a conformant value.

The Content Constraints table on this column already says `JSON`, so the narrow fix is the column bullet:

```diff
-* RequesterDetails MUST be of type JSON Object (serialized as a String where necessary).
+* RequesterDetails MUST be of type JSON (serialized as a String where necessary).
```

That leaves RequesterDetails as the only one of nine JSON columns not reading `MUST be of type JSON Object`. I would take that over leaving the collision, though whether the other eight follow is a group call. Four of them conform to KeyValueFormat rather than JsonObjectFormat, so they are a separate question against a different attribute.

specification/datasets/cost_and_usage/columns/requesterdetails.md
```diff
@@ -31,80 +31,93 @@ The following section details the normative requirements for the RequesterDetail
...
+
+* Entry in RequesterDetailsObject MUST include a `key` property.
+* Entry in RequesterDetailsObject MUST include a `value` property.
+* Entry.key in RequesterDetailsObject MUST be "Principal", "Credential", or a custom key.
+* Entry.key in RequesterDetailsObject MUST be unique within RequesterDetailsObject.
+* Entry.value in RequesterDetailsObject MUST be an object carrying the descriptive attributes of the *requester* level named by Entry.key.
+* Entry.value.Type MUST represent the kind of entity described by the entry.
+* Entry.value.Name MUST represent a readable display name for the entity described by the entry.
+* Entry.value.Email MUST represent an email address associated with the entity described by the entry.
+* Custom entry in RequesterDetailsObject MUST represent a level of the *requester* that is neither the *principal* nor the *credential*.
```
Matt-Cowsert:
```suggestion
* Custom entry in RequesterDetailsObject MUST represent a level of the *requester* that is neither the *principal* nor the *credential*.
* Custom Entry.key in RequesterDetailsObject MUST begin with the string "x_" followed by PascalCase.
* Custom property name in Entry.value in RequesterDetailsObject MUST begin with the string "x_" followed by PascalCase.
```

These two bullets add to the block you asked to shrink on 18 August, so worth saying why they differ from the value-property requirements. Those said what a generator must put in a value; these are the only control on what a custom key may be. Without them the section permits "a custom key" without saying what makes one, while the schema tests `^x_[A-Z][a-zA-Z0-9]*$` against both the entry key and the property name.

`MUST begin with the string "x_"` rather than "MUST be prefixed with" because AGENTS.md lists `prefix` among the prohibited process verbs; `skupricedetails.md` line 32 and `invoicedetailgrain.md` line 19 both state this constraint that way. Two bullets because each is a separate obligation.

This is also the half of the earlier `x_` thread that did not land. It was closed on 19 August as applied, and the casing went in while the move into Requirements did not.

specification/datasets/cost_and_usage/columns/requesterdetails.md
```diff
@@ -31,80 +31,93 @@ The following section details the normative requirements for the RequesterDetail
...
+| `Email` | String | False | Email address associated with the entity the entry describes. |
 
 ### Object Implementation Guidance
 
-<div class="h7-nonindex">Custom Properties</div>
+<div class="h7-nonindex">Custom Keys and Properties</div>
 
-To facilitate querying data across *principals* and across *service providers*, a data generator may include one or more custom properties. These may be placed at the top level of the object (alongside `Name`) or nested within the `Credential` object. Custom keys must be prefixed with "x_" followed by PascalCase format (e.g., `x_MyCustomKey`) to make them easy to identify as well as prevent collisions with FOCUS-defined keys.
+To facilitate querying data across *principals* and across [*service providers*](#glossary:service-provider), a data generator may include one or more custom entries, or custom properties within the `value` of a FOCUS-defined entry. Custom entry keys and custom property names MUST be prefixed with "x_" followed by PascalCase format (e.g., `x_MyCustomKey`) to make them easy to identify as well as prevent collisions with FOCUS-defined keys.
```
Matt-Cowsert:
```suggestion
To facilitate querying data across *principals* and across [*service providers*](#glossary:service-provider), a data generator may include one or more custom entries, or custom properties within the `value` of a FOCUS-defined entry. The "x_" prefix followed by PascalCase format (e.g., `x_MyCustomKey`) makes custom keys easy to identify and prevents collisions with FOCUS-defined keys.
```

AGENTS.md keeps capitalized BCP-14 keywords inside Requirements sections, and this paragraph sits under Object Implementation Guidance. This text carried a lowercase "must" before this PR, so raising it to MUST in place is what moved it out of bounds. Stating the obligation in Object Requirements and leaving the reasoning here keeps the normative force without the placement problem.

Matt-Cowsert (commented):
`JsonObjectFormat` now admits an array root while the bullets under it still take an object as their subject, so it is undefined which of them reach an array-rooted column. The loss lands on this column. `Entry.key in RequesterDetailsObject MUST be unique within RequesterDetailsObject` at line 43 has nothing enforcing it: `Key in Object in *FOCUS dataset column* MUST be unique` binds keys inside each entry object, not entry keys across the array, and the schema does not test it either. Two `Principal` entries carrying different values validate against `requesterdetailsobjectschema.json` as it stands.

Not every bullet goes inert. `Object in array in *FOCUS dataset column* MUST NOT be repeated` does reach the entries and rules out two identical ones, so the residual gap is same key, different value. Nesting is the other bullet worth a decision, since array to entry to value is three levels.

`specification/appendix/requester_attribution.md` line 56 describes `Type` as a key of "the `Credential` property". `Credential` is now an entry key and `Type` is a property of that entry's `value`, so the appendix documents a shape the column no longer has. That file is not in this PR:

```diff
-The Credential Type column is not a FOCUS column. It shows the value a data generator would carry in the `Type` key of the `Credential` property in [Requester Details](#datamodel.costandusage.requesterdetails), and is included here to make the kind of *credential* in each scenario explicit.
+The Credential Type column is not a FOCUS column. It shows the value a data generator would carry in the `Type` property of the `Credential` entry's `value` in [Requester Details](#datamodel.costandusage.requesterdetails), and is included here to make the kind of *credential* in each scenario explicit.
```

The description has drifted from the branch in four places. It names the column `CredentialDetails`. Its example carries a `Credential.value.Id` that the schema rejects under `additionalProperties: false`. It describes per-key validation "using `oneOf` with `const` on `key`", which the schema does not do, and that mechanism is what the description offers against the RecommendationDetails objection. And it says the service-provider removal already landed on #2553, while the diff still de-actorizes it across this column's three opening paragraphs and all five scenarios in the examples file. A reader sizing this PR from the description gets the wrong shape.

I will take the merge into #2553, the Requester Details and Credential ID PR this one targets, and can carry the `requester_attribution.md` line myself. Happy to take another pass ahead of Task Force 2.

specification/datasets/cost_and_usage/columns/credentialdetails.md
udam-stitcher: Will use 'JSON value'

specification/datasets/cost_and_usage/columns/requesterdetails.md
udam-stitcher: Will fix the ServiceProvider reference.

PR #2554, opened 2026-07-22
RM #2315: Add AllocatedServiceName requirements model rules

thecloudman:
## Summary

Adds requirements model rules for the `AllocatedServiceName` column introduced in spec PR #2395. Per RM TF agreement, requirements model updates are handled in separate PRs from spec content changes.

**Depends on PR #2395 merging first.**

Changes:
* New file: `allocatedservicename.json` — column rules `CAU-AllocatedServiceName-C-000-C` through `CAU-AllocatedServiceName-C-005-C` covering type, format, and nullability requirements.
* Updated: `costandusage.json` — adds presence rule `CAU-CostAndUsage-D-099-C` (AllocatedServiceName MUST be included when the operating model includes split cost allocation) and wires it into the dataset composite rule.

Part of #2315

## Type of Change

* [ ] New column or attribute
* [x] Requirements model update
* [ ] Editorial / non-normative

## Author Checklist

* [x] I have read the [AI Usage Guidelines](guidelines/contributors/ai-usage-guidelines.md) and this contribution complies with the attestation requirements.
* [x] This PR depends on #2395 and is noted as such.
* [x] RM rules follow the `<ArtifactName>-<Type>-<NumericId>-<Status>` format.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #2655, opened 2026-08-25
FR #2315: (Replace #2612) Enhance normative requirements for AllocatedServiceName

### LINKED ISSUE #2610: [AI] Submit a PR against PR #2395 covering the open AllocatedServiceName requirement threads

kk09v:
## Summary

Per the conversation 20260824, this replaces https://github.com/FinOps-Open-Cost-and-Usage-Spec/FOCUS_Spec/pull/2612.

- Adjusted nullability requirements to use AllocatedMethodId which is not null for any split cost allocated charge
- Added normative requirement (SHOULD) reflecting the desire for AllocatedServiceName to reflect the existing service name used by providers, where applicable
  - SHOULD represents that there may be exceptions such as not be an existing service name to use for this.
- Added normative requirement (SHOULD) reflecting that unallocated charges should reflect the same AllocatedServiceName as the allocated charges
  - SHOULD represents that there may be exceptions such as where one Origin Charge is split into multiple AllocatedServiceNames

### Type of Change
* [ ] Bug fix
* [X] Editorial / Formatting

### Author Checklist
* [ X ] **AI Usage Guidelines:** I have read the [FOCUS AI Usage Guidelines](/guidelines/contributors/ai-usage-guidelines.md).
* [ X ] **Self Review Attestation:** I have performed a self-review of my own contribution.
* [ NA ] **AI-Generated Examples:** If this PR includes examples generated by AI, I have manually verified that the data is mathematically accurate, logically consistent, and compliant with the FOCUS schema.

### DIFF
--- a/specification/datasets/cost_and_usage/columns/allocatedservicename.md
+++ b/specification/datasets/cost_and_usage/columns/allocatedservicename.md
@@ -9,8 +9,10 @@ AllocatedServiceName MUST adhere to the following requirements:
 * AllocatedServiceName MUST be of type String.
 * AllocatedServiceName MUST conform to [StringHandling](#attributes.stringhandling) requirements.
 * AllocatedServiceName MUST adhere to the following nullability requirements:
-  * AllocatedServiceName MUST be null when [AllocatedResourceId](#datamodel.costandusage.allocatedresourceid) is null.
-  * AllocatedServiceName MUST NOT be null when AllocatedResourceId is not null.
+  * AllocatedServiceName MUST be null when [AllocatedMethodId](#datamodel.costandusage.allocatedresourceid) is null.
+  * AllocatedServiceName MUST NOT be null when AllocatedMethodId is not null.
+* AllocatedServiceName SHOULD match the ServiceName used by the data generator for the equivalent stand-alone service.
+* AllocatedServiceName SHOULD match the AllocatedServiceName used for other [allocated charges](#glossary.allocated-charge) related to the same [origin charge](#glossary:origin-charge) when a *charge* represents the unallocated portion of the origin charge.
 
 ## Column ID
 
--- a/specification/datasets/cost_and_usage/columns/allocatedservicename.md
+++ b/specification/datasets/cost_and_usage/columns/allocatedservicename.md
@@ -9,7 +9,7 @@ AllocatedServiceName MUST adhere to the following requirements:
 * AllocatedServiceName MUST be of type String.
 * AllocatedServiceName MUST conform to [StringHandling](#attributes.stringhandling) requirements.
 * AllocatedServiceName MUST adhere to the following nullability requirements:
-  * AllocatedServiceName MUST be null when [AllocatedMethodId](#datamodel.costandusage.allocatedresourceid) is null.
+  * AllocatedServiceName MUST be null when [AllocatedMethodId](#datamodel.costandusage.allocatedmethodid) is null.
   * AllocatedServiceName MUST NOT be null when AllocatedMethodId is not null.
 * AllocatedServiceName SHOULD match the ServiceName used by the data generator for the equivalent stand-alone service.
 * AllocatedServiceName SHOULD match the AllocatedServiceName used for other [allocated charges](#glossary:allocated-charge) related to the same [origin charge](#glossary:origin-charge) when a *charge* represents the unallocated portion of the origin charge.
--- a/specification/datasets/cost_and_usage/columns/allocatedservicename.md
+++ b/specification/datasets/cost_and_usage/columns/allocatedservicename.md
@@ -12,7 +12,7 @@ AllocatedServiceName MUST adhere to the following requirements:
   * AllocatedServiceName MUST be null when [AllocatedMethodId](#datamodel.costandusage.allocatedresourceid) is null.
   * AllocatedServiceName MUST NOT be null when AllocatedMethodId is not null.
 * AllocatedServiceName SHOULD match the ServiceName used by the data generator for the equivalent stand-alone service.
-* AllocatedServiceName SHOULD match the AllocatedServiceName used for other [allocated charges](#glossary.allocated-charge) related to the same [origin charge](#glossary:origin-charge) when a *charge* represents the unallocated portion of the origin charge.
+* AllocatedServiceName SHOULD match the AllocatedServiceName used for other [allocated charges](#glossary:allocated-charge) related to the same [origin charge](#glossary:origin-charge) when a *charge* represents the unallocated portion of the origin charge.
 
 ## Column ID
 


kk09v:
Per the conversation today, this replaces #2612.

- Adjusted nullability requirements to use AllocatedMethodId which is not null for any split cost allocated charge
- Added normative requirement (SHOULD) reflecting the desire for AllocatedServiceName to reflect the existing service name used by providers, where applicable
  - SHOULD represents that there may be exceptions such as not be an existing service name to use for this.
- Added normative requirement (SHOULD) reflecting that unallocated charges should reflect the same AllocatedServiceName as the allocated charges
  - SHOULD represents that there may be exceptions such as where one Origin Charge is split into multiple AllocatedServiceNames

specification/datasets/cost_and_usage/columns/allocatedservicename.md
```diff
@@ -9,8 +9,10 @@ AllocatedServiceName MUST adhere to the following requirements:
 * AllocatedServiceName MUST be of type String.
 * AllocatedServiceName MUST conform to [StringHandling](#attributes.stringhandling) requirements.
 * AllocatedServiceName MUST adhere to the following nullability requirements:
-  * AllocatedServiceName MUST be null when [AllocatedResourceId](#datamodel.costandusage.allocatedresourceid) is null.
-  * AllocatedServiceName MUST NOT be null when AllocatedResourceId is not null.
+  * AllocatedServiceName MUST be null when [AllocatedMethodId](#datamodel.costandusage.allocatedresourceid) is null.
```
Arjun-Trident, line 12: Small fix to the linkage given the change.

```suggestion
  * AllocatedServiceName MUST be null when [AllocatedMethodId](#datamodel.costandusage.allocatedmethodid) is null.
```

flanakin (approved):

Arjun-Trident (approved):

specification/datasets/cost_and_usage/columns/allocatedservicename.md
```diff
@@ -9,8 +9,10 @@ AllocatedServiceName MUST adhere to the following requirements:
 * AllocatedServiceName MUST be of type String.
 * AllocatedServiceName MUST conform to [StringHandling](#attributes.stringhandling) requirements.
 * AllocatedServiceName MUST adhere to the following nullability requirements:
-  * AllocatedServiceName MUST be null when [AllocatedResourceId](#datamodel.costandusage.allocatedresourceid) is null.
-  * AllocatedServiceName MUST NOT be null when AllocatedResourceId is not null.
+  * AllocatedServiceName MUST be null when [AllocatedMethodId](#datamodel.costandusage.allocatedmethodid) is null.
+  * AllocatedServiceName MUST NOT be null when AllocatedMethodId is not null.
+* AllocatedServiceName SHOULD match the ServiceName used by the data generator for the equivalent stand-alone service.
+* AllocatedServiceName SHOULD match the AllocatedServiceName used for other [allocated charges](#glossary:allocated-charge) related to the same [origin charge](#glossary:origin-charge) when a *charge* represents the unallocated portion of the origin charge.
```
ijurica, line 15: Wasn't the original intent that on the unallocated portion row `AllocatedServiceName` should equal the origin `ServiceName` (i.e. the parent)? That's how I read @kk09v's example, where the expected value for the `NULL` (unallocated) row is `Virtual Machines` rather than `Kubernetes`.

The new SHOULD rule as currently written points in the opposite direction. I'd suggest changing it to:

```suggestion
* AllocatedServiceName MUST match the ServiceName when a *charge* represents the unallocated portion of the *origin charge* after split cost allocation.
```


Two reasons I'd prefer anchoring on origin `ServiceName` rather than on sibling `AllocatedServiceName`:

1. **Ambiguity when siblings disagree.** If a provider splits an origin charge into allocated charges with *different* `AllocatedServiceName` values ( e.g. Pod1 → Kubernetes, Pod2 → Databases), the current wording "match the AllocatedServiceName used for other allocated charges related to the same origin charge" does not define which sibling value to pick. Anchoring on origin `ServiceName` is unambiguous regardless of how many distinct consuming services the origin was split across.

2. **Semantics of the unallocated portion.** The unallocated portion represents capacity the provider *could* have allocated to specific consuming resources/services but chose to leave unallocated. It is therefore not "consumed by" any of the sibling consuming services — even when all siblings happen to share the same `AllocatedServiceName`. The only service identity that remains meaningful for that row is the origin service, so I'd keep origin `ServiceName` there consistently.

---

**Minor (should you decide to accept the above suggestion):** it might be worth adding one sentence to the column description clarifying that on the unallocated portion of a split cost allocation, `AllocatedServiceName` reflects the origin service rather than a consuming one — otherwise the description ("the display name of the service to which cost is allocated") reads a bit contradictory for that row.

Matt-Cowsert: Irena, Karl's description reads the unallocated row as carrying the sibling AllocatedServiceName, with a multi-service split as the exception; your reading anchors it on the origin ServiceName. The two give different values on the NULL row of the example, so the wording follows whichever way the example is meant to read. Worth settling on the Task Force 2 call today.

jpradocueva: The comment has been submitted outside of this thread. It was submitted as a comment, but I guess being on the same line triggered a request for change.

jpradocueva (dismissed):
@kk09v minor editorial change about link `charge`

specification/datasets/cost_and_usage/columns/allocatedservicename.md
Matt-Cowsert: @jpradocueva this isn't a blocking change, please adjust to a comment, I don't believe first linking should resolve to a change request and block merging (especially since we can cover in consistency review).

jpradocueva (commented):
### F-002 — error — [Line 15](https://github.com/FinOps-Open-Cost-and-Usage-Spec/FOCUS_Spec/blob/b26618ec30f84977baddc46a3a64091553eea55f/specification/datasets/cost_and_usage/columns/allocatedservicename.md#L15)

**Changed content:** `when a *charge* represents the unallocated portion of the origin charge`

**Rule:** `editorial-glossary-17` — Writing Specification Content & Review Guidelines > Editorial Conventions > Glossary terms; position 17. “Link with `[*term*](#glossary:term)` format.”

**Engineering context:** FOCUS uses glossary links as stable type-like references that connect a term in prose to its canonical definition.

**Problem:** The first generic charge glossary reference should link to its canonical glossary entry, but the changed line renders *charge* as unlinked emphasis.

**Impact:** Readers cannot navigate from this newly introduced term to the specification's precise definition, weakening traceability in a normative requirement.

**Behavioral effect of the correction:** No schema or normative behavior change. Adding the link changes navigation only.

**Suggested correction:** Change the first generic `*charge*` occurrence to `[*charge*](#glossary:charge)` while preserving the final unlinked repetition of origin charge.

jpradocueva (approved):
The PR validates correctly.

specification/datasets/cost_and_usage/columns/allocatedservicename.md
mike-finopsorg: I agree with @ijurica, anchoring on origin ServiceName rather than on sibling AllocatedServiceName

kk09v: This would revert the change to what we had prior to a discussing this situation.

In TF2, we determined that we want the unallocated cost for a Kubernetes cluster (which GCP's model does have) to reflect `GKE` (Google Kubernetes Engine) rather than to reflect the service name that GKE was deployed on of `Compute Engine`.

kk09v:
To summarize the conversation from today:
- In AWS, EKS unused charges are all allocated to a pod
  - There is a general agreement on the desire for the unused charges to have the AllocatedServiceName reflecting EKS, which reflects the current behavior.
- In GCP, GKE unused charges are not allocated to a pod
- As a result, there is no impact to EKS reporting by changing the ```SHOULD match the AllocatedServiceName``` to ```MUST match the ServiceName```. However, this would impact the reporting in GKE.

Ref:
https://docs.cloud.google.com/kubernetes-engine/docs/how-to/cost-allocations#view_breakdown

In GKE, there are 2 main types of unallocated charges
- `kube:system-overhead` this is the difference between the node's capacity and the allocatable resources.
- `kube:unallocated` this is the node's capacity that's not currently requested by a workload. 

Additionally, there are other special types of unallocated charges
- `goog-k8s-unknown` which captures situations such as while a node is starting up or shutting down. It's not clear how EKS handles this situation, but I could see an argument for this going either direction (or even not being covered by split-cost allocation).
- `goog-k8s-unsupported-sku` which is a GKE related SKU that is not being split. EKS does similar things with network and disk, so this would be

specification/datasets/cost_and_usage/columns/allocatedservicename.md
Matt-Cowsert: Irena, Mike, Task Force 2 went with Karl's wording today: the unallocated row carries the AllocatedServiceName of its siblings, the platform service in Karl's GKE and EKS reading, so line 15 stands as written.

The multi-service case still needs a sentence. Where one origin charge splits across services, "the AllocatedServiceName used for other allocated charges" names more than one value and the bullet does not pick one. @kk09v, your description treats that as the exception the SHOULD covers, so saying so in the bullet would close it.

specification/datasets/cost_and_usage/columns/allocatedservicename.md
```diff
@@ -9,8 +9,10 @@ AllocatedServiceName MUST adhere to the following requirements:
 * AllocatedServiceName MUST be of type String.
 * AllocatedServiceName MUST conform to [StringHandling](#attributes.stringhandling) requirements.
 * AllocatedServiceName MUST adhere to the following nullability requirements:
-  * AllocatedServiceName MUST be null when [AllocatedResourceId](#datamodel.costandusage.allocatedresourceid) is null.
-  * AllocatedServiceName MUST NOT be null when AllocatedResourceId is not null.
+  * AllocatedServiceName MUST be null when [AllocatedMethodId](#datamodel.costandusage.allocatedmethodid) is null.
+  * AllocatedServiceName MUST NOT be null when AllocatedMethodId is not null.
```
Matt-Cowsert, line 13: Moving the gate to AllocatedMethodId reaches the retained origin row as well as the unallocated portion. AllocatedMethodId is populated on an origin row that is related to a split, so this bullet asks for an AllocatedServiceName value there, and both worked examples show null. `split_cost_allocation_examples.md` says "AllocatedResourceId is null, AllocatedServiceName is null" for Row 1, and the field mapping example added in #2494 says the three allocated columns are "null on the origin row, since no *allocation* was made *to* the origin row itself" while AllocatedMethodId carries the method identifier on every row. Either this bullet needs a carve-out for a retained origin row, or the two examples need to follow it. Worth deciding alongside the unallocated portion wording, since a fix to one does not settle the other.

specification/datasets/cost_and_usage/columns/allocatedservicename.md
ijurica: ## Analysis of records tied to the same parent EC2 resource identifier

**Data caveat.** AWS does not emit split cost allocation data (SCAD) through the FOCUS export — SCAD is a CUR 2.0-only feature. The analysis below is based on one production CUR 2.0 export with SCAD activated; the record types listed are those observed in that dataset.

### CUR 2.0 → FOCUS mapping used in this analysis

- **`ServiceName`** ← `line_item_product_code` (display expansion, e.g. `AmazonEC2` → `Amazon Elastic Compute Cloud`)
- **`ResourceId`** ← `line_item_resource_id`

Note: `product_servicecode` was also considered as a candidate for `ServiceName`. Comparing per-service cost totals between an AWS FOCUS export and its underlying CUR 2.0 shows that FOCUS `ServiceName` totals reconcile against `line_item_product_code` rather than against `product_servicecode`.

### Records participating in or co-located with the SCAD split (per `i-xxx` per hour)

`operation` and `usage_type` are shown to identify record types; AWS itself relies on `operation` for its SCAD origin filter (`operation LIKE '%RunInstances%'` in the reference [`scad_cca_summary_view`](https://github.com/aws-solutions-library-samples/cloud-intelligence-dashboards-framework) — the production QuickSight SCAD dashboard SQL) and exposes both columns in its FOCUS export as [`x_Operation`](https://docs.aws.amazon.com/cur/latest/userguide/table-dictionary-focus-1-0-aws-preview.html) / `x_UsageType`. CUR 2.0 column names shortened below (`line_item_product_code` → `product_code`, `product_servicecode` → `servicecode`, etc.).

`operation` and `usage_type` are shown to identify record types; AWS itself relies on `operation` for its SCAD origin filter (`operation LIKE '%RunInstances%'` in the reference [`scad_cca_summary_view`](https://github.com/aws-solutions-library-samples/cloud-intelligence-dashboards-framework)) and exposes both columns in its FOCUS export as `x_Operation` / `x_UsageType`. CUR 2.0 column names shortened below (`line_item_product_code` → `product_code`, `product_servicecode` → `servicecode`).

image


**Reading the table:**

- Row 1 is the SCAD split origin: the EC2 compute-time row that AWS's SCAD algorithm splits into per-pod child rows. Its `line_item_resource_id = i-xxx` is the parent identifier that Row 4 references.
- Rows 2 and 3 share `i-xxx` (same physical EC2 instance, same hour) but are outside SCAD scope — they don't satisfy AWS's SCAD origin filter and no child rows are generated for them.
- Row 4 is the SCAD output — 2 records per pod per hour (vCPU + GB). These do **not** have `i-xxx` as their `line_item_resource_id`; they carry the pod ARN. They point back to `i-xxx` through `split_line_item_parent_resource_id`.

### Answer to DoD items

**1. Record types tied to the same parent resource identifier that never split:**

- **Row 2** — `DataTransfer-Regional-Bytes` (`InterZone-In` / `InterZone-Out`)
  - `ServiceName` = **Amazon EC2** — **same as the split origin**
  - This is the case relevant to the `AllocatedServiceName` question raised in this thread: a charge that FOCUS classifies under the split origin's own service, yet AWS SCAD does not split.
- **Row 3** — `EKS-Auto:*-management-hours`
  - `ServiceName` = **Amazon EKS** — **different from the split origin**
  - Not the same scenario: an independent EKS Auto Mode charge that shares only the physical `line_item_resource_id` with the split origin. It never sits under the split origin's service classification, so it doesn't feed the same debate.

Neither row satisfies AWS's SCAD origin filter (`operation LIKE '%RunInstances%'` AND `product_family LIKE '%Compute Instance%'`).

**2. Which of them carry EC2 as the service:**

- **Row 2** — **yes**: `line_item_product_code = AmazonEC2` → `ServiceName = Amazon EC2`, identical to the split origin. On the same rows `product_servicecode = AWSDataTransfer` and `product_family = Data Transfer` — an alternative service classification that AWS's own FOCUS mapper does not propagate to `ServiceName`.
- **Row 3** — **no**: carries `Amazon EKS`, a distinct service classification from the split origin.

---

### Note in context of the TF2 decision on the unallocated row

Following the TF2 decision that the unallocated portion row carries the `AllocatedServiceName` of its allocated siblings (i.e. the platform service — `Amazon EKS` in the EKS scenario, mirroring the GKE reading), the finding above surfaces a related but distinct case:

The 5,793 co-located non-split rows (Row 2 above) never enter the SCAD split at all. They carry no `AllocatedServiceName` and remain under their native FOCUS `ServiceName = Amazon EC2` — even though they materially sit on the same EKS-hosting node and, under the "platform service" logic behind the TF2 decision, could be argued to be workload-natural to `Amazon EKS`.

specification/datasets/cost_and_usage/columns/allocatedservicename.md
thecloudman, line 13: 🤖 [AI][Claude Code]

Reviewed by thecloudman

**Nullability: the `AllocatedMethodId` gate reaches the retained origin row.**

`working_draft` now makes this concrete. [`split_cost_allocation_field_mapping.md`](https://github.com/FinOps-Open-Cost-and-Usage-Spec/FOCUS_Spec/blob/working_draft/specification/appendix/split_cost_allocation_field_mapping.md) states that AllocatedMethodId is "set to the allocation method identifier on every row, including the origin row - the origin row is *related to* the data generator-calculated split cost allocation even though it is not itself an *allocated charge*", and on the line above it that AllocatedResourceId, AllocatedResourceName and AllocatedServiceName are "null on the origin row". `split_cost_allocation_examples.md` shows the same for Row 1. As written, `MUST NOT be null when AllocatedMethodId is not null` requires a value on a row both appendices show as null. This branch is 10 commits behind `working_draft`, so the conflict surfaces in the built spec once it is updated.

The suggestion keeps the static gate on the null side and scopes the non-null side by row type, mirroring the three-bullet pattern AllocatedResourceId already uses. It leaves the retained origin row unconstrained exactly as AllocatedResourceId does, and it makes the Task Force 2 decision on the unallocated row normative rather than resting only on the SHOULD below.

Trade-off worth recording: the two non-null bullets become `Dynamic` model rules, matching `CAU-AllocatedResourceId-C-004-C`/`-005-C`/`-006-C`, while the null side remains statically testable on AllocatedMethodId. This also resolves F-002 by placing the first generic `*charge*` glossary link here, so line 15 correctly stays unlinked.

```suggestion
* AllocatedServiceName MUST adhere to the following nullability requirements:
  * AllocatedServiceName MUST be null when [AllocatedMethodId](#datamodel.costandusage.allocatedmethodid) is null.
  * AllocatedServiceName MUST NOT be null when a [*charge*](#glossary:charge) represents the allocated portion of the [*origin charge*](#glossary:origin-charge).
  * AllocatedServiceName MUST NOT be null when a *charge* represents the unallocated portion of the *origin charge* after split cost allocation.
```

specification/datasets/cost_and_usage/columns/allocatedservicename.md
```diff
@@ -9,8 +9,10 @@ AllocatedServiceName MUST adhere to the following requirements:
...
+* AllocatedServiceName SHOULD match the ServiceName used by the data generator for the equivalent stand-alone service.
```
thecloudman: 🤖 [AI][Claude Code]

Reviewed by thecloudman

**Undefined term and an unitalicised glossary term.**

"stand-alone" appears nowhere else in the specification, so this bullet introduces a coinage that is not defined anywhere. *service* is a glossary term and is italicised throughout the rest of this file. Linking ServiceName on its first normative reference also matches the AllocatedMethodId links in the nullability bullets above.

```suggestion
* AllocatedServiceName SHOULD match the [ServiceName](#datamodel.costandusage.servicename) used by the data generator for the equivalent independently billed *service*.
```

specification/datasets/cost_and_usage/columns/allocatedservicename.md
thecloudman: 🤖 [AI][Claude Code]

Reviewed by thecloudman

**The multi-service split case is still open, and the glossary links are unitalicised.**

Where one origin charge splits across more than one consuming service, "the AllocatedServiceName used for other allocated charges" names more than one value and the bullet does not pick one. Your description treats that as the exception the SHOULD covers; `except when` is permitted conditional phrasing, so it can be stated directly in the bullet.

The same edit italicises the two glossary links to match `[*term*](#glossary:term)` and line 3 of this file. The suggestion assumes the nullability rewrite above lands, which places the first generic `*charge*` glossary link earlier in the file, so `*charge*` stays unlinked here.

```suggestion
* AllocatedServiceName SHOULD match the AllocatedServiceName used for other [*allocated charges*](#glossary:allocated-charge) related to the same [*origin charge*](#glossary:origin-charge) when a *charge* represents the unallocated portion of the origin *charge*, except when those *allocated charges* use more than one AllocatedServiceName value.
```

thecloudman (commented):
🤖 [AI][Claude Code]

Reviewed by thecloudman

Review of the current head against `working_draft`, the sibling `Allocated*` columns, both split cost allocation appendices, and the requirements model. CI is green and `enhanced_markdown_lint.py` exits 0 on the changed file.

Three per-line suggestions are attached. Two further items fall outside this diff and are noted here rather than as suggestions:

**Sample queries in `supported_features/data_generator_calculated_split_cost_allocation.md`.** "Get Total Effective Cost by Allocated Service Name" filters on `AllocatedMethodId IS NOT NULL AND AllocatedServiceName IS NOT NULL`. Under the new nullability rule the second predicate becomes a tautology, and the result now includes the unallocated portion in what the section presents as consumer-side totals. Separately, "Find Total Unallocated Split Costs by ResourceId" identifies the unallocated portion as `AllocatedMethodId IS NOT NULL AND AllocatedResourceId IS NULL`, which also matches the retained origin row. The second is pre-existing, but this PR makes AllocatedMethodId load-bearing, so both are worth either fixing here or filing alongside.

**Requirements model.** PR #2554 still encodes the prior anchoring in `CAU-AllocatedServiceName-C-004-C` and `-005-C` (`MUST be null when AllocatedResourceId is null` / `MUST NOT be null when AllocatedResourceId is not null`). Whichever of the two merges second needs rebasing onto the final wording here, and #2554 will also need rules for the two SHOULD requirements.

One point already raised by @ijurica remains open and is not covered by the suggestions: the Description section reads "the display name of the *service* to which cost is allocated", which sits awkwardly with the unallocated portion row, where no cost is allocated to a consuming *service*. A single clause would close it, or it could be left to consistency review.

specification/datasets/cost_and_usage/columns/allocatedservicename.md
thecloudman: 🤖 [AI][Claude Code]

Reviewed by thecloudman

@Matt-Cowsert confirmed, and `working_draft` has since made it explicit rather than inferred. `split_cost_allocation_field_mapping.md` from #2494 now states that AllocatedMethodId is "set to the allocation method identifier on every row, including the origin row - the origin row is *related to* the data generator-calculated split cost allocation even though it is not itself an *allocated charge*", and directly above it that AllocatedResourceId, AllocatedResourceName and AllocatedServiceName are "null on the origin row, since no *allocation* was made *to* the origin row itself". Row 1 of `split_cost_allocation_examples.md` agrees. So the bullet as written requires a value on a row that two merged appendices show as null. This branch is 10 commits behind `working_draft`, which is why the contradiction is not visible in the build yet.

I have posted a suggestion on the nullability block that takes the carve-out route rather than the amend-the-examples route, since the examples represent a representation the attribute explicitly permits. It keeps AllocatedMethodId as the gate on the null side and scopes the non-null side by row type, mirroring the three bullets AllocatedResourceId already uses:

* MUST be null when AllocatedMethodId is null
* MUST NOT be null when a *charge* represents the allocated portion of the *origin charge*
* MUST NOT be null when a *charge* represents the unallocated portion of the *origin charge* after split cost allocation

That leaves the retained origin row unconstrained, exactly as AllocatedResourceId leaves it, so both appendices stand unchanged. It also makes the Task Force 2 decision on the unallocated row normative instead of resting only on the SHOULD, which the AllocatedResourceId-anchored wording could not express.

The cost is that the two non-null bullets become `Dynamic` model rules rather than statically testable ones, matching how `CAU-AllocatedResourceId-C-004-C`, `-005-C` and `-006-C` are already modelled. The null side stays static on AllocatedMethodId, so Karl's contribution there is preserved.

One related item this does not settle, flagged separately in the review summary: the "Find Total Unallocated Split Costs by ResourceId" query in `data_generator_calculated_split_cost_allocation.md` identifies the unallocated portion as `AllocatedMethodId IS NOT NULL AND AllocatedResourceId IS NULL`, which matches the retained origin row as well. Pre-existing, but the same root cause.

specification/datasets/cost_and_usage/columns/allocatedservicename.md
thecloudman: 🤖 [AI][Claude Code]

Reviewed by thecloudman

@Matt-Cowsert on the multi-service sentence you asked for, I have posted it as a suggestion on line 15 rather than leaving it to prose, since the ambiguity is inside the normative bullet itself.

@kk09v your reading is that the multi-service split is the exception the SHOULD covers. `except when` is permitted conditional phrasing under the authoring guidelines, so that intent can be stated in the bullet without weakening it:

> AllocatedServiceName SHOULD match the AllocatedServiceName used for other *allocated charges* related to the same *origin charge* when a *charge* represents the unallocated portion of the origin *charge*, except when those *allocated charges* use more than one AllocatedServiceName value.

That makes the obligation verifiable in the single-service case, which is the GKE and EKS case the Task Force settled on, and explicitly discharges it where siblings disagree, which is the case the current wording leaves undefined.

The same suggestion italicises the two glossary links to match `[*term*](#glossary:term)` and line 3 of this file. It assumes the nullability suggestion above it lands, which places the first generic `*charge*` glossary link earlier in the file and so resolves @jpradocueva's F-002; if that one is not taken, `*charge*` needs linking here instead.

Not addressed by the suggestion, and still open from @ijurica's original point: the Description section reads "the display name of the *service* to which cost is allocated", which sits awkwardly with the unallocated portion row now that it carries a value. One clause would close it, or it could go to consistency review.

specification/datasets/cost_and_usage/columns/allocatedservicename.md
kk09v: The origin charge is what exists without a split cost allocation applied. Once the split cost allocation is applied, there is no longer an origin charge, only allocated charges. 
I take one apple (origin charge). I can eat it as is or cut it into 4 pieces and _allocate_ those to different people. Once I cut it into 4 pieces, I can put those 4 pieces together to reconstruct the _original_ apple, but the _original_ apple no longer exists apart from the _allocated_ pieces.

Definition of split cost allocation
https://github.com/FinOps-Open-Cost-and-Usage-Spec/FOCUS_Spec/blob/8cac9f81abeca9a5b2ac0a5b7c4e59da3d1e6fd2/specification/attributes/data_generator_calculated_split_cost_allocation_handling.md

Origin charge is `The charge that existed prior to an operation.`
https://github.com/FinOps-Open-Cost-and-Usage-Spec/FOCUS_Spec/blob/8cac9f81abeca9a5b2ac0a5b7c4e59da3d1e6fd2/specification/glossary.md

`AllocatedMethodId MUST be null when a charge is not related to a data generator-calculated split cost allocation.`
https://github.com/FinOps-Open-Cost-and-Usage-Spec/FOCUS_Spec/blob/8cac9f81abeca9a5b2ac0a5b7c4e59da3d1e6fd2/specification/datasets/cost_and_usage/columns/allocatedmethodid.md

So if an origin charge exists in the FOCUS data, that means that split cost allocation was not applied and therefore the AllocatedMethodId MUST be null

specification/datasets/cost_and_usage/columns/allocatedservicename.md
kk09v: I'm generally okay with this carve-out for multiple AllocatedServices, but I think it needs to be structured differently because there is no guidance of what should happen when there are multiple AllocatedServices

kk09v: The only substantial change to Graham's suggestion is that Irena and I would suggest that it can now be a MUST since we've created a carve out for the multiple AllocatedServiceNames.

We are intentionally not describing what happens when it's shared across multiple AllocatedServiceNames because we don't know what that looks like and our preference differs based on the hypothetical scenarios we've come up with where this would be used.

kk09v:
```suggestion
* AllocatedServiceName MUST match the AllocatedServiceName used for other [*allocated charges*](#glossary:allocated-charge) related to the same [*origin charge*](#glossary:origin-charge) when a *charge* represents the unallocated portion of the origin *charge*, except when those *allocated charges* use more than one AllocatedServiceName value.
```

specification/datasets/cost_and_usage/columns/allocatedservicename.md
```diff
@@ -9,8 +9,10 @@ AllocatedServiceName MUST adhere to the following requirements:
 * AllocatedServiceName MUST be of type String.
 * AllocatedServiceName MUST conform to [StringHandling](#attributes.stringhandling) requirements.
 * AllocatedServiceName MUST adhere to the following nullability requirements:
-  * AllocatedServiceName MUST be null when [AllocatedResourceId](#datamodel.costandusage.allocatedresourceid) is null.
-  * AllocatedServiceName MUST NOT be null when AllocatedResourceId is not null.
+  * AllocatedServiceName MUST be null when [AllocatedMethodId](#datamodel.costandusage.allocatedmethodid) is null.
+  * AllocatedServiceName MUST NOT be null when AllocatedMethodId is not null.
+* AllocatedServiceName SHOULD match the [ServiceName](#datamodel.costandusage.servicename) used by the data generator for the equivalent independently billed *service*.
+* AllocatedServiceName SHOULD match the AllocatedServiceName used for other [allocated charges](#glossary:allocated-charge) related to the same [origin charge](#glossary:origin-charge) when a *charge* represents the unallocated portion of the origin charge.
```
Matt-Cowsert: ## TF2 Action Item - September 16, 2026

The `AllocatedServiceName` bullet for the unallocated portion row does not say which value applies when one origin charge splits across more than one service. The bullet asks the column to match the value on the other allocated charges from the same origin charge, and in that case those charges carry more than one value. @thecloudman posted wording on September 16, 2026 that exempts the case with "except when". On the Task Force 2 (TF-2) call the same day, asked what remains before pull request (PR) #2655 goes to Member review, @kk09v named this change.

After the call, @kk09v posted @thecloudman's wording as a suggestion on line 15 with one change: the bullet becomes a MUST, which @kk09v and @ijurica proposed now that the multi-service case is exempted. @kk09v's note says the wording leaves the multi-service case undescribed on purpose, because that case is not yet understood and preferences differ across the scenarios considered so far. The approvals on PR #2655 were given while the bullet was a SHOULD.

**Scope:**
- [ ] Commit the multi-service exemption to the unallocated portion bullet on line 15
- [ ] Confirm on the line 15 thread whether the bullet stays SHOULD or becomes MUST

**Assignee:** @kk09v

kk09v: Irena and I discussed after the call and we're sticking with @thecloudman's suggestion to have a carve-out for if something gets split into multiple AllocatedServices. We did change it to a MUST because the SHOULD was mainly to allow for the situation of multiple AllocatedServices.
We intentionally did not specify what should be done because we have no examples of this to evaluate and in our hypothetical scenarios, it was an "it depends."

If Graham is good with this, I think we can send it.

specification/datasets/cost_and_usage/columns/allocatedservicename.md
```diff
@@ -9,8 +9,10 @@ AllocatedServiceName MUST adhere to the following requirements:
 * AllocatedServiceName MUST be of type String.
 * AllocatedServiceName MUST conform to [StringHandling](#attributes.stringhandling) requirements.
 * AllocatedServiceName MUST adhere to the following nullability requirements:
-  * AllocatedServiceName MUST be null when [AllocatedResourceId](#datamodel.costandusage.allocatedresourceid) is null.
-  * AllocatedServiceName MUST NOT be null when AllocatedResourceId is not null.
+  * AllocatedServiceName MUST be null when [AllocatedMethodId](#datamodel.costandusage.allocatedmethodid) is null.
+  * AllocatedServiceName MUST NOT be null when AllocatedMethodId is not null.
+* AllocatedServiceName SHOULD match the [ServiceName](#datamodel.costandusage.servicename) used by the data generator for the equivalent independently billed *service*.
+* AllocatedServiceName MUST match the AllocatedServiceName used for other [*allocated charges*](#glossary:allocated-charge) related to the same [*origin charge*](#glossary:origin-charge) when a *charge* represents the unallocated portion of the origin *charge*, except when those *allocated charges* use more than one AllocatedServiceName value.
```
ijurica, line 15: follow-up on [this thread](https://github.com/FinOps-Open-Cost-and-Usage-Spec/FOCUS_Spec/pull/2655#discussion_r4052413067)

---

Each of the options we've discussed in this PR has its own trade-offs. The more I think about it, the more I find myself relying on **providers to document** on their side how and why they set `AllocatedServiceName` on individual allocated records — that transparency is likely what actually helps consumers in practice. 🙂

One small refactor of the current bullet that I'd suggest anyway: rewriting the **"when ... except"** clause as a **"when" with two positive conditions**. Even though both versions remain dynamic requirements, I believe the refactored version reads more naturally — **"when ... and ..."** rather than a **"when ... except ..."**:

```suggestion
* AllocatedServiceName MUST match the AllocatedServiceName used for other [*allocated charges*](#glossary:allocated-charge) related to the same [*origin charge*](#glossary:origin-charge) when a *charge* represents the unallocated portion of the origin *charge* and those *allocated charges* share a single AllocatedServiceName value.
```

specification/datasets/cost_and_usage/columns/allocatedservicename.md
ijurica: see [this comment](https://github.com/FinOps-Open-Cost-and-Usage-Spec/FOCUS_Spec/pull/2655/files#r4061439344) for a wording refactor of the normative bullet — same semantics, "when ... except ..." replaced with "when ... and ..." for cleaner reading.

flanakin (approved):

thecloudman (approved):
PR looks good to me now, agreed with the final update of the normative requirements

Issue #2665, opened 2026-09-01
[AI] Decide the SKU Price dataset feature level for v1.5


Matt-Cowsert:
### Detailed Description

Pull request #2424, the SKU Price dataset (feature request #1057), lists the dataset as Conditional, gated on the IncludesSkuPrices condition. At the August 31 Maintainers meeting we reviewed a proposal to drop that condition, which follows from the unit-pricing decision described below, and we did not reach agreement. This action item lays out the options with pros and cons so Task Force 1 can work through them on September 1 and the Maintainers group can decide on September 8. Add pros, cons, or options I have missed as comments on this issue.

#### Background

* Earlier in August the Maintainers group agreed that unit pricing applies globally, so a condition gating on it carries no information. PR #2567, now in Task Force review, applies that decision: it removes the IncludesUnitPricing and IncludesListUnitPrices conditions (through PR #2585, merged to its branch) and moves SkuId, SkuMeter, SkuPriceId, and SkuPriceDetails in Cost and Usage to Mandatory.
* The August 31 question was whether the same reasoning extends to the dataset: a SKU is predicated on unit pricing, IncludesSkuPrices then reads as equivalent to IncludesUnitPricing, and dropping it lands the dataset at Mandatory, the first Mandatory dataset beyond Cost and Usage.
* For calibration: every dataset since Cost and Usage entered the spec as Conditional. Contract Commitment entered in 1.3, Billing Period and Invoice Detail in 1.4. The obligation itself is not new; 1.3 and 1.4 carry per-dataset requirements such as "ContractCommitment MUST be present when the service provider supports contract commitments." v1.5 moves these requirements into the Data Model section and restates them on operating-model conditions.
[...]

Issue #2509, opened 2026-07-01
[Process] 1.5 Consistency Review


Matt-Cowsert:
This is a placeholder task to group the work associated with the 1.5 consistency review phase, which will occur from 2026-10-08 to 2026-10-29.

PR #2666, opened 2026-09-01
AI #2665: Align the Unit Pricing and Price List glossary entries with the SKU Price condition

Matt-Cowsert:
## Summary

Suggestion on #2424, routed from AI #2665: three text changes that make the SKU Price dataset's Conditional designation self-consistent, following the September 1 TF-1 discussion of the dataset's feature level.

1. **Unit Pricing (glossary):** the unit price is now "defined per *SKU Price*, where the *operating model* includes *SKU Prices*." The unqualified parenthetical read as making unit pricing and SKU Prices inseparable, which conflicts with the *SKU Price* entry that excludes unit price amounts, and that tension fueled the reading of `IncludesSkuPrices` as equivalent to the retired `IncludesUnitPricing` condition.
2. **Price List (glossary):** "a comprehensive, standing list of prices offered by a service provider." One added word, so a one-off quote or bid does not read as a price list.
3. **IncludesSkuPrices (condition):** an informative note naming the operating models the condition excludes: those where no billing party maintains a standing catalog of SKU Prices (auction-based marketplaces, pure pass-through aggregation, exclusively bespoke-quoted capacity). These are the operating-model shapes identified in the census posted on #2665.

No requirement text changes: two glossary definitions are clarified and one informative note is added.

### Type of Change
* [ ] Bug fix
* [ ] Editorial / Formatting

### Author Checklist

### September 1, 2026: Initial proposal

* Qualified the *Unit Pricing* parenthetical, added "standing" to the *Price List* definition, and added the informative exclusion note to `IncludesSkuPrices`.

### DIFF
--- a/specification/glossary.md
+++ b/specification/glossary.md
@@ -272,7 +272,7 @@ An individual who performs FinOps within an organization to maximize the busines
 
 <a name="glossary:price-list"><b>Price List</b></a>
 
-A comprehensive list of prices offered by a service provider.
+A comprehensive, standing list of prices offered by a service provider.
 
 <a name="glossary:service-provider"><b>Service Provider</b></a>
 
@@ -328,7 +328,7 @@ An agreement specified on a [*contract*](#glossary:contract) or [*invoice*](#glo
 
 <a name="glossary:unit-pricing"><b>Unit Pricing</b></a>
 
-A billing concept where the cost of a [*charge*](#glossary:charge) is deterministically derived as the product of a unit price (defined per [*SKU Price*](#glossary:sku-price)) and the corresponding [Pricing Quantity](#datamodel.costandusage.pricingquantity), representing purchased or consumed volume of a [*SKU*](#glossary:sku).
+A billing concept where the cost of a [*charge*](#glossary:charge) is deterministically derived as the product of a unit price (defined per [*SKU Price*](#glossary:sku-price), where the [*operating model*](#glossary:operating-model) includes *SKU Prices*) and the corresponding [Pricing Quantity](#datamodel.costandusage.pricingquantity), representing purchased or consumed volume of a [*SKU*](#glossary:sku).
 
 <a name="glossary:virtual-currency"><b>Virtual Currency</b></a>
 
--- a/specification/operating_model_conditions/includesskuprices.md
+++ b/specification/operating_model_conditions/includesskuprices.md
@@ -2,14 +2,14 @@
 
 The Includes SKU Prices operating model condition represents a verifiable state indicating whether the [*operating model*](#glossary:operating-model) includes [*SKU Prices*](#glossary:sku-price).
 
+> **Note:** An *operating model* does not include *SKU Prices* when no [*service provider*](#glossary:service-provider) in it supplies a [*price list*](#glossary:price-list) (e.g., auction-based marketplaces, pure pass-through aggregation of third-party prices, or exclusively bespoke-quoted capacity).
+
 ## Requirements
 
 IncludesSkuPrices MUST adhere to the following requirements:
 
 * IncludesSkuPrices MUST evaluate to true when the *operating model* includes *SKU Prices*.
 * IncludesSkuPrices MUST evaluate to false when the *operating model* does not include *SKU Prices*.
-
-> **Note:** An *operating model* does not include *SKU Prices* when no billing party maintains a standing catalog of them (e.g., auction-based marketplaces, pure pass-through aggregation of third-party prices, or exclusively bespoke-quoted capacity).
 
 ## Operating Model Condition ID
 
--- a/specification/operating_model_conditions/includesskuprices.md
+++ b/specification/operating_model_conditions/includesskuprices.md
@@ -9,6 +9,8 @@ IncludesSkuPrices MUST adhere to the following requirements:
 * IncludesSkuPrices MUST evaluate to true when the *operating model* includes *SKU Prices*.
 * IncludesSkuPrices MUST evaluate to false when the *operating model* does not include *SKU Prices*.
 
+> **Note:** An *operating model* does not include *SKU Prices* when no billing party maintains a standing catalog of them (e.g., auction-based marketplaces, pure pass-through aggregation of third-party prices, or exclusively bespoke-quoted capacity).
+
 ## Operating Model Condition ID
 
 IncludesSkuPrices


ijurica (approved):

Arjun-Trident (approved):

mike-finopsorg (approved):

specification/operating_model_conditions/includesskuprices.md
```diff
@@ -2,6 +2,8 @@
 
 The Includes SKU Prices operating model condition represents a verifiable state indicating whether the [*operating model*](#glossary:operating-model) includes [*SKU Prices*](#glossary:sku-price).
 
+> **Note:** An *operating model* does not include *SKU Prices* when no [*service provider*](#glossary:service-provider) in it supplies a [*price list*](#glossary:price-list) (e.g., auction-based marketplaces, pure pass-through aggregation of third-party prices, or exclusively bespoke-quoted capacity).
```
flanakin, line 5: 🤖 [AI][Claude] ⚠️ **Should fix**

The note decides whether an *operating model* includes *SKU Prices* by asking whether anyone supplies a *price list*. Those are different things in the glossary:

* **SKU Price** — "a pricing construct that encompasses SKU properties ... **while excluding** dynamic or negotiable pricing elements **such as unit price amounts**"
* **Price List** — "a comprehensive, standing list of **prices** offered by a service provider" (as amended by this pull request)

So the note reaches the existence of *SKU Prices*, which carry no amounts, through the presence of price amounts. That is the coupling this pull request set out to break: the summary says the unqualified Unit Pricing parenthetical "fuelled the reading of `IncludesSkuPrices` as equivalent to the retired `IncludesUnitPricing` condition", and testing for a price list reintroduces the same equivalence one layer down.

A concrete case where the two come apart: a provider whose catalogue is entirely free offers has *SKU Prices* — SKU properties, meters, units — but publishes no list of prices, because there are none to list. Under this note that *operating model* reads as excluding *SKU Prices*, and every column gated on `IncludesSkuPrices` disappears from a dataset that could populate all of them.

A second, separable point on the same sentence: the note scopes the test to "no [*service provider*](#glossary:service-provider) in it", while the pull request description says "no billing party maintains a standing catalog". The *operating model* glossary entry names "service providers, host providers, invoice issuers, and data generators" as participating entities, so a marketplace where the host provider publishes the catalogue satisfies the description but not the note as written.

Rewriting the test in terms of the construct the condition actually names, and widening the subject to match the description:

```suggestion
> **Note:** An *operating model* does not include *SKU Prices* when no participating entity in it defines *SKU Prices* for the [*SKUs*](#glossary:sku) it bills (e.g., auction-based marketplaces, pure pass-through aggregation of third-party prices, or exclusively bespoke-quoted capacity).
```

specification/glossary.md
```diff
@@ -328,7 +328,7 @@ An agreement specified on a [*contract*](#glossary:contract) or [*invoice*](#glo
 
 <a name="glossary:unit-pricing"><b>Unit Pricing</b></a>
 
-A billing concept where the cost of a [*charge*](#glossary:charge) is deterministically derived as the product of a unit price (defined per [*SKU Price*](#glossary:sku-price)) and the corresponding [Pricing Quantity](#datamodel.costandusage.pricingquantity), representing purchased or consumed volume of a [*SKU*](#glossary:sku).
+A billing concept where the cost of a [*charge*](#glossary:charge) is deterministically derived as the product of a unit price (defined per [*SKU Price*](#glossary:sku-price), where the [*operating model*](#glossary:operating-model) includes *SKU Prices*) and the corresponding [Pricing Quantity](#datamodel.costandusage.pricingquantity), representing purchased or consumed volume of a [*SKU*](#glossary:sku).
```
flanakin, line 331: 🤖 [AI][Claude] 💡 **Suggestion**

Two small things on the added parenthetical.

**Connective.** `where` is not one of the four conditional connectives the repository standardises on — `when`, `unless`, `only when`, `except when`. This is a glossary entry rather than a requirement, so no rule is broken, but `where` reads ambiguously here: it can be parsed as "in the case that" or as "in operating models in which", and only the first is meant.

**Coverage.** The qualification does the decoupling work it is meant to, but it also leaves a gap. `IncludesUnitPricing` and `IncludesSkuPrices` are now independent conditions, so an *operating model* can include *unit pricing* without including *SKU Prices*. In that case this definition no longer says what the unit price is defined per, which makes Unit Pricing a partial definition precisely for the case the split was introduced to allow.

Saying that the *SKU Price* is one anchor rather than the only one closes the gap without asserting a second one:

```suggestion
A billing concept where the cost of a [*charge*](#glossary:charge) is deterministically derived as the product of a unit price (defined per [*SKU Price*](#glossary:sku-price) when the [*operating model*](#glossary:operating-model) includes *SKU Prices*, and otherwise per the pricing terms the [*service provider*](#glossary:service-provider) applies to the *charge*) and the corresponding [Pricing Quantity](#datamodel.costandusage.pricingquantity), representing purchased or consumed volume of a [*SKU*](#glossary:sku).
```

flanakin (approved):
🤖 [AI][Claude] **PR Review**

**Summary:** Approving. Three small text changes that do the decoupling work they set out to do. Two observations on the new note, one on the Unit Pricing parenthetical — none blocking.

### ⚠️ Should fix (1)

1. The new `IncludesSkuPrices` note tests for a *price list*, which is defined in terms of price amounts that a *SKU Price* explicitly excludes.

### 💡 Suggestions (1)

1. The Unit Pricing parenthetical uses `where` as a conditional and leaves the non-SKU-Price case unanchored.

### ✅ What's done well

* Note placement follows the existing precedent exactly. `includesvirtualcurrency.md` carries its scoping note at line 5, between the intro sentence and `## Requirements`, and this lands in the same position.
* Glossary link discipline is correct throughout the note: `service provider` and `price list` are linked as first occurrences in the file, while `operating model` and `SKU Prices` are left plain and italicised because line 3 already links them.
* No requirement text changes, so the `IncludesSkuPrices` true/false pair stays the strict negation the other operating model conditions use.

specification/glossary.md
```diff
@@ -272,7 +272,7 @@ An individual who performs FinOps within an organization to maximize the busines
 
 <a name="glossary:price-list"><b>Price List</b></a>
 
-A comprehensive list of prices offered by a service provider.
+A comprehensive, standing list of prices offered by a service provider.
```
jpradocueva, line 275: ### F-001 — error — [Line 275](https://github.com/FinOps-Open-Cost-and-Usage-Spec/FOCUS_Spec/blob/3cfcb0110006f269862a5a66c238e09c02ea8307/specification/glossary.md#L275)

**Changed content:** `A comprehensive, standing list of prices offered by a service provider.`

**Rule:** `editorial-linking-01` — Writing Specification Content & Review Guidelines > Editorial Conventions > Linking Rule; position 1. “For each distinct entity or glossary destination in a source Markdown file: Ignore occurrences in document titles and section headings. Link the first remaining occurrence in reading order. Leave all later occurrences unlinked. Exceptions: A link whose anchor text is not an entity name or glossary term does not count toward first occurrence. Content Constraints sections link every entity reference. Glossary entries apply this rule independently within each entry.”

**Engineering context:** Each glossary entry is its own linking scope, similar to resolving the first reference in each independently rendered record.

**Problem:** The first body occurrence of each glossary destination should link to that destination, but the revised Price List entry uses plain text for its first `service provider` reference.

**Impact:** Readers cannot navigate directly from the revised definition to the Service Provider definition, and the entry no longer follows the specification's deterministic first-occurrence linking convention.

**Suggested correction:** Link `service provider` to `#glossary:service-provider` in the revised sentence.

```suggestion
A comprehensive, standing list of prices offered by a [*service provider*](#glossary:service-provider).
```

jpradocueva (commented):

Snapshot totals: 6 conversation comments, 4 reviews, and 2 inline review comments.

| --- | --- | --- |
| F-001 | Submit inline comment | Missing first-occurrence glossary link; correcting the link does not change schema or normative behavior. |

Matt-Cowsert:
Closing this pull request without merging. Its three changes existed to give IncludesSkuPrices a meaning separate from IncludesUnitPricing, and #2424 no longer uses IncludesSkuPrices. The Maintainers gated the SKU Price dataset on IncludesUnitPricing on September 8 and kept that condition when they went forward with #2690 on September 14. Commit c5bd24b6 on #2424 points the Data Model requirement at IncludesUnitPricing and removes the IncludesSkuPrices condition.

* **IncludesSkuPrices note.** The condition file it annotates is gone, so neither the note nor Michael's suggested rewrite of it has a place to land.
* **Unit Pricing qualifier.** `includesunitpricing.md` links to this glossary entry, so the entry defines the condition that now gates the SKU Price dataset. Adding "where the operating model includes SKU Prices" would tie that definition to a state no condition covers, and it would leave the gap Michael raised: the entry would no longer say what a unit price is defined per when an operating model has no SKU Prices.
* **"Standing" in Price List.** The word was there to keep a one-off quote or an auction bid from counting as a price list, which only mattered for the exemption the note described. On its own, it would narrow the SkuPriceId and PricingUnit requirements in Cost and Usage that use the term, and that change would need its own discussion.

The missing link Joaquin flagged on the Price List entry is in the current wording too, so that fix needs a tracked home in the 1.5 consistency review (#2509). The decision comment on #2665 records the glossary reconciliation as needing no change.

PR #2667, opened 2026-09-02
FR #2414: Rename ServingScope to RegionScope (stacked on #2613)

Matt-Cowsert:
## Summary

This pull request (PR) renames the `SkuPriceDetails` property that #2613 introduces as `ServingScope` to **`RegionScope`**, and rewrites its description and appendix in pricing terms. `RegionScope` is the name @flanakin's Action Item (AI) #2672 research recommends and the one @marc-perreaut proposed independently on 2026-09-03. It goes to the Open Forum on 2026-09-18 for confirmation.

Stacked on #2613 so the rename reviews as one diff and merges into that branch with one click, following the pattern #2634 used for #2553.

### Why `RegionScope`

"Pricing" comes out of the key because the column already supplies it. None of the other 17 `SkuPriceDetails` properties carries a pricing qualifier: `Redundancy`, `StorageClass`, and `OperatingSystem` all describe the SKU price without saying so. That was the substance of @flanakin's objection to `PricingRegionScope`, and it holds.

What the shorter name gives up is the visual pairing with `PricingRegionId` on #2424. That pairing was nominal rather than real, and the shared stem argued against itself: `PricingRegionId` holds an identifier and this property holds a breadth, so identical prefixes on a joinable row invited the confusion the next section separates.

Task Force 2 (TF-2) on 2026-09-02 settled the rest. The property stays in `SkuPriceDetails` as a FOCUS-defined property, the name leaves `ServingScope`, and "region" is the base word. A bare `Scope` is ruled out for a location property, per @finops-DaveMoreau on 2026-09-02 and @mike-finopsorg on 2026-09-08. `GeographicScope` is ruled out by @ijurica's reservation of "geo" for a later geographic aspect of resource regions.

### `RegionScope`, `RegionId`, and `PricingRegionId`

Three questions sit close together here, and each has its own home:

| Property | Dataset | Question it answers | Value it holds |
|:---|:---|:---|:---|
| `RegionId` | Cost and Usage | Which single region was the *resource* provisioned in? | A region identifier, for example "us-east-1" |
| `RegionScope` | Cost and Usage, inside `SkuPriceDetails` | How broad an area does the price on this charge cover? | A breadth class, for example "Global" or "DataZone" |
| `PricingRegionId` | SKU Price (#2424) | Which area does this unit price apply to? | A boundary identifier, including a macro-region, for example "eu" |

#2424 already draws the first half of this distinction itself. Its `PricingRegionId` Implementation Guidance reads: "**Pricing Region ID** defines the geographic boundary for which the *rate itself* is valid. **Region ID** defines the physical location where a specific *resource* is provisioned." `RegionScope` is the Cost and Usage side of that same axis, carried as a class rather than as an identifier, because Cost and Usage has no column for the priced area.

**Neither one derives the other.** A `PricingRegionId` of "eu" does not say whether "eu" is a region, a macro-region, or the whole world, without a service-provider-specific lookup. A `RegionScope` of "DataZone" says the breadth without saying which data zone. On a join from Cost and Usage to SKU Price, which #2424's dataset definition makes through `SkuPriceId`, the two read together as the identifier and its class.

**If the property later moves to the SKU Price dataset**, the ask @ijurica made at TF-2 on 2026-09-16, it becomes a column named `RegionScope` standing beside `PricingRegionId`. There is no name collision, and the pair reads as identifier plus class. The SKU Price dataset carries no `SkuPriceDetails` column, so any property that moves there becomes a column.

The appendix describes the third row of that table without naming `PricingRegionId`, because the SKU Price dataset is not in `working_draft` yet and a link to it fails the rendered-anchor test. The full relationship, including the named cross-reference, is recorded in `supporting_content/datasets/cost_and_usage/columns/skupricedetails-regionscope.md`, which this PR adds. The specification cross-reference follows once #2424 merges.

### Consistency check against the 1.5 SKU Price columns

@ijurica asked at TF-2 on 2026-09-16 that any property that could later move to the SKU Price dataset be checked against the columns #2424 introduces. The check ran twice. The first pass, on 2026-09-17, found one item in the example data. The second, on 2026-09-18, reached `SkuPriceEligibility` and found that the supporting content claimed more than #2424's columns support.

Scenario B gave its data zone row a `RegionId` of "eu", the only macro-region `RegionId` value across the 93 example CSVs under `specification/data/`. `RegionId` is defined for "an isolated geographic area where a resource is provisioned or a service is provided", while `PricingRegionId` is the column that covers "a global or macro-region scope". The row now carries "northeurope", a single region inside the area CrestNode prices as one data zone. Two rows in Scenario B now carry two different regions, and only `RegionScope` shows that one was priced across the whole data zone and the other for its own region alone, which makes the point better than the macro-region did.

`SkuPriceEligibility` is the column the first pass did not reach. It carries a predicate over arbitrary dimensions, and five of the nine examples in its appendix constrain `RegionId`, so a reader of both PRs arrives at the comparison on their own. It is not a fourth geographic attribute, and the supporting content now says so rather than adding it to the relationship table. Three claims in that file were qualified as a result, listed in the changelog below. No specification text changes.

### What changes

* `SkuPriceDetails`: the property row is renamed and re-slotted after `Redundancy`, which sorts before `RegionScope`. The description reads "Breadth of the geographic area the SKU price applies to". Footnote 4 now separates `RegionScope` from `RegionId` as well as from `Redundancy`.
* Appendix: `serving_scope_examples.md` becomes `region_scope_examples.md`, anchor `#appendix.examples:regionscope`, rewritten in pricing terms. Every figure is unchanged. A new paragraph names the three questions above. The include and the overview row keep their slot, since "Region Scope" still sorts between Participating Entity Identification and SaaS.
* Data: `specification/data/serving_scope/` becomes `specification/data/region_scope/`, with both CSVs renamed and the JSON key updated. Keys stay alphabetical.
* Supporting content: a new `skupricedetails-regionscope.md` records the development background, covering the provider terminology behind the axis, every name considered and why each was set aside, the relationship with `PricingRegionId` and `SkuPriceEligibility` named directly, why neither of those recovers geographic breadth, and what the pair of dimensions would enable on a rate card if the property is later elevated to a SKU Price column.

### Open items

* **The value set.** The values are service-provider-defined: TF-2 on 2026-08-19 recorded that FOCUS defines the key and each service provider defines the values, and all eight String properties in the `SkuPriceDetails` table list examples rather than allowed values. Two lines in the appendix read as though the values were standardized, saying `RegionScope` "is what makes the scopes comparable across the two datasets" and that it "establishes that both rows describe global pricing". Those hold only where service providers converge on the same words, and today they do not: AWS names its inference profiles In-Region, Geo, and Global, Azure names its deployment types Global, Data zone, and Azure geography, and Google Cloud names its location types region, dual-region, and multi-region. @flanakin's AI #2672 research reaches the same place from the other side, that "DataZone" is Azure's word. Either the values get a normative set, or the two appendix lines soften to what service-provider-defined values support.
* **Value overlap with `Redundancy` on "Global".** `Redundancy` example values are "Local", "Zonal", "Global"; this property's are "Global", "Regional", "DataZone". Footnote 4 separates the two definitions and says a SKU may carry both, but it does not address the shared value. Raised as the second finding in @flanakin's AI #2672 research. Either the two value ladders differentiate, or footnote 4 states what each "Global" means.
* **SKU property or SKU price property.** Open question 2 from TF-2 on 2026-09-02. @ijurica was working through examples offline. No disposition posted since.

### Changelog

* **2026-09-18** (revision): Read the supporting content against #2424's SKU Price columns and qualified what it claimed. `SkuPriceEligibility` enters as a predicate over arbitrary dimensions rather than a fourth geographic attribute, since five of the nine examples in its appendix constrain `RegionId`. The `RegionScope` and `PricingRegionId` pair is no longer called lossless: which regions sit inside a boundary is recoverable from neither, so the pair is more informative than either alone without being complete. An eligibility inclusion set does not recover breadth either, because inclusions are absent on globally scoped prices, which is 24 of the 26 rows on #2424's rate card example, and a dimension omitted from them is defined as an implicit wildcard. The `SkuPriceId` join is qualified as one-to-many, resolved with the effective period, `ContractId`, quantity tier, and pricing currency. The elevation benefits are qualified on a value set consistent enough to group on, and the cross-provider comparison bullet now says so, since the provider table in the same file shows three vocabularies. The `IsGlobalScope` distinction moves out of the elevation benefits into present tense, because the two terms can be confused today. A new paragraph records that #2424's `scope_and_evolution.md` reaches the same availability diagnosis and names a companion SKU Properties dataset joined on `SkuId` as its resolution rather than a SKU Price column. No specification text changes.
* **2026-09-18** (base): Merged `working_draft` into #2613 and #2613 into this branch. Both branches now carry the `conditions` to `operating_model_conditions` rename, so the diff against `working_draft` no longer shows that rename as deletions. No content change.
* **2026-09-17** (revision): Added `supporting_content/datasets/cost_and_usage/columns/skupricedetails-regionscope.md`. Supporting content names real providers, so the AWS, Azure, and Google Cloud terminology behind the axis is recorded directly rather than through the fictitious providers the appendix uses, and it names `PricingRegionId`, which the appendix cannot until #2424 merges. The elevation section is written as a scenario rather than a commitment, and separates what the two dimensions enable on a rate card from what elevation would leave unsettled, which is the value set.
* **2026-09-17** (revision): Adopted `RegionScope` as the property name, per @flanakin's AI #2672 research and @marc-perreaut's independent proposal of 2026-09-03. The property row moves after `Redundancy`, the appendix and data directory are renamed, the anchor becomes `#appendix.examples:regionscope`, and footnote 4 gains the `RegionId` distinction alongside the `Redundancy` one. A new appendix paragraph names the three questions that sit close together: how broad an area a price covers, which single region a resource is provisioned in, and the identifier a service provider publishes for the priced area, which no Cost and Usage column carries. Every figure is unchanged.
* **2026-09-17** (revision): Qualified the model-identity claim in the Scenario B comparison. The closing bullet said the model-identity properties establish that the Aura Web and CrestNode rows describe the same model, which reads to include `ModelId`, and `ModelId` differs across those rows ("aura.solora-reasoning-pro" and "crestnode.solora-reasoning-pro"). `Examples: AI Model Identity` already settles this: the served `ModelId` is namespaced by the service provider, so `ModelDeveloper`, `ModelFamily`, and `ModelVersion` are what associate a charge with the underlying model across service providers. The bullet now reads "the shared model-identity properties".
* **2026-09-17** (revision): Gave the Scenario B data zone row a single-region `RegionId`. It carried "eu", a macro-region, where `RegionId` is defined for an isolated geographic area a resource is provisioned in. The row now carries "northeurope" and "North Europe", and the bullet drops the sentence calling "eu" a region identifier. Because two rows now carry two different regions, the bullet makes the point that `RegionId` names where a row ran and not how broad its price is. No figure changes.
* **2026-09-09** (revision): Dropped the applicability clause from footnote 4. The TF-2 record of 2026-09-02 has population following the same SHOULD every FOCUS-defined property carries, so a service provider may populate the property where it prices only one scope. The clause read narrower than that record.
* **2026-09-09** (revision): Applied @thecloudman's review. Both globally priced rows in Scenario A now carry `RegionId` "us-east-1" and `RegionName` "US East (N. Virginia)", the region Aura Web anchors the offering to, so only the scope property varies across the four rows. Scenario B keeps the null `RegionId` on its global row, so the two scenarios carry both conformant shapes, and its prose now names the shapes a globally priced row can take. The two `RegionId` bullets in Scenario A are rewritten. `ChargeDescription` on the Scenario A global rows reads "Globally priced" rather than "Globally routed", which was the last routing language in the diff. The appendix include moves to its rendered-title slot, after Participating Entity Identification and before SaaS, matching `appendix_overview.md`.
* **2026-09-02** (initial draft): Renamed `ServingScope` across the property row, footnote 4, the appendix file and its anchor, the data directory, and the JSON key in both CSVs, and rewrote the description and appendix prose in pricing terms. `region_scope_b.csv` also picked up LF line endings, matching its Scenario A counterpart; the repository sets no line-ending convention, and the file's diff shows five changed lines for three changed values as a result.

### Validation performed

Re-run on 2026-09-18 at head `f36dba32`. The 2026-09-18 revision touches only `supporting_content/`, so the specification checks below carry forward from `a302b86` unchanged.

* Linter clean on all three changed specification markdown files, and on the supporting content file, which sits outside the build lint scope.
* Every figure added on 2026-09-18 checked against #2424's branch: five of the nine `SkuPriceEligibility` appendix examples constrain `RegionId`, 24 of the 26 rows in `aura_web_rate_card_full.csv` carry `IsGlobalScope` with no inclusions, and the join members are quoted from that PR's `dataset.md`.
* `validate_includes.py` passes for the appendix directory.
* Every glossary anchor in the renamed appendix file resolves against `specification/glossary.md`, and no glossary term or data model target is linked twice, per the first-occurrence rule.
* Both CSVs return zero findings on the gap check. Every row's cost equals `PricingQuantity` times `ListUnitPrice`, the data zone and regional uplifts are 10% and 20%, and the three scoped rows in Scenario B sum to $48.48 against the $51.48 scenario total the prose states.
* `RegionName` is populated on every row whose `RegionId` is populated, and null on the two rows whose `RegionId` is null.
* `SkuPriceDetails` keys are alphabetical in both CSVs. The property row sits after `Redundancy`, and the appendix include and overview table row are each in their alphabetical slot.
* No remaining reference to `ServingScope`, `PricingRegionScope`, or either former path anywhere under `specification/`.
* The `gen_pdf` check covers the pandoc and wkhtmltopdf build and the rendered-anchor validation on each revision, and passes at `f36dba32`.

### Type of Change
* [ ] Bug fix
* [ ] Editorial / Formatting

### Author Checklist


specification/appendix/appendix.mdpp
```diff
@@ -6,13 +6,13 @@
 !INCLUDE "commitment_program_eligibility_details_examples.md",1
 !INCLUDE "contract_commitments/contract_commitments.mdpp",1
 !INCLUDE "correction_handling_examples/correction_handling_examples.mdpp",1
+!INCLUDE "pricing_region_scope_examples.md",1
 !INCLUDE "split_cost_allocation_examples.md",1
 !INCLUDE "invoice_detail/invoice_detail.mdpp",1
 !INCLUDE "json_object_examples/json_object_examples.mdpp",1
 !INCLUDE "metadata_examples/metadata_examples.mdpp",1
 !INCLUDE "participating_entity_identification.md",1
 !INCLUDE "saas_examples/saas_examples.mdpp",1
```
thecloudman: 🤖 [AI][Claude Code] Reviewed by thecloudman

The include order in this file follows the rendered section titles alphabetically, which is also the order the overview table uses. The new include is inserted before `split_cost_allocation_examples.md` (which renders as "Examples: Data Generator-Calculated Split Cost Allocation"), so the built spec will show "Examples: Pricing Region Scope" ahead of the Data Generator, Invoice Detail, JSON Object, Metadata, and Participating Entity sections, while the overview table lists it between Participating Entity Identification and SaaS. Moving the include to that slot keeps the rendered order and the overview table consistent.

```suggestion
!INCLUDE "split_cost_allocation_examples.md",1
!INCLUDE "invoice_detail/invoice_detail.mdpp",1
!INCLUDE "json_object_examples/json_object_examples.mdpp",1
!INCLUDE "metadata_examples/metadata_examples.mdpp",1
!INCLUDE "participating_entity_identification.md",1
!INCLUDE "pricing_region_scope_examples.md",1
!INCLUDE "saas_examples/saas_examples.mdpp",1
```

specification/appendix/pricing_region_scope_examples.md
```diff
@@ -0,0 +1,54 @@
...
+* The globally priced offering, which Aura Web may serve from any region, is priced at $2.50 per 1,000,000 input tokens and $12.50 per 1,000,000 output tokens.
+* The region-pinned offering of the same model is priced 20% higher, at $3.00 and $15.00 respectively.
+* The globally priced workload consumes 4,000,000 input tokens and 1,000,000 output tokens. The region-pinned workload consumes 2,000,000 input tokens and 500,000 output tokens.
+
+[**CSV Example**](/specification/data/pricing_region_scope/pricing_region_scope_a.csv)
+
+Note the following details in the example dataset:
+
+* PricingRegionScope carries the breadth of the priced area ("Global" or "Regional"), and the model-identity properties are identical across all four rows because every row describes the same model. Grouping by PricingRegionScope therefore isolates the cost of each scope for one model.
+* [RegionId](#datamodel.costandusage.regionid) and PricingRegionScope answer different questions. The region-pinned rows carry a RegionId of "us-east-1", while the globally priced rows carry no RegionId, because that price is not tied to a distinct region.
```
thecloudman: 🤖 [AI][Claude Code] Reviewed by thecloudman

Non-blocking. This bullet and the CSV model the globally priced rows with a null `RegionId`. In practice neither major provider leaves the region empty on a globally priced charge. AWS bills a global cross-region inference profile against the region the request entered, so a row for a globally priced model still carries a `RegionId` such as "us-east-1" alongside a meter name that marks it as global. Azure's FOCUS export emits a literal "global" region identifier for SKUs that have no regional deployment rather than leaving the column null. Downstream tooling commonly maps an empty region to "Global" for display, which is exactly the inference the next bullet warns against.

The definition does not depend on `RegionId` being null, so nothing here needs to change normatively. The example would make the `RegionId` / `PricingRegionScope` distinction more convincing, and closer to what practitioners will see in their own data, if at least one globally priced row carried a `RegionId` (for example the same "us-east-1" as the region-pinned rows, representing the region the request entered) with the bullet reworded along the lines of: "RegionId and PricingRegionScope answer different questions. Both the region-pinned rows and the globally priced rows carry a RegionId of \"us-east-1\", because that is where the request entered the service, and only PricingRegionScope shows that the globally priced rows were charged at the global rate rather than the regional one." This would also need the two global rows in `pricing_region_scope_a.csv` to carry `us-east-1` / `US East (N. Virginia)`.

Happy to leave the current shape if the intent is to show the null case specifically, in which case a sentence noting that some providers populate `RegionId` on globally priced rows would cover it.

thecloudman (commented):
🤖 [AI][Claude Code] **Review**

Reviewed by thecloudman

The rename to `PricingRegionScope` reads well and the pricing framing is the right one: billing data records which price point applied, not where a request was routed. Checked against both major providers' AI serving models (AWS Bedrock global / geographic / regional inference profiles; Azure OpenAI Global, Data Zone, and Standard deployments), the three example values map cleanly and neither provider needs any change to `RegionId` to carry the scope. No overlap or conflict with the existing `RegionId` / `RegionName` columns: those identify where the resource ran, this classifies how broad the price is.

Verified: linter clean on the three changed Markdown files, no residual `ServingScope` reference under `specification/`, `SkuPriceDetails` keys remain alphabetical in both CSVs, every cost in both CSVs equals `ListUnitPrice` × `PricingQuantity`, the data zone and regional uplifts are 10% and 20%, and the scoped rows sum to 48.48 of the 51.48 scenario total as the prose states.

Two comments inline: one ordering fix in `appendix.mdpp` (the overview table row is in the right slot but the include is not), and one non-blocking observation on how real provider data populates `RegionId` on globally priced rows.

specification/appendix/appendix.mdpp
Matt-Cowsert: Taking the include reorder as you have it, Graham. `appendix.mdpp` orders by rendered section title rather than by filename, and every other entry matches `appendix_overview.md`, so "Examples: Pricing Region Scope" belongs between `participating_entity_identification.md` and `saas_examples/saas_examples.mdpp`.

The filename is what misled the sort: `pricing_region_scope_examples.md` sorts before `split_cost_allocation_examples.md`, but that file renders as "Examples: Data Generator-Calculated Split Cost Allocation" and is already in its correct slot. `validate_includes.py` checks membership rather than order, so the build stays green either way and the mismatch shows only in the rendered document.

specification/appendix/pricing_region_scope_examples.md
Matt-Cowsert: Good catch Graham, taking the Scenario A half of the `RegionId` change. The `RegionId` requirements read "MUST NOT be null when a resource or service is operated in or managed from a distinct region" and "MAY be null when a resource or service is not operated in or managed from a distinct region", so an offering the service provider still anchors to an entry region falls in the first case and an offering with no home region falls in the second. Both are conformant; the example was showing one of them twice.

It also makes Scenario A's point better. With the same `RegionId` on all four rows, only `PricingRegionScope` moves, which shows the two answering different questions more sharply than a version where they happen to move together.

Keeping the null in Scenario B rather than changing both, so the two scenarios carry both shapes. CrestNode publishes no identifier for the area its global price covers, which the Scenario B bullet already says, and it now adds that some service providers anchor a globally priced offering to a region, as Aura Web does in Scenario A, and some publish an identifier for the global area itself.

The edit: both global rows in `pricing_region_scope_a.csv` pick up "us-east-1" and "US East (N. Virginia)", since `RegionName` MUST NOT be null when `RegionId` is not null, and the line 33 bullet takes your wording with one tweak. Udam Dewaraja asked on #2613 to keep request framing out of this property, so the bullet says the region Aura Web anchors the offering to rather than where the request entered. The `ChargeDescription` on those two rows also goes from "Globally routed" to "Globally priced" in the same commit, which was the last routing language left after the rename.

flanakin:
🤖 [AI][Claude] Suggest `RegionScope` instead of `PricingRegionScope`:

* **`Pricing` is redundant** — no other `SkuPriceDetails` key carries it; the column already sets that context.
* **Pairing with `PricingRegionId` is a collision** — that column holds an ID ("us-east-1"); this key holds a breadth ("Regional").

Alternatives: `RegionGranularity`, `RegionType`. Details: [#2672 research](https://github.com/FinOps-Open-Cost-and-Usage-Spec/FOCUS_Spec/pull/2613#issuecomment-5609210360)

Matt-Cowsert:
## TF2 Action Item - September 16, 2026

The SkuPriceDetails pricing region property, including its name, goes to the Open Forum on September 18, 2026 rather than to another Task Force 2 (TF-2) session. If the property is not sufficiently complete after the Open Forum, feature request (FR) #2414 may move to the 1.6 release. @Matt-Cowsert set that path on the September 16 call, after @flanakin proposed moving FR #2414 to 1.6 because of the work left before Member review.

Pull request (PR) #2667 stages the rename to PricingRegionScope on top of PR #2613, and the name is still open. @flanakin's research for Action Item #2672 recommends RegionScope. @ijurica asked on the call that any property that could later move to the SKU Price dataset stay consistent with the SKU Price columns being introduced in the 1.5 release, even when the property does not move now.

**Scope:**
- [ ] Update PR #2667 with the name and description to bring to the September 18 Open Forum
- [ ] Check the property against the SKU Price columns being introduced in the 1.5 release
- [ ] Set PR #2613 and PR #2667 to the 1.6 release if the property is not sufficiently complete after the Open Forum

**Assignee:** @Matt-Cowsert

specification/appendix/region_scope_examples.md
```diff
@@ -0,0 +1,56 @@
...
+
+* Global pricing is $2.40 per 1,000,000 input tokens, data zone pricing 10% higher at $2.64, and regional pricing 20% higher at $2.88.
+* Only input tokens are shown, since the input and output split is already illustrated in Scenario A.
+* A fourth row covers a second model, Solora Reasoning Lite, which CrestNode offers at a single scope and therefore prices without a scope distinction.
+
+[**CSV Example**](/specification/data/region_scope/region_scope_b.csv)
+
+Note the following details in the example dataset:
+
+* The SkuMeter values ("Glbl Std Inp Tokens", "DZ Std Inp Tokens", and "Rgnl Std Inp Tokens") are abbreviated in a way specific to this *service provider*, and they do not match the meter names Aura Web uses in Scenario A. SkuMeter has no FOCUS-defined value set, so it cannot be matched across *service providers*. RegionScope is what makes the scopes comparable across the two datasets.
```
flanakin: 🤖 [AI][Claude] 💡 **Suggestion**

nit: This reads as though RegionScope values were standardized. They are not — the SkuPriceDetails table lists "Global", "Regional", and "DataZone" as examples, not as an allowed set, so the cross-provider comparison only works when two service providers happen to pick the same word. AWS, Azure, and Google Cloud each use different ones today.

```suggestion
* The SkuMeter values ("Glbl Std Inp Tokens", "DZ Std Inp Tokens", and "Rgnl Std Inp Tokens") are abbreviated in a way specific to this *service provider*, and they do not match the meter names Aura Web uses in Scenario A. SkuMeter has no FOCUS-defined value set, so it cannot be matched across *service providers*. RegionScope is what makes the scopes comparable across the two datasets, as long as both *service providers* use the same value. FOCUS defines the property key; the values are *service-provider*-defined.
```

specification/appendix/region_scope_examples.md
```diff
@@ -0,0 +1,56 @@
...
+* The Solora Reasoning Lite row carries no RegionScope, because CrestNode prices that model without a scope distinction. It is not a globally priced charge; it is a charge to which the scope distinction does not apply. Measuring the share of spend at each scope counts it as unclassified rather than assigning it to a scope, which is why the three scoped rows sum to $48.48 against a scenario total of $51.48.
+* Comparing the two scenarios, global pricing of the same model costs $2.50 per 1,000,000 input tokens on Aura Web and $2.40 on CrestNode. RegionScope establishes that both rows describe global pricing, and the shared model-identity properties establish that both describe the same model, so neither fact requires interpreting a *service-provider*-specific *SKU* identifier. Pairing the rows as input token prices relies on SkuMeter, which carries the metered dimension as *service-provider*-specific text; a FOCUS-defined property for that dimension would remove the remaining interpretation step from this comparison.
```
flanakin: 🤖 [AI][Claude] 💡 **Suggestion**

nit: Same point as the sentence on the SkuMeter bullet. RegionScope values are service-provider-defined, so "Global" on both rows is an agreement between these two fictitious providers rather than something FOCUS guarantees. Worth saying so, since the bullet is making a cross-provider comparison.

```suggestion
* Comparing the two scenarios, global pricing of the same model costs $2.50 per 1,000,000 input tokens on Aura Web and $2.40 on CrestNode. RegionScope carries "Global" on both rows, which establishes that both describe global pricing because both *service providers* chose the same value, and the shared model-identity properties establish that both describe the same model, so neither fact requires interpreting a *service-provider*-specific *SKU* identifier. Pairing the rows as input token prices relies on SkuMeter, which carries the metered dimension as *service-provider*-specific text; a FOCUS-defined property for that dimension would remove the remaining interpretation step from this comparison.
```

specification/datasets/cost_and_usage/columns/skupricedetails.md
```diff
@@ -58,14 +58,14 @@ The following keys should be used when applicable to facilitate cross-SKU and cr
...
 <br><sup>3</sup> This is the operating system family of the SKU, if it's included with the SKU or the SKU only supports one type of operating system.
-<br><sup>4</sup> ServingScope applies when a service provider prices more than one serving scope for the same capability. ServingScope describes the breadth of the geographic area a request may be served from, which is distinct from Redundancy. Redundancy describes how widely a SKU is replicated for durability or availability, whereas ServingScope describes where the work is carried out. A SKU may carry both.
+<br><sup>4</sup> RegionScope describes the breadth of the geographic area a SKU price applies to, from a single region to any region. RegionScope is distinct from RegionId, which identifies the one region a resource is provisioned in or a service is provided from: a charge may carry a single RegionId while the price on that charge applies across a much broader area. RegionScope is also distinct from Redundancy, which describes how widely a SKU is replicated for durability or availability. A SKU may carry both.
```
flanakin: 🤖 [AI][Claude] 💡 **Suggestion**

nit: Redundancy and RegionScope both list "Global" among their example values, and this footnote says a SKU may carry both. So a SKU can carry "Global" twice meaning two different things, with nothing here to tell them apart. One clause naming each closes it.

```suggestion
<br><sup>4</sup> RegionScope describes the breadth of the geographic area a SKU price applies to, from a single region to any region. RegionScope is distinct from RegionId, which identifies the one region a resource is provisioned in or a service is provided from: a charge may carry a single RegionId while the price on that charge applies across a much broader area. RegionScope is also distinct from Redundancy, which describes how widely a SKU is replicated for durability or availability. A SKU may carry both. Both properties list "Global" among their example values: on Redundancy it describes replication across regions, and on RegionScope it describes a price that applies in any region.
```

flanakin (approved):
🤖 [AI][Claude] **PR Review**

**Summary:** Approving. The rename is clean and the evidence behind it is solid. Three wording nits below, all on open items you already flagged in the PR body.

### 💡 Suggestions (3)

1. `region_scope_examples.md:52` — "makes the scopes comparable across the two datasets" reads as though the values were standardized.
2. `region_scope_examples.md:56` — same, on "establishes that both rows describe global pricing".
3. `skupricedetails.md:68` — footnote 4 separates RegionScope from Redundancy but not their shared "Global" value.

### ✅ What's done well

* Both CSVs verified end to end: every row's cost equals PricingQuantity × ListUnitPrice, the 10% and 20% uplifts hold, and the three scoped rows in Scenario B sum to $48.48 against the $51.48 total the prose states.
* Every glossary and data model anchor in the renamed appendix resolves, each linked once on first occurrence.
* `RegionScope` sorts after `Redundancy` in the property table, and the appendix include and overview row are both in their rendered-title slot.
* Giving the Scenario B data zone row "northeurope" instead of the macro-region "eu" makes the RegionId-vs-RegionScope point better than the old data did, and no other example CSV uses a macro-region RegionId.
* No reference to `ServingScope` or either former path remains anywhere under `specification/`.

PR #2580, opened 2026-07-31
[AI] #2425: Add conditions to Content Constraints tables

jpradocueva:
## Summary
This PR updates the Content Constraints tables throughout the specification to explicitly identify conditional column dependencies by populating the Conditions column where applicable.

The objective is to make conditional requirements more visible and improve specification readability by allowing readers to quickly identify when a column depends on one or more conditions, without changing the normative behavior of the specification. This change is editorial in nature and improves consistency in documenting conditional dependencies across the specification.

As an initial implementation, this PR introduces three representative examples that demonstrate how the Conditions column should be populated:

- **No condition** – a column where no dependency exists, but the Conditions property is included for completeness and consistency.
   - **Suggestion**: Use “Not applicable” instead of “n/a” because it is explicit, accessible, and unambiguous. It clearly indicates that no condition applies, rather than suggesting the information is unavailable. Spelling out the phrase also avoids introducing an undefined abbreviation into the specification.
- **Single condition** – demonstrated by the Allocated Method Details column, where the column is applicable under one defined condition.
- **Multiple conditions** – demonstrated by the Pricing Currency Contracted Unit Price column, where the column depends on two independent conditions.

These examples establish the editorial pattern that can be consistently applied across the remainder of the specification in subsequent pull requests.

### Type of Change
* [x] Bug fix
* [ ] New feature / Specification update
* [x] Editorial / Formatting

### Author Checklist
* [x] **AI Usage Guidelines:** I have read the [FOCUS AI Usage Guidelines](/guidelines/contributors/ai-usage-guidelines.md).
* [x] **Self Review Attestation:** I have performed a self-review of my own contribution.
* [x] **AI-Generated Examples:** If this PR includes examples generated by AI, I have manually verified that the data is mathematically accurate, logically consistent, and compliant with the FOCUS schema.

PR #2557, opened 2026-07-24
FR #1974: add adjusted validation path for flexibility examples

izadorasobral:
## Summary
- add rule exclusion support to `specification/data/validate_examples.py`
- add a version-scoped exclusion list for `commitment_discount_flexibility` (FOCUS 1.2)
- document the adjusted validation workflow in contributor development processes
- update 4 flexibility CSV examples with the latest remediation edits

## Files changed
- `specification/data/validate_examples.py`
- `specification/data/commitment_discount_flexibility_excluded_rules_v1_2.txt`
- `guidelines/contributors/development-processes.md`
- `specification/data/commitment_discount_flexibility/one_hundred_percent_utilization_with_commitment_discount_flexibility_with_1_resource.csv`
- `specification/data/commitment_discount_flexibility/one_hundred_percent_utilization_with_commitment_discount_flexibility_with_2_resources.csv`
- `specification/data/commitment_discount_flexibility/one_hundred_percent_utilization_without_commitment_discount_flexibility.csv`
- `specification/data/commitment_discount_flexibility/zero_percent_utilization_without_commitment_discount_flexibility.csv`

## Context
Related to #1974.

### Full validator (`ALL` criteria)
- `one_hundred_percent_utilization_with_commitment_discount_flexibility_with_1_resource.csv`: `Fail=9`
- `one_hundred_percent_utilization_with_commitment_discount_flexibility_with_2_resources.csv`: `Fail=6`
- `one_hundred_percent_utilization_without_commitment_discount_flexibility.csv`: `Fail=6`
- `zero_percent_utilization_without_commitment_discount_flexibility.csv`: `Fail=9`

### Adjusted validator (approved exclusions for known OR-composite/cascading false-negatives)
- `AdjustedFail=0` for all 4 flexibility files using:
  - `--exclude-rules-file specification/data/commitment_discount_flexibility_excluded_rules_v1_2.txt`

PR #2670, opened 2026-09-03
AI #2669: Implement FOCUS RM Extraction in GitHub Actions

### LINKED ISSUE #2669: [AI] Implement FOCUS RM Extraction into the FOCUS Spec GitHub Actions
### Detailed Description

Runs validate and extract as hard gates on every PR push, then reports the
model test, verify, and diff audits through the "RM extracted", "RM curated",
and "RM match" labels plus a job summary table.

### Definition of Done

FOCUS RM Extraction executes on PR open, update 
Correct labels are applied to PRs based on the results of the extraction

mike-finopsorg:
## Summary

Adds .github/workflows/rm_extraction.yml, which runs the FOCUS_RM_Extraction action on every push to a PR branch.

* validate and extract are hard gates: malformed markdown or a generator that throws leaves nothing downstream worth reading, so a failure fails the job.
* The specification's own Requirements Model suite, verify, and diff are audits that do not pass against the published 1.5 model today, so they run with continue-on-error and report through labels rather than gating the PR.
* Three labels record three distinct claims: "RM extracted" (extract, model tests, and verify all pass), "RM curated" (verify reports zero warnings), and "RM match" (diff finds no content differences against the baseline).
* A job summary table names the first stage that did not pass, so folding four commands into one label costs no diagnostic detail.
* The extracted model is uploaded as a build artifact.

### Type of Change
* [ ] Bug fix
* [ ] Editorial / Formatting
* [X] Repo maintenance

### Author Checklist

### DIFF
--- a/.github/workflows/rm_extraction.yml
+++ b/.github/workflows/rm_extraction.yml
@@ -1,0 +1,229 @@
+# Requirements Model extraction on every push to a PR branch.
+#
+# What fails the build and what only sets a label follows what each step can assert. validate
+# and extract are preconditions: malformed markdown or a generator that throws leaves nothing
+# downstream worth reading, so they fail the job. The model suite, verify and diff are audits
+# that do not pass against the published 1.5 model today, so gating on them would block every
+# PR. They report through labels instead.
+#
+# Three labels, because they assert three different kinds of thing. Each is applied only when
+# its own condition holds:
+#
+#   RM extracted    validate, extract, the specification's own model suite and verify all pass
+#                   over the extracted model. Named for its subject rather than its suite:
+#                   build_json.py already runs those same tests over the published releases in
+#                   generate_draft_spec and its siblings, so "RM model tests" would not have
+#                   said which model was tested.
+#   RM curated      verify reports zero warnings. Warnings are curation left at its defaults,
+#                   which verify deliberately does not fail on, so a model can be entirely
+#                   sound with dozens outstanding. A different claim from soundness.
+#   RM match        diff finds no content differences against the baseline.
+#
+# They read as a ladder: a model has to extract soundly before its curation is worth counting,
+# and be both before matching the baseline. Folding four commands into RM extracted costs no
+# diagnostic detail, because the summary table names the first stage that did not pass.
+#
+# A PR earns none of the three today.
+name: Requirements Model
+
+on:
+  pull_request:
+    types: [opened, reopened, synchronize]
+
+# A rapid series of pushes should leave one run, for the tip.
+concurrency:
+  group: rm-${{ github.event.pull_request.number }}
+  cancel-in-progress: true
+
+permissions:
+  contents: read
+  pull-requests: write
+
+env:
+  ACTION_REPO: FinOps-Open-Cost-and-Usage-Spec/FOCUS_RM_Extraction
+  ACTION_REF: v1
+  SPEC: ./specification
+  BASELINE: ./specification/requirements_model/releases/latest
+  OUT: ./build/requirements_model
+
+jobs:
+  requirements-model:
+    name: Extract and audit
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v7
+      - uses: actions/checkout@v7
+        with:
+          repository: ${{ env.ACTION_REPO }}
+          ref: ${{ env.ACTION_REF }}
+          path: .rm-action
+
+      # Hard gate. No continue-on-error: if this fails the job fails, and the label step below
+      # still runs to strip any labels the PR was carrying from an earlier push.
+      - name: Validate and extract
+        id: extract
+        uses: ./.rm-action
+        with:
+          commands: validate,extract
+          specification: ${{ env.SPEC }}
+          baseline: ${{ env.BASELINE }}
+          output: ${{ env.OUT }}
+
+      - uses: actions/setup-python@v6
+        with:
+          python-version: '3.12'
+          cache: pip
+          cache-dependency-path: specification/requirements_model/requirements.txt
+      - name: Install model test dependencies
+        run: pip install -r specification/requirements_model/requirements.txt packaging
+
+      # The specification's own Requirements Model suite, run over the extracted model rather
+      # than over a published release. It asks a different question from verify: verify audits
+      # the transformation, this audits the model against the spec repo's own schema and
+      # conventions. The published 1.5 passes it cleanly, so a failure here is a real gap and
+      # not a quirk of the suite.
+      #
+      # The harness is staged in a scratch tree because build_json.py and tests/conftest.py
+      # both anchor to <root>/specification/requirements_model and discover releases by
+      # scanning releases/, so neither can be pointed at an arbitrary folder. Dropping the
+      # extraction into the real releases/ would either collide with the published release of
+      # the same ModelVersion, since both build to build/model-<version>.json and the last
+      # writer wins, or overwrite the baseline the diff step still needs.
+      - name: Model tests
+        id: modeltests
+        continue-on-error: true
+        env:
+          EXTRACTED: ${{ env.OUT }}
+        run: |
+          set -euo pipefail
+
+          EXTRACTED="$(cd "$EXTRACTED" && pwd)"
+          SOURCE="$PWD/specification/requirements_model"
+          STAGE="${RUNNER_TEMP:-/tmp}/rm-model-tests"
+
+          # conftest.py reads ROOT as parents[3] of itself, so both intermediate folders are
+          # load-bearing and the staged tree cannot be flattened.
+          MODEL="$STAGE/specification/requirements_model"
+          rm -rf "$STAGE"
+          mkdir -p "$MODEL/releases"
+          cp "$SOURCE/build_json.py" "$SOURCE/build_helpers.py" "$MODEL/"
+          cp -RL "$SOURCE/tests" "$MODEL/tests"
+
+          # Name the release for the model's own version, so pytest ids read v1.5 rather than
+          # something that never appears in the published model.
+          VERSION="$(python3 -c 'import json,sys; print(json.load(open(sys.argv[1]))["Details"]["ModelVersion"])' \
+            "$EXTRACTED/model_details.json")"
+
+          # -L, not -R: a release's json_schemas/datasets is a symlink out to
+          # specification/schemas, so copying it as a link leaves it dangling once staged and
+          # the build fails resolving the file(...) schema references.
+          cp -RL "$EXTRACTED" "$MODEL/releases/$VERSION"
+
+          echo "Testing extracted model $VERSION under the specification's own suite"
+          cd "$MODEL"
+          python3 build_json.py --logging-level WARNING
+
+      - name: Verify
+        id: verify
+        continue-on-error: true
+        uses: ./.rm-action
+        with:
+          commands: verify
+          specification: ${{ env.SPEC }}
+          baseline: ${{ env.BASELINE }}
+          output: ${{ env.OUT }}
+          node-version: ''
+
+      - name: Diff
+        id: diff
+        continue-on-error: true
+        uses: ./.rm-action
+        with:
+          commands: diff
+          baseline: ${{ env.BASELINE }}
+          output: ${{ env.OUT }}
+          summary: true
+          node-version: ''
+
+      # !cancelled() rather than always(): a run superseded by the next push should leave the
+      # labels alone, not race the run that replaced it. It does still run when extract failed,
+      # which strips whatever labels an earlier push left on the PR.
+      #
+      # The summary is written here rather than in its own step so the label decision has one
+      # source of truth: the table reports what was actually applied, not what the result
+      # implies should have been.
+      - name: Apply labels and report
+        if: ${{ !cancelled() }}
+        uses: actions/github-script@v9
+        env:
+          EXTRACT_OUTCOME: ${{ steps.extract.outcome }}
+          TESTS_OUTCOME: ${{ steps.modeltests.outcome }}
+          VERIFY_OUTCOME: ${{ steps.verify.outcome }}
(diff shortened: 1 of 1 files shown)


jpradocueva (approved):
@mike-finopsorg, I am really looking forward to this functionality. Thanks!

Arjun-Trident (approved):

.github/workflows/rm_extraction.yml
```diff
@@ -0,0 +1,229 @@
...
+
+env:
+  ACTION_REPO: FinOps-Open-Cost-and-Usage-Spec/FOCUS_RM_Extraction
+  ACTION_REF: v1
+  SPEC: ./specification
+  BASELINE: ./specification/requirements_model/releases/latest
+  OUT: ./build/requirements_model
+
+jobs:
+  requirements-model:
```
flanakin, line 50: 🤖 [AI][Claude] ⚠️ **Should fix**

`v1` is a mutable tag, and this action runs in a job holding `pull-requests: write`. Two consequences:

1. **Supply chain.** Whoever can push to `FOCUS_RM_Extraction` can change what executes here, retroactively, on every open pull request.
2. **Silent degradation.** The label logic reads `steps.verify.outputs.verify-warnings` and `steps.diff.outputs.diff-verdict` by name. If a `v1` retag renames or drops either output, the value arrives as an empty string. `RM curated` then evaluates `VERIFY_WARNINGS === '0'` as false and the label is *removed*, while the summary table reports `n/a`. Nothing fails, so the signal disappears without anyone noticing.

Pinning to a full commit SHA fixes both. The trailing comment keeps the human-readable version visible for Dependabot-style updates.

```suggestion
  ACTION_REF: <full-40-char-sha>  # v1
```

.github/workflows/rm_extraction.yml
```diff
@@ -0,0 +1,229 @@
...
+
+permissions:
+  contents: read
+  pull-requests: write
+
+env:
+  ACTION_REPO: FinOps-Open-Cost-and-Usage-Spec/FOCUS_RM_Extraction
+  ACTION_REF: v1
+  SPEC: ./specification
+  BASELINE: ./specification/requirements_model/releases/latest
```
flanakin, line 46: 🤖 [AI][Claude] ⚠️ **Should fix**

On a `pull_request` event from a fork, `GITHUB_TOKEN` is read-only regardless of the `permissions` block declared here. The label step's `catch` handles that gracefully and records `unavailable`, so nothing breaks — but the three labels, which the header comment presents as this workflow's primary output, will never appear on a fork pull request.

This is not hypothetical for this repository. Four of the last sixty pull requests came from forks (`izadoraluizalabs/FOCUS_Spec` on #2662, #2661, #2558, and `izadorasobral/FOCUS_Spec` on #2557), so external contributors are exactly the audience that gets no signal.

Worth deciding explicitly rather than discovering later. Two options:

* **Accept it**, and say so in the header comment alongside the existing note about the read-only fork token, so a reviewer seeing an unlabelled fork pull request knows it means "not measured" rather than "did not pass".
* **Cover it**, by splitting the labelling into a `workflow_run` job triggered on this workflow's completion. That runs in the base repository's context with a writable token, reading the audit results from the uploaded artifact this workflow already produces.

No suggestion block, since the fix is a design decision rather than a line edit.

.github/workflows/rm_extraction.yml
```diff
@@ -0,0 +1,229 @@
...
+
+on:
+  pull_request:
+    types: [opened, reopened, synchronize]
+
+# A rapid series of pushes should leave one run, for the tip.
+concurrency:
+  group: rm-${{ github.event.pull_request.number }}
+  cancel-in-progress: true
+
```
flanakin, line 37: 🤖 [AI][Claude] 💡 **Suggestion**

Without a `paths` filter, this runs on every pull request — including ones touching only `.github/`, `guidelines/`, or `supporting_content/`, none of which can change the extracted Requirements Model. Each run pays for a second repository checkout, a Python environment, and a full extract plus audit.

There is a real argument for keeping it unfiltered: a label that is sometimes absent because the workflow did not run is harder to read than one that is always present. If that is the intent, it is worth a line in the header comment so the cost is a recorded choice.

If not, a paths filter covering the inputs the extraction actually reads:

```suggestion
on:
  pull_request:
    types: [opened, reopened, synchronize]
    paths:
      - 'specification/**'
      - '.github/workflows/rm_extraction.yml'
```

Note that `types` as written restates the default (`opened`, `synchronize`, `reopened`), so it can be dropped either way.

flanakin (approved):
🤖 [AI][Claude] **PR Review**

**Summary:** Approving. The workflow is sound and unusually well-reasoned — the header comment explains the fail-vs-label split, the label ladder, and why `!cancelled()` beats `always()`, which is more design rationale than most workflows carry. Three non-blocking observations inline.

### ⚠️ Should fix (2)

1. `ACTION_REF: v1` is a mutable tag on an external action running with `pull-requests: write`.
2. Fork pull requests never receive labels, and forks are in active use in this repository.

### 💡 Suggestions (1)

1. No `paths` filter, so every pull request pays for an external checkout plus a full extraction.

### ✅ What's done well

* Every pinned action version resolves to a current release: `checkout@v7`, `setup-python@v6`, `github-script@v9`, `upload-artifact@v7` all exist upstream.
* The staged-tree rationale is correct. `tests/conftest.py` reads `ROOT = Path(__file__).resolve().parents[3]`, so staging at `$STAGE/specification/requirements_model/tests` does resolve `ROOT` to `$STAGE`, and both intermediate folders really are load-bearing.
* `build_json.py` accepts `--logging-level WARNING`, and its only local import is `build_helpers`, which the step copies. The staged harness will run.
* Reading every value through `env:` rather than interpolating into the `github-script` body is the right call for a workflow that runs on pull request content.
* Using `steps.<id>.outcome` rather than `conclusion` is correct for steps carrying `continue-on-error: true`.

.github/workflows/rm_extraction.yml
mike-finopsorg: The plan is to move to a SHA once we are comfortable there isn't any small edits to make to the external tool. Thanks for raising this I will open an AI to track that change. https://github.com/FinOps-Open-Cost-and-Usage-Spec/FOCUS_Spec/issues/2693

.github/workflows/rm_extraction.yml
```diff
@@ -0,0 +1,229 @@
...
+#                   said which model was tested.
+#   RM curated      verify reports zero warnings. Warnings are curation left at its defaults,
+#                   which verify deliberately does not fail on, so a model can be entirely
+#                   sound with dozens outstanding. A different claim from soundness.
+#   RM match        diff finds no content differences against the baseline.
+#
+# They read as a ladder: a model has to extract soundly before its curation is worth counting,
+# and be both before matching the baseline. Folding four commands into RM extracted costs no
+# diagnostic detail, because the summary table names the first stage that did not pass.
+#
```
mike-finopsorg, line 25:
```suggestion
#
# This workflow has been intentionally developed to run on only local PRs and forks will miss RM  
# execution. There is a low number of PRs that come in via fork and having fork PRs run this workflow was a 
# lot more tricky. I see this as a low issue at this stage as the TF-RM is unlikely to need all PRs to have a RM 
# match. We can review this later once we decide a more strict policy on RM labels.
#
```

.github/workflows/rm_extraction.yml
mike-finopsorg: Added comment to the header about this. its intentional to skip on forks at this point.

.github/workflows/rm_extraction.yml
```diff
@@ -0,0 +1,237 @@
+# Requirements Model extraction on every push to a PR branch.
```
Matt-Cowsert:
```suggestion
# Requirements Model extraction on every push to a PR that changes specification/ or this file.
```

With the new paths filter, pull requests that change neither `specification/` nor this file skip this workflow, so the opening line no longer describes when it runs.

.github/workflows/rm_extraction.yml
```diff
@@ -0,0 +1,237 @@
...
+
+# A rapid series of pushes should leave one run, for the tip.
+concurrency:
+on:
+  pull_request:
+    types: [opened, reopened, synchronize]
+    paths:
+      - 'specification/**'
+      - '.github/workflows/rm_extraction.yml'
+permissions:
```
Matt-Cowsert:
```suggestion
on:
  pull_request:
    types: [opened, reopened, synchronize]
    paths:
      - 'specification/**'
      - '.github/workflows/rm_extraction.yml'

# A rapid series of pushes should leave one run, for the tip.
concurrency:
  group: rm-${{ github.event.pull_request.number }}
  cancel-in-progress: true

permissions:
```

The second `on:` key makes GitHub reject the file, so the workflow never starts, and `concurrency:` has lost its `group` and `cancel-in-progress` values. This moves the paths filter from the earlier suggestion into the original trigger and restores those two lines from the first commit. The `!cancelled()` label step relies on `cancel-in-progress` so a superseded run leaves the labels alone.

.github/workflows/rm_extraction.yml
```diff
@@ -0,0 +1,237 @@
...
+  ACTION_REPO: FinOps-Open-Cost-and-Usage-Spec/FOCUS_RM_Extraction
+  ACTION_REF: v1
+  SPEC: ./specification
+  BASELINE: ./specification/requirements_model/releases/latest
+  OUT: ./build/requirements_model
+
+jobs:
+  requirements-model:
+    name: Extract and audit
+    runs-on: ubuntu-latest
```
Matt-Cowsert, line 60:
```suggestion
    name: Extract and audit
    runs-on: ubuntu-latest
    if: github.event.pull_request.head.repo.full_name == github.repository
```

You noted that skipping forks is intentional, but the workflow does not skip them yet: fork pull requests run the full job, and only the label writes fail. This condition skips the job for forks, so the header on lines 26 to 29 holds. If forks should run instead, the header needs rewording to say they run without labels.

.github/workflows/rm_extraction.yml
```diff
@@ -0,0 +1,237 @@
...
+      - uses: actions/checkout@v7
+      - uses: actions/checkout@v7
+        with:
+          repository: ${{ env.ACTION_REPO }}
+          ref: ${{ env.ACTION_REF }}
+          path: .rm-action
+
+      # Hard gate. No continue-on-error: if this fails the job fails, and the label step below
+      # still runs to strip any labels the PR was carrying from an earlier push.
+      - name: Validate and extract
```
Matt-Cowsert, line 72: This step fails against the current working_draft, so every pull request that triggers the workflow will show a failed check. The ecc50628 run failed on `ENOENT: no such file or directory, scandir '.../specification/conditions/'` ([log](https://github.com/FinOps-Open-Cost-and-Usage-Spec/FOCUS_Spec/actions/runs/35020805685/job/104555682008)).

#2580 made three changes that the extractor at `v1` does not handle yet: it renamed the folder to `specification/operating_model_conditions/`, renamed each file's `Condition ID` heading to `Operating Model Condition ID`, and changed condition links from `#conditions.` to `#operatingmodelconditions.`. This step needs an extractor release that handles all three, with `v1` moved to it. With only the folder updated, extraction would skip every condition file and mark every baseline condition rule as removed.

Matt-Cowsert (changes requested):
GitHub rejects `rm_extraction.yml` at 026d586a as an invalid workflow file ([run](https://github.com/FinOps-Open-Cost-and-Usage-Spec/FOCUS_Spec/actions/runs/35020922431)), so the Requirements Model check never ran on the latest commit and the pull request's status checks still report success. The file now has a second `on:` block where the `concurrency:` settings were; the suggestion on lines 34 to 46 moves the paths filter into the first `on:` block and restores those settings.

Once the file parses, the extract step will fail on every pull request that triggers it, because the extractor at `v1` still looks for `specification/conditions/`, which #2580 renamed after this pull request's first run. The comment on the Validate and extract step lists the three changes the extractor needs.

.github/workflows/rm_extraction.yml
mike-finopsorg: I believe I have now fixed this update in the upstream repo.

.github/workflows/rm_extraction.yml
```diff
@@ -0,0 +1,238 @@
...
+              c.state = c.want ? 'on' : 'off';
+              if (c.want === has) continue;          // only touch what actually changed
+              try {
+                if (c.want) await github.rest.issues.addLabels({ owner, repo, issue_number, labels: [c.label] });
+                else        await github.rest.issues.removeLabel({ owner, repo, issue_number, name: c.label });
+                core.info(`${c.want ? 'added' : 'removed'} "${c.label}"`);
+              } catch (err) {
+                // A pull_request run from a fork gets a read-only token. The audit still ran
+                // and reaches the summary table; only the label is unavailable, which the
+                // 'unavailable' state records. Not worth failing the job over.
```
Matt-Cowsert:
```suggestion
                // Fork PRs skip this job, so a failure here comes from the label API, such as
                // a label removed between the list and the write. The summary table still
                // reports each result; only the label is unavailable, which the 'unavailable'
                // state records. Not worth failing the job over.
```

These lines describe fork runs reaching this step, but the `if:` on line 61 now skips the whole job for forks, as the header note on lines 26 to 29 says. The rewording names what can still fail here and drops "The audit still ran", which is false when extract fails and this step only strips earlier labels.

Matt-Cowsert (approved):
The Requirements Model check passes now that FOCUS_RM_Extraction v1.1.0 reads the renamed operating model conditions and the `v1` tag points to it. Thanks for the release, Mike.

The description's first sentence says the workflow runs "on every push to a PR branch", but the workflow now skips pull requests from forks and pull requests that change neither `specification/` nor the workflow file. The description becomes the commit message on working_draft when this pull request is squash-merged, so could you update that sentence? For example:

> Adds .github/workflows/rm_extraction.yml, which runs the FOCUS_RM_Extraction action on pull requests from this repository that change specification/ or the workflow file.

Issue #2671, opened 2026-09-03
[AI] Research whether token caching and token direction generalize beyond tokens


Matt-Cowsert:
**Assignee(s):** @flanakin
**Parent Issue:** #2099

**Detailed Description**
Task Force 2 (TF-2) agreed on 2026-09-02 to replace the single TokenType property proposed on PR #2606 with two normalized FOCUS-defined properties in SkuPriceDetails: one for whether a token was served from or written to a cache, and one for whether a token is input or output. Working names are token caching and token direction. Both are scoped to tokens for now. @ijurica's reason: a token-bound property has a short set of allowed values the group can define, and a generic property would carry definitions the group would revise once other services supplied examples.

@flanakin took the question of whether either property should be generic instead. Direction has a candidate outside AI in network ingress and egress SKUs. Caching may apply to other services, and @flanakin said that should be checked before a generic caching property is proposed, without a deep investigation. @matt-cowsert is splitting the property on PR #2606 in parallel, so the findings can be reviewed against that split.

**Definition of Done**
- [ ] Findings on whether direction generalizes past tokens, with the network ingress and egress case as the first candidate
- [ ] Findings on whether caching generalizes past tokens, with the services examined named
- [ ] A recommendation on token-bound versus generic names for each property, posted on PR #2606

iwasinnam2:
@Matt-Cowsert on the 10 September TF-2 note: I am not asking FOCUS to adopt a format, and I will not repeat Mike Flanakin’s price-sheet work. He already closed the non-AI half of whether CacheAction should be generic.

Support @ijurica on the token prefix. A generator reads the property name, not footnote 4. I leave the “don’t open descriptions as questions” rewrite to the group.

The AI-side fact a generic name will lose is Dave Moreau’s parked third query on PR #2682: join cost to usage from a gateway that has no SKU. Out of scope for 1.5 is still a useful answer — as long as 1.5 does not absorb that event into CacheAction.

Two different zeros are already landing on the same month-end close, and neither is CacheAction Read:

-(1 Provider billed prompt-cache. SKU exists. Read/Write belong on the FOCUS row.

[...]

Issue #2488, opened 2026-06-25
[FR] Standardize cost representation for AI context cache storage charges


davidearney:
### Problem Statement

Multiple AI providers charge separately for **storing** a context cache — a distinct cost dimension from the inference tokens that read from or write to that cache. This storage charge is denominated in a compound unit (tokens × time) that has no normative representation in FOCUS:

- **Anthropic**: Context cache storage is billed at $3.00/MTok/hour (or fraction thereof). This is separate from cache write tokens (billed at 125% of base input price) and cache read tokens (billed at 10% of base input price). A single prompt-caching workflow generates up to three distinct charge types: write, read, and storage.
- **Google Vertex AI (Gemini)**: Context caching charges storage at a per-token-per-hour rate, similarly distinct from cache hit and cache miss inference charges.

The FOCUS `UnitFormat` attribute defines recommended unit names but contains no compound time-based unit (e.g., `"Token-Hours"` or `"K-Token-Hours"`). Without one, providers cannot represent cache storage pricing in a normalized, cross-comparable way. Current workarounds include provider-specific `SkuId` values, non-standard `PricingUnit` strings, or collapsing all cache charges into a single row that obscures the storage dimension. [#2363](https://github.com/FinOps-Open-Cost-and-Usage-Spec/FOCUS_Spec/issues/2363) expanded `UnitFormat` recommended values in v1.5 (adding `Session`, `Image`, `Character`, `Job`, `Page`) but did not address compound time-based units.

### Use Case / User Story

As a FinOps practitioner managing a RAG application that uses Anthropic context caching, I want cache storage charges to appear as a normalized, identifiable line item in my FOCUS data — with a cross-provider unit I can use to compare Anthropic and GCP Gemini cache storage costs — so that I can model the full cost of my caching strategy (write cost + storage cost + read savings) and make optimization decisions based on cache duration and hit rate.

### Success Criteria

- A FinOps practitioner can identify context cache storage rows in a FOCUS dataset by a normative `SkuMeter` value, without parsing provider-specific `SkuId` strings.
- Cache storage unit prices from Anthropic and GCP Vertex AI are expressed in the same `PricingUnit` format, enabling direct per-token-per-hour cost comparison.
- A practitioner can build a complete cache cost model: cache write cost + cache storage cost − cache read savings = net cost of caching strategy, using only normative FOCUS columns.
[...]

Issue #2099, opened 2026-03-21
[FR] Distinguish cached from uncached token costs in FOCUS


Matt-Cowsert:
### Problem Statement

GenAI API providers price cached and uncached tokens differently, creating distinct pricing dimensions within the same token type. Both major providers offer significant discounts on cached input tokens, but their caching models differ structurally:

- **Anthropic:** Cache reads priced at 10% of base input price. Cache writes carry a premium: 125% of base for 5-minute cache, 200% of base for 1-hour cache. Caching is explicit (opt-in via `cache_control`).
- **OpenAI:** Cached input tokens priced at 10% of base input price. No cache write premium. Caching is automatic (no code changes required).

Practitioners managing GenAI costs cannot currently:
- Distinguish cached from uncached token costs in a standardized format
- Compare caching cost savings across providers
- Identify cache write investment vs. cache read returns (where applicable)

Without guidance, data generators will represent cache pricing inconsistently. Some may merge cached and uncached tokens into a single row. Others may use different SkuMeter values, SkuPriceDetails properties, or ad-hoc custom columns. The result is the same cross-provider normalization problem FOCUS exists to solve.

### Use Case / User Story

**As a** FinOps practitioner managing GenAI API costs across multiple providers
**I need** cached and uncached token costs represented as distinct pricing dimensions in FOCUS data
**So that** I can quantify prompt caching savings and optimize caching strategies without provider-specific parsing logic.

### Success Criteria

1. Practitioners can filter or group FOCUS data by cache status (cached vs. uncached) to calculate caching savings without provider-specific logic
2. Cache duration tiers (where providers offer multiple durations) are represented with distinct pricing, enabling cost comparison between duration options
3. Data generators have documented guidance on which FOCUS columns carry cache status and cache duration information
4. Cache write costs and cache read savings are visible as separate line items, enabling ROI analysis on caching investment

### Type of Request

Enablement/Supporting Content - Non-normative items like appendix explanations, examples, or supporting documentation

### Organizations Requesting This Feature

- **Coinbase** - Nice to Have - Needs direct visibility into cache vs. non-cache token costs for actionable cost analysis. Interested in contributing to FOCUS working group.
- **NatWest** - Nice to Have - Managing multi-provider AI costs (from #1941)
- **Shutterstock** - Nice to Have - 20+ AI vendor normalization need (from #1941)
[...]

PR #2682, opened 2026-09-09
FR #2099: Replace TokenType with CacheAction and TokenDirection (stacked on #2606)

Matt-Cowsert:
## Summary

`SkuPriceDetails` gains two FOCUS-defined token properties, `CacheAction` and `TokenDirection`, replacing the single `TokenType` property proposed in #2606 for FR #2099. This PR is based on that PR's branch, so the diff shows only the replacement, and merging it carries the change into #2606. Nothing here targets `working_draft` directly.

Task Force 2 decided on September 2, 2026 to replace `TokenType` with two properties scoped to tokens, one for caching and one for direction, each with allowed values, under the working names "token caching" and "token direction". `TokenDirection` keeps the working name's shape. `CacheAction` drops the token binding from the name, for the reason given under open question 1.

**What changes**

* `SkuPriceDetails`: `TokenType` is replaced by two FOCUS-defined properties, each with allowed values. `CacheAction` carries "None", "Read", "Write", or "Other". `TokenDirection` carries "Input" or "Output". Both apply to SKUs metered in tokens. Population is SHOULD, as for every FOCUS-defined property; a SKU Price not metered in tokens carries neither, under the existing rule that properties not applicable to a `SkuPriceId` are not included.
* One requirement is added so allowed values are normative: a property value MUST be one of the allowed values when allowed values are specified for that property. The properties table's fourth column is retitled from "example values" to "values" so it can hold either kind. These are the first FOCUS-defined properties to carry allowed values, which @ijurica flagged on September 2, 2026.
* Footnote 4 defines the values, states that each token a service provider bills is counted on one row (so `ConsumedQuantity` on a "None" row excludes tokens billed on "Read" or "Write" rows), keeps the model-developer definition source from #2606, and lists what the properties do not carry: token modality, reasoning effort, cache retention duration, context window size, and retention charges metered in token-hours.
* The AI Prompt Caching appendix and both CSVs carry the two properties, and the appendix introduction points to the new supported-feature queries. The Scenario B context cache storage row carries neither property.
* The AI Model Identity appendix and both CSVs, which #2606 had labeled with `TokenType`, now carry `TokenDirection` and `CacheAction` "None".
* Resource Usage gains two example queries, cache hit rate per workload and input-to-output token ratio per model, and Cost Comparison gains a cache cost efficiency query. No new supported feature. Each feature's supporting column list gains the columns the new queries read (Resource Usage: ChargeClass, SkuPriceDetails, SubAccountId, Tags; Cost Comparison: ChargeClass, ConsumedUnit, ContractedUnitPrice, ListUnitPrice, PricingQuantity, SkuPriceDetails), and both features record 1.5 as the version updated. The hit rate and the ratio are two of the three queries @finops-DaveMoreau named on September 2, 2026; the third, joining cost data to usage from a gateway that has no SKU, is out of scope here. Cost efficiency is added because it catches what the hit rate cannot (below).

**Why two properties**

On September 2, 2026, @flanakin's objection to `TokenType` was that a value folding direction and caching into one string overlaps what `SkuMeter` and `ChargeDescription` already carry, so a practitioner still parses it to filter, while every other property in `SkuPriceDetails` names one attribute. @finops-DaveMoreau and @ijurica argued the same from the practitioner side: granular normalized attributes are what a practitioner filters on across providers.

Two questions a practitioner asks of cache billing each need one axis. What share of my input tokens came from the cache needs direction, to define "all input", and cache interaction, to find the reads. What did caching save me, net of what it cost, needs both axes plus the cost columns FOCUS already has. With `TokenType`, "all input" is a hand-maintained list of values that grows with every new token kind. With `TokenDirection`, it is one predicate.

The same pair of measures is converging outside FOCUS: a bounded cache hit rate rather than an unbounded reuse ratio, plus a cost-based companion that catches what the hit rate misses. The hit rate carries directly. There is no settled formula for the cost measure, so the Cost Comparison query follows the same form as the Discount Effectiveness by Service query already in that feature, and the shape holds whichever denominator practitioners land on.

**What the queries show on the example datasets**

All three queries were extracted from the supported-feature files and executed against the four CSVs in this PR, with `JSON_VALUE` mapped to its DuckDB equivalent and with `SubAccountId` and `Tags`, which the example datasets omit, supplied as null.

| Dataset | Cache hit rate | Input-to-output ratio | Cache cost efficiency |
|---|---|---|---|
| AI Prompt Caching, Scenario A | 76.2% | 7.0 | 63.8% |
| AI Prompt Caching, Scenario B | 76.2% | 7.0 | 49.5% |
| AI Model Identity, Scenarios A and B (no cache rows) | 0% | 3.3 | 0% |

Hit rate is identical across the two caching scenarios and cost efficiency is not, so the two measures are not redundant. Scenario B's figure includes its $6.00 context cache storage row; without it the query would report 68.6% and rank B above A, which is why the query carries retention. Scenario B remains, as the #2606 description says, a composite of two billing behaviors that do not occur together at one named service provider.

**Open questions for Task Force 2**

For the September 9, 2026 call.

1. **Names.** `TokenDirection` is token-bound. `CacheAction` is not, so a retention value could be added later without a rename, while its definition here is scoped to tokens. Direction stays token-bound here because the nearest non-token analog, network ingress and egress, uses a different vocabulary and has no feature request; it is also the case @flanakin named as the candidate for direction generalizing, and #2671 decides that. Symmetric token-bound names change nothing else in this PR.
2. **"None" as an allowed value.** No FOCUS allowed-value list uses "None" today. The alternative, omitting the property on uncached input rows, makes "not cached" indistinguishable from "not populated" and leaves the cost-efficiency query with no row to read the uncached unit price from.
3. **Retention charges.** Carried by the cost-efficiency query, which selects token-hour rows, rather than by a value. #2488 (context cache storage) is the durable home. If the group prefers a value, "Retain" is the candidate and the property's definition widens past tokens.
4. **A composite or vendor-named token type alongside the two properties.** Open question 3 from September 2, 2026. The case for keeping one is that a vendor-provided type name stays useful even though it does not normalize on its own. `SkuMeter` carries the service provider's meter name today; the question is whether a separate normalized composite is wanted or a concatenation query covers it.
5. **Placement** of the three queries under Resource Usage and Cost Comparison rather than a new supported feature.
6. **The allowed-values mechanism.** If Maintainers decline enumerated values on a FOCUS-defined property, the same values ship as examples and the requirement bullet and column header revert; the shape does not change.

**Verification**

* `make STYLE=working_draft` completed with no linter or validation errors, including the internal anchor checks, and the 140 spec tests and 180 requirements-model tests passed.
* The three supported-feature queries produced the table above.
* Every CSV row was checked programmatically: `SkuPriceDetails` keys in alphabetical order, both properties on every token-metered row and neither on the token-hour row, `PricingQuantity` x `ListUnitPrice` = `ListCost`, and `ListCost`, `ContractedCost`, `BilledCost`, and `EffectiveCost` equal on every row.
* `SkuPriceDetails` is the only column that changed in any CSV; every cost, quantity, and identifier value is unchanged.
* The FOCUS validator at version 1.3 reports the same pass and fail counts for the Scenario A CSV before and after this change (211 pass, 254 fail, the fails being column-absence cascades on an example dataset). At version 1.4 the validator fails with an internal rule-graph error on every CSV, including unchanged `working_draft` examples, so that layer is a tooling issue rather than a data finding.

> Note: Requirements-model rules are not included; those are updated downstream per release.

### Changelog

* **2026-09-08** (initial draft): Replaced `TokenType` with `CacheAction` and `TokenDirection` across the `SkuPriceDetails` table, footnote 4, both appendix examples, and all four CSVs. Added the allowed-values requirement and retitled the table column. Added the cache hit rate, input-to-output ratio, and cache cost efficiency queries to Resource Usage and Cost Comparison, with the columns they read added to each feature's supporting columns.

### Type of Change

* [ ] Bug fix
* [ ] Editorial / Formatting

### Author Checklist

* [ ] **Self Review Attestation:** I have performed a self-review of my own contribution.


specification/datasets/cost_and_usage/columns/skupricedetails.md
```diff
@@ -35,13 +38,23 @@ SkuPriceDetails MUST adhere to the following requirements:
...
+  * CacheAction MUST be "Other" when the *SKU Price* meters a cache-related token charge to which none of the other allowed values apply.
+  * TokenDirection MUST be "Input" when the tokens metered by the *SKU Price* are consumed from a request.
+  * TokenDirection MUST be "Output" when the tokens metered by the *SKU Price* are generated in a response.
+  * [ConsumedQuantity](#datamodel.costandusage.consumedquantity) on a row with CacheAction "Uncached" MUST NOT include tokens counted in ConsumedQuantity on a row with CacheAction "Read" or "Write" (e.g., a cached token that a usage report also counts inside a prompt token total is counted on the "Read" row only).
 
 ## FOCUS-Defined Properties
 
 The following keys should be used when applicable to facilitate cross-SKU and cross-service-provider queries for the same conceptual property. FOCUS-defined keys will appear in the list below and custom (e.g., service-provider-defined) keys will be prefixed with "x_" to make them easy to identify as well as prevent collisions.
```
ijurica, line 54: ## Rationale for retaining token-specific properties

I initially leaned toward the generic approach, as `CacheAction` and `DataDirection` could potentially be reused for other metering scenarios. However, after considering the semantics more closely, I now lean toward retaining the token-specific properties `TokenCacheAction` and `TokenDirection`.

The main reason is **semantic precision**. The current properties describe characteristics of token-metered SKUs, and their values have well-defined meanings in that context. Using token-specific names makes this scope explicit and avoids introducing broader abstractions whose semantics may differ across other metering scenarios.

For example, `Input` and `Output` have a clear meaning for token metering, but their semantics would not necessarily be the same for network traffic, storage operations, or other types of metered data. Similarly, `CacheAction` would require a consistent definition of cache interactions across potentially unrelated metering models.

I would therefore prefer to keep the properties token-specific, at least for now.

### Property definitions in the two options

#### Option A — Token-specific

| Key | Description | Data Type | Values |
|---|---|---|---|
| TokenCacheAction | Interaction of tokens with a prompt cache | String | "None", "Read", "Write", "Other" |
| TokenDirection | Direction of token flow | String | "Input", "Output" |

#### Option B — Generic

| Key | Description | Data Type | Values |
|---|---|---|---|
| CacheAction | Kind of interaction with a cache | String | "None", "Read", "Write", "Other" |
| DataDirection | Direction of the data flow | String | "Input", "Output" |

### Footnotes

Consider splitting the current footnote into **two separate footnotes**, one for each property.

specification/datasets/cost_and_usage/columns/skupricedetails.md
```diff
@@ -22,6 +22,9 @@ SkuPriceDetails MUST adhere to the following requirements:
   * SkuPriceDetails MUST include the FOCUS-defined SKU Price property when an equivalent property is included as a custom property.
   * SkuPriceDetails MUST NOT include properties that are not applicable to the corresponding SkuPriceId.
   * SkuPriceDetails SHOULD include all FOCUS-defined SKU Price properties listed below that are applicable to the corresponding SkuPriceId.
+  * SkuPriceDetails SHOULD include CacheAction when the *SKU Price* is metered in tokens.
+  * SkuPriceDetails SHOULD include TokenDirection when the *SKU Price* meters only tokens consumed from a request or only tokens generated in a response.
+  * SkuPriceDetails MUST NOT include CacheAction or TokenDirection when the *SKU Price* is not metered in tokens (e.g., a charge for retaining cached content metered in token-hours).
```
flanakin: 🤖 [AI] Two issues in these three bullets.

**1. `CacheAction` should not apply to output rows.** Generated tokens cannot interact with a cache, so a value there is noise rather than information. It also produces a misleading result: the AI Model Identity example is a service with no caching capability at all, and labelling its rows makes the cache hit rate query return 0% instead of no row. FOCUS's existing convention for not-applicable is to omit — see ChargeClass, CommitmentDiscountStatus, and CapacityReservationStatus, all of which are null rather than carrying a sentinel.

**2. The `MUST NOT` bullet is already covered.** Line 23 of this file states that SkuPriceDetails MUST NOT include properties that are not applicable to the corresponding SkuPriceId, and the property definitions scope both properties to token-metered SKU Prices. A per-property MUST NOT sets a precedent that will not scale as more FOCUS-defined properties gain scope conditions.

```suggestion
  * SkuPriceDetails SHOULD include CacheAction when the *SKU Price* meters tokens consumed from a request.
  * SkuPriceDetails SHOULD include TokenDirection when the *SKU Price* meters only tokens consumed from a request or only tokens generated in a response.
```

specification/datasets/cost_and_usage/columns/skupricedetails.md
```diff
@@ -35,13 +38,23 @@ SkuPriceDetails MUST adhere to the following requirements:
...
   * Property value MUST be of the type specified for that property.
   * Property value MUST represent the value for a single PricingUnit, denominated in the unit of measure specified for that property when the property holds a numeric value.
+  * Property value MUST be one of the allowed values specified for that property when allowed values are specified.
+* When included, CacheAction and TokenDirection MUST adhere to the following requirements:
+  * CacheAction MUST be "Read" when the *SKU Price* meters only request tokens served from a cache.
+  * CacheAction MUST be "Write" when the *SKU Price* meters only request tokens placed into a cache (e.g., a charge a service provider meters as cache creation).
+  * CacheAction MUST be "Uncached" when the *SKU Price* meters tokens without distinguishing a cache interaction (e.g., tokens generated in a response, or request tokens that a service provider bills on one meter whether or not they were placed into a cache).
+  * CacheAction MUST be "Other" when the *SKU Price* meters a cache-related token charge to which none of the other allowed values apply.
```
flanakin: 🤖 [AI] `"Uncached"` carries three different meanings and the appendix reads it two contradictory ways.

The bullet below defines the value against the **meter** ("meters tokens without distinguishing a cache interaction"), but the value name asserts a fact about the **tokens**. The appendix then splits on which reading applies:

* Scenario A: the row "holds only the input tokens that were neither served from nor placed into the cache" — a claim about the tokens.
* Scenario B: the row explicitly includes the tokens that populated the cache — a claim about the meter only.

Both cannot be true of one value. `"None"` states that no cache action is metered, which is accurate in every case including the mixed Scenario B meter, and it leaves the tokens themselves unclaimed.

Also dropping `"Other"`. The axis is closed: a metered request token was either served from a cache, placed into one, or the meter does not say. No surveyed provider bills a fourth thing. An escape hatch on the first FOCUS-defined property to carry allowed values undercuts the normalization it exists to provide, and new values can be added by version the way ServiceSubcategory does.

The `(e.g., tokens generated in a response, ...)` clause is removed here because output rows omit the property.

```suggestion
  * CacheAction MUST be "Read" when the *SKU Price* meters only request tokens served from a cache.
  * CacheAction MUST be "Write" when the *SKU Price* meters only request tokens placed into a cache (e.g., a charge a service provider meters as cache creation).
  * CacheAction MUST be "None" when the *SKU Price* meters request tokens without distinguishing a cache interaction (e.g., request tokens that a service provider bills on one meter whether or not they were placed into a cache).
```

specification/datasets/cost_and_usage/columns/skupricedetails.md
```diff
@@ -35,13 +38,23 @@ SkuPriceDetails MUST adhere to the following requirements:
...
+  * TokenDirection MUST be "Input" when the tokens metered by the *SKU Price* are consumed from a request.
+  * TokenDirection MUST be "Output" when the tokens metered by the *SKU Price* are generated in a response.
+  * [ConsumedQuantity](#datamodel.costandusage.consumedquantity) on a row with CacheAction "Uncached" MUST NOT include tokens counted in ConsumedQuantity on a row with CacheAction "Read" or "Write" (e.g., a cached token that a usage report also counts inside a prompt token total is counted on the "Read" row only).
```
flanakin: 🤖 [AI] The intent here is right — without it a provider that reports cached tokens inside a prompt token total would double count — but the bullet does not fit where it sits.

The section anchor above reads "When included, CacheAction and TokenDirection MUST adhere to the following requirements", and this bullet's subject is ConsumedQuantity. It also anchors on "a row", which is not the entity the rest of this section constrains; every other bullet constrains a *SKU Price*.

Restating it as a CacheAction requirement keeps the anti-double-count rule, matches the anchor, and uses a schema-level subject.

```suggestion
  * CacheAction MUST NOT be "None" for a *SKU Price* that meters tokens also metered by a *SKU Price* with CacheAction "Read" or "Write" (e.g., a cached token that a usage report also counts inside a prompt token total is metered only by the "Read" *SKU Price*).
```

specification/datasets/cost_and_usage/columns/skupricedetails.md
```diff
@@ -35,13 +38,23 @@ SkuPriceDetails MUST adhere to the following requirements:
   * Property key MUST match the spelling and casing specified for the FOCUS-defined property.
   * Property value MUST be of the type specified for that property.
   * Property value MUST represent the value for a single PricingUnit, denominated in the unit of measure specified for that property when the property holds a numeric value.
+  * Property value MUST be one of the allowed values specified for that property when allowed values are specified.
```
flanakin: 🤖 [AI] This requirement is enforceable only if a reader can tell which properties are closed, and right now that signal lives in prose inside a table cell — "Allowed values: ..." in the same column that otherwise reads "Examples: ...". Two properties in the same table behave differently with only a word distinguishing them, and the requirements model has no structure to bind to.

Every other closed value set in FOCUS gets three things: a `MUST be one of the allowed values` requirement, an `## Allowed Values` table with a description per value, and `Value format: Allowed values` in Content Constraints. See ChargeCategory, ChargeClass, CommitmentDiscountStatus, and CapacityReservationStatus.

Suggest matching that shape here: add a `Value Format` column to the FOCUS-Defined Properties table carrying either "Allowed values" or "Examples", and add a `## FOCUS-Defined Property Allowed Values` section with one `Value | Description` table per closed property.

If Maintainers would rather not introduce a per-property allowed-value mechanism in this PR, the fallback in the description works — ship the values as examples and drop this bullet — but the current middle position gives a normative requirement nothing concrete to point at.

specification/datasets/cost_and_usage/columns/skupricedetails.md
```diff
@@ -35,13 +38,23 @@ SkuPriceDetails MUST adhere to the following requirements:
...
 
 ## FOCUS-Defined Properties
 
 The following keys should be used when applicable to facilitate cross-SKU and cross-service-provider queries for the same conceptual property. FOCUS-defined keys will appear in the list below and custom (e.g., service-provider-defined) keys will be prefixed with "x_" to make them easy to identify as well as prevent collisions.
 
-| Key                      | Description                                                              | Data Type        | Unit of Measure (numeric) or example values (string)  |
+| Key                      | Description                                                              | Data Type        | Unit of Measure (numeric) or values (string)          |
 | :----------------------- | :----------------------------------------------------------------------- | :--------------- | :---------------------------------------------------- |
+| CacheAction              | How the tokens metered by the SKU interacted with a prompt cache<sup>4</sup> | String           | Allowed values: "Uncached", "Read", "Write", "Other"  |
```
flanakin: 🤖 [AI] Aligning this row with the value changes recommended on the requirements above: `"Uncached"` becomes `"None"` so the value describes the meter rather than making a claim about the tokens, and `"Other"` comes out because no surveyed provider bills a cache token charge that is not a read, a write, or neither.

```suggestion
| CacheAction              | How the tokens metered by the SKU interacted with a prompt cache<sup>4</sup> | String           | Allowed values: "None", "Read", "Write"               |
```

specification/datasets/cost_and_usage/columns/skupricedetails.md
```diff
@@ -59,13 +72,13 @@ The following keys should be used when applicable to facilitate cross-SKU and cr
...
+<br><sup>4</sup> CacheAction and TokenDirection apply to SKUs metered in tokens. The requirements above state when each property applies, what each value identifies, and that a token billed on a "Read" or "Write" row is not also counted on an "Uncached" row. The set of token kinds a model distinguishes is defined by the model developer. Which of those kinds a service provider meters as its own charge, and the meter name it uses, vary across service providers and across model versions, so these properties identify the kind of token independently of how a given service provider names its meters. Dimensions that qualify a charge without describing the direction of its tokens or their interaction with a cache, such as token modality, reasoning effort, cache retention duration, and context window size, are not values of these properties. [Examples: AI Prompt Caching](#appendix.examples:aipromptcaching) shows both properties on two service providers that meter cache activity differently.
```
flanakin: 🤖 [AI] Worth naming reasoning tokens explicitly in this footnote. Reasoning and thinking tokens are billed at the output rate and are covered by the output direction, so they need no property of their own — but #2487 is live and currently reads as if it might propose one. A single sentence here closes that off.

The footnote already lists reasoning effort as a dimension these properties do not carry, which is correct and a different point: effort is a setting, the tokens it produces are still output.

specification/appendix/ai_prompt_caching_examples.md
```diff
@@ -29,9 +29,9 @@ For this scenario, Acme Corp purchases the model directly from the model develop
...
-* Every token-metered row carries a TokenType, including the output row, so grouping the dataset by TokenType produces a breakdown that sums to the total token cost.
-* An input row processed without a cache carries "Input" rather than a distinct uncached value. Cache reads and cache writes are identified by their own values, so the remaining input tokens are the uncached ones.
+* CacheAction does not create that distinction; it normalizes it. The property carries "Uncached", "Write", and "Read" on those same three rows, and TokenDirection carries "Input" on all three, so a query can select cache reads, or every input token, without matching on the SkuMeter text.
+* Every token-metered row carries both properties, including the output row, which carries TokenDirection "Output" and CacheAction "Uncached", so grouping the dataset by either property produces a breakdown that sums to the total token cost.
+* The uncached input row carries CacheAction "Uncached". Cache reads and cache writes are identified by their own values, and each token is counted on one row, so the "Uncached" row holds only the input tokens that were neither served from nor placed into the cache.
```
flanakin: 🤖 [AI] This bullet reads the value as a claim about the tokens ("neither served from nor placed into the cache"), while the Scenario B bullet later in this file reads the same value as a claim about the meter, on a row that explicitly does include tokens that populated the cache. One of the two has to change.

Recommending the meter-based reading throughout, paired with renaming the value to `"None"` on the requirements. That reading is true in both scenarios; the token-based reading is true only in Scenario A.

```suggestion
* The uncached input row carries CacheAction "None", meaning the *SKU Price* does not meter a cache interaction. Cache reads and cache writes are metered by their own *SKU Prices*, so on this *service provider* the "None" row holds the input tokens that were neither served from nor placed into the cache.
```

specification/appendix/ai_prompt_caching_examples.md
```diff
@@ -53,10 +53,10 @@ For this scenario, the same model is served by a cloud provider, LatticeScale, w
...
-* The context cache storage row has its own SkuId and a SkuMeter value of "Context Cache Storage", and omits TokenType. It is denominated in token-hours rather than tokens, so it does not describe a kind of token consumed.
+* The SkuMeter wording differs from Scenario A for the same conceptual charge. Cache reads appear on a meter named "Cached Input Tokens" here and "Cache Read Input Tokens" in Scenario A, while both rows carry CacheAction "Read". Matching on CacheAction selects both; matching on SkuMeter text selects neither consistently.
+* No row carries CacheAction "Write". This *service provider* does not meter cache writes as their own charge, so no *SKU Price* holds that value.
+* The "Input Tokens" row covers 2,500,000 tokens, comprising both the tokens that populated the cache and those processed without caching. The *service provider* does not meter them separately, so the row carries CacheAction "Uncached" and the dataset cannot separate them either. The cost of populating the cache is not separable on this *service provider*, which is a property of its billing model rather than of the dataset.
```
flanakin: 🤖 [AI] This is the bullet that shows why the value name matters: the row is described as containing the tokens that populated the cache, and is then labelled "Uncached". Under the value name `"None"` the sentence is simply true, because the claim is about what the meter distinguishes rather than about the tokens.

Worth being explicit that this is why a query cannot treat the value as "definitely not cached" on every provider.

```suggestion
* The "Input Tokens" row covers 2,500,000 tokens, comprising both the tokens that populated the cache and those processed without caching. The *service provider* does not meter them separately, so the row carries CacheAction "None", meaning no cache interaction is metered, rather than a claim that none occurred. The cost of populating the cache is not separable on this *service provider*, which is a property of its billing model rather than of the dataset.
```

specification/supported_features/resource_usage.md
```diff
@@ -40,6 +48,60 @@ GROUP BY
...
+FROM focus_data_table
+WHERE ChargeCategory = 'Usage'
+  AND ChargeClass IS NULL
+  AND ConsumedUnit = 'Tokens'
+  AND JSON_VALUE(SkuPriceDetails, '$.TokenDirection') = 'Input'
+  AND ChargePeriodStart >= ? AND ChargePeriodEnd <= ?
+GROUP BY
+  ServiceProviderName,
+  SubAccountId,
+  Tags
```
flanakin: 🤖 [AI] Two problems with grouping this query by SubAccountId and Tags.

**`GROUP BY Tags` does not group by workload.** Tags is a JSON object, so this groups by the entire serialized tag set — every distinct combination of every tag becomes its own group. Getting a per-workload rate means extracting one tag key, which is provider- and customer-specific and not something a spec example can pick. No other supported-feature query in the spec puts Tags in SQL; it appears only as a directly dependent column of Cost and Usage Attribution.

**Neither column exists in the example datasets.** The PR description notes both were supplied as null to make the query run. An example query that cannot be executed against any dataset in the repo is hard for a reader to verify.

Grouping by ServiceProviderName and ModelId fixes both, matches the other two queries added in this PR, and runs against the shipped CSVs unchanged.

```suggestion
  ServiceProviderName,
  JSON_VALUE(SkuPriceDetails, '$.ModelId') AS ModelId,
  COALESCE(SUM(CASE WHEN JSON_VALUE(SkuPriceDetails, '$.CacheAction') = 'Read'
                    THEN ConsumedQuantity END), 0)
    / NULLIF(SUM(ConsumedQuantity), 0) AS CacheHitRate
FROM focus_data_table
WHERE ChargeCategory = 'Usage'
  AND ChargeClass IS NULL
  AND ConsumedUnit = 'Tokens'
  AND JSON_VALUE(SkuPriceDetails, '$.TokenDirection') = 'Input'
  AND ChargePeriodStart >= ? AND ChargePeriodEnd <= ?
GROUP BY
  ServiceProviderName,
  ModelId
```

specification/supported_features/resource_usage.md
```diff
@@ -14,12 +14,20 @@ FOCUS enables tracking of resource consumption by providing information about wh
...
 
 * ChargeCategory
+* ChargeClass
 * ChargePeriodEnd
 * ChargePeriodStart
 * ServiceProviderName
 * ServiceName
+* SkuPriceDetails
+* SubAccountId
+* Tags
```
flanakin: 🤖 [AI] SubAccountId and Tags are here only because the cache hit rate query groups by them, and that grouping does not work as intended — Tags is a JSON object, so grouping by it produces one group per distinct tag set rather than one per workload.

Supporting columns signal what a feature depends on, so adding two columns that Resource Usage does not otherwise need widens the feature's footprint for a query detail. Recommending they come out along with the grouping change.

SkuPriceDetails stays: the two new queries genuinely read it.

```suggestion
* SkuPriceDetails
```

specification/supported_features/resource_usage.md
```diff
@@ -40,6 +48,60 @@ GROUP BY
   ConsumedUnit
 ```
 
+### Cache Hit Rate for Token-Metered SKUs
+
+Computes the share of input tokens served from a prompt cache, per workload, using the TokenDirection and CacheAction properties. The denominator is every input token row, so it holds whether or not a service provider meters cache writes as their own charge. Where a service provider emits cache read rows but no uncached input row, the denominator loses that bucket and the ratio overstates the hit rate.
```
flanakin: 🤖 [AI] This paragraph flags one coverage gap; there is a second worth stating, because it fails silently rather than skewing the number.

A *service provider* that bills request and response tokens on a single blended meter has no direction to report, so TokenDirection is omitted on that row and the `= 'Input'` predicate drops it from both numerator and denominator. The query then returns a rate for a subset of the workload without indicating that it did.

```suggestion
Computes the share of input tokens served from a prompt cache, per model, using the TokenDirection and CacheAction properties. The denominator is every input token row, so it holds whether or not a service provider meters cache writes as their own charge. Where a service provider emits cache read rows but no uncached input row, the denominator loses that bucket and the ratio overstates the hit rate. Where a service provider bills request and response tokens on a single meter, TokenDirection is not populated and those rows are excluded entirely.
```

specification/datasets/cost_and_usage/columns/skupricedetails.md
flanakin: 🤖 [AI] Output tokens can't hit a cache, so leave CacheAction off those rows. The MUST NOT bullet repeats the existing "not applicable" rule.

```suggestion
  * SkuPriceDetails SHOULD include CacheAction when the *SKU Price* meters tokens consumed from a request.
  * SkuPriceDetails SHOULD include TokenDirection when the *SKU Price* meters only tokens consumed from a request or only tokens generated in a response.
```

specification/datasets/cost_and_usage/columns/skupricedetails.md
flanakin: 🤖 [AI] "Uncached" is defined by the meter but reads as a claim about the tokens. "None" fits every case. "Other" has no real use.

```suggestion
  * CacheAction MUST be "Read" when the *SKU Price* meters only request tokens served from a cache.
  * CacheAction MUST be "Write" when the *SKU Price* meters only request tokens placed into a cache (e.g., a charge a service provider meters as cache creation).
  * CacheAction MUST be "None" when the *SKU Price* meters request tokens without distinguishing a cache interaction (e.g., request tokens that a service provider bills on one meter whether or not they were placed into a cache).
```

specification/datasets/cost_and_usage/columns/skupricedetails.md
flanakin: 🤖 [AI] This bullet's subject is ConsumedQuantity, but it sits under the CacheAction/TokenDirection anchor. Suggest restating it as a CacheAction rule.

```suggestion
  * CacheAction MUST NOT be "None" for a *SKU Price* that meters tokens also metered by a *SKU Price* with CacheAction "Read" or "Write" (e.g., a cached token that a usage report also counts inside a prompt token total is metered only by the "Read" *SKU Price*).
```

specification/datasets/cost_and_usage/columns/skupricedetails.md
flanakin, line 41: 🤖 [AI] This depends on "Allowed values" text inside a table cell. Suggest one rule per property, plus an Allowed Values table for each (like ChargeCategory).

```suggestion
  * When CacheAction is included, CacheAction MUST be one of the allowed values.
  * When TokenDirection is included, TokenDirection MUST be one of the allowed values.
```

specification/datasets/cost_and_usage/columns/skupricedetails.md
flanakin: 🤖 [AI] Match the value changes above.

```suggestion
| CacheAction              | How the tokens metered by the SKU interacted with a prompt cache<sup>4</sup> | String           | Allowed values: "None", "Read", "Write"               |
```

specification/datasets/cost_and_usage/columns/skupricedetails.md
flanakin: 🤖 [AI] Suggest one sentence noting reasoning tokens are "Output", so #2487 doesn't add a new property.

specification/appendix/ai_prompt_caching_examples.md
flanakin: 🤖 [AI] Scenario A says this value means "not cached." Scenario B uses it on a row that includes cache writes. Pick one meaning.

```suggestion
* The uncached input row carries CacheAction "None", meaning the *SKU Price* does not meter a cache interaction. Cache reads and cache writes are metered by their own *SKU Prices*, so on this *service provider* the "None" row holds the input tokens that were neither served from nor placed into the cache.
```

specification/appendix/ai_prompt_caching_examples.md
flanakin: 🤖 [AI] This row includes cache writes but is labeled "Uncached". With "None" (no cache action metered), the sentence holds.

```suggestion
* The "Input Tokens" row covers 2,500,000 tokens, comprising both the tokens that populated the cache and those processed without caching. The *service provider* does not meter them separately, so the row carries CacheAction "None", meaning no cache interaction is metered, rather than a claim that none occurred. The cost of populating the cache is not separable on this *service provider*, which is a property of its billing model rather than of the dataset.
```

specification/supported_features/resource_usage.md
flanakin, line 72: 🤖 [AI] `GROUP BY Tags` groups by the whole JSON object, not by workload. Neither column exists in the example data. Group by model instead.

```suggestion
  ServiceProviderName,
  JSON_VALUE(SkuPriceDetails, '$.ModelId') AS ModelId,
  COALESCE(SUM(CASE WHEN JSON_VALUE(SkuPriceDetails, '$.CacheAction') = 'Read'
                    THEN ConsumedQuantity END), 0)
    / NULLIF(SUM(ConsumedQuantity), 0) AS CacheHitRate
FROM focus_data_table
WHERE ChargeCategory = 'Usage'
  AND ChargeClass IS NULL
  AND ConsumedUnit = 'Tokens'
  AND JSON_VALUE(SkuPriceDetails, '$.TokenDirection') = 'Input'
  AND ChargePeriodStart >= ? AND ChargePeriodEnd <= ?
GROUP BY
  ServiceProviderName,
  ModelId
```

specification/supported_features/resource_usage.md
flanakin, line 24: 🤖 [AI] Only the Tags grouping needed these. Remove them with it.

```suggestion
* SkuPriceDetails
```

specification/supported_features/resource_usage.md
flanakin: 🤖 [AI] Add: rows from a single combined input+output meter have no TokenDirection, so this query silently skips them.

```suggestion
Computes the share of input tokens served from a prompt cache, per model, using the TokenDirection and CacheAction properties. The denominator is every input token row, so it holds whether or not a service provider meters cache writes as their own charge. Where a service provider emits cache read rows but no uncached input row, the denominator loses that bucket and the ratio overstates the hit rate. Where a service provider bills request and response tokens on a single meter, TokenDirection is not populated and those rows are excluded entirely.
```

flanakin (approved):
🤖 [AI] Approving. Splitting TokenType into two properties is right. Please address before merge:

**Blocking**
* Appendix uses "Uncached" two ways: Scenario A means not cached, Scenario B includes cache writes.
* `GROUP BY Tags` groups by the whole JSON object, not by workload.

**Needs a decision**
* How allowed values work. See the inline suggestion.

**Naming (for TF-2):** `TokenCacheAction` + `TokenDirection`. An Azure price sheet (753K rows) has no read/write cache meters outside AI. Every cache meter bills by time or capacity.

Matt-Cowsert:
## TF2 Action Item - September 9, 2026

PR #2682 replaces the single TokenType property proposed on PR #2606 with two FOCUS-defined properties carrying allowed values. Task Force 2 (TF-2) reviewed it on September 9, 2026 and agreed the shape is the direction for the 1.5 release. @Matt-Cowsert said he would merge PR #2682 into PR #2606 within about a day, and @flanakin agreed to hold his own follow-on work until after that merge.

Two naming items came out of the session and are unsettled. @ijurica proposed prefixing the cache property with the token scope, since the property description binds it to tokens. She also proposed rewriting both property descriptions so they do not open as questions, which is how every other property description is written. @flanakin proposed a generic data direction property instead of the token-scoped one, and the group did not settle that in session. Action Item #2671 owns the recommendation.

**Scope:**
- [x] Merge PR #2682 into PR #2606
- [x] Record a decision on @ijurica's proposal to prefix the cache property with the token scope
- [ ] Record a decision on @ijurica's proposed rewrite of both property descriptions (carried to the September 16 PR Action on PR #2606: https://github.com/FinOps-Open-Cost-and-Usage-Spec/FOCUS_Spec/pull/2606#discussion_r4032128401)
- [x] Record a decision on whether the direction property stays token-scoped or becomes generic, once Action Item #2671 reports
---

**Assignee:** @Matt-Cowsert

specification/datasets/cost_and_usage/columns/skupricedetails.md
flanakin: I'm good with either. My reasoning:

**Caching**: After looking at the Azure price sheet, I don't see any action-specific cache SKUs, so that difference may not even be needed from a price perspective. I do suspect there's a larger-scala analysis that could be done to drive efficiency, which falls into the FinOps domain, but this wouldn't be driven based on the usage data, as far as I can tell. That would require obswervability metrics.

**Direction**: While we def need DataDirection, the token usage of this can be argued to not be about data flow/movement. The token case is more like a "parameter" (input) and "return value" (output), which is slightly different than data flow.

